PR #1079, opened 2020-11-10
Fix to do not filter out service manifests

nakabonne:
Currently, define [service name](https://pipecd.dev/docs/user-guide/configuration-reference/#kubernetesservice):
```yaml
kind: KubernetesApp
spec:
  service:
    name: foo
```

Then that manifest was being filtered out. Where all we have to do filter out is Istio's `VirtualService`, isn't it? If so, this PR will make it proper behavior.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/executor/kubernetes/primary.go
+++ b/pkg/app/piped/executor/kubernetes/primary.go
@@ -47,32 +47,25 @@ func (e *Executor) ensurePrimaryRollout(ctx context.Context) model.StageStatus {
 	}
 	e.LogPersister.Successf("Successfully loaded %d manifests", len(manifests))
 
-	// Find traffic routing manifests and filter out it from primary manifests.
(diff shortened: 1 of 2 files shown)


nakabonne:
Just added test cases for those.

pkg/app/piped/executor/kubernetes/primary.go
```diff
@@ -47,32 +47,25 @@ func (e *Executor) ensurePrimaryRollout(ctx context.Context) model.StageStatus {
...
-		e.LogPersister.Errorf("Failed while finding traffic routing manifest: (%v)", err)
-		return model.StageStatus_STAGE_FAILURE
-	}
-
-	if len(trafficRoutingManifests) > 1 {
-		e.LogPersister.Infof(
-			"Detected %d traffic routing manifests but only the first one (%s) will be used",
-			len(trafficRoutingManifests),
-			trafficRoutingManifests[0].Key.ReadableString(),
-		)
```
nakabonne, line 62: And I feel like no need to tell them.

PR #1181, opened 2020-12-02
Use nodejs scripts instead of sed to modify some imports

cakecatz:

Use JS to strip unnecessary imports from generated files(`:api_client`, `:models`).
I think it's better than using `sed` because `nodejs_binary` is added as Bazel dependency(`sed` is not).

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/BUILD.bazel
+++ b/pkg/app/web/BUILD.bazel
@@ -21,9 +21,7 @@ nodejs_binary(
 
 genrule(
     name = "build_api",
-    srcs = [
-        ":api_proto",
-    ],
+    srcs = [":api_proto"],
     outs = ["api_client"],
     cmd = "$(locations strip_imports) $(OUTS) $(SRCS)",
     output_to_bindir = True,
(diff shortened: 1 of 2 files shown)


pkg/app/web/BUILD.bazel
```diff
@@ -13,42 +14,31 @@ ts_proto_library(
...
+    name = "strip_imports",
+    entry_point = ":strip-imports.js",
+)
+
 genrule(
     name = "build_api",
-    srcs = [":api_proto"],
+    srcs = [
+        ":api_proto",
+    ],
```
khanhtc1202, line 26: nits, looks like some lint on local config changed 👀

pkg/app/web/strip-imports.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
khanhtc1202, line 1: I think we could remove this line 👀

cakecatz: Ah, it throws an error by ESLint currently.
But I think it does not need this rule for `.js` files.
I will change `.eslintrc` to exclude `.js` files from that rule later. 
Thank you!

khanhtc1202: I see, thank you :pray:

PR #1246, opened 2020-12-14
Compare image in the container registry and one in git

nakabonne:

Fixes https://github.com/pipe-cd/pipe/issues/1209

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/imagewatcher/watcher.go
+++ b/pkg/app/piped/imagewatcher/watcher.go
@@ -60,7 +60,8 @@ func NewWatcher(cfg *config.PipedSpec, gitClient gitClient, logger *zap.Logger) 
 	}
 }
 
-// Run spawns goroutines for each image provider.
+// Run spawns goroutines for each image provider. They periodically pull the image
+// from the container registry to compare the image with one in the git repository.
 func (w *watcher) Run(ctx context.Context) error {
 	// Pre-clone to cache the registered git repositories.
 	for _, r := range w.config.Repositories {
@@ -134,7 +135,7 @@ func (w *watcher) determineUpdates(ctx context.Context, repoID string, repo git.
 		return nil, fmt.Errorf("failed to fetch from and integrate with a local branch: %w", err)
 	}
 
-	// Load Image Watcher Config
+	// Load Image Watcher Config for the given repo.
 	includes := make([]string, 0)
 	excludes := make([]string, 0)
 	for _, target := range w.config.ImageWatcher.Repos {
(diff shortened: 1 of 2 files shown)


pkg/app/piped/imagewatcher/watcher.go
```diff
@@ -96,90 +100,99 @@ func (w *watcher) run(ctx context.Context, provider imageprovider.Provider, inte
...
-				zap.Error(err),
-			)
+// determineUpdates gives back target images to be updated for a given repo.
+func (w *watcher) determineUpdates(ctx context.Context, repoID string, repo git.Repo, provider imageprovider.Provider) ([]config.ImageWatcherTarget, error) {
+	branch := repo.GetClonedBranch()
+	w.mu.Lock()
+	err := repo.Pull(ctx, branch)
+	w.mu.Unlock()
+	if err != nil {
+		return nil, fmt.Errorf("failed to update repository branch: %w", err)
```
khanhtc1202, line 134: nit: "failed to get update from branch" or "failed to pull branch"

nakabonne: Yup, you're right, thank you!

pkg/app/piped/imagewatcher/watcher.go
```diff
@@ -96,90 +100,99 @@ func (w *watcher) run(ctx context.Context, provider imageprovider.Provider, inte
...
 
-// filterTargets gives back the targets corresponding to the given provider.
-func filterTargets(provider string, targets []config.ImageWatcherTarget) (filtered []config.ImageWatcherTarget) {
-	for _, t := range targets {
-		if t.Provider == provider {
-			filtered = append(filtered, t)
-		}
+// checkOutdated checks if the image defined in the given target is identical to the one in image provider.
+func checkOutdated(ctx context.Context, target config.ImageWatcherTarget, repo git.Repo, provider imageprovider.Provider) (bool, error) {
+	if provider.Name() != target.Provider {
```
khanhtc1202, line 170: nit: could you add notes to clarify why do we treat this case as a not outdated image 🙏

nakabonne: As you feel, it's unclear. gonna update them.

nakabonne: Updated them.
The Run() spawns runners for each image provider, so the checkOutdated function is called in the context for image provider. We should ignore the targets for other providers.

khanhtc1202:
Nice 👍 
/approve

pkg/app/piped/imagewatcher/watcher.go
khanhtc1202: 🙆

PR #1248, opened 2020-12-14
Implement insight page

cakecatz:

image

This screenshot uses dummy data.

Ref: https://github.com/pipe-cd/pipe/issues/1160

Fixes #

```release-note
Add insight page
```

### DIFF
--- a/pkg/app/web/src/modules/deployment-frequency.test.ts
+++ b/pkg/app/web/src/modules/deployment-frequency.test.ts
@@ -1,0 +1,16 @@
+import { deploymentFrequencySlice } from "./deployment-frequency";
+
+describe("deploymentFrequencySlice reducer", () => {
+  it("should handle initial state", () => {
+    expect(
+      deploymentFrequencySlice.reducer(undefined, {
+        type: "TEST_ACTION",
+      })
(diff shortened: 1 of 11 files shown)


pkg/app/web/src/modules/deployment-frequency.test.ts
```diff
@@ -0,0 +1,16 @@
...
+
+describe("deploymentFrequencySlice reducer", () => {
+  it("should handle initial state", () => {
+    expect(
+      deploymentFrequencySlice.reducer(undefined, {
+        type: "TEST_ACTION",
+      })
+    ).toMatchInlineSnapshot(`
+      Object {
+        "data": Array [],
```
khanhtc1202, line 11: nit: should it be `[]` 🤔

cakecatz: This code is generated by `jest`, so we'll leave it as is.

khanhtc1202: I see 😅 thanks

khanhtc1202:
/approve
🚀🚀🚀🚀🚀

PR #1280, opened 2020-12-17
Refactor deployment list page

### LINKED ISSUE #1178: Refresh page on switch between Application and Deployment tabs
**What happened**:
On ControlPlane, move to Deployments tab from Applications tab currently does not refresh deployments list 👀 I have to click `refresh` button directly.

Auto refresh items list on tab change events.

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

cakecatz:

* This refactoring will fix a bug that not refreshed deployment list 

Fixes #1178

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/types/module.ts
+++ b/pkg/app/web/src/types/module.ts
@@ -1,0 +1,1 @@
+export type LoadingStatus = "idle" | "loading" | "succeeded" | "failed";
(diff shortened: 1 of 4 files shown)


pkg/app/web/src/types/module.ts
```diff
@@ -0,0 +1 @@
+export type LoadingStatus = "idle" | "loading" | "succeeded" | "failed";
```
khanhtc1202, line 1: nit, sorry for the dumb question but I don't understand when do we need "idle" state 🤔

cakecatz: @khanhtc1202 
`idle` is used to represent the state before fetching.

https://redux.js.org/tutorials/essentials/part-5-async-logic#loading-state-for-requests

It is not used currently, but it will be used in other modules/components.

khanhtc1202: ah, got it, thank you 🙏

nghialv (approved):

PR #1297, opened 2020-12-21
Update commit message generated by Image Watcher

nakabonne:
It used to too long.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/imagewatcher/watcher.go
+++ b/pkg/app/piped/imagewatcher/watcher.go
@@ -35,6 +35,7 @@ import (
 )
 
 const (
+	// image ref, file name and new tag are supposed.
 	defaultCommitMessageFormat = "Update %s in %s to %s"
 	defaultCheckInterval       = 5 * time.Minute
 )
--- a/pkg/app/piped/imagewatcher/watcher.go
+++ b/pkg/app/piped/imagewatcher/watcher.go
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	defaultCommitMessageFormat = "Update image %s to %s defined at %s in %s"
+	defaultCommitMessageFormat = "Update %s in %s to %s"
 	defaultCheckInterval       = 5 * time.Minute
 )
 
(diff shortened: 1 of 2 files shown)


pkg/config/testdata/.pipe/image-watcher.yaml
```diff
@@ -5,8 +5,8 @@ spec:
     - image: gcr.io/pipecd/foo
       provider: my-gcr
       filePath: foo/deployment.yaml
-      field: spec.containers[0].image
+      field: $.spec.containers[0].image
```
nakabonne, line 8: This does nothing. Just for example.

pkg/app/piped/imagewatcher/watcher.go
```diff
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	defaultCommitMessageFormat = "Update image %s to %s defined at %s in %s"
+	defaultCommitMessageFormat = "Update %s in %s to %s"
```
khanhtc1202, line 38: nits, I think the previous message format is more readable/useable than the new one since I could fulfill data to that template without checking for sample use 😅 
how about `Update image: %s in path: %s to registry: %s` 🤔

nakabonne: @khanhtc1202 Thank you for your suggestion. To be honest I want to make this message as short and clear as possible. It actually has been already too long...
image

And this constant is used only for image watcher (not for piped operator). So how about just adding a comment on this line?

khanhtc1202: it would be nice 👍 and it's still too long btw 🤣

nakabonne: Yep, maybe I'll make a change on here again 😙

khanhtc1202: okay, I will approve this PR, thank you 🙏

PR #1337, opened 2020-12-25
Update fields of optional image watcher config

nakabonne:
The three changes are included:
- Rename repos to gitRepos (some people can confuse and mistake it for container repository)
- Make `checkInterval` global (most cases need only global settings. Let's add a field for each repo when really needed)
- Add some validations

The current structure is:
```yaml
kind: Piped
spec:
  imageWatcher:
    repos:
      - repoId: foo
        checkInterval: 10m
        commitMessage: Update image
        includes:
          - imagewatcher-dev.yaml
          - imagewatcher-stg.yaml
```

From now:
```yaml
kind: Piped
spec:
  imageWatcher:
    checkInterval: 10m
    gitRepos:
      - repoId: foo
        commitMessage: Update image
        includes:
          - imagewatcher-dev.yaml
          - imagewatcher-stg.yaml
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/config/piped.go
+++ b/pkg/config/piped.go
@@ -22,10 +22,6 @@ import (
 	"github.com/pipe-cd/pipe/pkg/model"
 )
 
-const (
-	defaultImageWatcherCheckInterval = Duration(5 * time.Minute)
-)
-
 var DefaultKubernetesCloudProvider = PipedCloudProvider{
 	Name:             "kubernetes-default",
 	Type:             model.CloudProviderKubernetes,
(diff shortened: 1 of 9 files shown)


pkg/config/piped.go
```diff
@@ -590,31 +586,34 @@ func (p *SealedSecretManagement) UnmarshalJSON(data []byte) error {
...
-// And it populates default value if not set.
+// Validate checks if:
+// - empty repo ids exist
+// - duplicated repository settings exist
 func (p *PipedImageWatcher) Validate() error {
-	repos := make(map[string]struct{}, len(p.Repos))
-	for i := 0; i < len(p.Repos); i++ {
-		if _, ok := repos[p.Repos[i].RepoID]; ok {
-			return fmt.Errorf("duplicated repo id (%s) found in the imageWatcher directive", p.Repos[i].RepoID)
+	repos := make(map[string]struct{}, len(p.GitRepos))
```
khanhtc1202, line 600: should we check for the `p.CheckInterval` value is set here?

PR #1367, opened 2021-01-07
Fix example to use quickSync instead of sync

nakabonne:

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice catch ⚾ 
/lgtm

PR #1390, opened 2021-01-11
Create determineCronJobHealth

### LINKED ISSUE #355: Check health state of CronJob resource
We can borrow around [here](https://github.com/argoproj/gitops-engine/tree/master/pkg/health).

https://github.com/pipe-cd/pipe/blob/285525ca8c9340d12b0a6dfcaf81ecad6e502ea6/pkg/app/piped/cloudprovider/kubernetes/state.go#L293-L296

sanposhiho:

Fixes #355

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/cloudprovider/kubernetes/state.go
+++ b/pkg/app/piped/cloudprovider/kubernetes/state.go
@@ -291,9 +291,9 @@ func determineReplicaSetHealth(obj *unstructured.Unstructured) (status model.Kub
 	return
 }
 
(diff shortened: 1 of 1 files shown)


sanposhiho:
I couldn't imagine a case where a cronjob is not healthy, so I made it to always return KubernetesResourceState_HEALTHY.

nakabonne:
Yep, we finely watch the Job's health. So I feel like it's enough to give back "HEALTHY" as well. Thanks 👍 
/lgtm

PR #1398, opened 2021-01-12
Create Role, RoleBinding, ClusterRole, ClusterRoleBinding health check

### LINKED ISSUE #1136: Check health state of Role resources
We can borrow around [here](https://github.com/argoproj/gitops-engine/tree/master/pkg/health).

https://github.com/pipe-cd/pipe/blob/c5ff4e52d789671274fc14138ee645f94ba2ffd5/pkg/app/piped/cloudprovider/kubernetes/state.go#L72-L75

We need to check these resources:

- [ ] Role
- [ ] RoleBinding
- [ ] ClusterRole
- [ ] ClusterRoleBinding

**Why is this needed**:

sanposhiho:
Create Role, RoleBinding, ClusterRole, ClusterRoleBinding health check

Fixes #1136

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/cloudprovider/kubernetes/resourcekey.go
+++ b/pkg/app/piped/cloudprovider/kubernetes/resourcekey.go
@@ -72,6 +72,10 @@ const (
 	KindService               = "Service"
 	KindIngress               = "Ingress"
 	KindServiceAccount        = "ServiceAccount"
+	KindRole                  = "Role"
+	KindRoleBinding           = "RoleBinding"
+	KindClusterRole           = "ClusterRole"
+	KindClusterRoleBinding    = "ClusterRoleBinding"
 
 	DefaultNamespace = "default"
 )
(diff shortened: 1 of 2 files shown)


sanposhiho:
I checked [the definitions of structures](
https://github.com/kubernetes/api/blob/18112a7b933b885640bf9fabf41bb497d9cef620/rbac/v1/types.go#L104) of Role, but I couldn't figure out how to specify the criteria for determining healthy.
I felt that as long as the role is applied successfully, it will not become unhealthy in the middle of the process, so I implemented functions in this PR so that it always returns healthy.
The same goes for Rolebinding, ClusterRole, and ClusterRoleBinding.

nakabonne:
@sanposhiho Thank you for your investigation! To be honest we've put off that check the whole time. As you mentioned, it looks like never getting unhealthy once applied. Let's deal with it later if we have to finely control for in the future 👍 
/lgtm

khanhtc1202:
thank you
/approve

PR #1399, opened 2021-01-12
Add configurations for Event watcher

nakabonne:
Most of the core logic is identical to ImageWatcher's one.

```release-note
NONE
```


pkg/config/event_watcher.go
```diff
@@ -0,0 +1,137 @@
...
+	"path/filepath"
+)
+
+type EventWatcherSpec struct {
+	Events []EventWatcherEvent `json:"events"`
+}
+
+// EventWatcherEvent defines which file will be replaced when the given event happened.
+type EventWatcherEvent struct {
+	// The event name. The combination of this name and labels must be unique within a project.
```
nghialv, line 30: Do we need this statement: "The combination of this name and labels must be unique within a project."
Because piped is only handling the events of its project.

pkg/config/event_watcher.go
```diff
@@ -0,0 +1,137 @@
...
+	Name string `json:"name"`
+	// List of places where will be replaced when the new event found.
```
nghialv, line 32: nit: "new event matches"

pkg/config/event_watcher.go
```diff
@@ -0,0 +1,137 @@
...
+		if _, ok := blackList[f.Name()]; !ok {
+			filtered = append(filtered, f)
+		}
+	}
+	return filtered, nil
+}
+
+func (s *EventWatcherSpec) Validate() error {
+	if len(s.Events) == 0 {
+		return ErrNotFound
```
nghialv, line 125: Events are required?

pkg/config/event_watcher.go
nakabonne: Sorry for the confusion, as you feel this shouldn't be here.

I just mean that "if you use the same name across repositories, they affect each other. So you have to make that combination unique if you want to explicitly distinguish them".

pkg/config/event_watcher.go
```diff
@@ -0,0 +1,137 @@
...
+
+// EventWatcherEvent defines which file will be replaced when the given event happened.
+type EventWatcherEvent struct {
+	// The event name. The combination of this name and labels must be unique within a project.
+	Name string `json:"name"`
+	// List of places where will be replaced when the new event found.
+	Replacements []EventWatcherReplacement `json:"replacements"`
+	// Additional attributes of event. This can make an event definition
+	// unique even if the one with the same name exists.
+	Labels map[string]string `json:"labels"`
```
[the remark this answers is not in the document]
nakabonne: I don't exactly get what you said. Soffy but could you show me some examples of that?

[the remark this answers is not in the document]
nghialv: I want to say that our configuration will have 2 groups:
- first group is about how to match the event. This contains "name" and "labels"
- second group is about how to handle when the event matches. This contains "replacements"

So I think we should change the order to group them for more readable.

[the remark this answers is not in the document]
nakabonne: That makes sense! I placed the `Label` field at the end because this is optional, but putting after `Name` is more readable in such case.

pkg/config/event_watcher.go
nakabonne: I thought so because it looks wierd that no events despite the `EventWatcher` config file exists.

But on second thought, I realized it's not a problem to give back an EventWatcherSpec with no Events 👍

nakabonne:
@nghialv Thank you for your polite review. Fixed them.

nghialv:
Nice.
/lgtm

khanhtc1202:
🚀 
/approve

PR #1424, opened 2021-01-13
Fix application sync status displaying

cakecatz:

Create a component to display application sync status and use it on application list/detail.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/app-sync-status.tsx
+++ b/pkg/app/web/src/components/app-sync-status.tsx
@@ -6,6 +6,7 @@ import {
   ApplicationSyncState,
   ApplicationSyncStatus,
 } from "../modules/applications";
+import { UI_TEXT_NOT_AVAILABLE_TEXT } from "../constants/ui-text";
 
 const useStyles = makeStyles((theme) => ({
   statusText: {
@@ -19,8 +20,6 @@ interface Props {
   className?: string;
   size?: "medium" | "large";
 }
-
-const NOT_AVAILABLE_TEXT = "N/A";
 
 export const AppSyncStatus: FC<Props> = ({
   syncState,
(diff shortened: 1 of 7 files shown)


pkg/app/web/src/components/app-sync-status.tsx
```diff
@@ -0,0 +1,63 @@
...
+}));
+
+interface Props {
+  syncState?: ApplicationSyncState;
+  deploying: boolean;
+  className?: string;
+  size?: "medium" | "large";
+}
+
+const NOT_AVAILABLE_TEXT = "N/A";
```
khanhtc1202, line 23: nit, similar to [pkg/app/web/src/components/application-list-item.tsx#L52](https://github.com/pipe-cd/pipe/blob/master/pkg/app/web/src/components/application-list-item.tsx#L52) should we move it to `constants/text.ts` :eyes:

pkg/app/web/src/components/application-detail.tsx
```diff
@@ -63,10 +61,12 @@ const useStyles = makeStyles((theme) => ({
     color: theme.palette.text.secondary,
     marginLeft: theme.spacing(1),
   },
-  syncStatusText: {
-    marginLeft: theme.spacing(0.5),
+  AppSyncState: {
```
khanhtc1202, line 64: nit, do we have conventions for CSS classes' naming, looks like others start with normal character (not capital) 👀

cakecatz: Nice catch!

pkg/app/web/src/components/app-sync-status.tsx
cakecatz: You're right, thanks😄

PR #1429, opened 2021-01-14
Add apistore for caching the latest events

nakabonne:
This PR adds a new `apistore` for Events according to @nghialv 's [suggestion](https://github.com/pipe-cd/pipe/pull/1410#issuecomment-759202852). It allows us to reduce communication number by caching the latest events.

This will be used by https://github.com/pipe-cd/pipe/pull/1411 at first.

```release-note
NONE
```


pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,154 @@
...
+			return nil
+		}
+	}
+}
+
+// sync fetches a list of events inside the range between own milestone and the current local time.
+// Only this function takes responsibility for updating the cache.
+func (s *store) sync(ctx context.Context) error {
+	// TODO: Use UTC and let control-plane convert it into the control-plane's Local time
+	// Unexpected behavior can be happened if Piped uses different timezone from the control-plane.
```
nakabonne, line 105: I will look into this concern when working on another PR.

pkg/model/event.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// EventDefinitionID builds a unique identifier based on the given name and labels.
+// It returns the exact same string as long as both are the same.
+func EventDefinitionID(name string, labels map[string]string) string {
```
nakabonne, line 25: I've been thinking about what we should call something whose uniqueness is consists of its Name and Labels. I just settled on calling it Event Definition.

An Event is like an Instance In Object-oriented world, and an event definition is like a Class. Each Event has its own ID and at the same time, each Event definition has its own ID. The event definition ID must be the exact same string as long as Name and Labels are the same.

nakabonne: I just figured out it would be nice if we build this when saving and keep it as a field of Event.

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,154 @@
...
+	resp, err := s.apiClient.GetLatestEvent(ctx, &pipedservice.GetLatestEventRequest{
+		Name:   name,
+		Labels: labels,
+	})
+	if err != nil {
+		s.logger.Error("failed to get the latest event", zap.Error(err))
+		return nil, false
+	}
+
+	// NOTE: Don't update the cache to prevent it from being overwritten by slightly older ones.
```
nghialv, line 152: Why we don't update the list?
The event has `CreatedAt` field, so we can use that to check whether the event should be updated or not.

And I think in the microservices list of our users, there are always some applications that have not been actively developed compared to others. Their events will be very sparse. So without saving into the list, piped will always make the call to control-plane to get the same one.

pkg/app/api/service/pipedservice/service.proto
```diff
@@ -362,9 +362,16 @@ message GetLatestEventResponse {
     pipe.model.Event event = 1 [(validate.rules).message.required = true];
 }
 
+enum ListEventsOrder {
+    NONE = 0;
+    FROM_OLDEST = 1;
+    FROM_NEWEST = 2;
+}
```
nghialv, line 369: How about `ListOrder` enum that contains `NONE`, `DESC`, `ASC`?

pkg/model/event.go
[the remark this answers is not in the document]
nakabonne: Actually it will be used by https://github.com/pipe-cd/pipe/issues/1406. It could be increased the need to build a unique identity made of Name and Labels, that's why I decided to make a helper function.

pkg/app/api/service/pipedservice/service.proto
nakabonne: That sounds good, that's more versatile. I initially thought about making it but I wasn't sure which should be general, CreatedAt or UpdatedAt, that's why I made a special enum for now. but what do you think?

pkg/model/event.go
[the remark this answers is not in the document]
nghialv: I see. Now we are having our cache/store for the event. So maybe we don't need to merge them for simplicity.

The user can merge by changing their configuration if needed.
We implicitly merge them and combine their changes into a single commit that may cause the behavior they were not expected.
(I am not sure that I understood the merging correctly.)

pkg/app/api/service/pipedservice/service.proto
nghialv: I think it is fine to make `ListOrder`. It can be used by other RPCs too.
Let's do it.

pkg/app/piped/apistore/eventstore/store.go
nakabonne: The `sync` function periodically updates the cache, and it works in another goroutine. What I'm most worried about is, while processing `GetLatest`, `sync` can update the cache. I mean:

```go
	// NOTE: Don't update the cache to prevent it from being overwritten by slightly older ones.
	eventMap = s.latestEventMap.Load().(map[string]*model.Event)
	eventMap[id] = resp.Event        // I'm most worried about the "sync()" function updates the cache at this point.
	s.latestEventMap.Store(eventMap) // Then the latest cache is overwritten by the old one here.
```

pkg/app/api/service/pipedservice/service.proto
nakabonne: Fine, I'll do so!

pkg/app/piped/apistore/eventstore/store.go
nakabonne: If Piped was run on the multi-core processor, I think it can happen, but let me know if I'm wrong.

pkg/model/event.go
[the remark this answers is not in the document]
nakabonne: >(I am not sure that I understood the merging correctly.)

Your understanding is correct. I attempted to merge it to combine their changes into a single commit.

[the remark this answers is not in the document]
nakabonne: Okay, I agree with all of them. So let's this function private and think it if needed in the future!

pkg/app/piped/apistore/eventstore/store.go
nghialv: The data race can be resolved by some ways:
- change from using atomic to using a mutex. (fetches the list from control-plane, lock, and update). I think it is fine enough for our case: 1 call/m and EventWatcher.
- use 2 separate maps, one for LIST API, one for GET API

nakabonne: Ah, not fancy 😄  The first one looks enough! Only two goroutines can access this and it rarely happens to wait for unlocking. Thank you for telling me!

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (s *store) Getter() Getter {
+	return s
+}
+
+func (s *store) GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool) {
+	eventMap := s.latestEventMap.Load().(map[string]*model.Event)
+	id := model.EventDefinitionID(name, labels)
+	event, ok := eventMap[id]
```
nghialv, line 137: One thing I want to ask about our event matching, in case we have an event 

```
name = foo
labels: { label-1: value-1, label-2: value-2}
```

That event will match the following case or not?

```
piVersion: pipecd.dev/v1beta1
kind: EventWatcher
spec:
  events:
    - name: foo
      labels:
           label-1: value-1
      replacements:
        - file: dev/app1/deployment.yaml
          yamlField: $.spec.template.spec.containers[0].image
```

nghialv: Or all labels must be specified to be matched.

nakabonne: I totally assume exact matches are needed, but on second thought, seems like that case should be treated as a matching case. Let me think, just a moment please...

pkg/app/piped/apistore/eventstore/store.go
nghialv: I think the timezone difference is not a problem, because we are using Unix time (the number of seconds elapsed since January 1, 1970 UTC.).
But the `clock drift` may cause the problem. But that is often rare.

Btw, maybe we can leave the `to` parameter as zero to use `time.Now` of the server.

nakabonne: >I think the timezone difference is not a problem, because we are using Unix time (the number of seconds elapsed since January 1, 1970 UTC.).

I put off thinking about it but after thinking, it's a basic thing and no problem as you mentiond.

nakabonne: >Btw, maybe we can leave the to parameter as zero to use time.Now of the server.

Cool, I'll try it.

nakabonne: >Btw, maybe we can leave the to parameter as zero to use time.Now of the server.

After thinking, it makes derivating the next's `from` tougher. As you mentioned, clock drifts is reraly happened. I guess it's enough to be as-is.

nakabonne: Not tougher, I came up with a better solution just now.

pkg/app/piped/apistore/eventstore/store.go
nakabonne: @nghialv I'm still debating but which one do you prefer?

nghialv: If possible, the event should be matched when the name is equal AND the event's labels contain the specified ones.
(Because they are labels and users will have more options.)

So I think we can change the map to a new data structure like `map[event-name]ListOfEventsHaveSameName`.

nghialv: > But one thing we have to consider is what happens when multiple events match the specified label list.

So in order to avoid this complexity and avoid adding the dynamic label by users (e.g. version number),
I think we should not allow subset matching.

Name and all labels must be matched.

nakabonne: My feeling is exactly the same as yours. Some people can be helped by subset matching, but too complicated to handle for us. Thank you for a great opportunity to re-think it and letting me know your cool thought!

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,180 @@
...
+
+	s.mu.Lock()
+	s.latestEventMap[id] = resp.Event
+	s.mu.Unlock()
+	return resp.Event, true
+}
+
+// eventDefinitionID builds a unique identifier based on the given name and labels.
+// It returns the exact same string as long as both are the same.
+func eventDefinitionID(name string, labels map[string]string) string {
```
nghialv, line 162: nit: How about renaming this to `makeEventKey`?

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+
+		case <-ctx.Done():
+			s.logger.Info("event store has been stopped")
+			return nil
+		}
+	}
+}
+
+// sync fetches a list of events newly created after its own milestone.
```
nghialv, line 106: nit: `sync fetches ... and updates ...`

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,180 @@
...
+		Order: pipedservice.ListOrder_ASC,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list events: %w", err)
+	}
+
+	// Make the cache up-to-date by traversing events sorted by oldest first.
+	var latestTime int64
+	s.mu.Lock()
+	for _, event := range resp.Events {
```
nghialv, line 119: `resp.Events` may contain multiple events for the same `key/definition-id`, so I think we should build their `key/definition-id` and filter duplicates before locking the mutex to update the list.

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,180 @@
...
+		id := eventDefinitionID(event.Name, event.Labels)
+		s.latestEventMap[id] = event
```
nghialv, line 121: We are doing asynchronous operations so this event can be older than the one stored in the map.
Let's check the CreatedAt before overriding.

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,180 @@
...
+		Name:   name,
+		Labels: labels,
+	})
+	if err != nil {
+		s.logger.Error("failed to get the latest event", zap.Error(err))
+		return nil, false
+	}
+
+	s.mu.Lock()
+	s.latestEventMap[id] = resp.Event
```
nghialv, line 155: Here, check CreatedAt too.

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,180 @@
...
+	syncInterval time.Duration
+	gracePeriod  time.Duration
+	logger       *zap.Logger
+
+	// Mark that it has handled all events that was created before this UNIX time.
+	milestone int64
+	mu        sync.RWMutex
+	// The key is supposed to be a string consists of name and labels.
+	// And the value is the address to the latest Event.
+	latestEventMap map[string]*model.Event
```
nghialv, line 61: nit: `events`

pkg/app/piped/apistore/eventstore/store.go
nakabonne: Good. That looks enough.

pkg/app/piped/apistore/eventstore/store.go
nakabonne: Good point. I initially attempted to do so, but I didn't. That's necessary to keep the time to lock mutex as short as possible. But don't we really have to care about that performance? I guess we should prioritize to keep simplicity than performance at this time, that's why I settled on just sorting by oldest first and traversing all events.

pkg/app/piped/apistore/eventstore/store.go
nakabonne: At the beginning of the loop, it is more likely to happen as you mentioned. But eventually, all Events on the cache will be up-to-date because the list of events is sorted by oldest first. I'd say it's enough but what do you think?

pkg/app/piped/apistore/eventstore/store.go
nakabonne: For this case, I totally agree with you 👍

pkg/app/piped/apistore/eventstore/store.go
nakabonne: you mean that should be `latestEventsMap` right?

pkg/app/piped/apistore/eventstore/store.go
nakabonne: Ah, now I could get what you're most worried about. Okay, let me fix them.

pkg/app/piped/apistore/eventstore/store.go
nghialv: Sorry for the unclear message. Just a nit to rename `latestEventMap`.
I think it can just be `latestEvents` or `events`.

nakabonne: Thank you. I feel the same way, the current one is too redundant. Let me rename it to latestEvents

pkg/app/piped/apistore/eventstore/store.go
nghialv: > But don't we really have to care about that performance?

Actually, I didn't want to say that we don't really care about the performance.
What I wanted to say before is that the complexity of the concurrency model is not worth the performance it brings.

But calculating the ID and filtering the duplicates before locking can be done easily. So it's worth it.

pkg/app/piped/apistore/eventstore/store.go
nghialv: Yes. The list is updated by `GetLatest` too.

pkg/app/piped/apistore/eventstore/store.go
nakabonne: You're quite right. Adding filtering is not complicated enough to worry about complexity.

nakabonne:
Thank you for your polite review. I've fixed all of them.

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,194 @@
...
+// and updates the cache of latest events.
+func (s *store) sync(ctx context.Context) error {
+	resp, err := s.apiClient.ListEvents(ctx, &pipedservice.ListEventsRequest{
+		From:  s.milestone,
+		Order: pipedservice.ListOrder_ASC,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list events: %w", err)
+	}
+
```
nghialv, line 116: nit: Return fast if no events.

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,194 @@
...
+	// Eliminate events that have duplicated key.
+	var latestTime int64
```
nghialv, line 118: `latestTime := resp.Events[len(resp.Events)-1].CreatedAt`

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -0,0 +1,194 @@
...
+		cached, ok := s.latestEvents[key]
+		if ok && cached.CreatedAt > event.CreatedAt {
+			continue
+		}
+		s.latestEvents[key] = event
+	}
+	s.mu.Unlock()
+
+	// Set the latest one within the result as the next time's "from".
+	s.milestone = latestTime + 1
```
nghialv, line 137: nit: `s.milestone := resp.Events[len(resp.Events)-1].CreatedAt + 1`

nghialv:
Nice. Thank you.
Looks good.

pkg/app/piped/apistore/eventstore/store.go
nakabonne: Right, the previous one could override the milestone to zero. I appreciate you.

nakabonne:
Fixed done.

nghialv:
Nice.
/lgtm

khanhtc1202:
:rocket:
/approve

PR #1435, opened 2021-01-14
Fix to use config's timeout on Prometheus

### LINKED ISSUE #1244: Prometheus Provider doesn't respect a timeout in an Analysis Template Configuration
**What happened**:

I set a very small timeout (`1µs`) in the `analysis-template.yaml`, but Prometheus query didn't time out.

https://pipecd.dev/docs/user-guide/configuration-reference/#analysis-template-configuration

It's because Prometheus Provider always uses `defaultTimeout`.

https://github.com/pipe-cd/pipe/blob/master/pkg/app/piped/analysisprovider/metrics/prometheus/prometheus.go#L31-L67

Prometheus Provider respects a timeout in the `analysis-template.yaml`, and correctly times out.

**How to reproduce it**:

Set a timeout like below, and set an `ANALYSIS` deployment  step in a pipeline configuration, then execute a deployment.

```
kind: AnalysisTemplate
spec:
  metrics:
    grpc_error_rate_percentage:
      interval: 1m
      provider: prometheus-dev
      expected:
        max: 10
      query: |
        100 - sum(
            rate(
                grpc_server_handled_total{
                  grpc_code!="OK",
                  kubernetes_namespace="{{ .K8s.Namespace }}",
                  kubernetes_pod_name=~"{{ .App.Name }}-[0-9a-zA-Z]+(-[0-9a-zA-Z]+)"
                }[{{ .Args.interval }}]
            )
        )
        /
        sum(
            rate(
                grpc_server_started_total{
                  kubernetes_namespace="{{ .K8s.Namespace }}",
                  kubernetes_pod_name=~"{{ .App.Name }}-[0-9a-zA-Z]+(-[0-9a-zA-Z]+)"
                }[{{ .Args.interval }}]
            )
        ) * 100
      timeout: 1µs
```

**Environment**:
- `piped` version: v0.9.0
- `control-plane` version: v0.9.0

sanposhiho:
Currently, Always used the default timeout value, so, changed to use the value in the configuration.

Fixes #1244

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/analysisprovider/metrics/factory.go
+++ b/pkg/app/piped/analysisprovider/metrics/factory.go
@@ -35,10 +35,10 @@ func NewFactory(logger *zap.Logger) *Factory {
 }
 
 // NewProvider generates an appropriate provider according to analysis provider config.
-func (f *Factory) NewProvider(providerCfg *config.PipedAnalysisProvider) (provider Provider, err error) {
+func (f *Factory) NewProvider(analysisTempCfg *config.TemplatableAnalysisMetrics, providerCfg *config.PipedAnalysisProvider) (provider Provider, err error) {
 	switch providerCfg.Type {
 	case model.AnalysisProviderPrometheus:
-		provider, err = prometheus.NewProvider(providerCfg.PrometheusConfig.Address, f.logger)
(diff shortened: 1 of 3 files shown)


nakabonne:
@sanposhiho Great work! Really helped a lot!
/approve

PR #1453, opened 2021-01-20
Use code splitting for web frontend

cakecatz:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/tsconfig.json
+++ b/pkg/app/web/tsconfig.json
@@ -5,7 +5,7 @@
     },
     "allowUnreachableCode": false,
     "baseUrl": ".",
-    "module": "CommonJS",
+    "module": "esnext",
     "moduleResolution": "node",
     "noImplicitReturns": true,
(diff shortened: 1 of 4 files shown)


pkg/app/web/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     },
     "allowUnreachableCode": false,
     "baseUrl": ".",
-    "module": "CommonJS",
+    "module": "esnext",
```
khanhtc1202, line 8: nits, is it safe to use `esnext` over a specific version of js 👀

cakecatz: @khanhtc1202 Yes, it's safe because we use `webpack` after compiling with `typescript`.
If it is broken, the webpack build should fail.

khanhtc1202: got it, thanks for your explaining 🙆‍♀️

khanhtc1202:
🚀 
/approve

PR #1488, opened 2021-01-26
Add tooltip to show full id at the piped list row

cakecatz:

https://user-images.githubusercontent.com/6136383/105808649-76cc8400-5feb-11eb-8225-629894376c5d.mp4

Fixes #

```release-note
The full ID of piped will be shown while hovering over the name
```


khanhtc1202:
Maybe it's out of scope, but should we add `click to copy` function 👀  the long UUID is quite hard to confirm if we only just look at it 👀 
( never mind if we still merge this PR and make another for that function if we're going to make it 😄 )

khanhtc1202:
🚀 
/approve

PR #1556, opened 2021-02-05
Use pipecd-base as a base image of pipecd

nakabonne:

Ref https://github.com/pipe-cd/pipe/issues/1471

```release-note
NONE
```


khanhtc1202:
Nice
/lgtm

PR #1586, opened 2021-02-15
Add pipeline mock data for the web

cakecatz:

Fixes #

```release-note
NONE
```


pkg/app/web/src/components/pipeline/index.test.tsx
```diff
@@ -21,8 +21,8 @@ it("should dispatch updateActiveState when first rendering", () => {
     {
       type: updateActiveStage.type,
       payload: {
-        name: dummyDeployment.stagesList[0].name,
-        stageId: dummyDeployment.stagesList[0].id,
+        name: dummyDeployment.stagesList[1].name,
+        stageId: dummyDeployment.stagesList[1].id,
```
khanhtc1202, line 25: just to confirm, is that neccessary to make this change 👀

cakecatz: @khanhtc1202 
This is a test to make sure that the running stage is automatically selected when the page is opened.
I had to change the stage to refer to because I changed the pipeline mock data like this:

before: `[running] ` 
after: `[success, running, not started]`

khanhtc1202: got it, thank you 🙇

khanhtc1202:
🚀 
/approve

PR #1605, opened 2021-02-19
Rollback Lambda application configuration in case traffic change has not been committed

### LINKED ISSUE #1600: Rollback Lambda application not working in case deployment failed before traffic routing stage
**What happened**:

In case of deployment failed before the traffic routing stage, the current rollback implementation will always skip running since we can not get the previous traffic routing config from metadata store.
ref: https://github.com/pipe-cd/pipe/blob/master/pkg/app/piped/executor/lambda/rollback.go#L93-L99

**How to reproduce it**:

**Environment**:
- `piped` version: 0.9.5
- `control-plane` version:
- Others:

khanhtc1202:

Fixes #1600 

```release-note
None
```

### DIFF
--- a/pkg/app/piped/executor/lambda/rollback.go
+++ b/pkg/app/piped/executor/lambda/rollback.go
@@ -90,11 +90,19 @@ func rollback(ctx context.Context, in *executor.Input, cloudProviderName string,
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


nakabonne:
🚀 
/approve

PR #1612, opened 2021-02-22
[RFC] SQL database as Control-Plane datastore

khanhtc1202:

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/rfcs/0003-sql-datastore.md
+++ b/docs/rfcs/0003-sql-datastore.md
@@ -96,7 +96,7 @@ From MySQL docs
 
 PostgreSQL has UUID type to store this kind of data and support it as primary key as well.
 
-MySQL: 👍 PostgreSQL: 👍
+MySQL: 👍 PostgreSQL: 👍👍
 
 2. Create indexes for JSON attributes without affecting the schemaless advantage
 
--- a/docs/rfcs/0003-sql-datastore.md
+++ b/docs/rfcs/0003-sql-datastore.md
@@ -11,7 +11,7 @@ The background of this proposal is we are going to add complex queries and more 
 
 # Detailed design
 
-As mention is the motivation section, the root cause of this change is: we're going to add complex queries with those indexes to the datastore of PipeCD control-plane. The use-case of those queries requires:
+As mention in the motivation section, the root cause of this change is: we're going to add complex queries with those indexes to the datastore of PipeCD control-plane. The use-case of those queries requires:
 - easy way to pagination query result
 - easy way to ordering query result
 - easy way to add indexes for future adding queries
@@ -20,7 +20,7 @@ Besides, to keep `simplicity on installing` characteristic of PipeCD, the chosen
 - easy to be installed on cloud provider environment (GCP, AWS, etc)
 - can use the updated version easily
 
-We're using NoSQL for now due to its schemaless characteristic is good for the deployment process, so we focus on NoSQL support features on the chosen SQL database.
+We're using NoSQL for now because of its schemaless characteristic is good for the deployment process, so we focus on NoSQL support features on the chosen SQL database. In order to accomplish the schema flexibility on SQL database, we use a JSON field to store our model data containing fluctuant structure fields, just some stable fields like ID, CreatedAt, UpdatedAt will be explicitly declared while defining the table schema.
 
 The sample table creates commands for the `project` model and `application` model (other models than project model is as same as the application) as below.
 
@@ -43,7 +43,7 @@ CREATE TABLE applications (
 );
 
 # MySQL
-CREATE TABLE `projects` (
+CREATE TABLE projects (
   id BINARY(16) PRIMARY KEY,
   data JSON NOT NULL,
   disabled BOOL NOT NULL,
@@ -51,7 +51,7 @@ CREATE TABLE `projects` (
   updated_at INT(11) NOT NULL
 ) ENGINE=InnoDB;
 
-CREATE TABLE `applications` (
+CREATE TABLE applications (
   id BINARY(16) PRIMARY KEY,
   project_id BINARY(16) NOT NULL,
   data JSON NOT NULL,
@@ -61,12 +61,9 @@ CREATE TABLE `applications` (
 ) ENGINE=InnoDB;
 ```
 
-For indexing issue:
-- PostgreSQL indexing for the attributes of Jsonb data type is as follow: [Json indexing](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING)
-- MySQL also support indexing for attributes of Json data type as follow: [Json indexing by secondary indexes](https://dev.mysql.com/doc/refman/8.0/en/create-table-secondary-indexes.html)
-
-(note: both Jsonb and Json point to the type of storing JSON data in those databases. The key difference is that Json data is stored as an exact copy of the JSON input text, whereas Jsonb stores data in a decomposed binary form. Jsonb seems a better choice since it supports indexing with significant performance increases)
-
+Note:
+- Both Jsonb and Json point to the type of storing JSON data in those databases. The key difference is that Json data is stored as an exact copy of the JSON input text, whereas Jsonb stores data in a decomposed binary form. Jsonb seems a better choice since it supports indexing with significant performance increases.
+- Save those data redundantly (both as table columns and as attributes of the JSON data) to reduce model object initialization costs.
 # Alternatives
 
 Currently, we consider between MySQL and PostgreSQL for those support for NoSQL features, PostgreSQL has a longer time in this field. Some considering factors:
@@ -77,11 +74,11 @@ Currently, we consider between MySQL and PostgreSQL for those support for NoSQL 
 
 ## Ability to index a specific attribute of the JSON field
 
-For `MySQL 8.0`, we have 2 points which have to be considered.
+We have 2 points which have to be considered.
 
-1. UUID data for `id` and `project_id` fields
+1. UUID as primary keys should not affect queries performance
 
-Since we're using `application side UUID generate` pattern, (which mean to MySQL, those ids are true random UUID) store those ids under `VARCHAR(32)` data type is costly for both read and write operation due to the indexes does not work, using `UUID_TO_BIN` & `BIN_TO_UUID` with `swap_flag` and store data under `BINARY(16)` would help.
+While MySQL does not have UUID data type, since we're using `application side UUID generate` pattern, (which mean to MySQL, those ids are true random UUID) store those ids under `VARCHAR(32)` data type is costly for both read and write operation due to the indexes does not work, using `UUID_TO_BIN` & `BIN_TO_UUID` with `swap_flag` and store data under `BINARY(16)` would help.
 
 From MySQL docs
 ```
@@ -97,12 +94,16 @@ From MySQL docs
 
 ```
 
-2. Create indexes for JSON attribute without affecting the schemaless advantage
+PostgreSQL has UUID type to store this kind of data and support it as primary key as well.
 
-We could use `CREATE INDEX` to create secondary indexes on generated columns which stored JSON attributes as an indirect way to indexes JSON attributes.
+MySQL: 👍 PostgreSQL: 👍
+
+2. Create indexes for JSON attributes without affecting the schemaless advantage
+
+For MySQL, we could use `CREATE INDEX` to create secondary indexes on generated columns which stored JSON attributes as an indirect way to indexes JSON attributes.
 In order to keep the advantage of the schemaless pattern, we will use `virtual generated columns` instead of `stored generated columns` (which will physically store along with other columns of table). __The virtual generated columns wouldn't be generated on READ as long as we keep all generated columns as part of some secondary indexes__, which reduce the cost of recomputing from READ operation (note that computing virtual columns value cost on WRITE remains).
 
-Sample secondary indexes creation as follow
+Sample secondary indexes creation commands as follow
 
 ```sql
 mysql> CREATE INDEX idx ON applications ( ( CAST( data->>"$.name" AS CHAR(10) ) ) );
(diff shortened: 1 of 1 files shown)


nghialv:
@khanhtc1202 Nice. I think we could start adding the thoughts about the implementation details.
For examples,
- what indexes will be needed
- how do we create the index
- what kind of queries do we need to fit our datastore interface
- what fully-managed SQL service we should recommend to user for each cloud provider

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,61 @@
...
+	created_at TIMESTAMP NOT NULL,
+	updated_at TIMESTAMP NOT NULL,
+);
+
+CREATE TABLE application (
+	id VARCHAR ( 32 ) PRIMARY KEY,
+	project_id VARCHAR ( 32 ) NOT NULL,
+	data JSON NOT NULL,
+	created_at TIMESTAMP NOT NULL,
+	updated_at TIMESTAMP NOT NULL,
```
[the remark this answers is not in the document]
khanhtc1202: Sure 👍 Thanks for your comment 🙏 
This is the draft for postgres tables
```sql
CREATE TABLE projects (
	id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
	data JSONB NOT NULL,
	disabled BOOL NOT NULL,
	created_at BIGINT NOT NULL,
	updated_at BIGINT NOT NULL
);
CREATE TABLE applications (
	id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
	project_id UUID NOT NULL,
	data JSONB NOT NULL,
	disabled BOOL NOT NULL,
	created_at BIGINT NOT NULL,
	updated_at BIGINT NOT NULL
);
```
How do you think? 🙏

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,62 @@
...
+
+For indexing issue:
+- PostgreSQL indexing for the attributes of Jsonb data type is as follow: [Json indexing](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING)
+- MySQL also support indexing for attributes of Json data type as follow: [Json indexing by secondary indexes](https://dev.mysql.com/doc/refman/8.0/en/create-table-secondary-indexes.html)
+
+(note: both Jsonb and Json point to the type of storing JSON data in those databases. The key difference is that Json data is stored as an exact copy of the JSON input text, whereas Jsonb stores data in a decomposed binary form. Jsonb seems a better choice since it supports indexing with significant performance increases)
+
+# Alternatives
+
+Currently, we consider between MySQL and PostgreSQL for those support for NoSQL features, PostgreSQL has a longer time in this field.
```
nghialv, line 52: (Some thoughts about what should be chosen: MySQL or PostgresSQL)
Based on our data size and the way of using the database (just a simple feature) I think performance and feature-set are not big factors used to consider when choosing.
Better than those, the popularity would be a good factor. The popularity means which is more familiar to our users, supported well by many cloud providers, has a good library to use...

Btw, let's make this decision after having a clearer look at the implementation detail.
https://github.com/pipe-cd/pipe/pull/1612#issuecomment-784726947

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Unresolved questions
+
+Currently, we have 2 points which need to investigate more
+1. The support of each cloud providers (GCP and AWS for now) for the chosen SQL database, since it's not the native service of those cloud providers (not as firestore of GCP and dynamodb of AWS for instance).
+- AWS supported MySQL 8 recently (the latest version).
+- GCP supported MySQL 8 for now (the latest version).
+- AWS is supporting PostgreSQL 12 for now (the latest version is 13).
+- GCP supported PostgreSQL 13 for now (the latest version).
```
nghialv, line 61: PostgreSQL on Azure https://azure.microsoft.com/en-us/services/postgresql/

docs/rfcs/0003-sql-datastore.md
nghialv: We majorly depend on the JSON field, so the ability of the JSON field supported by the database would be a good factor too. (The ability to index a specific attribute of the JSON field, the supported operator on that field, and the performance of operating on that field.)

docs/rfcs/0003-sql-datastore.md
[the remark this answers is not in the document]
nghialv: This looks good. Just one nit is that I think we don't need to configure the default value for the ID field.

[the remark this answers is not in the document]
khanhtc1202: Nice catch, thank you 🙏

docs/rfcs/0003-sql-datastore.md
nakabonne: So interesting. I'm just curious about whether you guys are looking into supporting both in the end if both have enough abilities.

docs/rfcs/0003-sql-datastore.md
khanhtc1202: https://docs.microsoft.com/en-us/azure/postgresql/concepts-supported-versions
It's version 11 currently, how disappointing 😞

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,199 @@
...
+
+This RFC proposes adding SQL database as datastore for PipeCD Control-plane. Currently, we're considering between PostgreSQL and MySQL for those NoSQL support features.
+
+# Motivation
+
+The background of this proposal is we are going to add complex queries and more indexes to the control-plane datastore for some new features of PipeCD, the current in-use NoSQL datastore is blocking those requirements.
+
+# Detailed design
+
+As mention is the motivation section, the root cause of this change is: we're going to add complex queries with those indexes to the datastore of PipeCD control-plane. The use-case of those queries requires:
```
nghialv, line 14: typo: "is the" -> "in the"

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,199 @@
...
+- easy way to pagination query result
+- easy way to ordering query result
+- easy way to add indexes for future adding queries
+
+Besides, to keep `simplicity on installing` characteristic of PipeCD, the chosen must be:
+- easy to be installed on cloud provider environment (GCP, AWS, etc)
+- can use the updated version easily
+
+We're using NoSQL for now due to its schemaless characteristic is good for the deployment process, so we focus on NoSQL support features on the chosen SQL database.
```
nghialv, line 23: nit: "due to" -> "because"

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,199 @@
...
+  data JSON NOT NULL,
+  disabled BOOL NOT NULL,
+  created_at INT(11) NOT NULL,
+  updated_at INT(11) NOT NULL
+) ENGINE=InnoDB;
+```
+
+For indexing issue:
+- PostgreSQL indexing for the attributes of Jsonb data type is as follow: [Json indexing](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING)
+- MySQL also support indexing for attributes of Json data type as follow: [Json indexing by secondary indexes](https://dev.mysql.com/doc/refman/8.0/en/create-table-secondary-indexes.html)
```
nghialv, line 66: "support" -> "supports"

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,199 @@
...
+- easy way to pagination query result
+- easy way to ordering query result
+- easy way to add indexes for future adding queries
+
+Besides, to keep `simplicity on installing` characteristic of PipeCD, the chosen must be:
+- easy to be installed on cloud provider environment (GCP, AWS, etc)
+- can use the updated version easily
+
+We're using NoSQL for now due to its schemaless characteristic is good for the deployment process, so we focus on NoSQL support features on the chosen SQL database.
+
```
nghialv, line 24: How about adding something like this to emphasize the method we use?

`In order to accomplish the schema flexibility on SQL database, we use a JSON field to store our model data containing fluctuant structure fields, just some stable fields like ID, CreatedAt, UpdatedAt will be explicitly declared while defining the table schema.`

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,199 @@
...
+(note: both Jsonb and Json point to the type of storing JSON data in those databases. The key difference is that Json data is stored as an exact copy of the JSON input text, whereas Jsonb stores data in a decomposed binary form. Jsonb seems a better choice since it supports indexing with significant performance increases)
+
+# Alternatives
+
+Currently, we consider between MySQL and PostgreSQL for those support for NoSQL features, PostgreSQL has a longer time in this field. Some considering factors:
+- ability to index a specific attribute of the JSON field
+- supported operators on specific JSON field
+- performance of query operation on indexed JSON field
+- able to keep the advantage of schemaless pattern
```
nghialv, line 76: Another question is that will we create multiple indexes or a single index combined multiple columns?

nghialv:
@khanhtc1202 This is a really nice RFC. Great!

docs/rfcs/0003-sql-datastore.md
khanhtc1202: Nice, I will use your suggestion 👍

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,280 @@
...
+
+```sql
+postgres=> CREATE INDEX idx ON test.applications ((data->>'name'));
+```
+
+ref: https://www.postgresql.org/docs/13/datatype-json.html
+
+Since both MySQL and PostgreSQL have multi-columns indexes, it's okay if we want to add index to multi attributes of JSON data.
+
+MySQL: 👍 PostgreSQL: 👍
```
nghialv, line 130: Maybe PostgreSQL should have 2 👍  (👍 👍 ) because of its simplicity.

khanhtc1202: then it should be in this `Community supports` section 😄 
https://github.com/pipe-cd/pipe/pull/1612/files#diff-61adf68d03c8a5d48c38d1ccad3f1f95bf0b6948f603525cebb5cb5fdcf78e1bR276

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,281 @@
...
+
+# Unresolved questions
+
+Currently, we have 2 points which need to investigate more
+## The support of each cloud providers
+
+With GCP and AWS (for now), since it's not the native service of those cloud providers, we have to use them fully-managed SQL services (SQL for GCP and RDS for AWS).\
+Besides, both AWS and GCP support the latest versions of PostgreSQL(v13) and MySQL(v8), but other cloud providers such as Azure only support up-to-date MySQL(v8) and a little behind PostgreSQL(v11).
+
+MySQL: 👍 PostgreSQL: 👎
```
nghialv, line 275: I think we should unify the way of using those emojis.
(👎  means bad, 👍  means good, 👍 👍  means very good.)

Maybe the latest PostgreSQL will be supported soon, so instead of 👎 , it should be 👍  for PostgreSQL and 👍 👍  for MySQL.

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,281 @@
...
+## Community supports
+
+In case `PostgreSQL` is chosen, there is no official driver for golang currently, we have a list of candidates: [pgx](https://github.com/jackc/pgx), [go-pg/pg](https://github.com/go-pg/pg).\
+`MySQL` is better in that field since it has a wider range of users/services and also has an official golang driver.
+
+MySQL: 👍 PostgreSQL: 👎
```
nghialv, line 281: MySQL: 👍 👍 , PostgreSQL: 👍

docs/rfcs/0003-sql-datastore.md
khanhtc1202: I see 🙏 , updated by [`7c0b806` (#1612)](https://github.com/pipe-cd/pipe/pull/1612/commits/7c0b806b8971e8cb70862192d85ee479b7bd4d34)

docs/rfcs/0003-sql-datastore.md
```diff
@@ -0,0 +1,281 @@
...
+    time-low and time-high parts (the first and third groups of
+    hexadecimal digits, respectively) are swapped. This moves the more
+    rapidly varying part to the right and can improve indexing
+    efficiency if the result is stored in an indexed column.
+
+```
+
+PostgreSQL has UUID type to store this kind of data and support it as primary key as well.
+
+MySQL: 👍 PostgreSQL: 👍
```
nghialv, line 99: I vote for PostgreSQL to have 👍 👍  because of its native UUID type.

khanhtc1202: sure 👍 will do 🙆‍♂️

nghialv:
@khanhtc1202 Thank you for a great RFC. 💯 
/lgtm

IMO, both MySQL and PostgreSQL are good, both of them allow us to accomplish the schema flexibility on SQL database.
(Although I prefer PostgreSQL personally, I think many of our users will find MySQL more familiar.)

So I give my vote for supporting MySQL at this time.

@khanhtc1202 @nakabonne Please leave your votes and merge this PR.
@khanhtc1202  After we decided what to use, please create a batch of issues for the implementation tasks. It would be nice if we can break it into multiple small issues. (e.g. add client lib, add ops to ensure index, implement datastore package...)

nghialv:
Just a note: this could help in the future when we write a blog to guide our user to deploy a stateful MySQL in Kubernetes.
https://kubernetes.io/docs/tasks/run-application/run-single-instance-stateful-application/

khanhtc1202:
> @khanhtc1202 @nakabonne Please leave your votes and merge this PR.

Thanks for your vote! IMO, since PipeCD is an OSS, I think using well-supported and accessible dependencies from the community will benefit in the long run. So, my vote will go for MySQL at this time, we may consider supporting other RDSs in the future.

> @khanhtc1202 After we decided what to use, please create a batch of issues for the implementation tasks. It would be nice if we can break it into multiple small issues. (e.g. add client lib, add ops to ensure index, implement datastore package...)

Sure, will do it 🙏

nakabonne:
>the popularity would be a good factor. The popularity means which is more familiar to our users, supported well by many cloud providers, has a good library to use...

As @nghialv mentioned, it would be appropriate to pay attention to the popularity. In that sense, I feel like MySQL is suitable. (It's just an opinion of me who is unfamiliar with RDB products.)

PR #1621, opened 2021-02-25
Add support for building web assets it has API mocking

cakecatz:

Add environemnet variable `ENABLE_MOCK` for building, it will output web assets that has API Mocking. 
It use like this.
```
bazelisk build :public_files --action_env=ENABLE_MOCK=true
```

The API mock does not work by itself.
It needs `mockServiceWorker.js` that generated by `msw`.
We need to put that file in the same location as `index.html` so that it can be accessed from the browser.

Fixes #

```release-note
NONE
```


khanhtc1202:
🆒 
/approve

PR #1659, opened 2021-03-07
Fix wrong error message and docs of the func

sanposhiho:
~~Support multiple field on event watcher
Currently, the implementation does not expect yamlprocessor.GetValue to return []interface{}.
So, when the user register the event yamlField as `$.bar[*].hoge` and yamlprocessor.GetValue return slice, `convertStr` return error.~~

- Fix wrong error message
- remove the statement that yamlprocessor.GetValue supported []interface{}

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/eventwatcher/eventwatcher.go
+++ b/pkg/app/piped/eventwatcher/eventwatcher.go
@@ -265,7 +265,7 @@ func modifyYAML(path, field, newValue string) ([]byte, bool, error) {
 	}
 	value, err := convertStr(v)
 	if err != nil {
-		return nil, false, fmt.Errorf("a value of unknown type is defined at %s in %s: %w", err, field, path)
+		return nil, false, fmt.Errorf("a value of unknown type is defined at %s in %s: %w", field, path, err)
 	}
 	if newValue == value {
 		// Already up-to-date.
--- a/pkg/yamlprocessor/yamlprocessor.go
+++ b/pkg/yamlprocessor/yamlprocessor.go
@@ -25,7 +25,7 @@ import (
 )
 
 // GetValue gives back the value placed at a given path. The type of
-// returned value can be string, int64, uint64, float64, bool, and []interface{}.
+// returned value can be string, int64, uint64, float64 and bool.
 //
 // The path requires to start with "$" which represents the root element.
 // Available operators are:
@@ -33,7 +33,6 @@ import (
 // .     : child operator
 // ..    : recursive descent
 // [num] : object/element of array by number
-// [*]   : all objects/elements for array.
 //
 // e.g. "$.foo.bar[0].baz"
 func GetValue(yml []byte, path string) (interface{}, error) {
--- a/pkg/yamlprocessor/yamlprocessor_test.go
+++ b/pkg/yamlprocessor/yamlprocessor_test.go
@@ -112,16 +112,6 @@ foo:
 			want:    []interface{}{map[string]interface{}{"bar": uint64(1)}, map[string]interface{}{"baz": uint64(2)}},
 			wantErr: false,
 		},
-		{
-			name: "given an entire array path with wildcard",
-			yml: `
-foo:
-- bar: 1
-- baz: 2`,
-			path:    "$.foo[*]",
-			want:    []interface{}{map[string]interface{}{"bar": uint64(1)}, map[string]interface{}{"baz": uint64(2)}},
-			wantErr: false,
-		},
 	}
 	for _, tc := range testcases {
 		t.Run(tc.name, func(t *testing.T) {


pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -277,26 +279,45 @@ func modifyYAML(path, field, newValue string) ([]byte, bool, error) {
...
-		out = strconv.FormatBool(v)
+// isUpToDate check if all targets is up-to-date or not
+func isUpToDate(targets []string, newVal string) bool {
+	for _, t := range targets {
+		if t != newVal {
+			return false
+		}
+	}
+	return true
+}
```
nakabonne, line 294: it depends on whether we allow sequences as a `newVal` or not. If we support sequences, we need to parse `newVal` and compare each element precisely.
In the case that we allow only scalar value as a `newVal`, will be a kind of like:

```go
func isUpToDate(targets []string, newVal string) bool {
	if len(targets) != 1 {
		return false
	}
	return targets[0] == newVal
}
```

nakabonne (commented):
@sanposhiho Thank you! No wonder you feel a little bit odd. It confuses you because it doesn't handle the case of []inerface{} even though it's showing the comment `// The type of returned value can be string, int64, uint64, float64, bool, and []interface{}.`.

I'm sorry that I wasn't leaving any comment but honestly I'm still debating if we should support Sequences or not, and was putting it off for a while.
- As commented, it gets a bit more complicated if we support sequences as new values as well.
- In Most cases, it's enough to replace Scalars, for instance:
  -  ```yaml
      foo:
        - 1
        - bar
      ```
  - Even in such a case, we just need to replace one by one. `foo[0] => 2` `foo[1] => baz`

In that sense, I'm feeling we just have to add support for it if we get a couple of requests.

Therefore, I'm guessing what we have to do in this patch is just adding a comment which shows why it doesn't handle Sequences, on the `convertStr` function. (Also, your nice catch should be kept as-is.)

I'm opening to any opinions. What do you think?

sanposhiho:
@nakabonne Thanks for checking PR.

> it gets a bit more complicated if we support sequences as new values as well.

Sorry, I cannot understand this comment well.
Which specific parts of the implements will be complex?

nakabonne:
@sanposhiho For sure. On second thought, it does not get that so complex! All we have to do is parsing the new value and compare it with the current one.

Let's say we have a sequence value kind of like `[1, "hello", false]` as a new value. We have to parse it as below and compare it with the current one to check if the value is up-to-date or not.

```go
[]interface {}{
  0x0000000000000001,
  "hello",
  false,
}
```

That is, what we end up with is the `isUpToDate()` function that does:
- parse the given new value
- compare the current one and the given new value even if the value cannot be represented as a simple []string{} (we have to consider the nested sequences because supporting sequences means dealing with something like below:
  ```yaml
  foo: 
  - 2
  - 
    bar: 
      - 
        baz: 
          - a
          - b
          - c
      - Hello
      - false
  ```

sanposhiho:
I have a question.
Do we need to check if it's up-to-date with new value(s)? Isn't it enough to compare newYml and yml like below?

```
func modifyYAML(path, field, newValue string) ([]byte, bool, error) {
	yml, err := ioutil.ReadFile(path)
	if err != nil {
		return nil, false, fmt.Errorf("failed to read file: %w", err)
	}

	// Modify the local file and put it into the change list.
	newYml, err := yamlprocessor.ReplaceValue(yml, field, newValue)
	if err != nil {
		return nil, false, fmt.Errorf("failed to replace value at %s with %s: %w", field, newValue, err)
	}
	return newYml, yml == newYml, nil
}
```

nakabonne:
You're exactly right. We should not rewrite the user-defined YAML file as much as possible, so I was initially worried about it treats the ones that are different format but with the same meaning as different ones. For instance, I was fear Event watcher modifies it kind of like:

```diff
+ foo: "bar"
- foo: bar
```

```diff
+ foo:
+   - 1
+   - 2
- foo: [1, 2]
```

But who cares. Fundamentally, we just need to care about it only at first. Let's implement to do so for now and careful test, then apply it!

sanposhiho:
@nakabonne 
change it to directly compare bytes directly to see if they are up-to-date.
As you expected, there was a case where the yaml value was the same, but the format was changed.([this test case](https://github.com/pipe-cd/pipe/pull/1659/files#diff-6426f2ee7888d7206ce67b88301e9c6c41e6f6763ec2f063938f2fffa23e8600R80))
What do you think?

nakabonne:
@sanposhiho As mentioned earlier, I was initially worried about ones that are different format (e.g. "foo" and foo) with the same meaning are handled as different ones. I little poked around it and I confirmed that the formats are definitely holding as a [Token](https://github.com/goccy/go-yaml/blob/8a300506fb3fc075a6b31b106c69bff066eab28c/token/token.go#L677-L686) provided by well-tested `goccy/go-yaml` package for the basic case. So for most cases, we don't have to worry about it thought...

>As you expected, there is a case where the yaml value was the same, but the format was changed.(this test case)

Way to pick that up! I could repro it and actually we can run into this issue regardless of this PR. However, the thing I feared most has occurred. Piped does keep committing endlessly.

image

I wonder why but Piped recognizes their format as the different ones even though all change contents are exactly the same as:
image

nakabonne:
There you go. It turns out to become such unexpected behavior thanks to your implementation. Even if we can deal with this `foo[*].bar` case, we're like to be unable to deal with all cases with the simple comparison of the YAML bytes. Hence, let's stop comparing the YAML bytes.

nakabonne:
@sanposhiho A big thank you goes out to your efforts! Let's loop back to our initial aim. I would say we shouldn't allow the sequences values as long as it can't replace a kind of like `foo[*].bar` correctly. To support it, we have to:
- Send a patch to [goccy/go-yaml](https://github.com/goccy/go-yaml) so that it replaces even in the case of `foo[*].bar`
- After merging and releasing it, re-start our discussion (we have to consider the case of `foo[*]` and whether we support Sequences as new values)

To avoid being a long-lived branch, you can just put the TODO comment and include [this fix](https://github.com/pipe-cd/pipe/pull/1659#discussion_r589124320) in this PR for now.

sanposhiho:
> To avoid being a long-lived branch, you can just put the TODO comment and include this fix in this PR for now.

Okey, fixed.

---

BTW, I have another idea.
You are talking about replacing multiple fields with multiple values, right?

I think it is also a simple and good idea to only support rewriting multiple fields with the single value at the same time.
As you said before, if you want to rewrite multiple values, you can just rewrite them one by one.
The good thing is that this way can be implemented with the current goccy/go-yaml.

If you think it's also good, keep it in mind as one of the ideas.
(As you said, continue discussion on this PR is not good. So, let's keep this PR itself to this error fix)

nakabonne:
Thank you. Let us merge it for now.
/approve

PR #1660, opened 2021-03-07
Fix wrong argument passed on error builder

khanhtc1202:

Fixes #

```release-note
NONE
```


nakabonne:
@khanhtc1202 Thank you for reading the source code around Eventwatcher and fixing it!

Sorry but seems like that will be fixed on https://github.com/pipe-cd/pipe/pull/1659 which opened earlier than this PR. To prevent conflict, would you mind prioritize sanposhiho's one and close this?

khanhtc1202:
> Sorry but seems like that will be fixed on #1659 which opened earlier than this PR. To prevent conflict, would you mind prioritize sanposhiho's one if we settled on merging that PR? If that rejected then let me approve your patch!

Sorry, I didn't notice #1659 🙏 In that case, please go ahead with that PR, I will close this one 👍

PR #1693, opened 2021-03-12
Refactor settings table styling

cakecatz:

Made other tables look the same as Piped tables.

## Before

image
image

## After

image
image

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/utils/is-nullish.ts
+++ b/pkg/app/web/src/utils/is-nullish.ts
@@ -1,0 +1,7 @@
+export function isNullish(obj: unknown): boolean {
+  if (obj === undefined || obj === null) {
+    return true;
+  }
+
+  return false;
+}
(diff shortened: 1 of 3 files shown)


pkg/app/web/src/utils/is-nullish.ts
```diff
@@ -0,0 +1,7 @@
+export function isNullish(obj: unknown): boolean {
```
khanhtc1202, line 1: looks like this function is not be used anywhere else 👀 should we remove it?
otherwise, I think `??` is good enough, no need to defined check `isNullish` ourselves. What do you think? 🤔 
ref https://mariusschulz.com/blog/nullish-coalescing-the-operator-in-typescript#providing-fallback-values-with-the-operator

cakecatz: Ah, this file is not for this PR.
I'll remove this.
Thank you!

khanhtc1202:
🚀 🚀 🚀 
/approve

PR #1730, opened 2021-03-16
Show application counts on the application list

cakecatz:

image

Ref #1436

```release-note
Show application counts on the application list
```

### DIFF
--- a/pkg/app/web/src/components/application-count/index.tsx
+++ b/pkg/app/web/src/components/application-count/index.tsx
@@ -1,4 +1,3 @@
-import React, { FC, memo, useState } from "react";
 import {
   Box,
   Card,
@@ -7,8 +6,8 @@ import {
   Popover,
   Typography,
 } from "@material-ui/core";
-import { ApplicationKind } from "../../modules/applications";
-import { APPLICATION_KIND_TEXT } from "../../constants/application-kind";
+import clsx from "clsx";
+import React, { FC, memo, useState } from "react";
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -32,12 +31,19 @@ const useStyles = makeStyles((theme) => ({
 export interface ApplicationCountProps {
   totalCount: number;
   disabledCount: number;
-  kind: ApplicationKind;
+  kind: string;
   onClick: () => void;
+  className?: string;
 }
 
 export const ApplicationCount: FC<ApplicationCountProps> = memo(
-  function ApplicationCount({ totalCount, disabledCount, kind, onClick }) {
+  function ApplicationCount({
+    totalCount,
+    disabledCount,
+    kind,
+    onClick,
+    className,
+  }) {
     const classes = useStyles();
 
     const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null);
@@ -48,7 +54,7 @@ export const ApplicationCount: FC<ApplicationCountProps> = memo(
     };
 
     return (
-      <Card raised className={classes.root}>
+      <Card raised className={clsx(classes.root, className)}>
         <CardActionArea
           className={classes.actionArea}
           onClick={onClick}
@@ -58,7 +64,7 @@ export const ApplicationCount: FC<ApplicationCountProps> = memo(
           onMouseLeave={handlePopoverClose}
         >
           <Typography variant="h6" component="div" color="textSecondary">
-            {APPLICATION_KIND_TEXT[kind]}
+            {kind}
           </Typography>
           <Box display="flex" justifyContent="center" alignItems="baseline">
             <Typography variant="h4" component="span">
--- a/pkg/app/web/src/components/application-count/index.tsx
+++ b/pkg/app/web/src/components/application-count/index.tsx
@@ -29,7 +29,7 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 export interface ApplicationCountProps {
-  totalCount: number;
+  enabledCount: number;
   disabledCount: number;
   kindName: string;
   onClick: () => void;
@@ -38,7 +38,7 @@ export interface ApplicationCountProps {
 
 export const ApplicationCount: FC<ApplicationCountProps> = memo(
   function ApplicationCount({
-    totalCount,
+    enabledCount,
     disabledCount,
     kindName,
     onClick,
@@ -68,7 +68,7 @@ export const ApplicationCount: FC<ApplicationCountProps> = memo(
           </Typography>
           <Box display="flex" justifyContent="center" alignItems="baseline">
             <Typography variant="h4" component="span">
-              {totalCount}
+              {enabledCount}
             </Typography>
             {disabledCount > 0 ? (
               <Typography
@@ -110,8 +110,8 @@ export const ApplicationCount: FC<ApplicationCountProps> = memo(
           disableRestoreFocus
         >
           <div>
-            <b>{totalCount}</b>
-            {" total applications"}
+            <b>{enabledCount}</b>
+            {" enabled applications"}
           </div>
           <div>
             <b>{disabledCount}</b>
--- a/pkg/app/web/src/components/application-count/index.tsx
+++ b/pkg/app/web/src/components/application-count/index.tsx
@@ -31,7 +31,7 @@ const useStyles = makeStyles((theme) => ({
 export interface ApplicationCountProps {
   totalCount: number;
   disabledCount: number;
-  kind: string;
+  kindName: string;
   onClick: () => void;
   className?: string;
 }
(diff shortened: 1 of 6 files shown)


pkg/app/web/src/components/application-counts/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+          }
+
+          return (
+            <ApplicationCount
+              key={kindName}
+              kind={kindName}
+              totalCount={
+                counts[kindName][
+                  APPLICATION_ACTIVE_STATUS_NAME[
+                    ApplicationActiveStatus.ENABLED
```
nghialv, line 61: The total number should be `ENABLED` + `DISALBED` + `DELETED`.

khanhtc1202: I think `DELETED` should not be count here 🤔  (agree with it's not just `ENABLED` but `ENABLED` + `DISABLED`)

pkg/app/web/src/components/application-count/index.tsx
```diff
@@ -32,12 +31,19 @@ const useStyles = makeStyles((theme) => ({
 export interface ApplicationCountProps {
   totalCount: number;
   disabledCount: number;
-  kind: ApplicationKind;
+  kind: string;
```
khanhtc1202, line 34: nits, not sure the point of changing the type of `kind` from enum `ApplicationKind` to `string` 🤔

pkg/app/web/src/components/application-counts/index.tsx
nghialv: ah. You are right. enabled + disabled

pkg/app/web/src/components/application-count/index.tsx
```diff
@@ -58,7 +64,7 @@ export const ApplicationCount: FC<ApplicationCountProps> = memo(
           onMouseLeave={handlePopoverClose}
         >
           <Typography variant="h6" component="div" color="textSecondary">
-            {APPLICATION_KIND_TEXT[kind]}
+            {kind}
```
khanhtc1202, line 67: as if we keep this `kind` type as ApplicationKind (enum) I think we could make string key name from enum by `ApplicationKind[kind]` (tested with [playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBAggDnANgSwMYEMAuyD2IDSyIAJlAN5QBQUU+AqgEICiASgHJMAqTAylALxQADABpqUbixYwAYgHkWAWQFQAjGJoBhFnJ48ACgBkYHFQCYNUY4oYARGCoDMlzYbl1bLOmxUAWS0yafIIArGIAvpSUqHgAzjiIwAB0iDgA5gAU8EhoWLgERMQA2tkoGNh4hCRJ9MzsXLwAugL8ggBEtawc3DxtAJRAA)) and `APPLICATION_KIND_TEXT` will safe to be removed? 🤔

cakecatz: @khanhtc1202 That is a TypeScript feature, but `ApplicationKind` is generated by `protoc` from proto file.
The actual code looks like this.
```js
/**
 * @enum {number}
 */
proto.pipe.model.ApplicationKind = {
  KUBERNETES: 0,
  TERRAFORM: 1,
  CROSSPLANE: 2,
  LAMBDA: 3,
  CLOUDRUN: 4,
  ECS: 5
};
```

So we should define that mapping own 😢.

khanhtc1202: Ah, yes I understand the need of our own defined `ApplicationKind` type (as you mentioned 🙏 ) I just wonder, instead of making this `kind` here as primitive `string` type, keeping it as `ApplicationKind` and at the time when we need to pass a string ApplicationKind name to `Typography`, we could make it by using `ApplicationKind[kind]` as 
```typescript
import { ApplicationKind } from "../../modules/applications";
...
<Typography variant="h6" component="div" color="textSecondary">
  {ApplicationKind[kind]}
</Typography>
```
That what I guess 😄 just to clarify, if we could not do that, then please don't mind of this suggestion 😁

pkg/app/web/src/components/application-count/index.tsx
cakecatz: https://github.com/pipe-cd/pipe/pull/1730/files/329239fc425c7614d22031d330e0ea7ffba04026#diff-0efde055bddb2f8b7015e7b2cdbbf5f83ad0cb0a9ade472851219cd14bea75e3R55-R58
```ts
            <ApplicationCount
              key={kindName}
              kind={kindName}
              totalCount={
```

This is because `string` is easier to pass when actually using this component.
It can also reduce dependency.

pkg/app/web/src/components/application-count/index.tsx
khanhtc1202: 👋 I just realize that this `kind` here is a string since we make it as it from here
```typescript
const createInitialCounts = (): Record<string, Record<string, number>> => ({
  [APPLICATION_KIND_TEXT[ApplicationKind.KUBERNETES]]: createInitialCount(),
  [APPLICATION_KIND_TEXT[ApplicationKind.TERRAFORM]]: createInitialCount(),
  [APPLICATION_KIND_TEXT[ApplicationKind.CROSSPLANE]]: createInitialCount(),
  [APPLICATION_KIND_TEXT[ApplicationKind.LAMBDA]]: createInitialCount(),
  [APPLICATION_KIND_TEXT[ApplicationKind.CLOUDRUN]]: createInitialCount(),
  [APPLICATION_KIND_TEXT[ApplicationKind.ECS]]: createInitialCount(),
});

const initialState: ApplicationCounts = {
  updatedAt: 0,
  counts: createInitialCounts(),
};
```
ref [modules/application-counts.ts](https://github.com/pipe-cd/pipe/blob/master/pkg/app/web/src/modules/application-counts.ts#L31-L43)
In that case, rename this `kind` variable as `kindName` (as we do in the caller side [components/application-counts/index.tsx ](https://github.com/pipe-cd/pipe/pull/1730/files/329239fc425c7614d22031d330e0ea7ffba04026#diff-0efde055bddb2f8b7015e7b2cdbbf5f83ad0cb0a9ade472851219cd14bea75e3R55)) may avoid the confusion I caught above 😂 
Sorry for my dumb suggestion 🙏

pkg/app/web/src/components/application-count/index.tsx
khanhtc1202: Thank you for clarifying it 🙏 in that case, my suggestion is to rename this `ApplicationCount.kind` as `ApplicationCount.kindName` to avoid confusing 😅 this `kind` here makes it looks like it should be of type `ApplicationKind` as previously. How do you think about it? 😄

cakecatz: @khanhtc1202 You're right! It seems better to change this.
Thank you! 👍

pkg/app/web/src/components/application-count/index.tsx
cakecatz: @khanhtc1202 Sorry for my confusing explanation.
```js
proto.pipe.model.ApplicationKind = {
  KUBERNETES: 0,
  TERRAFORM: 1,
  CROSSPLANE: 2,
  LAMBDA: 3,
  CLOUDRUN: 4,
  ECS: 5
};
```
This code is an actual enum code generated by `protoc`.
It just a simple `object`. It not has [reverse mappings](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings).
So we cannot get a name by using this code `ApplicationKind[ApplicationKind.KUBERNETES]` (This is `undefined`).

khanhtc1202: 👀 I see, thank you for clarifying it 🙏

pkg/app/web/src/components/application-counts/index.tsx
khanhtc1202: In the second thought, I guess if we make this "total apps" here as `ENABLED` + `DISABLED`, the below Lambda will be `2/2 apps`. 👀 
Screen Shot 2021-03-18 at 14 56 43
IMO, I think the current `0/2apps` pattern looks better to understand, how do you think about it? 🤔

cakecatz: If I look at `2/2`, it will look like there are two active applications, so I think the `enabled/disabled` display will be easier to understand.

khanhtc1202: You're right 👍 please go ahead with `enabled/disabled` pattern 🙆‍♀️

khanhtc1202:
Nice 🚀 
/approve

PR #1757, opened 2021-03-22
Refine ListFilter value for MySQL datastore

khanhtc1202:

Currently, `database/sql` package from Golang does not support values of type Slide, which leads to errors: `{"error": "sql: converting argument $2 type: unsupported type []model.DeploymentStatus, a slice of int32"}` on query `IN` operator in our current `Find` query implementation of MySQL client.

ref: https://golang.org/src/database/sql/driver/types.go#L280

Avoid the above error by build query values of type `slide` or `array` as `(v1,v2,...)` pattern which satisfied the Golang SQL interface.

Fixes #

```release-note
NONE
```


nakabonne:
Looks neat.
/lgtm

PR #1874, opened 2021-04-16
Firestore pagination direction using ordering direction

khanhtc1202:

Related #1555

```release-note
NONE
```

### DIFF
--- a/pkg/datastore/firestore/firestore.go
+++ b/pkg/datastore/firestore/firestore.go
@@ -112,7 +112,13 @@ func (s *FireStore) Find(ctx context.Context, kind string, opts datastore.ListOp
 		if err != nil {
 			return nil, err
 		}
-		q = q.StartAfter(values)
+		// Using the first order direction as pagination direction.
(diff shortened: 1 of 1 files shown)


nakabonne:
@khanhtc1202 To be quite honest, I don't get your point exactly. Tell me a little bit more about what case you're supposing.

nakabonne:
I mean I think it should give back entities after the given cursor in both cases. So it would be great if you could let me know.

khanhtc1202:
@nakabonne @nghialv 
Sorry, I was wrong 🙇 This should be handled by Firestore server itself so no need to care about the direction here 👍 
ref: https://github.com/googleapis/google-cloud-go/blob/master/firestore/query.go#L191-L194
```go
// Otherwise, StartAt should be called with one field value for each OrderBy clause,
// in the order that they appear. For example, in
//   q.OrderBy("X", Asc).OrderBy("Y", Desc).StartAt(1, 2)
// results will begin at the first document where X = 1 and Y = 2.
```

PR #1877, opened 2021-04-16
Enable to specify UID that Piped runs as

nakabonne:
Once https://github.com/pipe-cd/pipe/pull/1871 got merged, I will open up this one after populating the image digest.

Fixes https://github.com/pipe-cd/pipe/issues/1870

```release-note
Enable to specify UID that Piped runs as
```


khanhtc1202:
No wonder 👍 
/lgtm

PR #1891, opened 2021-04-20
Ensure to use /etc/pipecd as a home directory

nakabonne:

Fixes https://github.com/pipe-cd/pipe/issues/1885

```release-note
NONE
```


nakabonne:
It turns out such work isn't needed
/close

PR #1899, opened 2021-04-22
Honor $HOME so that any user can write things in a unified way

nakabonne:
We settled on putting all files to be touched by Piped into the home directory. The piped-base image is already changed to use them: https://github.com/pipe-cd/pipe/pull/1890

I know this PR includes a known issue. It can happen only if the Piped process was owned by a kind of user not in `/etc/passwd`. I just settled on putting this issue to https://github.com/pipe-cd/pipe/issues/1905. And I'd love to merge this PR for now because the latest version of piped doesn't work even if it's running by a user existing in `/ect/passwd`. We had better fix it as soon as possible.

Fixes https://github.com/pipe-cd/pipe/issues/1885

```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -129,10 +129,10 @@ load(
 
 container_pull(
     name = "piped-base",
-    digest = "sha256:41175b8a7cf8acf425808f6770cabcc43fef423a31b647957341ac2de4161a9e",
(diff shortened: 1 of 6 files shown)


khanhtc1202:
Way to go 🚀 
/approve

PR #1907, opened 2021-04-22
Remove redundant firestore indexes

khanhtc1202:
Those indexes should be covered by the newly defined `ProjectId - UpdatedAt - Id` indexes.

Follow PR #1906 

```release-note
NONE
```


nakabonne:
I'm guessing `ProjectId - ASC & UpdatedAt - DESC` is totally different from `ProjectId - ASC & UpdatedAt - DESC & Id - ASC`. But I will look up it a little bit more.

nakabonne:
From skimming the entire code, looks like `ProjectId - ASC & UpdatedAt - DESC` is used for only listing Application/Deployment. Let us eliminate them from indexes.json, existing ones in Firestore don't need to be removed though.

PR #1916, opened 2021-04-23
Change to reset the page when the filter is changed

cakecatz:

Currently, changing the filter does not change the page so the application may not be displayed, so I fixed that.

Fixes #

```release-note
Change to reset the page when the filter is changed
```


khanhtc1202:
Nice 👍 
/approve

PR #1921, opened 2021-04-26
Update installation instruction for single machine

nakabonne:
It's because `--tools-dir` and `sshConfigFilePath` will be non-required from the next version

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/content/en/docs/operator-manual/piped/installation.md
+++ b/docs/content/en/docs/operator-manual/piped/installation.md
@@ -105,28 +105,24 @@ Most part is identical to the previous way, but some part is slightly different.
   apiVersion: pipecd.dev/v1beta1
   kind: Piped
   spec:
-    projectID: YOUR_PROJECT_ID
-    pipedID: YOUR_PIPED_ID
-    pipedKeyFile: PATH_TO_PIPED_KEY_FILE
+    projectID: {YOUR_PROJECT_ID}
+    pipedID: {YOUR_PIPED_ID}
+    pipedKeyFile: {PATH_TO_PIPED_KEY_FILE}
     # Write in a format like "host:443" because the communication is done via gRPC.
-    apiAddress: YOUR_CONTROL_PLANE_ADDRESS
-    webAddress: http://YOUR_CONTROL_PLANE_ADDRESS
+    apiAddress: {YOUR_CONTROL_PLANE_ADDRESS}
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/operator-manual/piped/installation.md
```diff
@@ -105,28 +105,24 @@ Most part is identical to the previous way, but some part is slightly different.
   apiVersion: pipecd.dev/v1beta1
   kind: Piped
   spec:
-    projectID: YOUR_PROJECT_ID
-    pipedID: YOUR_PIPED_ID
-    pipedKeyFile: PATH_TO_PIPED_KEY_FILE
+    projectID: {YOUR_PROJECT_ID}
```
khanhtc1202, line 108: to keep all config file syntax consistent, I think all values should wrap in `{ }` or not at this time, so without `{ }` is better to me. We can make it by another PR which change all other configuration files 👀

docs/content/en/docs/operator-manual/piped/installation.md
```diff
@@ -105,28 +105,24 @@ Most part is identical to the previous way, but some part is slightly different.
...
     syncInterval: 1m
   ```
 
 - Start running the `piped`
 
   ``` console
-  ./piped piped \
-  --config-file=PATH_TO_PIPED_CONFIG_FILE \
-  --tools-dir=PATH_TO_TOOLS_DIR
+  ./piped piped --config-file=PATH_TO_PIPED_CONFIG_FILE
```
khanhtc1202, line 126: Nice 😁

docs/content/en/docs/operator-manual/piped/installation.md
nakabonne: >I think all values should wrap in { } or not 

I feel exactly the same way. But I'm guessing the ones enclosed with `{ }` is better because there are a couple of fields to be all capitalized kind of like `KUBERNETES`, it can be confusing.

khanhtc1202: Sure 👍 Then let's address it with another PR that refine all other configuration files 🙏

nakabonne: Yep 👍

khanhtc1202:
🚀 
/lgtm

nghialv (approved):

PR #1930, opened 2021-04-27
Empower Piped to add login user to /etc/passwd

nakabonne:

Fixes https://github.com/pipe-cd/pipe/issues/1905

```release-note
Add the `--add-login-user-to-passwd` flag to add login user to /etc/passwd when starting Piped
```


nakabonne:
Closing since I settled on taking a four-stage approach like said [here](https://github.com/pipe-cd/pipe/issues/1905#issuecomment-839388119)
/close

PR #1957, opened 2021-05-12
Add a button to copy the environment id

cakecatz:

Create and use the copy button component.

Add a copy button to the environment id as well as the piped id.
Screen Shot 2021-05-12 at 20 40 47

Fixes #

```release-note
Add a button to copy the environment id
```

### DIFF
--- a/pkg/app/web/src/components/copy-icon-button/index.tsx
+++ b/pkg/app/web/src/components/copy-icon-button/index.tsx
@@ -1,0 +1,34 @@
+import { IconButton } from "@material-ui/core";
+import copy from "copy-to-clipboard";
+import { FileCopyOutlined as CopyIcon } from "@material-ui/icons";
+import { FC, useCallback, memo } from "react";
+import { useDispatch } from "react-redux";
+import { addToast } from "../../modules/toasts";
+
+export interface CopyIconButtonProps {
+  name: string;
+  value: string;
+  size?: "small" | "medium";
+  className?: string;
+}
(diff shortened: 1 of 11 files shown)


pkg/app/web/src/components/copy-icon-button/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }, [dispatch, value, name]);
+
+    return (
+      <IconButton
+        className={className}
+        aria-label={`Copy ${name}`}
+        onClick={handleCopy}
+        size={size}
+      >
+        <CopyIcon fontSize={size === "small" ? "small" : "default"} />
```
khanhtc1202, line 30: I guess the type of size should be `"small" | "default"` so that we could remove this ternary operator. (make the default value for this `size` in line 16)

cakecatz: @khanhtc1202 `size` is also used by `IconButton`, and is a different type from `fontSize` in `CopyIcon`. 
(`size = 'small' | 'medium' | 'large'` and `fontSize = 'inherit' | 'default' | 'small' | 'large'`)
To support both sizes, I use the ternary operator here.

khanhtc1202: I get your point 👍 I missed line 28 🙏

khanhtc1202:
🚀 
/lgtm

PR #1964, opened 2021-05-13
Use native clipboard API for copy text on the web

cakecatz:

Use `Clipboard API` instead of `copy-to-clipboard` to copy text.

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

Browser compatibility seems to be fine.

image

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice 👍 
/lgtm

PR #1973, opened 2021-05-17
Update installation instruction of pipectl image

nakabonne:
Current most users can be supposed to Docker users. Even if they use podman, all flags can be pass-through.
Anyway, the current installation instruction is a little bit unkind.

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
🚀

PR #2019, opened 2021-05-27
Fix applications fetching at the Application List page

cakecatz:

Fix a bug that filter options were not applied when fetching applications after the application action. (EDIT, DELETE, DISABLE, etc.)

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/add-application-drawer/index.tsx
+++ b/pkg/app/web/src/components/add-application-drawer/index.tsx
@@ -21,8 +21,8 @@ import {
 
 export interface AddApplicationDrawerProps {
   open: boolean;
-  onClose: () => void;
-  onAdded: () => void;
+  onClose?: () => void;
+  onAdded?: () => void;
 }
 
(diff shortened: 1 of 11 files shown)


pkg/app/web/src/components/add-application-drawer/index.tsx
```diff
@@ -22,14 +22,15 @@ import {
 export interface AddApplicationDrawerProps {
   open: boolean;
   onClose: () => void;
+  onAdded: () => void;
```
khanhtc1202, line 25: how about making this `onAdded` optional and make a default value for it on L33 👀 
In most of cases, looks like we only pass `onAdded={() => null}` for that 🤔

cakecatz: Sure, that might be a good idea.
Thank you!

khanhtc1202:
Thank you 🚀 
/approve

PR #2021, opened 2021-05-27
Add TaskSetDefinition as required configuration for ECS deployment

### LINKED ISSUE #2011: Find better way to get those 2 values instead of set it via service def.
https://github.com/pipe-cd/pipe/blob/8808feb81ca5eafb6b84b585fb4701a0dbba5d9b/pkg/app/piped/cloudprovider/ecs/client.go#L149-L152

khanhtc1202:

Fixes #2011 

```release-note
NONE
```


nakabonne:
Nice
/lgtm

PR #2031, opened 2021-05-31
Make it possible to delete environment from web client

### LINKED ISSUE #2026: Add a capability to disable environment on the web
On the environment settings page:
- add a disable button on a row
- before calling the API, show a warning popup like `it will delete all [applications that belong to this environment](https://{API_ADDRESS}/applications?envId=xxx&page=1)`
- Then call `DeleteEnvironment`
- If `DeleteEnvironment` gives back a `FailedPrecondition` (or something else, I'm debating) error, that indicates there are Pipeds that have permission for the environment to be deleted. It should be shown as a popup

**Why is this needed**:

cakecatz:

image

---

image

Fixes #2026

```release-note
Make it possible to delete environment from web client
```

### DIFF
--- a/pkg/app/web/src/pages/settings/environment/components/delete-confirm-dialog.tsx
+++ b/pkg/app/web/src/pages/settings/environment/components/delete-confirm-dialog.tsx
@@ -4,9 +4,11 @@ import {
   DialogActions,
   DialogContent,
   DialogTitle,
+  Link,
   makeStyles,
   Typography,
 } from "@material-ui/core";
+import { Link as RouterLink } from "react-router-dom";
 import { Alert } from "@material-ui/lab";
 import { FC, memo, useCallback } from "react";
 import { UI_TEXT_CANCEL, UI_TEXT_DELETE } from "../../../../constants/ui-text";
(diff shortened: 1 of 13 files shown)


pkg/app/web/src/pages/settings/environment/components/delete-confirm-dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    return (
+      <Dialog open={env !== null} onClose={handleClose} fullWidth>
+        <form>
+          <DialogTitle>{DIALOG_TITLE}</DialogTitle>
+          <Alert severity="warning" className={classes.description}>
+            {DESCRIPTION}
+          </Alert>
+          <DialogContent>
+            <Typography variant="caption">Delete Applications</Typography>
```
nakabonne, line 86: "Applications to be deleted" might be more appropriate.

pkg/app/web/src/pages/settings/environment/components/delete-confirm-dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+            <Typography variant="body1" className={classes.targetName}>
+              {apps.length > 0 ? apps.map((app) => app.name).join(", ") : "-"}
```
nakabonne, line 88: I'm a bit worried about for Environment where Applications are quite a lot. That is why I suggested just displaying `/applications?envId=xxx` is enough, but what do you think about it?

pkg/app/web/src/pages/settings/environment/components/delete-confirm-dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+            </Typography>
+            <Typography variant="caption">Delete Environment</Typography>
```
nakabonne, line 90: Environment to be deleted?

pkg/app/web/src/pages/settings/environment/components/delete-confirm-dialog.tsx
cakecatz: @nakabonne I thought it would be okay if I make it scrollable, we might need to see other information than just the name, so a link would be better.

cakecatz: Fixed it.

nghialv (approved):

PR #2034, opened 2021-06-02
Using array-contains instead of array-contains-any

khanhtc1202:

Using `array-contains` instead of `array-contains-any` since the value which be used is a simple string. 

ref: 
- https://github.com/pipe-cd/pipe/blob/master/pkg/app/api/grpcapi/web_api.go#L203-L207
- https://firebase.google.com/docs/firestore/query-data/queries#array_membership

Follow PR #2029 

```release-note
NONE
```


nakabonne:
That's the case. At this point we just want to specify one value as a comparison value. Thank you!
/approve

PR #2042, opened 2021-06-03
Support contains operator for MySQL

nakabonne:
Indexing on a JSON column didn't go well. Though I'm in the middle of trying to do it, the query works fine without the index.
Let me merge it to support the environment deletion feature in the next version. It's no problem to add index after that version.

Fixes https://github.com/pipe-cd/pipe/issues/2033

```release-note
NONE
```

### DIFF
--- a/pkg/datastore/mysql/query.go
+++ b/pkg/datastore/mysql/query.go
@@ -75,7 +75,7 @@ func buildWhereClause(filters []datastore.ListFilter) string {
 			valLength := reflect.ValueOf(filter.Value).Len()
 			conds[i] = fmt.Sprintf("%s %s (?%s)", filter.Field, filter.Operator, strings.Repeat(",?", valLength-1))
 		case "JSON_CONTAINS":
-			conds[i] = fmt.Sprintf("%s(%s, '%q', '$')", filter.Operator, filter.Field, filter.Value)
+			conds[i] = fmt.Sprintf("%s(%s, ?, '$')", filter.Operator, filter.Field)
 		default:
 			conds[i] = fmt.Sprintf("%s %s ?", filter.Field, filter.Operator)
 		}
--- a/pkg/datastore/mysql/query.go
+++ b/pkg/datastore/mysql/query.go
@@ -74,6 +74,8 @@ func buildWhereClause(filters []datastore.ListFilter) string {
 			// Make string of (?,...) which contains the number of `?` equal to the element number of filter.Value
 			valLength := reflect.ValueOf(filter.Value).Len()
 			conds[i] = fmt.Sprintf("%s %s (?%s)", filter.Field, filter.Operator, strings.Repeat(",?", valLength-1))
+		case "JSON_CONTAINS":
+			conds[i] = fmt.Sprintf("%s(%s, '%q', '$')", filter.Operator, filter.Field, filter.Value)
 		default:
 			conds[i] = fmt.Sprintf("%s %s ?", filter.Field, filter.Operator)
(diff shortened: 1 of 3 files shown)


pkg/datastore/mysql/query.go
```diff
@@ -74,6 +74,8 @@ func buildWhereClause(filters []datastore.ListFilter) string {
 			// Make string of (?,...) which contains the number of `?` equal to the element number of filter.Value
 			valLength := reflect.ValueOf(filter.Value).Len()
 			conds[i] = fmt.Sprintf("%s %s (?%s)", filter.Field, filter.Operator, strings.Repeat(",?", valLength-1))
+		case "JSON_CONTAINS":
+			conds[i] = fmt.Sprintf("%s(%s, '%q', '$')", filter.Operator, filter.Field, filter.Value)
```
khanhtc1202, line 78: We shouldn't pass the `filter.Value` here, it will automatically be passed by [here](https://github.com/pipe-cd/pipe/blob/master/pkg/datastore/mysql/mysql.go#L98), what we do here is to make the string query that mark every value on the query as `?` character. 😄

nakabonne:
hmm, it's killing me that it takes a bit long time to reflect on the local environment, and plus i wonder it really reflects the changes. i'm in the middle of looking up what's going on.

nakabonne:
>and plus i wonder if it really reflects the changes. 

Okay, first of all, for embed files, it turns out they don't get updated with `make push`, it requires `make build` to re-generate code.
Besides, to apply source code changes, we seem to have to `make kind-down` every time before `make push`, which takes more than 20min or more on my laptop (Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz with 16GB of RAM on macOS 10.15.7) 😭

khanhtc1202:
> Besides, to apply source code changes, we seem to have to make kind-down every time before `make push`

I didn't know that 👀 I'm using docker4mac k8s cluster instead of `kind` on my local 😂

nakabonne:
>I didn't know that 👀 I'm using docker4mac k8s cluster instead of kind on my local 😂

I see. Let me talk about it in the near future.

nakabonne:
Indexing on a JSON column didn't go well. Though I'm in the middle of trying to do it, the query works fine without the index.
Let me merge it to support the environment deletion feature in the next version. It's no problem to add index after that version.

nakabonne:
Sorry, hold on
/hold

pkg/datastore/mysql/query.go
```diff
@@ -74,6 +74,8 @@ func buildWhereClause(filters []datastore.ListFilter) string {
 			// Make string of (?,...) which contains the number of `?` equal to the element number of filter.Value
 			valLength := reflect.ValueOf(filter.Value).Len()
 			conds[i] = fmt.Sprintf("%s %s (?%s)", filter.Field, filter.Operator, strings.Repeat(",?", valLength-1))
+		case "MEMBER OF":
+			conds[i] = fmt.Sprintf("? %s (%s)", filter.Operator, filter.Field)
```
khanhtc1202, line 78: Please add test for this in [query_test.go/TestBuildFindQuery](https://github.com/pipe-cd/pipe/blob/master/pkg/datastore/mysql/query_test.go#L386) 😁

nakabonne: Right right

nakabonne:
PTAL ;)

khanhtc1202:
Way to go 🙌 
/lgtm

PR #2043, opened 2021-06-03
Don't use Deleted as a filtering field

nakabonne:
Deleted field is a newly added one; anything doesn't match with any queries because some entities doesn't have that field. This should be filtered by the web client.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/api/grpcapi/web_api.go
+++ b/pkg/app/api/grpcapi/web_api.go
@@ -158,11 +158,6 @@ func (a *WebAPI) ListEnvironments(ctx context.Context, req *webservice.ListEnvir
 				Operator: datastore.OperatorEqual,
 				Value:    claims.Role.ProjectId,
 			},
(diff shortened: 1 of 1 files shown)


pkg/app/api/grpcapi/web_api.go
```diff
@@ -254,11 +249,6 @@ func (a *WebAPI) DeleteEnvironment(ctx context.Context, req *webservice.DeleteEn
 				Operator: datastore.OperatorEqual,
 				Value:    req.EnvironmentId,
 			},
-			{
-				Field:    "Deleted",
-				Operator: datastore.OperatorEqual,
-				Value:    false,
```
nakabonne, line 260: We can delete logically deleted applications, it's definitely a waste though.

khanhtc1202:
Nice catch
/lgtm

PR #2049, opened 2021-06-08
Disable EnvIds column

nakabonne:
In order to maintain consistency, commented out the newly added column where type is likely to be varied in the near future. After fixing the type, it will be reverted.

If you try to delete an environment where datastore is MySQL, it gets something like:

image

Fixes #

```release-note
NONE
```


khanhtc1202:
Cool, thank you 🚀 
/lgtm

PR #2082, opened 2021-06-16
Show ECS traffic stage detail on web view

khanhtc1202:

Follow PR #2077 

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/executor/ecs/ecs.go
+++ b/pkg/app/piped/executor/ecs/ecs.go
@@ -37,8 +37,8 @@ const (
 	canaryTaskSetARNKeyName = "canary-taskset-arn"
 	canaryServiceKeyName    = "canary-service-object"
 	// Stage metadata keys.
-	trafficRoutePrimaryMetadataKey = "primary-traffic"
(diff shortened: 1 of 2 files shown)


pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -37,8 +37,8 @@ const (
 	canaryTaskSetARNKeyName = "canary-taskset-arn"
 	canaryServiceKeyName    = "canary-service-object"
 	// Stage metadata keys.
-	trafficRoutePrimaryMetadataKey = "primary-traffic"
-	trafficRouteCanaryMetadataKey  = "canary-traffic"
+	trafficRoutePrimaryMetadataKey = "primary-percentage"
+	trafficRouteCanaryMetadataKey  = "canary-percentage"
```
khanhtc1202, line 41: Change the key name so that we can reuse the web view defined key name
ref: https://github.com/pipe-cd/pipe/pull/2082/files#diff-f1edd0e8d855531a0d162b13744ae393436bcf4a6f2e9b795254aaa333c44a14R62-R67

PR #2095, opened 2021-06-18
Make Popover render only when needed

cakecatz:

I set the KeepMounted option to false.
This will mount the DOM only when the menu is opened.
https://material-ui.com/api/modal/#props

**Before**

<image width="300" src="https://user-images.githubusercontent.com/6136383/122552546-63d94f80-d071-11eb-89ca-05d90c1685ce.png" />

**After**

<image width="300" src="https://user-images.githubusercontent.com/6136383/122552490-56bc6080-d071-11eb-9a17-6db29f81e1cb.png" />

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice catch ⚾ 
/approve

PR #2115, opened 2021-06-23
Add minimum required dashboards

nakabonne:
This PR adds three Grafana dashboards.

image

Because all dashboards got removed due to https://github.com/pipe-cd/pipe/pull/2191, let me merge once though I'm still in the middle of building.

Ref https://github.com/pipe-cd/pipe/issues/1867

```release-note
NONE
```


nakabonne:
After discussing, we will apply them.
/hold

manifests/pipecd/values.yaml
```diff
@@ -382,15 +382,17 @@ grafana:
           folder: 'Cluster'
           type: file
           disableDeletion: false
-          editable: false
+          # TODO: Disallow to update Grafana UI from the web once got fixed
+          # The following Control-Plane's one should be reverted as well.
+          allowUiUpdates: true
```
nakabonne, line 387: We can build dashboards by copying the original. But it's a bit tedious especially because of the current number of dashboards.

khanhtc1202:
🤟🤟🤟
/approve

PR #2155, opened 2021-06-30
Enable Node exporter and Kube state metrics

nakabonne:
This PR enables [Node exporter](https://github.com/prometheus/node_exporter) and [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) so that it can fetch node metrics and cluster-level metrics.

Ref https://github.com/pipe-cd/pipe/issues/1867

```release-note
NONE
```


khanhtc1202:
🚀 
/approve

PR #2197, opened 2021-07-07
Do not render json files under grafana-dashboards

nakabonne:
To avoid something like: https://github.com/pipe-cd/pipe/pull/2115#issuecomment-875304232

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice catch
/approve

PR #2217, opened 2021-07-12
Disable to collect cluster-level stats by default

nakabonne:
This PR allows us to switch monitoring scope using `.Values.monitoring.clusterStats`.
To make Prometheus config possible to template, I moved it to configmap.yaml.

If `.Values.monitoring.clusterStats` true:
- Prometheus scrape cluster stats such as Nodes, Pods and Prometheus itself
- Grafana shows cluster dashboards

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/pipecd/templates/configmap.yaml
+++ b/manifests/pipecd/templates/configmap.yaml
@@ -11,9 +11,9 @@ data:
 {{- end }}
 
 {{- if .Values.monitoring.enabled }}
+{{- if .Values.monitoring.clusterStats }}
 ---
 
-{{- if .Values.monitoring.clusterStats }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


manifests/pipecd/templates/configmap.yaml
```diff
@@ -13,6 +13,7 @@ data:
 {{- if .Values.monitoring.enabled }}
 ---
 
+{{- if .Values.monitoring.clusterStats }}
```
khanhtc1202, line 16: Lets keep the convention of nearest divider `---` belong to the doc by move this line be for L14 and the close ( `{{ end }}` ) at line L32 to L31 👀  How do you think?

nakabonne: for sure. I gonna apply it!

nakabonne:
Fixed

khanhtc1202:
Thx
/lgtm

PR #2227, opened 2021-07-13
Change to manage Grafana datasources as a ConfigMap

nakabonne:
We used to fix the Prometheus service name as `pipecd-prometheus-server` to discover it from Grafana statically. This PR resolves that problem by changing to manage the datasources config as our ConfigMap, which can template with the release name.

On the other hand, still we have to fix that ConfigMap name. It is `pipecd-grafana-datasources` by default. If you want to change the name, you can change it by setting `.Values.grafana.extraConfigmapMounts[0].configMap`.

Fixes https://github.com/pipe-cd/pipe/issues/2226

```release-note
NONE
```

### DIFF
--- a/manifests/pipecd/templates/configmap.yaml
+++ b/manifests/pipecd/templates/configmap.yaml
@@ -50,6 +50,29 @@ data:
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


khanhtc1202:
Nice
/approve

PR #2233, opened 2021-07-13
Use declared TTL for stat cache instead of default value for normal cache

khanhtc1202:

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/pipecd/server.go
+++ b/cmd/pipecd/server.go
@@ -65,7 +65,7 @@ var (
 
 const (
 	defaultPipedStatHashKey = "HASHKEY:PIPED:STATS"
-	defaultPipedStatTTL     = 2 * time.Minute
+	pipedStatTTL            = 2 * time.Minute
 )
 
 type httpHandler interface {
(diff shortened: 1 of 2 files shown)


cmd/pipecd/server.go
```diff
@@ -65,6 +65,7 @@ var (
 
 const (
 	defaultPipedStatHashKey = "HASHKEY:PIPED:STATS"
+	defaultPipedStatTTL     = 2 * time.Minute
```
nakabonne, line 68: This shouldn't be changeable even in the future; `pipedStatTTL` may be better to prevent us confusion.

nakabonne:
This change avoids showing wrong metrics for a long time. LGTM except for a tiny comment!
/lgtm

cmd/pipecd/server.go
khanhtc1202: Nice point 👍 lets me adopt that 🙆‍♂️

khanhtc1202: Done

nghialv (approved):

PR #2281, opened 2021-07-21
Filter outdated piped stat metrics

khanhtc1202:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/ops/pipedstatsbuilder/builder.go
+++ b/pkg/app/ops/pipedstatsbuilder/builder.go
@@ -19,11 +19,16 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"time"
 
 	"go.uber.org/zap"
 
 	"github.com/pipe-cd/pipe/pkg/cache"
 	"github.com/pipe-cd/pipe/pkg/model"
+)
+
+const (
+	liveStateExceedDuration = 2 * time.Minute
 )
 
 type PipedStatsBuilder struct {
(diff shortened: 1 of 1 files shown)


pkg/app/ops/pipedstatsbuilder/builder.go
```diff
@@ -19,13 +19,18 @@ import (
...
+	"time"
 
 	"go.uber.org/zap"
 
 	"github.com/pipe-cd/pipe/pkg/cache"
 	"github.com/pipe-cd/pipe/pkg/model"
 )
 
+const (
+	liveStateExceedDuration = 2 * time.Minute
```
nakabonne, line 31: nit: How about something like:
```suggestion
	pipedStatsRetention = 2 * time.Minute
```

nakabonne: Looks like It isn't only for liveState. And `retention` is often used for how long the data is stored in the monitoring context.

khanhtc1202: Thanks, lets me apply this 🙏

khanhtc1202: Addressed by https://github.com/pipe-cd/pipe/pull/2281/commits/bfccaa3ec393d757f1d0d5282c135469a1d0dcd9 🙏

nakabonne:
Nice
/approve

PR #2287, opened 2021-07-26
Implement Pipeds stat cleaner

khanhtc1202:

Ensure remove staled Pipeds' stat using cleaner.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/ops/staledpipedstatcleaner/cleaner.go
+++ b/pkg/app/ops/staledpipedstatcleaner/cleaner.go
@@ -1,0 +1,107 @@
+// Copyright 2021 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package staledpipedstatcleaner
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/pipe-cd/pipe/pkg/cache"
+	"github.com/pipe-cd/pipe/pkg/model"
+)
+
+var (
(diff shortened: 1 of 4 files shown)


pkg/app/ops/staledpipedstatcleaner/cleaner.go
```diff
@@ -0,0 +1,110 @@
...
+		select {
+		case <-ctx.Done():
+			s.logger.Info("staledPipedStatCleaner has been stopped")
+			return nil
+
+		case <-t.C:
+			start := time.Now()
+			if err := s.flushStaledPipedStat(); err == nil {
+				s.logger.Info("successfully cleaned staled pipeds stat", zap.Duration("duration", time.Since(start)))
+			}
```
nakabonne, line 61: doesn't handle the error?

nakabonne: I'd say stop emitting log within flushStaledPipedStat, instead, wrap them using fmt.Errorf like `fmt.Errorf("ailed to fetch piped stats from cache", err)`. Then emit them as a log here

pkg/app/ops/staledpipedstatcleaner/cleaner.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/pipe-cd/pipe/pkg/cache"
+	"github.com/pipe-cd/pipe/pkg/model"
+)
+
+var (
```
nakabonne, line 30: should be const

pkg/app/ops/staledpipedstatcleaner/cleaner.go
khanhtc1202: Okay, I hear you 👍 Just wanna keep this cleaner the same as other cleaners but I like your suggestion. Will adopt that 👌

nakabonne:
Nice one
/approve

pkg/app/ops/staledpipedstatcleaner/cleaner.go
khanhtc1202: @nakabonne sorry, I mean I will adopt your suggestion in this PR 🙏 
Addressed by https://github.com/pipe-cd/pipe/pull/2287/commits/c92b4db970abc14f761ce8ad313be0ea42da357a 🙏 
PTAL 👀

nakabonne: i got wrong

PR #2307, opened 2021-07-27
Fix flaky piped stat builder test

khanhtc1202:

Currently, the bytes stream returned by `PipedStatsBuilder.Build` has inconsistent order (cause by Golang hash map[string]interface which provided by its backend `hashcache.GetAll` call). Since we don't care about the order of that bytes stream, lets update the test assert to make it compares in a more flexible way.

ref: https://github.com/pipe-cd/pipe/blob/master/pkg/app/ops/pipedstatsbuilder/builder.go#L47

Related to https://github.com/pipe-cd/pipe/pull/2299#issuecomment-886516252

```release-note
NONE
```

### DIFF
--- a/pkg/app/ops/pipedstatsbuilder/builder_test.go
+++ b/pkg/app/ops/pipedstatsbuilder/builder_test.go
@@ -65,10 +65,10 @@ func TestBuildPipedStat(t *testing.T) {
 
 	buf := new(strings.Builder)
 	io.Copy(buf, rc)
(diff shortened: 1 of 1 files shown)


nakabonne:
ElementsMatch is better. Thank you
/approve

khanhtc1202:
@nghialv @nakabonne My miss on the trailing newline, PTAL 🙏

PR #2312, opened 2021-07-27
Use labels to distinguish what component's metrics

nakabonne:
This PR stops distinguishing component by using other metrics name, instead settled on using labels like `go_goroutines{pipecd_component="ops"} 23`

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/pipecd/server.go
+++ b/cmd/pipecd/server.go
@@ -131,7 +131,7 @@ func NewServerCommand() *cobra.Command {
 
 func (s *server) run(ctx context.Context, t cli.Telemetry) error {
 	// Register all metrics.
-	registerMetrics()
+	reg := registerMetrics()
 
 	group, ctx := errgroup.WithContext(ctx)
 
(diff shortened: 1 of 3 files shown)


cmd/pipecd/server.go
```diff
@@ -465,8 +465,16 @@ func createFilestore(ctx context.Context, cfg *config.ControlPlaneSpec, logger *
 	}
 }
 
-func registerMetrics() {
-	r := prometheus.WrapRegistererWithPrefix("pipecd_server_", prometheus.DefaultRegisterer)
+func registerMetrics() *prometheus.Registry {
+	r := prometheus.NewRegistry()
+	wrapped := prometheus.WrapRegistererWith(map[string]string{
```
khanhtc1202, line 470: I think this wrapped registerer should be used in L470,L410 instead of the r registerer 🤔

nakabonne: good catch!

nakabonne:
Fixed

khanhtc1202:
🚀 
/lgtm

nghialv (approved):

PR #2325, opened 2021-08-02
Ensure to register grpc server metrics to custom Prometheus registerer

nakabonne:
The `go_grpc_prometheus` package registers its counters to the default registerer in advance, but for custom one, it's needed to register manually. After adopting the custom one, we're losing grpc metrics until now.

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice catch 👍 
/approve

PR #2349, opened 2021-08-12
Change Prometheus pod update strategy to Recreate

nakabonne:
Using `--storage.tsdb.no-lockfile` is a bit dangerous because possibly a newly created process could try to write the disk where the old one is writing.

This change causes a bit down time but seems like it's better as a workaround at least compared to using that flag.

Fixes https://github.com/pipe-cd/pipe/issues/2348

```release-note
NONE
```


khanhtc1202:
Way to go 🚀 
/approve

PR #2361, opened 2021-08-17
Add sql service account

khanhtc1202:

Add sqlServiceAccount as helm values. It will be used to enable connect from PipeCD workload to GCP Cloud SQL using Cloud SQL Auth proxy.

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/pipecd/values.yaml
+++ b/manifests/pipecd/values.yaml
@@ -83,6 +83,9 @@ secret:
   gcsServiceAccount:
     fileName: "gcs-service-account"
     data: ""
+  sqlServiceAccount:
+    fileName: "sql-service-account"
+    data: ""
   minioAccessKey:
     fileName: "minio-access-key"
     data: ""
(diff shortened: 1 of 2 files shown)


manifests/pipecd/values.yaml
```diff
@@ -83,6 +83,9 @@ secret:
   gcsServiceAccount:
     fileName: "gcs-service-account"
     data: ""
+  sqlServiceAccount:
```
nakabonne, line 86: nit: `cloudSQLServiceAccount`? We're going to support the authentication for Amazon RDS, right?

khanhtc1202: Well, this change is for authorization, not authentication. And I've not looked at Amazon RDS connect method yet, but basically, we only support authentication RDS by username/password as we do for now, this service account is specific for connect to GCP cloud SQL only 🙏 
About the name change, I will adopt your suggestion since it would be better to clearly mention that this is only for cloud SQL as you said 🙆‍♀️

khanhtc1202: addressed by https://github.com/pipe-cd/pipe/pull/2361/commits/48309fffe4394196e14c266694924d9b0bc704f0 🙏

nakabonne:
Thank you
/approve

PR #2420, opened 2021-09-03
Add ECS livestate store

khanhtc1202:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/livestatestore/ecs/store.go
+++ b/pkg/app/piped/livestatestore/ecs/store.go
@@ -1,0 +1,51 @@
+// Copyright 2021 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


pkg/app/piped/livestatestore/ecs/store.go
```diff
@@ -0,0 +1,51 @@
...
+		With(zap.String("cloud-provider", cloudProvider))
+
+	return &Store{
+		logger: logger,
+	}
+}
+
+func (s *Store) Run(ctx context.Context) error {
+	s.logger.Info("start running ecs app state store")
+
```
nakabonne, line 48: You look to add the actual implementation here right?

khanhtc1202: Yeh, currently just add this thing to unify with other deployment kinds. Just released I missed it while implemented it in the past.

nakabonne: Gotcha

PR #2439, opened 2021-09-09
Add the explanation to the document to add bazelisk to user's PATH

ono-max:

When I setup my environment for pipe-cd with the document, the error is occurred as follows.
```shell
s15236@CA-20014576 pipe % bazelisk version
zsh: command not found: bazelisk
```

The reason of this was that bazelisk had not been added to my PATH. So, I added the explanation to the document.

Fixes #

<!--
If no, just write "NONE" in the release-note block below.
```release-note
NONE
```

### DIFF
--- a/docs/content/en/docs/contribution-guidelines/development.md
+++ b/docs/content/en/docs/contribution-guidelines/development.md
@@ -16,6 +16,11 @@ You can install `bazelisk` by `go get` as the following:
 go get github.com/bazelbuild/bazelisk
 ```
 
+Then Add it to your PATH
+```
+export PATH=$PATH:$(go env GOPATH)/bin
+```
+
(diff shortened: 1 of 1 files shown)


khanhtc1202:
This setup should be a part of Golang installation so I don't think we need that in our docs.

ono-max:
@khanhtc1202 

In the [bazelisk](https://github.com/bazelbuild/bazelisk#requirements) project, there is the explanation about user's PATH. That's why I think I need it.

what do you think?

khanhtc1202:
This PR is okay to be closed after this update.
https://github.com/pipe-cd/pipe/pull/2440#discussion_r704939605

ono-max:
I don't need change because bazelisk doc will be referred.

PR #2466, opened 2021-09-10
Reset grafana provisioning board id

khanhtc1202:

As mentioned in the official [docs](https://grafana.com/docs/grafana/latest/administration/provisioning/)
> Note: The JSON definition in the input field when using Copy JSON to Clipboard or Save JSON to file will have the id field automatically removed to aid the provisioning workflow.

We need to reset those boards' id to `null` so that exported JSON configuration could be reused by other Grafana pods. I still concern about the behavior of the Grafana pod from which we export the board configuration, since the configuration contains no id (set to `null`), it may create a completely new board with the same `uid`?

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/pipecd/grafana-dashboards/cluster/pod.json
+++ b/manifests/pipecd/grafana-dashboards/cluster/pod.json
@@ -1,4 +1,40 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "8.1.2"
+    },
+    {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph (old)",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "stat",
+      "name": "Stat",
+      "version": ""
+    }
+  ],
   "annotations": {
     "list": [
       {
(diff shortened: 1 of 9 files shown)


manifests/pipecd/grafana-dashboards/cluster/pod.json
```diff
@@ -672,268 +698,16 @@
         "label": null,
         "multi": false,
         "name": "service",
-        "options": [
```
nakabonne, line 675: Why variable options get removed?

nakabonne:
Not sure why it's not enough just to change id to null. Can I check the whole dashboards in any environment?

khanhtc1202:
> Not sure why it's not enough just to change id to null. Can I check the whole dashboards in any environment?

I get your point, currently, the only way to test it is using the local cluster with monitoring value set to `true` or on dev. Besides, those configurations are export from dev, the changes contain diff cause by Grafana version being updated as well so it quite big. Should I separate this PR into 3 parts (for piped, control-plane and cluster)?

nakabonne:
Yep, it's great if you can separate this into several parts. Too huge to see what will happen exactly (actually unexpected change made). If it solves this problem by just changing id to null, you can first make that change, then make others little by little.

manifests/pipecd/grafana-dashboards/cluster/pod.json
nakabonne: Makes sense. Options aren't needed to export because they are determined dynamically.

nakabonne:
Well, it's fine to merge because seems like those JSON files are exported from the current dev dashboards with `Export for sharing externally` button.

/lgtm

manifests/pipecd/grafana-dashboards/cluster/pod.json
khanhtc1202: Ah get it, thanks for your explanation 👍

khanhtc1202:
> Well, it's fine to merge because seems like those JSON files are exported from the current dev dashboards with `Export for sharing externally` button.
> 
> /lgtm

Thanks 👍  Since we updated Grafana version recently so those changes on auto-generation configs are expected. Let's check it again on dev after it been merged, will revert or make fixes if necessary.

nghialv (approved):

PR #2467, opened 2021-09-10
Insert new lines in `pkg/app/piped/controller/scheduler.go` to consolidate the format

ono-max:
Add the comment and insert a new line

Fixes #

```release-note
NONE
```


pkg/app/piped/controller/scheduler.go
```diff
@@ -642,12 +643,14 @@ func (s *scheduler) reportMostRecentlySuccessfulDeployment(ctx context.Context)
 		retry = pipedservice.NewRetry(10)
 	)
 
+	// Update deployment status on remote.
```
khanhtc1202, line 646: the called function name is self-explanatory so I don't think we need this comment.

nghialv (approved):

PR #2468, opened 2021-09-13
Add a new analyzer implementation to address the new ADA strategies

nakabonne:
This PR makes two changes:

- add configuration fields for the new ADA strategies
- add new analyzer implementation so that we can implement for each strategy after this.

This PR doesn't change the code that will be run in any context. Just preparing for the next step.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/livestatestore/BUILD.bazel
+++ b/pkg/app/piped/livestatestore/BUILD.bazel
@@ -8,7 +8,6 @@ go_library(
     deps = [
         "//pkg/app/piped/cloudprovider/kubernetes:go_default_library",
         "//pkg/app/piped/livestatestore/cloudrun:go_default_library",
-        "//pkg/app/piped/livestatestore/ecs:go_default_library",
         "//pkg/app/piped/livestatestore/kubernetes:go_default_library",
         "//pkg/app/piped/livestatestore/lambda:go_default_library",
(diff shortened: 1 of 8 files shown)


nakabonne:
Supplement:

Currently, the analysis executor performs static ADA by just letting an analysis provider evaluate. (see [here](https://github.com/pipe-cd/pipe/blob/814794b442bac526472e3ff5ce392c94ca4b9275/pkg/app/piped/executor/analysis/analyzer.go#L61-L96))

But in this way, it's a bit difficult to compare two variants. So I settled on adding a new analyzer that evaluates itself and looking to replace the above legacy implementation with it.

pkg/app/piped/livestatestore/BUILD.bazel
```diff
@@ -8,7 +8,6 @@ go_library(
     deps = [
         "//pkg/app/piped/cloudprovider/kubernetes:go_default_library",
         "//pkg/app/piped/livestatestore/cloudrun:go_default_library",
-        "//pkg/app/piped/livestatestore/ecs:go_default_library",
```
khanhtc1202, line 11: why do we have this change? 🤔

nakabonne: For sure. Let me undo it.

PR #2492, opened 2021-09-17
Remove fake client for Piped service

nakabonne:

Fixes #

```release-note
Remove the "--use-fake-api-client" option from Piped
```


khanhtc1202:
@nakabonne We have it here too
https://github.com/pipe-cd/pipe/blob/master/manifests/piped/templates/deployment.yaml#L33

nakabonne:
Right

nakabonne:
I `make clean`-ed just now so just a moment

khanhtc1202:
すっきりした 😌
/approve

Issue #2448, opened 2021-09-09
Fix repeated name application on list deployment page filter


khanhtc1202:
**What would you like to be added**:

<img width="357" alt="Screen Shot 2021-09-09 at 14 14 15" src="https://user-images.githubusercontent.com/32532742/132626793-346a0dff-c1cf-44a0-b08b-fa4a90e299f4.png">

**Why is this needed**:

khanhtc1202:
Fixed by PR #2496 
/close

PR #2493, opened 2021-09-17
API add ability to filter deployments by application name

khanhtc1202:

Currently, we only have a way to filter deployments by Env, Kind, Status, and ApplicationId (display as application name on UI) which is hard to use in case there are many applications with the same name. This change adds ApplicationName as a filter, the change for the client will be addressed by another PR.

Part of #2448

```release-note
NONE
```

### DIFF
--- a/pkg/app/api/grpcapi/web_api.go
+++ b/pkg/app/api/grpcapi/web_api.go
@@ -1041,6 +1041,13 @@ func (a *WebAPI) ListDeployments(ctx context.Context, req *webservice.ListDeploy
 				Value:    o.EnvIds[0],
 			})
 		}
+		if o.ApplicationName != "" {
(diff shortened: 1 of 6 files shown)


nakabonne:
Honestly I'm not sure why this change solves that issue. I think the root cause is that it contains disabled applications as filter options.

nakabonne:
That makes sense. Now I get it.
/lgtm

khanhtc1202:
> Honestly I'm not sure why this change solves that issue. I think the root cause is that it contains disabled applications as filter options.

Has your question been answered yet?

nakabonne:
It has 👍

khanhtc1202:
> It has 👍

nice 🤝

PR #2567, opened 2021-10-04
Add support zip packing function code for Lambda deployment

### LINKED ISSUE #2551: Support zip packing for Lambda deployment kind

Currently, PipeCD only supports Lambda function which uses package type `Image` (container as function). We're going to support zip packing as well.

ref: https://github.com/aws/aws-sdk-go-v2/blob/main/service/lambda/types/types.go#L341-L366
```go
// The code for the Lambda function. You can specify either an object in Amazon S3,
// upload a .zip file archive deployment package directly, or specify the URI of a
// container image.
type FunctionCode struct {

	// URI of a container image
	// (https://docs.aws.amazon.com/lambda/latest/dg/lambda-images.html) in the Amazon
	// ECR registry.
	ImageUri *string

	// An Amazon S3 bucket in the same Amazon Web Services Region as your function. The
	// bucket can be in a different Amazon Web Services account.
	S3Bucket *string

	// The Amazon S3 key of the deployment package.
	S3Key *string

	// For versioned objects, the version of the deployment package object to use.
	S3ObjectVersion *string

	// The base64-encoded contents of the deployment package. Amazon Web Services SDK
	// and Amazon Web Services CLI clients handle the encoding for you.
	ZipFile []byte

	noSmithyDocumentSerde
}
```

With the current implementation, we can simply address this issue by supporting other kinds of FunctionCode, the official SDK supports: container image, s3 object, zip file directly as FunctionCode and I'm going to add s3 object as PipeCD supporting FunctionCode.

ref: https://github.com/pipe-cd/pipe/blob/master/pkg/app/piped/cloudprovider/lambda/client.go#L112-L114
```go
		Code: &types.FunctionCode{
		},
```

PipeCD supporting features for Lambda deployment kind (such as canary deployment) get no hurt due to this update but I worried about the versioning control problem in case of using zip packing mode.

**Why is this needed**:

khanhtc1202:

Fixes #2551 

```release-note
Add ability to deploy Lambda zip packing function code which stored in AWS S3
```

### DIFF
--- a/pkg/app/piped/cloudprovider/lambda/client.go
+++ b/pkg/app/piped/cloudprovider/lambda/client.go
@@ -109,16 +109,31 @@ func (c *client) IsFunctionExist(ctx context.Context, name string) (bool, error)
 
 func (c *client) CreateFunction(ctx context.Context, fm FunctionManifest) error {
 	input := &lambda.CreateFunctionInput{
-		Code: &types.FunctionCode{
-			ImageUri: aws.String(fm.Spec.ImageURI),
-		},
-		PackageType:  types.PackageType("Image"),
-		Role:         aws.String(fm.Spec.Role),
 		FunctionName: aws.String(fm.Spec.Name),
+		Role:         aws.String(fm.Spec.Role),
+		MemorySize:   aws.Int32(fm.Spec.Memory),
+		Timeout:      aws.Int32(fm.Spec.Timeout),
 		Tags:         fm.Spec.Tags,
 		Environment: &types.Environment{
 			Variables: fm.Spec.Environments,
 		},
+	}
+	// Container image packing.
+	if fm.Spec.ImageURI != "" {
+		input.PackageType = types.PackageTypeImage
+		input.Code = &types.FunctionCode{
+			ImageUri: aws.String(fm.Spec.ImageURI),
+		}
+	}
+	// Zip packing which stored in s3.
+	if fm.Spec.S3Bucket != "" {
+		input.PackageType = types.PackageTypeZip
+		input.Code = &types.FunctionCode{
+			S3Bucket:        aws.String(fm.Spec.S3Bucket),
+			S3Key:           aws.String(fm.Spec.S3Key),
+			S3ObjectVersion: aws.String(fm.Spec.S3ObjectVersion),
+		}
+		input.Handler = aws.String(fm.Spec.Handler)
 	}
 	_, err := c.client.CreateFunction(ctx, input)
 	if err != nil {
@@ -131,7 +146,16 @@ func (c *client) UpdateFunction(ctx context.Context, fm FunctionManifest) error 
 	// Update function code.
 	codeInput := &lambda.UpdateFunctionCodeInput{
 		FunctionName: aws.String(fm.Spec.Name),
-		ImageUri:     aws.String(fm.Spec.ImageURI),
+	}
+	// Container image packing.
+	if fm.Spec.ImageURI != "" {
+		codeInput.ImageUri = aws.String(fm.Spec.ImageURI)
+	}
+	// Zip packing which stored in s3.
+	if fm.Spec.S3Bucket != "" {
+		codeInput.S3Bucket = aws.String(fm.Spec.S3Bucket)
+		codeInput.S3Key = aws.String(fm.Spec.S3Key)
+		codeInput.S3ObjectVersion = aws.String(fm.Spec.S3ObjectVersion)
 	}
 	_, err := c.client.UpdateFunctionCode(ctx, codeInput)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/app/piped/cloudprovider/lambda/client.go
```diff
@@ -109,17 +109,32 @@ func (c *client) IsFunctionExist(ctx context.Context, name string) (bool, error)
...
+			ImageUri: aws.String(fm.Spec.ImageURI),
+		}
+	}
+	// Zip packing which stored in s3.
+	if fm.Spec.S3Bucket != "" {
+		input.PackageType = types.PackageTypeZip
+		input.Code = &types.FunctionCode{
+			S3Bucket:        aws.String(fm.Spec.S3Bucket),
+			S3Key:           aws.String(fm.Spec.S3Key),
+			S3ObjectVersion: aws.String(fm.Spec.S3ObjectVersion),
```
nghialv, line 134: I want to ask about this object version. Is it configurable or predictable from the developer?
I mean our designed workflow is as
1. developer runs CI tasks to upload their source code to their S3 bucket 
2. developer updates (or by our Event Watcher) `s3ObjectVersion` field in function.yaml file to deploy

At the first step, can they easily know the `s3ObjectVersion` of the newly uploaded data?
The version can be specified or generated automatically with a random string?

khanhtc1202: > The version can be specified or generated automatically with a random string?
 
 It will be generated by S3 due to this [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/versioning-workflows.html)
 
 > can they easily know the s3ObjectVersion of the newly uploaded data?
 
 At the CI step, the `sam package ...` command which does (1) as you mentioned will return required value to detect where to retrieve the object (key, version).

khanhtc1202: Besides, one more thing to mention is that the `s3Objectversion` is not a required field, in case users do not enable versioning feature on s3, they still can upload their zip function code to the s3 and use only s3Bucket and s3Key (the path to object) to deploy their lambda function.

nghialv:
Lets merge this to support S3 and make another support for deploying Zip directly.

/lgtm

pkg/app/piped/cloudprovider/lambda/client.go
```diff
@@ -109,17 +109,32 @@ func (c *client) IsFunctionExist(ctx context.Context, name string) (bool, error)
...
 	}
+	// Container image packing.
+	if fm.Spec.ImageURI != "" {
+		input.PackageType = types.PackageTypeImage
+		input.Code = &types.FunctionCode{
+			ImageUri: aws.String(fm.Spec.ImageURI),
+		}
+	}
+	// Zip packing which stored in s3.
+	if fm.Spec.S3Bucket != "" {
```
nakabonne, line 129: It seems to be assumed that only one of `ImageURI` or `S3Bucket` is specified.
I'm gonna say we need to add a check to `FunctionManifestSpec.validate()` as below:

```go
	if fmp.ImageURI != "" && fmp.S3Bucket != "" {
		return fmt.Errof("error")
	}
```

khanhtc1202: Thanks for the comment, I understand your thought. I did the same way before but the AWS SDK itself validates this created `input` when we pass it to CreateFunction API and returns error which emphasizes what is missing in detail so I guess we don't need to do more validation stuff.

khanhtc1202: On the second thought, I think it's good, lets me adopt that 🙏

khanhtc1202: addressed by https://github.com/pipe-cd/pipe/pull/2567/commits/f8992ba25df3a72ebbac51b3a759e02b29c6e582 🙏

pkg/app/piped/cloudprovider/lambda/function.go
```diff
@@ -53,21 +53,25 @@ func (fm *FunctionManifest) validate() error {
...
 }
 
 func (fmp FunctionManifestSpec) validate() error {
 	if len(fmp.Name) == 0 {
 		return fmt.Errorf("lambda function is missing")
 	}
-	if len(fmp.ImageURI) == 0 {
-		return fmt.Errorf("image uri is missing")
+	if len(fmp.ImageURI) == 0 && len(fmp.S3Bucket) == 0 {
+		return fmt.Errorf("one of image or s3 bucket is required to be configured")
```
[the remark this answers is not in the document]
khanhtc1202: I don't think the above suggestion can ensure that only one of them is specified, since if both have length `> 0` it still passes the if condition. Besides, CreateFunction (and also other AWS SDK APIs) has its own validation so passes them there and lets SDK does its work is easier for us, what I want to do here is to ensure at least one of the required fields had configured. How do you think?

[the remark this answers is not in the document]
khanhtc1202:
```go
	if len(fmp.ImageURI) == 0 && len(fmp.S3Bucket) == 0 {
		return fmt.Errorf("one of image or s3 bucket is required to be configured")
	}
	if len(fmp.ImageURI) != 0 && len(fmp.S3Bucket) != 0 {
		return fmt.Errorf("at least one of image or s3 bucket is required to be configured")
	}
```
or is it your suggestion 😄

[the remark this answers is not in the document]
nakabonne: Right, that is what I was trying to say.

But if the SDK does it then it's fine to not check it anymore 👍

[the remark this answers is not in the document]
khanhtc1202: Thanks 👍

PR #2576, opened 2021-10-06
Prevent a panic by checking if notification is not nil

ono-max:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/controller/scheduler.go
+++ b/pkg/app/piped/controller/scheduler.go
@@ -662,6 +662,7 @@ func (s *scheduler) getMentionedAccounts(ctx context.Context, event model.Notifi
 	}
 
 	if ds.GenericDeploymentConfig.DeploymentNotification == nil {
+		// There is no event to mention users.
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/app/piped/controller/scheduler.go
```diff
@@ -661,6 +661,10 @@ func (s *scheduler) getMentionedAccounts(ctx context.Context, event model.Notifi
 		return nil, err
 	}
 
+	if ds.GenericDeploymentConfig.DeploymentNotification == nil {
+		return nil, nil
+	}
+
```
khanhtc1202, line 667: nits
- should be done before line L654 to avoid unnecessary targetDSP fetch.
- add log (using s.logger) to show why does this return with no error is better.

nakabonne: Or just having a comment works fine.

ono-max: Sorry, I can't understand the below one.

> should be done before line L654 to avoid unnecessary targetDSP fetch.

To check if `DeploymentNotification` is not nil, we need targetDSP fetch, right?

khanhtc1202: @ono-max  Ah, that was my missed 🙏

ono-max: 87ccd80

nakabonne:
Thanks for addressing!
/lgtm

nakabonne:
Why doesn't it get merged even though this build was completed first?

nakabonne:
well, what we can do now is just to wait.

PR #2595, opened 2021-10-07
Add an api to encrypt using registered public key

nakabonne:
This aims to make it possible to encrypt plaintext via pipectl.

Fixes #

```release-note
NONE
```


nakabonne:
Applied them!

PR #2606, opened 2021-10-08
Avoid panic while trigger deployment for kind that not k8s

khanhtc1202:

Follow PR #2581 

```release-note
NONE
```


nakabonne:
Good catch
/lgtm

nakabonne:
@ono-max you had better look over it just in case!

PR #2626, opened 2021-10-13
Fix the error message that a user is not included in approvers

ono-max:

Fixes #

```release-note
NONE
```


nakabonne:
@nghialv We're on the same page

PR #2665, opened 2021-10-20
Avoid stacked toast of the same reason

### LINKED ISSUE #2411: Make toast notifications non-stackable.
Currently toast notifications are stacked on top of each other, I would like them to not stack on top of each other and instead remove the previous and add a new one.

**Why is this needed**:
So I don't need to press +5 times when doing testing to get rid of error toast messages.

khanhtc1202:

Before

https://user-images.githubusercontent.com/32532742/138092609-4a40aac1-b4c3-4e03-b167-a8bfb8c4127c.mp4

After

https://user-images.githubusercontent.com/32532742/138092652-2d2ce35f-bdbd-4207-b1d6-ba7d580c562b.mp4

Due to the change in this PR, toast caused by the same reason (same issuer) and has the same message with the latest toast in the toast list will not be added to the toast list, that avoids annoying stacked toasts in case of errors occurred intervally.

Fixes #2411 

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/middlewares/thunk-error-handler.ts
+++ b/pkg/app/web/src/middlewares/thunk-error-handler.ts
@@ -12,7 +12,7 @@ export const thunkErrorHandler: Middleware = ({
   let res;
   try {
     res = await next(action);
-  } catch (err) {
+  } catch (err: any) {
     if (process.env.NODE_ENV === "development") {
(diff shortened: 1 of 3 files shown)


nakabonne:
It has driven me crazy. This is really well done!
/lgtm

PR #2673, opened 2021-10-22
Add ability to notify all events to users when an event field is empty

ono-max:

Fixes #

```release-note
Add ability to mention the specified Slack accounts for all events when the event name was not given
```


ono-max:
See https://github.com/pipe-cd/pipe/pull/2673#discussion_r734192014

PR #2712, opened 2021-10-28
Add a new Tag model

nakabonne:

Fixes #

```release-note
NONE
```


nakabonne:
We decided to use Labels instead of Tags.
Furthermore, we made sure to define Labels declaratively in the configuration file, so we don't need to complete them; we don't need to store them for ACLs either, so we don't need a Label model.

/close

PR #2751, opened 2021-11-05
Replace Tags with Labels to idendify applications/deployments

nakabonne:
This PR adds labels field to Applications/Deployments. And stop addingthem when Add/UpdateApplication.
It just makes changes on API side so no user-facing change will happen.

Ref: https://github.com/pipe-cd/pipe/issues/2750

```release-note
NONE
```


khanhtc1202:
🚀 
/approve

PR #2772, opened 2021-11-10
Add a piped component that watches app configs

nakabonne:
This PR adds a new Piped component that primarily watches two types of Applications that is:
- already registered
- unregistered

For the first one (already registered):
- Piped sends only configs that have been changed since then
- Contol-plane updates Applications stored in Datastore

For the second one (unregistered):
- Piped sends all the latest configs every time whatever the repo is changed or not. The reason is the cache in control-plane is likely to be removed, and configs in Git is also likely to be removed
- Control-plane caches them for suggesting when you attempt to add a new Application

Once this gets merged, I will drill into the detailed implementation.

Fixes https://github.com/pipe-cd/pipe/issues/2755
Ref https://github.com/pipe-cd/pipe/pull/2772
Ref https://github.com/pipe-cd/pipe/issues/2750

```release-note
NONE
```

### DIFF
--- a/pkg/app/api/service/pipedservice/service.proto
+++ b/pkg/app/api/service/pipedservice/service.proto
@@ -164,8 +164,8 @@ service PipedService {
 
     // UpdateApplicationConfigurations updates application configurations.
     rpc UpdateApplicationConfigurations(UpdateApplicationConfigurationsRequest) returns (UpdateApplicationConfigurationsResponse) {}
-    // PutLatestUnusedApplicationConfigurations puts the latest configurations of applications that isn't registered yet.
-    rpc PutUnregisteredApplicationConfigurations(PutUnregisteredApplicationConfigurationsRequest) returns (PutUnregisteredApplicationConfigurationsResponse) {}
+    // ReportLatestUnusedApplicationConfigurations puts the latest configurations of applications that isn't registered yet.
+    rpc ReportUnregisteredApplicationConfigurations(ReportUnregisteredApplicationConfigurationsRequest) returns (ReportUnregisteredApplicationConfigurationsResponse) {}
 }
 
 enum ListOrder {
@@ -449,28 +449,20 @@ message GetDesiredVersionResponse {
     string version = 1;
 }
 
-message ApplicationConfiguration {
-    // TODO: Make it possible to update application configs via Piped
-    //   To do so, first we need to enable application configs to be defined in Git repository
-    //string name = 1 [(validate.rules).string.min_len = 1];
-    //string kind = 2 [(validate.rules).string.min_len = 1];
-    //string env_id = 3 [(validate.rules).string.min_len = 1];
-    map<string, string> labels = 4;
-}
 
 message UpdateApplicationConfigurationsRequest {
     // The application configurations that should be updated.
-    repeated ApplicationConfiguration applications = 1;
+    repeated pipe.model.ApplicationInfo applications = 1;
 }
 
 message UpdateApplicationConfigurationsResponse {
 }
 
-message PutUnregisteredApplicationConfigurationsRequest {
+message ReportUnregisteredApplicationConfigurationsRequest {
     // All the latest application configurations that isn't registered yet.
     // Note that ALL configs are always be contained every time.
-    repeated ApplicationConfiguration applications = 1;
+    repeated pipe.model.ApplicationInfo applications = 1;
 }
 
-message PutUnregisteredApplicationConfigurationsResponse {
+message ReportUnregisteredApplicationConfigurationsResponse {
 }
--- a/pkg/app/api/service/pipedservice/service.proto
+++ b/pkg/app/api/service/pipedservice/service.proto
@@ -162,7 +162,10 @@ service PipedService {
     // GetDesiredVersion returns the desired version of the given Piped.
     rpc GetDesiredVersion(GetDesiredVersionRequest) returns (GetDesiredVersionResponse) {}
 
-    // TODO: Add an rpc to update application info based on one defined in the application config
+    // UpdateApplicationConfigurations updates application configurations.
+    rpc UpdateApplicationConfigurations(UpdateApplicationConfigurationsRequest) returns (UpdateApplicationConfigurationsResponse) {}
+    // PutLatestUnusedApplicationConfigurations puts the latest configurations of applications that isn't registered yet.
+    rpc PutUnregisteredApplicationConfigurations(PutUnregisteredApplicationConfigurationsRequest) returns (PutUnregisteredApplicationConfigurationsResponse) {}
 }
 
 enum ListOrder {
@@ -444,4 +447,30 @@ message GetDesiredVersionRequest {
 
 message GetDesiredVersionResponse {
     string version = 1;
+}
+
+message ApplicationConfiguration {
+    // TODO: Make it possible to update application configs via Piped
+    //   To do so, first we need to enable application configs to be defined in Git repository
+    //string name = 1 [(validate.rules).string.min_len = 1];
+    //string kind = 2 [(validate.rules).string.min_len = 1];
+    //string env_id = 3 [(validate.rules).string.min_len = 1];
+    map<string, string> labels = 4;
+}
+
+message UpdateApplicationConfigurationsRequest {
+    // The application configurations that should be updated.
+    repeated ApplicationConfiguration applications = 1;
+}
+
+message UpdateApplicationConfigurationsResponse {
+}
+
+message PutUnregisteredApplicationConfigurationsRequest {
+    // All the latest application configurations that isn't registered yet.
+    // Note that ALL configs are always be contained every time.
+    repeated ApplicationConfiguration applications = 1;
+}
+
+message PutUnregisteredApplicationConfigurationsResponse {
 }
--- a/pkg/app/piped/appconfigreporter/appconfigreporter.go
+++ b/pkg/app/piped/appconfigreporter/appconfigreporter.go
@@ -1,0 +1,220 @@
+// Copyright 2021 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package appconfigreporter
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+
+	"github.com/pipe-cd/pipe/pkg/app/api/service/pipedservice"
+	"github.com/pipe-cd/pipe/pkg/cache"
+	"github.com/pipe-cd/pipe/pkg/cache/memorycache"
+	"github.com/pipe-cd/pipe/pkg/config"
+	"github.com/pipe-cd/pipe/pkg/git"
+	"github.com/pipe-cd/pipe/pkg/model"
+)
+
+const (
+	defaultLastFetchedCommitCacheSize = 500
+)
+
+type apiClient interface {
+	UpdateApplicationConfigurations(ctx context.Context, in *pipedservice.UpdateApplicationConfigurationsRequest, opts ...grpc.CallOption) (*pipedservice.UpdateApplicationConfigurationsResponse, error)
+	PutUnregisteredApplicationConfigurations(ctx context.Context, in *pipedservice.PutUnregisteredApplicationConfigurationsRequest, opts ...grpc.CallOption) (*pipedservice.PutUnregisteredApplicationConfigurationsResponse, error)
+}
+
+type gitClient interface {
+	Clone(ctx context.Context, repoID, remote, branch, destination string) (git.Repo, error)
+}
+
+type applicationLister interface {
+	List() []*model.Application
+}
+
+type Reporter struct {
+	apiClient         apiClient
+	gitClient         gitClient
+	applicationLister applicationLister
+	config            *config.PipedSpec
+	gitRepos          map[string]git.Repo
+	gracePeriod       time.Duration
+	// Cache for the last fetched commit for each repository.
+	lastFetchedCommitCache cache.Cache
+	logger                 *zap.Logger
+}
+
+func NewReporter(
+	apiClient apiClient,
+	gitClient gitClient,
+	appLister applicationLister,
+	cfg *config.PipedSpec,
+	gracePeriod time.Duration,
+	logger *zap.Logger,
+) (*Reporter, error) {
+	cache, err := memorycache.NewLRUCache(defaultLastFetchedCommitCacheSize)
+	if err != nil {
+		return nil, err
+	}
+	return &Reporter{
+		apiClient:              apiClient,
+		gitClient:              gitClient,
+		applicationLister:      appLister,
+		config:                 cfg,
+		gracePeriod:            gracePeriod,
+		lastFetchedCommitCache: cache,
+		logger:                 logger.Named("app-config-reporter"),
+	}, nil
+}
+
+func (r *Reporter) Run(ctx context.Context) error {
+	r.logger.Info("start running app-config-reporter")
+
+	// Pre-clone to cache the registered git repositories.
+	r.gitRepos = make(map[string]git.Repo, len(r.config.Repositories))
+	for _, repoCfg := range r.config.Repositories {
+		repo, err := r.gitClient.Clone(ctx, repoCfg.RepoID, repoCfg.Remote, repoCfg.Branch, "")
+		if err != nil {
+			r.logger.Error("failed to clone repository",
+				zap.String("repo-id", repoCfg.RepoID),
+				zap.Error(err),
+			)
+			return err
+		}
+		r.gitRepos[repoCfg.RepoID] = repo
+	}
+
+	// FIXME: Think about sync interval of app config reporter
+	ticker := time.NewTicker(r.config.SyncInterval.Duration())
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := r.checkApps(ctx); err != nil {
+				r.logger.Error("failed to check application configurations defined in Git", zap.Error(err))
+			}
+		case <-ctx.Done():
+			r.logger.Info("app-config-reporter has been stopped")
+			return nil
+		}
+	}
+}
+
+// checkApps checks and reports two types of applications.
+// One is applications registered in Control-plane already, and another is ones that aren't registered yet.
+func (r *Reporter) checkApps(ctx context.Context) (err error) {
+	if len(r.gitRepos) == 0 {
+		r.logger.Info("no repositories were configured for this piped")
+		return
+	}
+
+	var (
+		unusedApps      = make([]*pipedservice.ApplicationConfiguration, 0)
+		appsToBeUpdated = make([]*pipedservice.ApplicationConfiguration, 0)
+		appsMap         = r.listApplications()
+	)
+	for repoID, repo := range r.gitRepos {
+		if err = repo.Pull(ctx, repo.GetClonedBranch()); err != nil {
+			r.logger.Error("failed to update repo to latest",
+				zap.String("repo-id", repoID),
+				zap.Error(err),
+			)
+			return
+		}
+
+		// TODO: Collect unused application configurations that aren't used yet
+		//   Currently, it could be thought the best to open files that suffixed by .pipe.yaml
+
+		var headCommit git.Commit
+		// Get the head commit of the repository.
+		headCommit, err = repo.GetLatestCommit(ctx)
+		if err != nil {
+			return
+		}
+		var lastFetchedCommit interface{}
+		lastFetchedCommit, err = r.lastFetchedCommitCache.Get(repoID)
+		if err != nil && !errors.Is(err, cache.ErrNotFound) {
+			r.logger.Error("failed to get the last fetched commit from cache", zap.Error(err))
+		}
+		if headCommit.Hash == lastFetchedCommit.(string) {
+			continue
+		}
+		apps, ok := appsMap[repoID]
+		if !ok {
+			continue
+		}
+		for _, app := range apps {
+			gitPath := app.GetGitPath()
+			_ = filepath.Join(repo.GetPath(), gitPath.Path, gitPath.ConfigFilename)
+			// TODO: Collect applications that need to be updated
+		}
+
+		defer func() {
+			if err == nil {
+				r.lastFetchedCommitCache.Put(repoID, headCommit)
+			}
+		}()
+	}
+	if len(unusedApps) > 0 {
+		_, err = r.apiClient.PutUnregisteredApplicationConfigurations(
+			ctx,
+			&pipedservice.PutUnregisteredApplicationConfigurationsRequest{
+				Applications: unusedApps,
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("failed to put the latest unregistered application configurations: %w", err)
+		}
+	}
+
+	if len(appsToBeUpdated) == 0 {
+		return nil
+	}
+	_, err = r.apiClient.UpdateApplicationConfigurations(
+		ctx,
+		&pipedservice.UpdateApplicationConfigurationsRequest{
+			Applications: appsToBeUpdated,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to update application configurations: %w", err)
+	}
+
+	return
+}
+
+// listApplications retrieves all applications that should be handled by this piped
+// and then groups them by repoID.
+func (r *Reporter) listApplications() map[string][]*model.Application {
+	var (
+		apps       = r.applicationLister.List()
+		repoToApps = make(map[string][]*model.Application)
+	)
+	for _, app := range apps {
+		repoId := app.GitPath.Repo.Id
+		if _, ok := repoToApps[repoId]; !ok {
+			repoToApps[repoId] = []*model.Application{app}
+		} else {
+			repoToApps[repoId] = append(repoToApps[repoId], app)
+		}
+	}
+	return repoToApps
+}
--- a/pkg/app/piped/appconfigreporter/appconfigreporter.go
+++ b/pkg/app/piped/appconfigreporter/appconfigreporter.go
@@ -17,7 +17,9 @@ package appconfigreporter
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"go.uber.org/zap"
@@ -98,7 +100,7 @@ func (r *Reporter) Run(ctx context.Context) error {
 	for {
 		select {
 		case <-ticker.C:
-			if err := r.checkApps(ctx); err != nil {
+			if err := r.scanAppConfigs(ctx); err != nil {
 				r.logger.Error("failed to check application configurations defined in Git", zap.Error(err))
 			}
 		case <-ctx.Done():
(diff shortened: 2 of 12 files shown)


pkg/app/api/service/pipedservice/service.proto
```diff
@@ -445,3 +448,29 @@ message GetDesiredVersionRequest {
...
 }
+
+message ApplicationConfiguration {
+    // TODO: Make it possible to update application configs via Piped
+    //   To do so, first we need to enable application configs to be defined in Git repository
+    //string name = 1 [(validate.rules).string.min_len = 1];
+    //string kind = 2 [(validate.rules).string.min_len = 1];
+    //string env_id = 3 [(validate.rules).string.min_len = 1];
+    map<string, string> labels = 4;
+}
```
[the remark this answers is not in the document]
nghialv: And that message should contain the path of application directory and config filename as well.

pkg/app/api/service/pipedservice/service.proto
```diff
@@ -162,7 +162,10 @@ service PipedService {
     // GetDesiredVersion returns the desired version of the given Piped.
     rpc GetDesiredVersion(GetDesiredVersionRequest) returns (GetDesiredVersionResponse) {}
 
-    // TODO: Add an rpc to update application info based on one defined in the application config
+    // UpdateApplicationConfigurations updates application configurations.
+    rpc UpdateApplicationConfigurations(UpdateApplicationConfigurationsRequest) returns (UpdateApplicationConfigurationsResponse) {}
+    // PutLatestUnusedApplicationConfigurations puts the latest configurations of applications that isn't registered yet.
+    rpc PutUnregisteredApplicationConfigurations(PutUnregisteredApplicationConfigurationsRequest) returns (PutUnregisteredApplicationConfigurationsResponse) {}
```
nghialv, line 168: nit: `Put` -> `Report`

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,220 @@
...
+
+type Reporter struct {
+	apiClient         apiClient
+	gitClient         gitClient
+	applicationLister applicationLister
+	config            *config.PipedSpec
+	gitRepos          map[string]git.Repo
+	gracePeriod       time.Duration
+	// Cache for the last fetched commit for each repository.
+	lastFetchedCommitCache cache.Cache
```
nghialv, line 60: This is just for marking the hash value of the previously scanned commit, so to be simple we can use a normal map.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,220 @@
...
+		r.logger.Info("no repositories were configured for this piped")
+		return
+	}
+
+	var (
+		unusedApps      = make([]*pipedservice.ApplicationConfiguration, 0)
+		appsToBeUpdated = make([]*pipedservice.ApplicationConfiguration, 0)
+		appsMap         = r.listApplications()
+	)
+	for repoID, repo := range r.gitRepos {
```
nghialv, line 134: We can break this function to be simpler ones. For example

``` golang
func scan(cox context.Context) error 
  for _, repo := range r.gitRepos {
     // 1. Fetch to update the local data of each repo to ensure they are up-to-date.
     // 2. Continue if there was no newly added commits.
     // 3. shouldScanRepos = append(shouldScanRepos, repo)
  }

  for _, repo := range shouldScanRepos {
     if err := scanRepo(cox context.Context, repo, previousHash, currentHash); err == nil {
        // mark the new hash to the map
     }
  }
}

func scanRepo(ctx context.Context, repo Repo, previousHash, currentHash string) error {
   ...
}
```

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		apps, ok := appsMap[repoID]
+		if !ok {
+			continue
+		}
+		for _, app := range apps {
+			gitPath := app.GetGitPath()
+			_ = filepath.Join(repo.GetPath(), gitPath.Path, gitPath.ConfigFilename)
+			// TODO: Collect applications that need to be updated
+		}
```
nghialv, line 168: Instead of iterate apps to check like this, we can use `git diff previousHash currentHash --name-only` to have a list of changed files. And then find the updated ones and un-registered ones based on that list.

pkg/app/api/service/pipedservice/service.proto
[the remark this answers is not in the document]
nakabonne: For sure

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Right

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Obviously more readable than before

nakabonne: But to be honest, since we need to send all unregisteredApps each time, we have to scan all repositories anyway.

nakabonne: >Piped sends all the latest configs every time whatever the repo is changed or not. The reason is the cache in control-plane is likely to be removed, and configs in Git is also likely to be removed

As mentioned in the PR description, I'm looking to letting Piped send all the latest configs every time whatever the repo is changed or not. The reasons are:
- There are no new commits to the repository, but the Control-plane cache may have already been removed.
- If you send only the configuration to be updated, you can't address the case that the file itself is deleted from the repo.

pkg/app/api/service/pipedservice/service.proto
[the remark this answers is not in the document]
nakabonne: One thing that I'm worried about is duplication. The fields of the Application model are partly duplicated with this model.
Hopefully we can embed this to the Application model, but we can't due to model consistency.

pkg/config/deployment.go
```diff
@@ -29,6 +29,7 @@ const (
 )
 
 type GenericDeploymentSpec struct {
+	model.ApplicationInfo `json:",inline"`
```
nghialv, line 32: This will also include unnecessary files such as `kind` `config_filename`. So maybe we should define them explicitly here.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,209 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/pipe-cd/pipe/pkg/app/api/service/pipedservice"
+	"github.com/pipe-cd/pipe/pkg/config"
+	"github.com/pipe-cd/pipe/pkg/git"
+	"github.com/pipe-cd/pipe/pkg/model"
+)
+
+const (
+	defaultLastFetchedCommitCacheSize = 500
```
nghialv, line 33: This is no longer needed.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,209 @@
...
+
+type Reporter struct {
+	apiClient         apiClient
+	gitClient         gitClient
+	applicationLister applicationLister
+	config            *config.PipedSpec
+	gitRepos          map[string]git.Repo
+	gracePeriod       time.Duration
+	// Cache for the last fetched commit for each repository. Not goroutine safe.
+	lastFetchedCommitCache map[string]string
```
nghialv, line 57: nit: `lastScannedCommits`

pkg/app/api/service/pipedservice/service.proto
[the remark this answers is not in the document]
nakabonne: I settled on moving it to the model package and renameit , and embed it to the ApplicationConfiig on https://github.com/pipe-cd/pipe/pull/2772/commits/390aa68e4fe85faa863180b513f47880f7c5e556

`ApplicationConfiguration` means the entire config including others like pipeline, so I choose the name `ApplicationInfo`, I'm still debating though

pkg/app/piped/appconfigreporter/appconfigreporter.go
nghialv: I get your points.

Checking by the diff still be valid in those cases. Scanning the full repository can be done for only the first time and then just the diff from the second time. Of course, we should send the full list of "un-registered" ones and the changed "registered" ones to the control-plane.

About the deleted files I think we can ignore them since app deletion only be allowed from web.

pkg/config/deployment.go
nakabonne: for sure

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Good catch!

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Ah, you mean to separate updating the repository from reading the files. Cool, I have no objection.

>About the deleted files I think we can ignore them since app deletion only be allowed from web.

Sorry for confusing you. I was afraid that even though the files of unregistered applications were deleted, they were still in the cache and would be suggested when registering an application.

nghialv: >  I was afraid that even though the files of unregistered applications were deleted, they were still in the cache and would be suggested when registering an application.

After having the list of changed files, we still have to read their content to get app_name, kind... So we can know that case at parsing time to handle as normal.

nakabonne: Sorry, my words were possibly not enough. I was talking about unregistered applications.
Could you tell me about your thoughts on always sending ALL unregistered apps each time?

nghialv: > Could you tell me about your thoughts on always sending ALL unregistered apps each time?

Yes, I think that is ok, we have to send the full list of unregistered apps because they are saved in the cache.
We are on the same page about that topic.

pkg/model/common.go
```diff
@@ -22,3 +24,8 @@ func ApplicationKindStrings() []string {
 	}
 	return out
 }
+
+// BuildGitPathID builds a unique path between repositories.
+func BuildGitPathID(repoID, path, configFileName string) string {
```
nakabonne, line 29: I just want to look into if the application is registered, using GitPath. I'm still wondering if we can do that in other ways.

nakabonne:
@nghialv I am very grateful for your careful confirmation. Based on your suggestion, I replaced the part that used to be the TODO comment with the implementation that actually updates it.

It's not exactly what you suggested, but I would say it's more readable than before. Please take a look at it when you have time.

nghialv:
> It's not exactly what you suggested, but I would say it's more readable than before. Please take a look at it when you have time.

Sure. Let me check it.

pkg/model/application.go
```diff
@@ -20,7 +20,11 @@ import (
 	"strings"
 )
 
-const DefaultDeploymentConfigFileName = ".pipe.yaml"
+const (
+	// TODO: Consider changing the default application config name
+	DefaultDeploymentConfigFileName      = ".pipe.yaml"
+	DefaultDeploymentConfigFileExtension = ".pipe.yaml"
```
nghialv, line 26:
```suggestion
	DefaultDeploymentConfigFileExtension = ".pipecd.yaml"
```

pkg/config/deployment.go
```diff
@@ -29,6 +29,10 @@ const (
 )
 
 type GenericDeploymentSpec struct {
+	Name   string            `json:"name"`
+	EnvID  string            `json:"env_id"`
```
nghialv, line 33:
```suggestion
	EnvID  string            `json:"envId"`
```

nghialv: Btw, isn't the name better than the id?

nghialv:
Nice. It became more readable.
I have just left some nits but I am still reviewing this pull request.
And let me rewrite my suggestion this evening.

pkg/config/deployment.go
nakabonne: I was worried about the uniqueness of env_name but let's leave it to the user to maintain the uniqueness.

It's obvious that the application name shouldn't be unique, but ideally the env name should be unique because Environment is a project-wide concept.
To do that with a document-oriented database, we need to prepare some kind of extra table to ensure uniqueness.

But we're planning to migrate Env to Labels in the future, again, for now let's leave it to the user to maintain the uniqueness.

nakabonne: If we proceed with the Name as non-unique, we will have to add a not neat method to `environmentstore`. I'd like to hear your opinion on uniqueness before implementing it.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,280 @@
...
+			Applications: apps,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to update application configurations: %w", err)
+	}
+	return
+}
+
+func (r *Reporter) readApplicationInfo(repoID, path, cfgFilename string, registeredAppPaths map[string]struct{}, wantRegistered bool) (appInfo *model.ApplicationInfo, skip bool) {
```
nghialv, line 242: To be simple, `registeredAppPaths map[string]struct{}, wantRegistered bool` should not be a part of this function.

nghialv:
I have just left some comments. 👍 
Please check out them and I will take another look at this PR later once they got resolved.

pkg/config/deployment.go
```diff
@@ -29,6 +29,10 @@ const (
 )
 
 type GenericDeploymentSpec struct {
+	Name   string            `json:"name"`
+	EnvID  string            `json:"envId"`
```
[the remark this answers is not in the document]
nakabonne: Sorry, I was worried about is that an environment name isn't unique.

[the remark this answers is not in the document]
nakabonne: It is possible to have multiple environments with the same name within a project, and I meant to say that in order to guarantee uniqueness of fields other than ID in Firestore, we need to have another document that stores only the environment name.
However, I think it is sufficient to leave that responsibility to the user for now.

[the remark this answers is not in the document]
nghialv: > However, I think it is sufficient to leave that responsibility to the user for now.

Yes. I think that is fine since the environment is going to be just a normal label eventually.
We will remove all environment requirements from Piped adding form, application adding/editing form.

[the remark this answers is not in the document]
nakabonne: Totally!

nakabonne:
@nghialv Thank you for your reviewing. I just applied your suggestions:
- Separating the finding part and the reporting part to have some tests for finding logic.
- Use env_name instead of env_id

I settled on using `fs.FS` so that we can test for the `finding` part using the file system in memory.

nghialv:
Okay, let me check it now.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+		id := model.BuildGitPathID(app.GitPath.Repo.Id, app.GitPath.Path, app.GitPath.ConfigFilename)
+		registeredAppPaths[id] = struct{}{}
+	}
+
+	if err := r.updateUnregisteredApps(ctx, registeredAppPaths); err != nil {
+		return fmt.Errorf("failed to update unregistered applications: %w", err)
+	}
+	if err := r.updateRegisteredApps(ctx, registeredAppPaths); err != nil {
+		return fmt.Errorf("failed to update registered applications: %w", err)
+	}
```
nghialv, line 156: Registered applications are more important than unregistered ones so let's change the order of updating.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// updateUnregisteredApps sends all unregistered application configurations to the control-plane.
+func (r *Reporter) updateUnregisteredApps(ctx context.Context, registeredAppPaths map[string]struct{}) error {
+	apps := make([]*model.ApplicationInfo, 0)
+	for repoID, repo := range r.gitRepos {
+		as, err := r.findUnregisteredApps(ctx, repo.GetPath(), repoID, registeredAppPaths)
+		if err != nil {
+			return err
+		}
```
nghialv, line 168: nit: Add a log message that could help for debugging by our users.

`found out %d unregistered applications in repository %s`

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+
+		appInfo, err := r.readApplicationInfo(ctx, repoPath, path)
+		if err != nil {
+			r.logger.Error("failed to read application info",
+				zap.String("repo-id", repoID),
+				zap.String("config-file-path", path),
+				zap.Error(err),
+			)
+			return nil
```
nghialv, line 209: Returning nil because we want to return the found list as many as possible, right?
So let's add a comment to tell why we are returning nil instead of the received error.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+		if err != nil {
+			return err
+		}
+		apps = append(apps, as...)
+		id := repoID
+		defer func() {
+			if err == nil {
+				r.lastScannedCommits[id] = headCommit.Hash
+			}
+		}()
```
nghialv, line 240: I think to be simple we should avoid `defer` in the loop.
So instead of that, we can mark the commit list in a variable and save them in a single `defer`.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+	ChangedFiles(ctx context.Context, from, to string) ([]string, error)
+}
+
+// findRegisteredApps finds out registered application info in the given git repository.
+func (r *Reporter) findRegisteredApps(ctx context.Context, repoID string, repo gitRepo, headCommitHash string, registeredAppPaths map[string]struct{}) ([]*model.ApplicationInfo, error) {
+	// Skip if the head commit is already scanned.
+	lastScannedCommit, ok := r.lastScannedCommits[repoID]
+	if ok && headCommitHash == lastScannedCommit {
+		return []*model.ApplicationInfo{}, nil
+	}
```
nghialv, line 269: Moving this part outside of this function (to be at caller) to avoid this function from depending on `lastScannedCommits` cache. Tests could become easier.
I mean that retrieving from and saving to that cache should not be part of this function.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+				zap.Error(err),
+			)
+			continue
+		}
+		apps = append(apps, appInfo)
+	}
+	return apps, nil
+}
+
+func shouldSkip(repoID, path, cfgFilename string, registeredAppPaths map[string]struct{}, wantRegistered bool) bool {
```
nghialv, line 298: Does `cfgFilename` contain the relative path or just a name?

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+	return &Reporter{
+		apiClient:          apiClient,
+		gitClient:          gitClient,
+		applicationLister:  appLister,
+		envGetter:          envGetter,
+		config:             cfg,
+		gracePeriod:        gracePeriod,
+		lastScannedCommits: make(map[string]string),
+		// FIXME: Find another way to specify the root dir for a file system
+		fsys:   os.DirFS("/"),
```
nghialv, line 85: Initializing this with `/` is not a best practice so I think we should avoid this.

How about defining our own FS like

```
type fileSystem struct {}

func (s *fileSystem) Open(name string) (File, error) {
   return os.Open(name)
}

```

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,337 @@
...
+	gitClient         gitClient
+	applicationLister applicationLister
+	envGetter         environmentGetter
+	config            *config.PipedSpec
+	gitRepos          map[string]git.Repo
+	gracePeriod       time.Duration
+	// Cache for the last scanned commit for each repository.
+	// Not goroutine safe.
+	lastScannedCommits map[string]string
+	fsys               fs.FS
```
nghialv, line 63:
```suggestion
	fileReader               fs.FS
```

pkg/model/common.go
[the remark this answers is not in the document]
nakabonne: Yeah, totally agree with you!

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: For sure 👍

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Nice. It is definite that currently it's quite hard to see what's going on.

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: There you go.

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Cool. It will reduce the amount of code.

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Or, filieSystem?

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: For sure

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Nice catch. I found some issues around this function. The current implementation is unable to keep the application path unique.

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Love it! Let me apply that simpler one.

nakabonne:
@nghialv Applied everything you suggested! PTAL!

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,351 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to update application configurations: %w", err)
+	}
+	return
+}
+
+// findRegisteredApps finds out registered application info in the given git repository.
+func (r *Reporter) findRegisteredApps(ctx context.Context, repoID string, repo gitRepo, lastScannedCommit, headCommitHash string, registeredAppPaths map[string]struct{}) ([]*model.ApplicationInfo, error) {
+	files, err := repo.ChangedFiles(ctx, lastScannedCommit, headCommitHash)
```
nghialv, line 217: What happens when Piped has just been started and this `lastScannedCommit` was empty?

nakabonne: Good catch! I assumed it will be like `git diff --name-only  <headCommitHash>` when it was empty. But Go's exec package doesn't depend on any shell, and either way, that is incorrect.
Let me fix it.

nghialv: 👍

nakabonne:
@nghialv I addressed the case you were worried about and revised some issues on file traversing. PTAL when you have time.
Again, a bit thank you goes out to you!

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,366 @@
...
+func shouldSkip(repoID, cfgFilePath string, registeredAppPaths map[string]struct{}, wantRegistered bool) bool {
+	if !strings.HasSuffix(cfgFilePath, model.DefaultDeploymentConfigFileExtension) {
+		return true
+	}
+	gitPathKey := makeGitPathKey(repoID, cfgFilePath)
+	if _, registered := registeredAppPaths[gitPathKey]; registered != wantRegistered {
+		return true
+	}
+	return false
+}
```
nghialv, line 330: How about removing this function and using its code directly?
Because handling for both "registered" and "unregistered" cases makes it harder to read.

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,366 @@
...
+	for repoID, repo := range r.gitRepos {
+		as, err := r.findUnregisteredApps(ctx, repo.GetPath(), repoID, registeredAppPaths)
+		if err != nil {
+			return err
+		}
+		r.logger.Info(fmt.Sprintf("found out %d unregistered applications in repository %s", len(as), repoID))
+		apps = append(apps, as...)
+	}
+	if len(apps) == 0 {
+		return nil
```
nghialv, line 260: Do you think that we should send this empty list to the control plane to delete the cache?

nghialv: Or we can have a simple mark in this Reporter to avoid sending continuous this kind of request.

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Right. It should not be made common.

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: To avoid sending meaningless requests simply, I applied this way. Cache in Control-plane gets periodically removed, so I thought it's fine. But it's worth considering.

nakabonne:
@nghialv Revised!

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,371 @@
...
+	gitRepos          map[string]git.Repo
+	gracePeriod       time.Duration
+	// Cache for the last scanned commit for each repository.
+	// Not goroutine safe.
+	lastScannedCommits map[string]string
+	fileSystem         fs.FS
+	logger             *zap.Logger
+
+	// Whether it already swept all unregistered apps from control-plane.
+	// Not goroutine safe.
```
nghialv, line 76: Do we need this kind of comment since we are not spawning any new goroutines in this package?

pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+
+		cfgRelPath, err := filepath.Rel(repoRoot, path)
+		if err != nil {
+			return err
+		}
+
+		if !strings.HasSuffix(cfgRelPath, model.DefaultDeploymentConfigFileExtension) {
+			return nil
+		}
```
nghialv, line 312: nit: We can move this to before line 305 and use `path` to check the extension. It will reduce a lot of unnecessary `filepath.Rel` calls.

nghialv (commented):
Finished my review. Just some nits. 👍

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: I just left these comments for when we need to spawn multiple goroutines for performance improvement in the future. But it's completely YAGNI 👍

pkg/app/piped/appconfigreporter/appconfigreporter.go
nakabonne: Good catch

nakabonne:
@nghialv Fixed

nghialv:
Looks good! Go ahead.
/lgtm

nakabonne:
I just added a TODO comment

khanhtc1202:
Way to go 🚀 
/approve

PR #2798, opened 2021-11-18
Add mentions to trigger deployment failed notification

khanhtc1202:

Fixes #

```release-note
NONE
```


ono-max:
Wow, nice idea 👍 
/lgtm

nakabonne:
Nice. Do we need to address the case of `NotificationEventDeploymentRollingBack` as well?

nakabonne:
In either way, it is able to fall into another PR.

khanhtc1202:
> Nice. Do we need to address the case of `NotificationEventDeploymentRollingBack` as well?

Not yet, since we do not return that event anywhere. Should make a separated PR for that (include where we will notify that event to slack) cc @ono-max

nakabonne:
Alright 👍 
/approve

PR #2800, opened 2021-11-18
Move TriggeredDeploymentID metadata key to command model

khanhtc1202:

`triggeredDeploymentIDKey` used in both trigger(piped) and pipectl but must be defined to use could lead to unexpected error in case the key name changed in one end. This PR moves the key to the Command model (since it is the key in command metadata) and explicitly refers it in the used end.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/pipectl/client/application.go
+++ b/pkg/app/pipectl/client/application.go
@@ -51,8 +51,6 @@ func SyncApplication(
 	defer ticker.Stop()
 
 	check := func() (deploymentID string, shouldRetry bool) {
-		const triggeredDeploymentIDKey = "TriggeredDeploymentID"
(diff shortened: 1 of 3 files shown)


nakabonne:
Even if the size of PR is tiny, It would help us to understand what context is behind that if you could write a short note in the PR body. Just a little bit is enough 🙏

khanhtc1202:
> Even if the size of PR is tiny, It would help us to understand what context is behind that if you could write a short note in the PR body. Just a little bit is enough 🙏

Sure, updated the PR description 🙏

nakabonne:
Thank you. Looks neat!
/lgtm

PR #2801, opened 2021-11-18
Add trigger application grpc to piped api

khanhtc1202:

To implement trigger strategy in the deployment chain, this gRPC is needed. The plan is
- Piped1 triggers deployment for application1
- Piped1 load the application1 configuration, passes the postSync configuration section then use this gRPC to make a sync_application command for other applications.
- Pipedx (in-charge of applications other than application1) listens to the command and handles it as usual.

Thing to consider:
I created a new command type named `IN_CHAIN_SYNC_APPLICATION` in this PR since this command should be handled in a different way than the other commands (with OnChainDeterminer instead of OnCommandDeterminer). About the command name, `InChainTriggerApplication` and `InChainSyncApplication` which one do you guys prefer?

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/api/grpcapi/piped_api.go
+++ b/pkg/app/api/grpcapi/piped_api.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/google/uuid"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


khanhtc1202:
We're not going this way 👋 
/close

PR #2812, opened 2021-11-19
Make EnvId in all models optional

nakabonne:
This PR makes all EnvID(s) optional since we're going to migrate Environment to Labels.
I looked up places where use EnvId and confirmed all seem to be able to work well without EnvId.

Close https://github.com/pipe-cd/pipe/issues/2805

```release-note
NONE
```


ono-max:
There you go.

/approve

PR #2819, opened 2021-11-19
Change to treat .pipe.yaml as unregistered app config

nakabonne:
I feel like it's better to use the default file name (`.pipe.yaml`) as well as the file extension (`.pipecd.yaml`) as a collection condition for UnregisteredApps.

Currently, ConfigFilename in the Application model is optional and is [considered to be ".pipe.yaml" if empty](https://github.com/pipe-cd/pipe/blob/d234a79e15f188c4b30d2917e4fe74bffaad9875/pkg/model/application.go#L28-L34). So changing the default ConfigFilename is a bit complicated.

It is confusing to recommend using the extension `.pipecd.yaml` without changing the default from `.pipe.yaml`, so I would say `pipe.yaml` should also be treated as a collection condition for UnregisteredApps.

~I don't think there are many people who would be happy with being able to freely decide the name of the application config file, so I would say it would be a good idea to fix the name from now on. Alternatively, you can also decide that a .pipe.yaml at the end is all you need, like: `if strings.HasSuffix(path, ".pipe.yaml")`~

This PR also removed the check the file extension for RegisteredApps, since it can be determined without checking them.

Fixes https://github.com/pipe-cd/pipe/issues/2808

```release-note
NONE
```

### DIFF
--- a/pkg/app/pipectl/cmd/application/add.go
+++ b/pkg/app/pipectl/cmd/application/add.go
@@ -43,7 +43,7 @@ type add struct {
 func newAddCommand(root *command) *cobra.Command {
 	c := &add{
 		root:           root,
-		configFileName: model.DefaultDeploymentConfigFileName,
+		configFileName: model.DefaultApplicationConfigFilename,
 	}
 	cmd := &cobra.Command{
 		Use:   "add",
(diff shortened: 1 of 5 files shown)


ono-max:
Do we plan to update the doc(`DeploymentConfig` -> `ApplicationConfig`), too?

nakabonne:
I updated to allow both `.pipe.yaml` and `*.pipecd.yaml` and modified the PR description.

nakabonne:
@ono-max 
>Do we plan to update the doc(DeploymentConfig -> ApplicationConfig), too?

Yes. The two are mixed up right now, but we are going to replace them all at some point.

PR #2826, opened 2021-11-22
Update CreateDeployment RPC to handle create in chain deployment

khanhtc1202:

For deployment which be created to sync application in chain, we should add the ref to it in the chain model it belongs to, otherwise is keep as is for standalone deployment (deployment that does not belong to any chain called standalone deployment)

Follow PR #2815 

```release-note
NONE
```


nakabonne:
found typo in PR title: `Update CreateDeployment RCP` -> `Update CreateDeployment RPC`

nakabonne:
makes sense
/lgtm

PR #2829, opened 2021-11-24
Enable to convert config kind into model.ApplicationKind

nakabonne:
In order to get to know the kind of Application config defined in Git without any additional field, we need kind of like this a bit dirty method.

Fixes https://github.com/pipe-cd/pipe/issues/2807

```release-note
NONE
```


khanhtc1202:
Nice
/lgtm

PR #2833, opened 2021-11-24
Empower Piped to change how often to sync app config in Git

nakabonne:
Since an application config file is updated much less frequently than the deployment frequency, the settings should be separated.

Fixes https://github.com/pipe-cd/pipe/issues/2809

```release-note
NONE
```


khanhtc1202:
Something big has come 😌 
/lgtm

ono-max (approved):

PR #2841, opened 2021-11-25
Fix small part in state.go

ono-max:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/cloudprovider/kubernetes/state.go
+++ b/pkg/app/piped/cloudprovider/kubernetes/state.go
@@ -351,7 +351,6 @@ func determineJobHealth(obj *unstructured.Unstructured) (status model.Kubernetes
 			failed = true
 			completed = true
 			message = condition.Message
-			break
 		case batchv1.JobComplete:
 			completed = true
 			message = condition.Message
--- a/pkg/app/piped/cloudprovider/kubernetes/state.go
+++ b/pkg/app/piped/cloudprovider/kubernetes/state.go
@@ -472,6 +472,7 @@ func determineServiceHealth(obj *unstructured.Unstructured) (status model.Kubern
 	if len(s.Status.LoadBalancer.Ingress) == 0 {
 		status = model.KubernetesResourceState_OTHER
 		desc = "Ingress points for the load-balancer are in progress"
+		return
 	}
 	return
 }
--- a/pkg/app/piped/cloudprovider/kubernetes/state.go
+++ b/pkg/app/piped/cloudprovider/kubernetes/state.go
@@ -356,6 +356,9 @@ func determineJobHealth(obj *unstructured.Unstructured) (status model.Kubernetes
 			completed = true
 			message = condition.Message
 		}
+		if failed {
+			break
+		}
 	}
 
 	switch {
--- a/pkg/app/piped/cloudprovider/kubernetes/state.go
+++ b/pkg/app/piped/cloudprovider/kubernetes/state.go
@@ -351,9 +351,11 @@ func determineJobHealth(obj *unstructured.Unstructured) (status model.Kubernetes
 			failed = true
 			completed = true
 			message = condition.Message
+			break
 		case batchv1.JobComplete:
 			completed = true
 			message = condition.Message
+			break
 		}
 	}
 
@@ -450,7 +452,6 @@ func determineIngressHealth(obj *unstructured.Unstructured) (status model.Kubern
 	status = model.KubernetesResourceState_OTHER
 	desc = fmt.Sprintf("Unexpected error while calculating: unable to convert %T to neither %T, %T, nor %T: %v", obj, v1Ingress, betaIngress, extensionIngress, err)
 	return
-
 }
 
 func determineServiceHealth(obj *unstructured.Unstructured) (status model.KubernetesResourceState_HealthStatus, desc string) {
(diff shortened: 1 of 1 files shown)


pkg/app/piped/cloudprovider/kubernetes/state.go
```diff
@@ -469,7 +470,6 @@ func determineServiceHealth(obj *unstructured.Unstructured) (status model.Kubern
 	if len(s.Status.LoadBalancer.Ingress) == 0 {
 		status = model.KubernetesResourceState_OTHER
 		desc = "Ingress points for the load-balancer are in progress"
-		return
```
nakabonne, line 472: You're correct. In the current implementation, it's definitely an unneeded statement. But it explicitly shows us it must return here even if the additional check is appended following this in the future.

pkg/app/piped/cloudprovider/kubernetes/state.go
```diff
@@ -351,9 +351,11 @@ func determineJobHealth(obj *unstructured.Unstructured) (status model.Kubernetes
 			failed = true
 			completed = true
 			message = condition.Message
+			break
```
nakabonne, line 354: Nice

pkg/app/piped/cloudprovider/kubernetes/state.go
```diff
@@ -351,9 +351,11 @@ func determineJobHealth(obj *unstructured.Unstructured) (status model.Kubernetes
...
 		case batchv1.JobComplete:
 			completed = true
 			message = condition.Message
+			break
```
nakabonne, line 358: It shouldn't break here because what we want to do here is check if any one of the conditions gets failed, 

Here is a comment of `JobStatus.Conditions`.

>	// The latest available observations of an object's current state. When a Job
	// fails, one of the conditions will have type "Failed" and status true. When

nakabonne: Ref: https://github.com/kubernetes/api/blob/master/batch/v1/types.go#L202-L203

pkg/app/piped/cloudprovider/kubernetes/state.go
nakabonne: but without a label, it continues the for loop.

pkg/app/piped/cloudprovider/kubernetes/state.go
ono-max: I see 👍 
I'll revert this change.

pkg/app/piped/cloudprovider/kubernetes/state.go
ono-max: Sorry, I couldn't understand what you mean.
Could you explain in more detail?

nakabonne: I mean: https://go.dev/play/p/m5qecum0nVV

nakabonne: This break just gets out from the switch statement. To break the wrapped For loop, you need to use Labels.

pkg/app/piped/cloudprovider/kubernetes/state.go
nakabonne: Sorry I am not sure what you mean. What are you stuck with?

ono-max: You think about future additional checks, right? For example:
```go
...
     if len(s.Status.LoadBalancer.Ingress) == 0 {
		status = model.KubernetesResourceState_OTHER
		desc = "Ingress points for the load-balancer are in progress"
                 return
      }
      if express {
          status = ...
          desc = ...
          return
     }
      return
}

ono-max: My opinion is that `if len(s.Status.LoadBalancer.Ingress) == 0 {...` should return if we have additional check under it.

ono-max: I think we don't need `return` now because there is no check under `if len(s.Status.LoadBalancer.Ingress) == 0 {...`.

nakabonne: >My opinion is that if len(s.Status.LoadBalancer.Ingress) == 0 {... should return if we have additional check under it.

We're on the same page.

>I think we don't need return now because there is no check under if len(s.Status.LoadBalancer.Ingress) == 0 {....

I feel you. Currently the result is the same whether it is there or not, but I think explicitly indicating it will prevent us from forgetting to add a return statement here in the future.

I'd say it is more important to reduce that risk as much as possible than to shorten just one line, but what do you think?

pkg/app/piped/cloudprovider/kubernetes/state.go
ono-max: Oh, it does make sense to me.
Thanks!

pkg/app/piped/cloudprovider/kubernetes/state.go
ono-max: Thank you for explaining in detail!
It does make sense to me. I'll revert this change.

nakabonne:
Great job 👍 
/lgtm

khanhtc1202:
Well done 👍 
/approve

PR #2850, opened 2021-11-26
Fix small part in state.go to make it consistency

ono-max:
As https://github.com/pipe-cd/pipe/pull/2841#discussion_r756537702, it should return to prevent us from forgetting to add a return statement in the future.

Fixes #

```release-note
NONE
```


nakabonne (approved):
Nice point

PR #2861, opened 2021-11-29
Hide database migration post from table of content list

khanhtc1202:

Hide the database migration post from blog table of content, since that version is staled and we're not going to support that migration script in the future.

Fixes #

```release-note
NONE
```


nakabonne:
This is a blog post, not documentation. I feel like you basically shouldn't delete it once you publish it as someone is still likely to refer to this page even if it gets outdated.

nakabonne:
I don't have a strong opinion so if you guys are fine, proceed as is.

khanhtc1202:
> This is a blog post, not documentation. I feel like you basically shouldn't delete it once you publish it even if it gets outdated as someone is still likely to refer to this page.

I feel you, but the change I made here does not delete the post content. It's possible to access the post by URL directly, just hide it from the table of content list, since migration post may cause users to misunderstand that our PipeCD highly depends on a database, that may not good for the product image.

nakabonne:
That makese total sense. In that case fine to me 👍 
There you go /approve

nakabonne (approved):

PR #2862, opened 2021-11-29
Add deployment chain status

khanhtc1202:

This PR adds the status field to the deployment chain model, we will use that status to show the deployment chain status on the client. In my image, the deployment chain only has 2 states: Running (status: running) and Finished (status: success, failure, cancelled). Once the deployment chain is created, we add a new deployment to the deployment chain blocks[0] immediately and mark the deployment chain status as Running.

I also add the logic to update the deployment chain status on the deployment status report working process in this PR, PTAL.

Fixes #

```release-note
NONE
```


khanhtc1202:
We will use `block` status to determine deployment chain behavior instead of `chain` status
/close

PR #2871, opened 2021-12-01
Apply tabs to app adding form to add from Git

nakabonne:
This PR adds Tabs to the application adding form so that we can support adding an app from config files defined in Git.
You can give it a try by simply using `make web-dev`.

Kapture 2021-12-01 at 19 05 08

Ref: https://github.com/pipe-cd/pipe/issues/2778

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/application-form/index.tsx
+++ b/pkg/app/web/src/components/application-form/index.tsx
@@ -100,7 +100,7 @@ function TabPanel(props: TabPanelProps): ReactElement {
           aria-labelledby={`simple-tab-${props.index}`}
       >
         {props.selected && (
-            <Box >
+            <Box>
               <Typography>{props.children}</Typography>
             </Box>
         )}
@@ -124,7 +124,7 @@ export const ApplicationFormTabs: React.FC<ApplicationFormProps> = (props) => {
 
   return (
       <Box width={600}>
-        <Box >
+        <Box>
           <Tabs value={selectedTabIndex} onChange={handleChange} aria-label="basic tabs example">
             <Tab label="Add manually" {...a11yProps(0)} />
             <Tab label="Add from Git (Alpha)" {...a11yProps(1)} />
@@ -137,7 +137,7 @@ export const ApplicationFormTabs: React.FC<ApplicationFormProps> = (props) => {
           Coming soon...
           {/** TODO: Show unregistered applications on the ADD FROM GIT tab */}
         </TabPanel>
-      </Box >
+      </Box>
   );
 }
 
--- a/pkg/app/web/src/components/application-form/index.tsx
+++ b/pkg/app/web/src/components/application-form/index.tsx
@@ -92,7 +92,6 @@ interface TabPanelProps {
 }
 
 function TabPanel(props: TabPanelProps) {
-
   return (
       <div
           role="tabpanel"
@@ -109,8 +108,6 @@ function TabPanel(props: TabPanelProps) {
   );
 }
 
-
-
 function a11yProps(index: number) {
   return {
     id: `simple-tab-${index}`,
@@ -119,10 +116,10 @@ function a11yProps(index: number) {
 }
 
 export const ApplicationFormTabs: React.FC<ApplicationFormProps> = (props) => {
-  const [selectedTabIndex, setValue] = useState(0);
+  const [selectedTabIndex, setSelectedTabIndex] = useState(0);
 
   const handleChange = (event: React.ChangeEvent<{}>, newValue: number) => {
-    setValue(newValue);
+    setSelectedTabIndex(newValue);
   };
 
   return (
--- a/pkg/app/web/src/components/application-form/index.tsx
+++ b/pkg/app/web/src/components/application-form/index.tsx
@@ -7,10 +7,13 @@ import {
   MenuItem,
   TextField,
   Typography,
+  Tabs,
+  Tab,
 } from "@material-ui/core";
 import { FormikProps } from "formik";
-import { FC, memo, ReactElement } from "react";
+import { FC, memo, ReactElement, useState } from "react";
 import * as yup from "yup";
+import PropTypes from "prop-types";
 import { APPLICATION_KIND_TEXT } from "~/constants/application-kind";
 import { UI_TEXT_CANCEL, UI_TEXT_SAVE } from "~/constants/ui-text";
 import { useAppSelector } from "~/hooks/redux";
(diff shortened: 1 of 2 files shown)


khanhtc1202:
Minor thing, could you revise the UI for `ADD MANUALLY` tab, looks like the tab content is surrounded by a blank border.

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +86,66 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+type TabPanelProps = {
+  children: any
+  value: any
+  index: any
+}
```
khanhtc1202, line 93: should avoid using `any` type, the check `props.value !== props.index` at line 100 is unsafe due to comparing between two values of type any, if it's a number, use type `number` instead.

khanhtc1202: For the `children` field, is type `HTMLElement` good enough?

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +86,66 @@ const useStyles = makeStyles((theme) => ({
...
+        <Box p={3}>{props.children}</Box>
+      </Typography>
+  );
+}
+
+TabPanel.propTypes = {
+  children: PropTypes.node,
+  index: PropTypes.any.isRequired,
+  value: PropTypes.any.isRequired
+};
```
khanhtc1202, line 113: nits, we may not need this, if only `children` is unrequired; adding a `?` after the field name in the type declaration is enough.

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +86,66 @@ const useStyles = makeStyles((theme) => ({
...
+
+function a11yProps(index: number) {
+  return {
+    id: `simple-tab-${index}`,
+    'aria-controls': `simple-tabpanel-${index}`,
+  };
+}
+
+export const ApplicationFormTabs: React.FC<ApplicationFormProps> = (props) => {
+  const [value, setValue] = useState(0);
```
khanhtc1202, line 124: nits, looks like this `value` is used to determine if this tab is selected or not. Not a convention, but in that case, instead of using a generic name like `value`, `selectedTabIndex` is more readable.

pkg/app/web/src/components/application-form/index.tsx
nakabonne: For sure, as long as we use Typescript, PropTypes doesn't come into play basically (except for [this](https://github.com/Microsoft/TypeScript/issues/4833). Let me revise it.

nakabonne:
@khanhtc1202 Thank you. 

>Minor thing, could you revise the UI for ADD MANUALLY tab, looks like the tab content is surrounded by a blank border.

Would you mind filling me? I'd like to know a little bit more about what to do.

pkg/app/web/src/components/application-form/index.tsx
nakabonne: Right. Let me look into it.

pkg/app/web/src/components/application-form/index.tsx
nakabonne: Like it 👍

khanhtc1202:
> @khanhtc1202 Thank you.
> 
> > Minor thing, could you revise the UI for ADD MANUALLY tab, looks like the tab content is surrounded by a blank border.
> 
> Would you mind filling me up? I'd like to know a little bit more about what to do.

Screen Shot 2021-12-02 at 13 03 16

I mean those spaces make me feel like the body is surrounded by an unnecessary blank border 🤔 

Maybe, it's caused by this line https://github.com/pipe-cd/pipe/pull/2871/files#diff-6448ead9affdf94b255e6401de0af28cd594cbcaa74071b1faf4fd0169034f51R104

Of course, it's just a minor thing, we can revise it later when you implement the autoload `ADD FROM GIT` tab content 😄

nakabonne:
>I mean those spaces make me feel like the body is surrounded by an unnecessary blank border 🤔

Makes sense. Considering that it is shown on a quite tiny display, we don't need any kind of unnecessary space, I would revert it 👍

nakabonne:
@khanhtc1202 I made a couple of changes based on your suggestion. PTAL 😄

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +85,65 @@ const useStyles = makeStyles((theme) => ({
...
+
+function a11yProps(index: number) {
+  return {
+    id: `simple-tab-${index}`,
+    'aria-controls': `simple-tabpanel-${index}`,
+  };
+}
+
+export const ApplicationFormTabs: React.FC<ApplicationFormProps> = (props) => {
+  const [selectedTabIndex, setValue] = useState(0);
```
khanhtc1202, line 122:
```suggestion
  const [selectedTabIndex, setSelectedTabIndex] = useState(0);
```
nits, as React convention, the setter should have a name that contains the value name it controls.

nakabonne: Got it!

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +85,65 @@ const useStyles = makeStyles((theme) => ({
...
+            <Tab label="Add from Git (Alpha)" {...a11yProps(1)} />
+          </Tabs>
+        </Box>
+        <TabPanel value={selectedTabIndex} index={0}>
+          <ApplicationForm {...props} />
+        </TabPanel>
+        <TabPanel value={selectedTabIndex} index={1}>
+          Comming soon...
+          {/** TODO: Show unregistered applications on the ADD FROM GIT tab */}
+        </TabPanel>
```
khanhtc1202, line 139:
```suggestion
        <TabPanel selected={selectedTabIndex === 0} index={0}>
          <ApplicationForm {...props} />
        </TabPanel>
        <TabPanel selected={selectedTabIndex === 1} index={1}>
          Comming soon...
          {/** TODO: Show unregistered applications on the ADD FROM GIT tab */}
        </TabPanel>
```
nits, I just feel like we can avoid this `value` name used here and use a bit more context provided variable `selected` since we use that value to determine should we show the tab children or not right? I usually make the component only focus on what to show and move the condition logic to the caller (in your component the part `props.value === props.index` is what I mean). When we do that way, we will have a single place where we can control display logic with the value we use (inside this `ApplicationFormTabs` component) that's easier for tracking and debugging. Besides, the props which need to be provided in order to use this `TabPanel` is easier to read too, it will be kind of
```
<TabPanel
   index={1}            // type number, index of the tab
   selected={true} // type bool, flag to control display based on is this Tab is selected or not
>
```

nakabonne: > I usually make the component only focus on what to show and move the condition logic to the caller 

I feel the same way. Love it!

nakabonne: Fixed at b9b20d7

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +85,62 @@ const useStyles = makeStyles((theme) => ({
...
+function TabPanel(props: TabPanelProps) {
+  return (
+      <div
+          role="tabpanel"
+          hidden={!props.selected}
+          id={`simple-tabpanel-${props.index}`}
+          aria-labelledby={`simple-tab-${props.index}`}
+      >
+        {props.selected && (
+            <Box >
```
khanhtc1202, line 103:
```suggestion
            <Box>
```
nits, and other `<Box >` too

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +85,62 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+interface TabPanelProps {
+  children?: React.ReactNode;
+  index: number;
+  selected: boolean;
+}
+
+function TabPanel(props: TabPanelProps) {
```
khanhtc1202, line 94:
```suggestion
function TabPanel(props: TabPanelProps): ReactElement {
```
nits

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -83,6 +85,62 @@ const useStyles = makeStyles((theme) => ({
...
+          <Tabs value={selectedTabIndex} onChange={handleChange} aria-label="basic tabs example">
+            <Tab label="Add manually" {...a11yProps(0)} />
+            <Tab label="Add from Git (Alpha)" {...a11yProps(1)} />
+          </Tabs>
+        </Box>
+        <TabPanel selected={selectedTabIndex === 0} index={0}>
+          <ApplicationForm {...props} />
+        </TabPanel>
+          <TabPanel selected={selectedTabIndex === 1} index={1}>
+          Comming soon...
```
nghialv, line 137:
```suggestion
          Coming soon...
```

nakabonne:
@khanhtc1202 @nghialv Thank you for having a look. PTAL ;)

khanhtc1202:
Nice work 🚀 
/lgtm

nghialv:
Nice.
/approve

PR #2896, opened 2021-12-07
Show the piped connecting status on the setting page

khanhtc1202:

https://user-images.githubusercontent.com/32532742/144983200-b7c37948-e669-461f-bcbd-d6e5677a34aa.mp4

Fixes #

```release-note
The Piped connection status is now showing in the piped tab on the setting page
```

### DIFF
--- a/pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
+++ b/pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
@@ -54,7 +54,7 @@ const useStyles = makeStyles((theme) => ({
     },
   },
   connectionStatus: {
-    paddingLeft: "10px",
+    paddingLeft: theme.spacing(1.5),
   },
 }));
 
--- a/pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
+++ b/pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
@@ -1,6 +1,7 @@
 import {
   Box,
   Button,
+  Badge,
   Dialog,
   DialogActions,
   DialogContent,
@@ -13,6 +14,7 @@ import {
   TableCell,
   TableRow,
   Typography,
+  Tooltip,
 } from "@material-ui/core";
 import { MoreVert as MoreVertIcon } from "@material-ui/icons";
 import clsx from "clsx";
@@ -20,6 +22,7 @@ import dayjs from "dayjs";
 import * as React from "react";
 import { FC, memo, useCallback, useState } from "react";
 import { CopyIconButton } from "~/components/copy-icon-button";
+import { PIPED_CONNECTION_STATUS_TEXT } from "~/constants/piped-connection-status-text";
 import { DELETE_OLD_PIPED_KEY_SUCCESS } from "~/constants/toast-text";
 import {
   UI_TEXT_ADD_NEW_KEY,
@@ -33,6 +36,7 @@ import {
   addNewPipedKey,
   deleteOldKey,
   fetchPipeds,
+  Piped,
   selectPipedById,
 } from "~/modules/pipeds";
 import { addToast } from "~/modules/toasts";
@@ -48,6 +52,9 @@ const useStyles = makeStyles((theme) => ({
     "&:hover button": {
       visibility: "visible",
     },
+  },
+  connectionStatus: {
+    paddingLeft: "10px",
   },
 }));
 
@@ -131,6 +138,17 @@ export const PipedTableRow: FC<Props> = memo(function PipedTableRow({
     onDisable(pipedId);
   }, [pipedId, onDisable]);
 
+  const connectionStatusColor = useCallback((status: Piped.ConnectionStatus) => {
+    switch (status) {
+      case Piped.ConnectionStatus.UNKNOWN:
+        return "secondary";
+      case Piped.ConnectionStatus.ONLINE:
+        return "primary";
+      case Piped.ConnectionStatus.OFFLINE:
+        return "error";
+    }
+  }, [piped?.status]);
+
   if (!piped) {
     return null;
   }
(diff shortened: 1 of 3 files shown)


pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
```diff
@@ -131,6 +138,17 @@ export const PipedTableRow: FC<Props> = memo(function PipedTableRow({
     onDisable(pipedId);
   }, [pipedId, onDisable]);
 
+  const connectionStatusColor = useCallback((status: Piped.ConnectionStatus) => {
+    switch (status) {
+      case Piped.ConnectionStatus.UNKNOWN:
+        return "secondary";
+      case Piped.ConnectionStatus.ONLINE:
+        return "primary";
```
nakabonne, line 146: I feel like Green looks more right to me but what do you think?
```suggestion
        return "success";
```

khanhtc1202: It's. but the Badge material does not provide that color type so we can't use that 😢 
Screen Shot 2021-12-07 at 16 46 31

nakabonne: You're right! Looks like the color "success" got introduced from v5: https://mui.com/api/badge/#props

nakabonne: I'm thinking about upgrading material-ui/core to v5, we're going to face a couple of issues to do so though.

khanhtc1202: Well, time to update our web dependencies 😂  lets me address it later 👍

nakabonne: Alright. For the color, let's change to it after that.

nakabonne:
Thrilled to see it's getting better gradually (also in this patch, there are lots of things I can learn from)
/lgtm

nghialv:
I feel it is a bit hard to see what status is showing.
How do you think about adding another column for it such as `Status` or `Connectivity`, which shows one fo the following text?

```diff
+ ONLINE
! OFFLINE
# UNKNOWN
```

nakabonne:
Basically there are only two statuses, Online or Offline, right? So I feel like showing color (green and red) is enough.
But either is fine to me because currently not crowded with the Piped list.

khanhtc1202:
Well, I can say that users will understand what it's where to look at once they hover to the icon the first time. Besides, just look at the name and the status right there and they can know what is connecting is easier than must to read the whole line for that information. IMO, I think the new column is only good if we have a kind of order/filter by status or something, in that case, we need a place that lists all the piped status and that the `Connection` or `Status` column as you said. How do you think?

khanhtc1202:
One more thing is, the view on a small screen (MacBook 13inches) is crammed already, so adding one more column is a bit difficult to see, I guess.

Screen Shot 2021-12-07 at 17 45 25

nghialv:
What color is being used for each of them?

e.g. `#c5f015`

- ONLINE ?
- OFFLINE ?
- UNKNOWN ?

pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
```diff
@@ -142,7 +160,19 @@ export const PipedTableRow: FC<Props> = memo(function PipedTableRow({
         className={clsx({ [classes.disabledItem]: piped.disabled })}
       >
         <TableCell>
-          <Typography variant="subtitle2">{piped.name}</Typography>
+          <Typography variant="subtitle2">
+            {piped.name}
```
nghialv, line 164: Let's show the status before the name because the name length is variable.

khanhtc1202:
> What color is being used for each of them?
> 
> e.g. `#c5f015`
> 
> * ONLINE ?
> * OFFLINE ?
> * UNKNOWN ?

Nice point, as @nakabonne mentioned here (https://github.com/pipe-cd/pipe/pull/2896#discussion_r763717174) we should use some other color that is more understandable, but the current material UI we're using is quite old and I will change the color later when we update web dependencies to a newer version 🙏

pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
nakabonne: Nice point

khanhtc1202: It will look like this. I just worry about do they missed it with the line header or something 😂  And since the name of the column is `name` so I guess the name should have a higher priority and should be placed first.

Screen Shot 2021-12-07 at 17 59 08

khanhtc1202: I don't have a strong opinion about that, but plz let me know you guys' thought 😄

nakabonne: For sure, seeing it actually, we can tell it's a little bit confusing.

nakabonne: Just started feeling that it's slightly better to put them on the right side.

nghialv:
> One more thing is, the view on a small screen (MacBook 13inches) is crammed already, so adding one more column is a bit difficult to see, I guess.

Yes, I see. Agree about this. Let's adopt the current implementation.

But to be honest I have to say that I am not a fan of the "hover" pattern, where an additional small pop-up will be floating on something. It gives me a feeling of having complicated things.

pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
nghialv: 👍

nghialv:
Great work!
/approve

khanhtc1202:
> > One more thing is, the view on a small screen (MacBook 13inches) is crammed already, so adding one more column is a bit difficult to see, I guess.
> 
> Yes, I see. Agree about this. Let's adopt the current implementation.
> 
> But to be honest I have to say that I am not a fan of the "hover" pattern, where an additional small pop-up will be floating on something. It gives me a feeling of having complicated things.

I feel you, lets revise the method to show list items on PipeCD console later, I guess we have many things to do to make it better 👍

pkg/app/web/src/components/settings-page/piped/components/piped-table-row.tsx
khanhtc1202: 🙆‍♀️

PR #2905, opened 2021-12-08
Fix launcher to use valid Printf verbs

nakabonne:
To prevent from being:
```
failed with error %!w(int=404))
```

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice catch
/approve

PR #2916, opened 2021-12-09
Fix what only one unregistered application is shown

nakabonne:
Since unregistered applications don't have an id, all entities in the state have been overwritten because id is used as a record key by default.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/application-form/index.tsx
+++ b/pkg/app/web/src/components/application-form/index.tsx
@@ -705,6 +705,7 @@ const UnregisteredApplicationList: FC<ApplicationFormProps> = memo(
                               pipedId: app.pipedId,
                               repo: {
                                 id: app.repoId,
+                                // TODO: Populate repo info from state
                                 remote: "",
                                 branch: "",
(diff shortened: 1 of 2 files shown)


khanhtc1202:
Nice catch 👍 But tbh, I don't really understand why this change fixes the issue mentioned in the PR description. Could you provide some more context 🙏

nakabonne:
@khanhtc1202 Originally, if you pass an array of ApplicationInfo to `addMany`, it will try to create a hashmap internally with the ID as the key. Since the ID of all application info is empty, there will always be one element. By defining it as a field of state in this way, it can be kept as an array.

Honestly, I'm not sure if this is the best way to prevent it thus it would be great if you had any other potential solutions.

khanhtc1202:
> Originally, if you pass an array of ApplicationInfo to addMany, it will try to create a hashmap internally with the ID as the key. Since the ID of all application info is empty, there will always be one element.

I see, makes sense to me 👍 Then let's go with your option, thank you.

PR #2918, opened 2021-12-10
Undo Env to mandatory again in App config

nakabonne:
This PR reverts to making EnvId a required field when registering an application, and also makes envName a required field in Application config.

The point is, the web front is going to check if the EnvName is valid or not.
Obviously, it would be better to check this when Piped reports an unregistered application, but it needs a little bit more code change.
Since [this is a workaround that will get removed as soon as the ACL feature](https://github.com/pipe-cd/pipe/issues/2915) is supported in the near future, I settled on adapting this way which requires less code modification and does not cause unnecessary communication.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/api/service/webservice/service.proto
+++ b/pkg/app/api/service/webservice/service.proto
@@ -223,7 +223,7 @@ message AddEnvironmentRequest {
 
 message AddApplicationRequest {
     string name = 1 [(validate.rules).string.min_len = 1];
-    string env_id = 2;
+    string env_id = 2 [(validate.rules).string.min_len = 1];
     string piped_id = 3 [(validate.rules).string.min_len = 1];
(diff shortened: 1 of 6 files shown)


khanhtc1202:
> Since this is a workaround that will get removed as soon as the ACL feature is supported in the near future, I settled on adapting this way which requires less code modification and does not cause unnecessary communication.

Totally agree, just one thing to confirm, we have to add env info in the application configuration file to enable auto register application configuration from git right? or we can add that value on the add application from git form?

nakabonne:
>we have to add env info in the application configuration file to enable auto register application configuration from git

This one.

khanhtc1202:
> > we have to add env info in the application configuration file to enable auto register application configuration from git
> 
> This one.

That's the easier option 👍

khanhtc1202:
Nice point
/approve

PR #2945, opened 2021-12-15
Make all environment in input form optional

nakabonne:
Now we can register an Application and Piped without environments.

image

image

I've confirmed Piped works well without env.
Now the environment is just a label to identify applications and Pipeds (that means easier to replace than before).

Fixes https://github.com/pipe-cd/pipe/issues/2813

```release-note
Environment is no longer required for Piped and application registration
```

### DIFF
--- a/pkg/app/web/src/components/application-form/index.tsx
+++ b/pkg/app/web/src/components/application-form/index.tsx
@@ -380,7 +380,7 @@ export const ApplicationForm: FC<ApplicationFormProps> = memo(
                   ...emptyFormValues,
                   name: values.name,
                   kind: values.kind,
-                  env: item.value == emptyEnvName ? "" : item.value,
+                  env: item.value === emptyEnvName ? "" : item.value,
                 });
               }}
               disabled={isSubmitting || disableApplicationInfo}
--- a/pkg/app/web/src/components/application-form/index.tsx
+++ b/pkg/app/web/src/components/application-form/index.tsx
@@ -203,6 +203,7 @@ function FormSelectInput<T extends { name: string; value: string }>({
   label,
   value,
   items,
+  required,
   onChange,
   disabled = false,
 }: {
@@ -210,6 +211,7 @@ function FormSelectInput<T extends { name: string; value: string }>({
   label: string;
   value: string;
   items: T[];
+  required: boolean;
   onChange: (value: T) => void;
   disabled?: boolean;
 }): ReactElement {
@@ -219,7 +221,7 @@ function FormSelectInput<T extends { name: string; value: string }>({
       name={id}
       label={label}
       fullWidth
-      required
+      required={required}
       select
       disabled={disabled}
       variant="outlined"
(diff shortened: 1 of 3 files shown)


khanhtc1202:
Nice improvement 👍
Tbh, I think if a field in a form is optional, it should have a lower priority and has a less notable place. So how about revise the Application add form, and change the place we show the Environment select box with Piped select box? I think the new place for Environment select box will also be removed in the near future and we will place Lables setting box there. How do you think?

nakabonne:
@khanhtc1202 
Yeah, if there is any better idea, I'd love to apply it. You mean, for example kind of like this?:
メモ 2021-12-15

nakabonne:
Either way, this PR aims to make them optional; it's great if you let me do it as another patch.

khanhtc1202:
> Either way, this PR aims to make them optional; it's great if you let me do it as another patch.

Yes, I get your point. We can address it by a separated PR 👍 

> Yeah, if there is any better idea, I'd love to apply it.

I mean just simply flipping the place where we show the Environment select box with the Piped select box is enough for me at this time. Since previously, we had a kind of flow: select the Environment first, then select the Piped which belongs to the selected Environment; that's why the Environment select box comes first, then the Piped select box in the same line. And, we remove the requirement of Environment in the form, so making Piped appear before the Environment is good enough, I guess. 

Screen Shot 2021-12-15 at 13 12 24

nakabonne:
@khanhtc1202 I see. Actually there is still a reason why the Environment form comes before Piped. If you choose an environment then the Piped select box shows only ones that belong to the specified environment. That is, it works as a filter for Piped.

khanhtc1202:
> @khanhtc1202 I see. Actually, there is still a reason why the Environment form comes before Piped. If you choose an environment then the Piped select box shows only ones that belong to the specified environment. That is, it works as a filter for Piped.

I see, I forgot that case 😓 Do we plan to release this Environment as an optional field in the form or we're going to completely remove it first and only release the `labels in the form` version? I feel like if it's safe to create an application without an environment and we already disable edit application via console, then should completely remove this unnecessary environment field from the form. How do you think about it?

nakabonne:
@khanhtc1202 
I would say some users still need this Environment concept because it works as a filter for applications. So it's not kind to remove it suddenly.
Let's migrate gradually. In the next release, I'm looking to announce that Environment is deprecated. After supporting Labels, we can remove Environment and prompt those who want to use Env to use Labels.

nakabonne:
`pipectl application add --env-id` was also required so I made a change on that.

nakabonne:
Fixed according to @khanhtc1202's suggestion!

pkg/app/web/src/components/application-form/index.tsx
```diff
@@ -365,13 +373,14 @@ export const ApplicationForm: FC<ApplicationFormProps> = memo(
...
-              items={environments.map((v) => ({ name: v.name, value: v.id }))}
+              items={envItems}
+              required={false}
               onChange={(item) => {
                 setValues({
                   ...emptyFormValues,
                   name: values.name,
                   kind: values.kind,
-                  env: item.value,
+                  env: item.value == emptyEnvName ? "" : item.value,
```
khanhtc1202, line 383:
```suggestion
                  env: item.value === emptyEnvName ? "" : item.value,
```
nits

khanhtc1202:
Nice works 👍 LGTM with small nits
/lgtm

nakabonne:
Thank you!

khanhtc1202:
👍 
/lgtm

nghialv:
Nice.
/approve

PR #2946, opened 2021-12-15
Add name field to all example app configs

nakabonne:
With it, those who try PipeCD using examples can register an app without manual input.
This PR also fixes places where use deprecated fields like `triggerPaths`.

Ref https://github.com/pipe-cd/pipe/pull/2935#issuecomment-993399168

```release-note
NONE
```


khanhtc1202:
🚀 
/approve

PR #2948, opened 2021-12-15
Don't show env when there isn't

nakabonne:
This PR removes unneeded and odd Skeletons.

image
image
image

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
+++ b/pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
@@ -212,9 +212,11 @@ export const ApplicationDetail: FC<ApplicationDetailProps> = memo(
             <Typography variant="h5">
               {app ? app.name : <Skeleton width={100} />}
             </Typography>
-            <Typography variant="subtitle2" className={classes.env}>
-              {env && env.name}
-            </Typography>
+            {env && (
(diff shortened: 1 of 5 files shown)


pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
```diff
@@ -213,7 +213,7 @@ export const ApplicationDetail: FC<ApplicationDetailProps> = memo(
               {app ? app.name : <Skeleton width={100} />}
             </Typography>
             <Typography variant="subtitle2" className={classes.env}>
-              {env ? env.name : <Skeleton width={100} />}
+              {env && env.name}
             </Typography>
```
khanhtc1202, line 217: nits,
```typescript
{env && (
  <Typography variant="subtitle2" className={classes.env}>
    {env.name}
  </Typography>
)}
```

nakabonne: Right. The same as the deployment detail page.

nakabonne: Fixed

khanhtc1202:
Nice, just the `web-test` on CI's taking so long
/approve

PR #2949, opened 2021-12-15
Ensure to inspect unregistered apps even no new commit

nakabonne:
This PR stops `appconfigreporter` from caching the latest unregistered app configs.
Originally, as long as no new commit was made, it has been using the cached unregistered app configs so that we can avoid unneeded file I/O. Due to this, the issue linked below happened.

If you think about it, we hardly have so many unregistered app configs in one repo at the same time. It's not so late to address the issue after facing that issue actually.

Fixes https://github.com/pipe-cd/pipe/issues/2947

```release-note
NONE
```


khanhtc1202:
Well, get your point. Premature optimization is the root of all evil 👍 
/approve

nakabonne:
>Premature optimization is the root of all evil 

That is so true

PR #2957, opened 2021-12-17
Fix miss kind in docs

khanhtc1202:

Fixes #

```release-note
NONE
```


nakabonne:
The two latest ones should be revised as well?

nakabonne:
I'm curious about how we handle this kind of obvious mistake.

khanhtc1202:
> I'm curious about how we handle this kind of obvious mistake.

Well, I can say that, for the miss spell issue like this, we have to update all versions too 😅 let me update it manually, this PR is just to test the `edit this document` feature from our document site.

nakabonne:
Nice
/approve

PR #2987, opened 2021-12-24
Add an ability to filter apps by labels

nakabonne:
Aside from `Autocomplete`I couldn't find out any material UI component which accepts multiple texts that look like those mentioned in the issue.

You can't select other than suggested options, that brings us a kind of validation for label format.

https://user-images.githubusercontent.com/19730728/147307661-f2736e04-13e7-42f8-bb78-25b34695e0fa.mov

Fixes https://github.com/pipe-cd/pipe/issues/2756
Ref https://github.com/pipe-cd/pipe/issues/2978

```release-note
Filtering applications by Labels is now available
```

### DIFF
--- a/pkg/app/web/src/api/applications.ts
+++ b/pkg/app/web/src/api/applications.ts
@@ -54,6 +54,9 @@ export const getApplications = ({
       enabled.setValue((options.enabled.value as unknown) as boolean);
       o.setEnabled(enabled);
     }
+    for (const label of options.labelsMap) {
+      o.getLabelsMap().set(label[0], label[1]);
+    }
     req.setOptions(o);
   }
   return apiRequest(req, apiClient.listApplications);
--- a/pkg/app/web/src/components/applications-page/application-filter/index.tsx
+++ b/pkg/app/web/src/components/applications-page/application-filter/index.tsx
@@ -4,8 +4,10 @@ import {
   makeStyles,
   MenuItem,
   Select,
+  TextField,
 } from "@material-ui/core";
-import { FC, memo } from "react";
+import Autocomplete from "@material-ui/lab/Autocomplete";
+import { FC, memo, useState } from "react";
 import { FilterView } from "~/components/filter-view";
 import { APPLICATION_KIND_TEXT } from "~/constants/application-kind";
 import { APPLICATION_SYNC_STATUS_TEXT } from "~/constants/application-sync-status-text";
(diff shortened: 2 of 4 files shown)


nakabonne:
This is ready to be reviewed, but it must not get shipped until the next version will be out.
/hold

nghialv:
> This is ready to be reviewed, but it must not get shipped until the next version will be out.

I think we don't have to wait for that time. This can be normally included in the next release (`v0.23.0`).

The idea is to break the "filtering by label" feature into smaller ones:
"support filtering by labels for the Applications page"
and
"support for the Deployments page"

This PR is for the first one and will be included in `v0.23.0`.
The second one will be scheduled later.

nakabonne:
@nghialv Technically we can 👍 
I was just worried about the next version will be way too feature-packed. Including this filtering feature means including the Labels concept as well. If you are fine, I will get started preparing documentations for Labels concept.

nghialv:
@nakabonne Nice point!
What kind of doc do you intend to have for the Label feature?

nakabonne:
Pages needed to be updated are:
- `concepts/` - add Labels description
- `user-guide/adding-an-application/` - add an example of setting labels, and mention that Labels can be set only via an app config file.
- Everywhere mentioning Environment- mark them deprecated

Seeing it actually, there is not that much to deal with.

nakabonne:
Well, either way, it's fine to keep this PR hold if we can release v0.23.0 as soon as the new year starts 👍

nghialv:
> Well, either way, it's fine to keep this PR hold if we can release v0.23.0 as soon as the new year starts 👍

Yea. I'm fine with that way.

nghialv:
> concepts/ - add Labels description

nit: Label is an obvious one so maybe we don't need its definition.

khanhtc1202:
Nice work, just one point I worried about. There is no way to make users know the format of labels they should use, so how about add things like this for Labels field

Screen Shot 2022-01-04 at 13 34 28

I mean, the title field is remained as `Labels` as you did, but for the placeholder, we can use text like `in format key:value` so that users could aware of what they should write there as a label, wdyt.

nakabonne:
@khanhtc1202 Good idea! That placeholder is definitely more valuable!

nakabonne:
@khanhtc1202 Applied your idea

khanhtc1202:
Nice work 👍 
/lgtm

nakabonne:
https://github.com/pipe-cd/pipe/releases/tag/v0.23.0 got shipped which made this mergeable.
/hold cancel

nghialv:
Great!
/approve

nakabonne:
It's rare for `web-test` to succeed anymore 😓

PR #2996, opened 2022-01-04
Enable to handle query parameter as an array even only one

nakabonne:
For the filtering apps and deployments by Labels, this is required.
If `arrayFormat` is set to default, we can't determine if it's an element of an array or a single string.
See: https://github.com/sindresorhus/query-string#arrayformat-1

Ref https://github.com/pipe-cd/pipe/issues/2978

```release-note
NONE
```


khanhtc1202:
Nice catch 👍 
Just want to confirm, does this change effect other query params than labels?

nakabonne:
@khanhtc1202 No, this `useSearchParams` is used by only the application filter and the deployment filter. And they don't accept any array field so far, that's why no need to worry about it.
image

khanhtc1202:
Get your point, thank you 👍 
/lgtm

PR #3011, opened 2022-01-07
Show label on the app detail and list page

nakabonne:
With it we can see labels on the application detail page as well as the application list page.

On the detail page:
image

On the list:
image

Fixes https://github.com/pipe-cd/pipe/issues/3009

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
+++ b/pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
@@ -221,9 +221,9 @@ export const ApplicationDetail: FC<ApplicationDetailProps> = memo(
                 {env.name}
               </Typography>
             )}
-            {app?.labelsMap.map((label, i) => (
+            {app?.labelsMap.map(([key, value], i) => (
               <Chip
-                label={label[0] + ":" + label[1]}
+                label={key + ": " + value}
                 className={classes.labelChip}
                 variant="outlined"
                 key={i}
--- a/pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
+++ b/pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
@@ -5,6 +5,7 @@ import {
   makeStyles,
   Paper,
   Typography,
+  Chip,
 } from "@material-ui/core";
 import SyncIcon from "@material-ui/icons/Cached";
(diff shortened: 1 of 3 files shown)


nakabonne:
Updated to show them on the application list page as well.

nakabonne:
Since we're looking into updating the Environment when adding a label that has `env` as a key
/hold

nakabonne:
The warning icon and header text couldn't be aligned. Let me correct it later.
image

pkg/app/web/src/components/application-detail-page/application-detail/index.tsx
```diff
@@ -217,6 +221,14 @@ export const ApplicationDetail: FC<ApplicationDetailProps> = memo(
                 {env.name}
               </Typography>
             )}
+            {app?.labelsMap.map((label, i) => (
+              <Chip
+                label={label[0] + ":" + label[1]}
```
nghialv, line 226: How about adding a space between the value and key as `key: value`?

nakabonne: For the app list page, you can't do that because the line breaks make it difficult to read. But for the detail page, looks nice:
image

nakabonne: Applied!

nghialv: Thank you. 👍

nghialv:
Can't wait to see it. Great job!
/lgtm

pkg/app/web/src/components/applications-page/application-list/index.tsx
```diff
@@ -137,7 +143,13 @@ export const ApplicationList: FC<ApplicationListProps> = memo(
                 <TableCell>Status</TableCell>
                 <TableCell>Name</TableCell>
                 <TableCell>Kind</TableCell>
-                <TableCell>Environment</TableCell>
+                <TableCell>
+                  Environment
+                  <Tooltip title="Deprecated" className={classes.tooltip}>
```
nghialv, line 148: How about this?

```suggestion
                  <Tooltip title="Deprecated. Please use Label instead." className={classes.tooltip}>
```

nakabonne: For sure

PR #3022, opened 2022-01-12
Put deprecated notes on the Environment tab

nakabonne:
Whenever you hover over the Environment tab at the settings page, you will see the notes "deprecated".

image

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/settings-page/index.tsx
+++ b/pkg/app/web/src/components/settings-page/index.tsx
@@ -84,7 +84,7 @@ export const SettingsIndexPage: FC = memo(function SettingsIndexPage() {
                 <ListItemText primary={text} />
                 {link === PAGE_PATH_SETTINGS_ENV && (
                   <ListItemIcon className={classes.listItemIcon}>
-                    <Tooltip title="Deprecated">
+                    <Tooltip title="Deprecated. Please use Label instead.">
                       <Warning fontSize="small" />
                     </Tooltip>
                   </ListItemIcon>
--- a/pkg/app/web/src/components/settings-page/index.tsx
+++ b/pkg/app/web/src/components/settings-page/index.tsx
@@ -3,9 +3,12 @@ import {
   List,
   ListItem,
   ListItemText,
+  ListItemIcon,
   makeStyles,
   Toolbar,
+  Tooltip,
 } from "@material-ui/core";
+import { Warning } from "@material-ui/icons";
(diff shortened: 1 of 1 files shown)


pkg/app/web/src/components/settings-page/index.tsx
```diff
@@ -76,6 +82,13 @@ export const SettingsIndexPage: FC = memo(function SettingsIndexPage() {
                 activeClassName={classes.activeNav}
               >
                 <ListItemText primary={text} />
+                {link === PAGE_PATH_SETTINGS_ENV && (
```
khanhtc1202, line 85: not sure why do we need this check? 🤔

pkg/app/web/src/components/settings-page/index.tsx
```diff
@@ -76,6 +82,13 @@ export const SettingsIndexPage: FC = memo(function SettingsIndexPage() {
...
+                  <ListItemIcon className={classes.listItemIcon}>
+                    <Tooltip title="Deprecated">
```
nghialv, line 87: How about this?

```suggestion
                    <Tooltip title="Deprecated. Please use Label instead.">
```

pkg/app/web/src/components/settings-page/index.tsx
nakabonne: Because this is only for the Environment tab, but am I correct?

pkg/app/web/src/components/settings-page/index.tsx
nakabonne: Looks nice

pkg/app/web/src/components/settings-page/index.tsx
khanhtc1202: If this condition check exists, when we click on other tabs on the left sidebar, the warning will disappear. I think it should keep be shown even if users click on other tabs on the left sidebar, wdyt?

Screen Shot 2022-01-12 at 18 50 54

nakabonne: The variable name may be a bit confusing, but it just tells it to show the Icon only if it is a ListItem for Environment.

khanhtc1202: Ah okay, I get the point 👍 Thanks you.

khanhtc1202:
Nice
/lgtm

PR #3026, opened 2022-01-13
Disable to change app description via Web client

nakabonne:
This PR does
- disable to update the app description via the web client
- ensure not to show the description box when it doesn't exist

When it has a description:

image

When it doesn't have it:

image

Fixes https://github.com/pipe-cd/pipe/issues/3002

```release-note
Disable to change app description via Web console
```


khanhtc1202:
Nice 👍 
/approve

PR #3051, opened 2022-01-14
Remove Env from the app additional form

nakabonne:
Unregistered apps no longer have envName (because of https://github.com/pipe-cd/pipe/pull/3050), that's why these codes will get unneeded.

Fixes #

```release-note
NONE
```


khanhtc1202:
Nice catch
/approve

PR #3077, opened 2022-01-20
Empower WebAPI to list events

nakabonne:
This PR primarily adds a new rpc named `ListEvents` which is about the same as `ListDeployment` in that:
- does paging with the cursor
- prevent it from scanning all entities by setting min_updated_at
- Filtering entities by labels by the application side

```release-note
NONE
```


khanhtc1202:
Way to go 👍 
/approve

PR #3093, opened 2022-01-21
Change events filter to suggest possible event names

nakabonne:
This PR applies the Autocomplete components that suggest Event names that can be known from the pre-fetched latest 50 events.

It needs to allow to be filled with a name that isn't in the options because the web client doesn't know the name of every event in the past.

Kapture 2022-01-21 at 16 11 46

Fixes https://github.com/pipe-cd/pipecd/issues/3080

```release-note
NONE
```

### DIFF
--- a/pkg/app/web/src/components/events-page/event-filter/index.tsx
+++ b/pkg/app/web/src/components/events-page/event-filter/index.tsx
@@ -53,9 +53,18 @@ export const EventFilter: FC<EventFilterProps> = memo(function EventFilter({
   const events = useAppSelector<Event.AsObject[]>((state) =>
     selectAllEvents(state.events)
   );
+
+  const [allNames, setAllNames] = useState(new Array<string>());
+  useEffect(() => {
+    const names = new Set<string>();
+    events.map((app) => {
+      names.add(app.name);
+    });
+    setAllNames(Array.from(names));
+  }, [events]);
+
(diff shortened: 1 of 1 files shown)


pkg/app/web/src/components/events-page/event-filter/index.tsx
```diff
@@ -53,9 +53,18 @@ export const EventFilter: FC<EventFilterProps> = memo(function EventFilter({
   const events = useAppSelector<Event.AsObject[]>((state) =>
     selectAllEvents(state.events)
   );
+
+  const [allNames, setAllNames] = useState(new Array<string>());
+  useEffect(() => {
+    const names = new Set<string>();
+    events.map((app) => {
+      names.add(app.name);
+    });
```
khanhtc1202, line 62:
```suggestion
    events.map((event) => {
      names.add(event.name);
    });
```
nits

khanhtc1202 (commented):
lgtm with a small nit
/lgtm

pkg/app/web/src/components/events-page/event-filter/index.tsx
nakabonne: good catch!

khanhtc1202:
Nice
/lgtm

knanao:
Super!
/approve

PR #3108, opened 2022-01-24
Fix event watcher not to handle already handled events

nakabonne:
Even though the status of Event was updated, the status of the latest event cached by Piped had not changed. Hence we've been facing an issue that Piped tries to handle handled Events unless we restarted Piped.

This PR addresses it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/apistore/eventstore/store.go
+++ b/pkg/app/piped/apistore/eventstore/store.go
@@ -190,12 +190,18 @@ func (s *store) UpdateStatuses(ctx context.Context, latestEvents []model.Event) 
 	for _, e := range latestEvents {
 		s.mu.Lock()
 		cached, ok := s.latestEvents[e.EventKey]
-		if ok && cached.Id != e.Id {
+		if !ok {
+			s.latestEvents[e.EventKey] = &e
+			s.mu.Unlock()
+			continue
+		}
+		if cached.Id != e.Id {
 			// There is already an event newer than the given one.
 			s.mu.Unlock()
 			continue
 		}
-		s.latestEvents[e.EventKey] = &e
+		s.latestEvents[e.EventKey].Status = e.Status
+		s.latestEvents[e.EventKey].StatusDescription = e.StatusDescription
 		s.mu.Unlock()
 	}
 	return nil
--- a/pkg/app/piped/apistore/eventstore/store.go
+++ b/pkg/app/piped/apistore/eventstore/store.go
@@ -188,14 +188,13 @@ func (s *store) UpdateStatuses(ctx context.Context, latestEvents []model.Event) 
 
 	// Update cached events.
 	for _, e := range latestEvents {
-		s.mu.RLock()
+		s.mu.Lock()
 		cached, ok := s.latestEvents[e.EventKey]
-		s.mu.RUnlock()
 		if ok && cached.Id != e.Id {
 			// There is already an event newer than the given one.
+			s.mu.Unlock()
 			continue
 		}
-		s.mu.Lock()
 		s.latestEvents[e.EventKey] = &e
 		s.mu.Unlock()
 	}
--- a/pkg/app/piped/apistore/eventstore/store.go
+++ b/pkg/app/piped/apistore/eventstore/store.go
@@ -29,22 +29,21 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/model"
 )
 
-// Getter helps get an event. All objects returned here must be treated as read-only.
-type Getter interface {
-	// GetLatest returns the latest event that meets the given conditions.
-	GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool)
-}
-
 type Store interface {
 	// Run starts syncing the event list with the control-plane.
 	Run(ctx context.Context) error
-	// Getter returns a getter for retrieving an event.
-	Getter() Getter
+	// GetLatest returns the latest event that meets the given conditions.
+	// All objects returned here must be treated as read-only.
+	GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool)
+	// GetLatest updates the status of the latest events.
+	// The second arg supposed to be the latest event. If it's not the latest, it will be ignored.
+	UpdateStatuses(ctx context.Context, latestEvents []model.Event) error
 }
 
 type apiClient interface {
 	GetLatestEvent(ctx context.Context, req *pipedservice.GetLatestEventRequest, opts ...grpc.CallOption) (*pipedservice.GetLatestEventResponse, error)
 	ListEvents(ctx context.Context, req *pipedservice.ListEventsRequest, opts ...grpc.CallOption) (*pipedservice.ListEventsResponse, error)
+	ReportEventStatuses(ctx context.Context, req *pipedservice.ReportEventStatusesRequest, opts ...grpc.CallOption) (*pipedservice.ReportEventStatusesResponse, error)
 }
 
 type store struct {
@@ -138,10 +137,6 @@ func (s *store) sync(ctx context.Context) error {
 	return nil
 }
 
-func (s *store) Getter() Getter {
-	return s
-}
-
 func (s *store) GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool) {
 	key := model.MakeEventKey(name, labels)
 	s.mu.RLock()
@@ -176,4 +171,33 @@ func (s *store) GetLatest(ctx context.Context, name string, labels map[string]st
 	}
 	s.latestEvents[key] = resp.Event
 	return resp.Event, true
+}
+
+func (s *store) UpdateStatuses(ctx context.Context, latestEvents []model.Event) error {
+	es := make([]*pipedservice.ReportEventStatusesRequest_Event, 0, len(latestEvents))
+	for _, e := range latestEvents {
+		es = append(es, &pipedservice.ReportEventStatusesRequest_Event{
+			Id:                e.Id,
+			Status:            e.Status,
+			StatusDescription: e.StatusDescription,
+		})
+	}
+	if _, err := s.apiClient.ReportEventStatuses(ctx, &pipedservice.ReportEventStatusesRequest{Events: es}); err != nil {
+		return fmt.Errorf("failed to report event statuses: %w", err)
+	}
+
+	// Update cached events.
+	for _, e := range latestEvents {
+		s.mu.RLock()
+		cached, ok := s.latestEvents[e.EventKey]
+		s.mu.RUnlock()
+		if ok && cached.Id != e.Id {
+			// There is already an event newer than the given one.
+			continue
+		}
+		s.mu.Lock()
+		s.latestEvents[e.EventKey] = &e
+		s.mu.Unlock()
+	}
+	return nil
 }
--- a/pkg/app/piped/apistore/eventstore/store.go
+++ b/pkg/app/piped/apistore/eventstore/store.go
@@ -32,7 +32,7 @@ import (
 // Getter helps get an event. All objects returned here must be treated as read-only.
 type Getter interface {
 	// GetLatest returns the latest event that meets the given conditions.
-	GetLatest(ctx context.Context, name string, labels map[string]string, forceRefresh bool) (event *model.Event, cacheUsed bool, ok bool)
+	GetLatest(ctx context.Context, name string, labels map[string]string, notUsingCache bool) (event *model.Event, cacheUsed bool, ok bool)
 }
 
 type Store interface {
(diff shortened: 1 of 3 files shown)


pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -142,13 +142,16 @@ func (s *store) Getter() Getter {
 	return s
 }
 
-func (s *store) GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool) {
+func (s *store) GetLatest(ctx context.Context, name string, labels map[string]string, forceRefresh bool) (event *model.Event, cacheUsed bool, ok bool) {
```
knanao, line 145: `noUsingCache` seems to more suitable in this case, but what about?

nakabonne: I feel you, that looks neat because of readability within `GetLatest()`. But one thing that I'd like to emphasize is refreshing the cache which will be used next time as well.

nakabonne: That being said, if you think about it, `notUsingCache` is enough in this case. Thank you! let me apply it.

nakabonne: Applied at 5bf7e27

knanao:
Nice:)
/lgtm

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -142,13 +142,16 @@ func (s *store) Getter() Getter {
 	return s
 }
 
-func (s *store) GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool) {
+func (s *store) GetLatest(ctx context.Context, name string, labels map[string]string, notUsingCache bool) (event *model.Event, cacheUsed bool, ok bool) {
```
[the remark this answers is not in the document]
knanao: Sorry if it's wrong.
I think `cacheUsed` indicate whether or not cache hit. Hence it always returns false when `notUsingCache` is true, 
but the other depends on  whether the cache exists.
Based on above, `cacheHit`  or something like that may be more suitable.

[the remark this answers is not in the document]
khanhtc1202: In such case, this one should be `return resp.Event, true, true`?
https://github.com/pipe-cd/pipecd/pull/3108/files/5bf7e27bf20b19d118e318b3f1d463105c0e580e#diff-2a38334e3554346c2d1bce6d910c5a78b67c638bf03a7279f58c15b586c1cee5R177 (this file at L177)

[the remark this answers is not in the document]
nakabonne: >but the other depends on whether the cache exists.

There you go 👍 

>Based on above, cacheHit or something like that may be more suitable.

Sounds better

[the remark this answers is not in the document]
nakabonne: Alternatively, it is also fine to separate into other methods, which can share core logic.

```go
type eventGetter interface {
	GetLatest(ctx context.Context, name string, labels map[string]string) (event *model.Event, ok bool)
	GetLatestWithCache(ctx context.Context, name string, labels map[string]string) (event *model.Event, cacheHit bool, ok bool)
}
```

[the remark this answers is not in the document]
nghialv: One vote from me for splitting down those functions. But the returned `cacheHit` value is redundant to me. So we can have simpler things like as below, WDYT?

``` golang
type eventGetter interface {
	GetLatest(ctx context.Context, name string, labels map[string]string) (event *model.Event, ok bool)
	GetLatestFromCache(ctx context.Context, name string, labels map[string]string) (event *model.Event, ok bool)
}
```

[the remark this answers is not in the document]
nghialv: In the case of using just one function, since `cacheUsed` and `ok` bring the same meaning then we can have

``` golang
type eventGetter interface {
	GetLatest(ctx context.Context, name string, labels map[string]string, onlyFromCache bool) (event *model.Event, ok bool)
}
```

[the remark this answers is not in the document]
khanhtc1202: I see, thanks for explaination 🙏 
In that case, I think making cache flush/update logic out of `GetLatest` maybe a better option. I mean currently, we use the GetLatest to
- Get value by key (for cache or not)
- Update cache if cache hit in case of using cache to get event status

IMO, the existing logic of GetLatest is get from cache and if there is not, fetch it and update the cache, that's simple and understandable so we should keep it as is. By add one more return value (cacheHit as you guy mentioned), and use that to control whether we should update the cache (by using cache update interface) could be more clear, wdyt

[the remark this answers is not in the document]
khanhtc1202: Or another option is to make the current GetLatest interface as is, and internaly of the cache hit, refresh it and return the updated latest. In such case, the caller of event store doesn't need to care about whether event store uses cache or not.

nakabonne:
@knanao @khanhtc1202 @nghialv Thank you guys! All of your comments made me aware that this interface is quite odd.

The status of the latest cached event is likely to be out of date because Piped updates statuses of the events store in Control-plane, without updating the cache. That was the initial motivation to force the cache to be updated here.

After careful consideration, I realized it goes well to update events via `eventstore` and update the cache as well whenever updating the status of an event. Then you can trust the cache when reading, right?

I'm sorry for that the change got totally different, but could you take a look again?

knanao (commented):
Nice:)
/lgtm

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -177,3 +172,31 @@ func (s *store) GetLatest(ctx context.Context, name string, labels map[string]st
...
+	// Update cached events.
+	for _, e := range latestEvents {
+		s.mu.Lock()
+		cached, ok := s.latestEvents[e.EventKey]
+		if ok && cached.Id != e.Id {
+			// There is already an event newer than the given one.
+			s.mu.Unlock()
+			continue
+		}
+		s.latestEvents[e.EventKey] = &e
```
nghialv, line 198: Let's update just the `status` field as what the function name is saying instead of the whole object.

nakabonne: For sure!

nakabonne:
@nghialv Revised!

nghialv:
> @knanao @khanhtc1202 @nghialv Thank you guys! All of your comments made me aware that this interface is quite odd.
> 
> The status of the latest cached event is likely to be out of date because Piped updates statuses of the events stored in Control-plane, without updating the cache. That was the initial motivation to force the cache to be updated when reading.
> 
> After careful consideration, I realized it goes well to update event statuses stored in control-plane via `eventstore` and update the cache at the same time. Then you can trust the cache when reading, right?
> 
> I'm sorry that the change got totally different, but could you take a look again?

Tbh, I want to avoid the way like this, where the data could be written from 2 paths (one comes from the control plane, one comes from the event-watcher component).
The data inside the local store could be overridden and then cause unexpected behavior.
In order to do it in the right way, I think we may need more effort in handling their version.

But this `event` store is currently used just by the event-watcher component,
So let's go with the current implementation for now.

nakabonne:
@nghialv I feel you. I'm also thinking about adding `RefreshEvent` which forces us to reload the latest event. It might be better to call this after reporting event statuses.

```go
type Store interface {
	GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool)
        // New
	RefreshEvent(ctx context.Context, name string, labels map[string]string) error
}

File: pkg/app/piped/apistore/eventstore/store.go [hunk 4]
```diff
@@ -177,3 +172,37 @@ func (s *store) GetLatest(ctx context.Context, name string, labels map[string]st
...
+			// There is already an event newer than the given one.
+			s.mu.Unlock()
+			continue
+		}
+		s.latestEvents[e.EventKey].Status = e.Status
+		s.latestEvents[e.EventKey].StatusDescription = e.StatusDescription
+		s.mu.Unlock()
+	}
+	return nil
+}
```
nakabonne, line 208: Or, how about overriding the cache by events fetched from control-plane?
```suggestion
func (s *store) UpdateStatuses(ctx context.Context, latestEvents []model.Event) error {
	es := make([]*pipedservice.ReportEventStatusesRequest_Event, 0, len(latestEvents))
	for _, e := range latestEvents {
		es = append(es, &pipedservice.ReportEventStatusesRequest_Event{
			Id:                e.Id,
			Status:            e.Status,
			StatusDescription: e.StatusDescription,
		})
	}
	if _, err := s.apiClient.ReportEventStatuses(ctx, &pipedservice.ReportEventStatusesRequest{Events: es}); err != nil {
		return fmt.Errorf("failed to report event statuses: %w", err)
	}

	// Update cached events.
	for _, e := range latestEvents {
		resp, err := s.apiClient.GetLatestEvent(ctx, &pipedservice.GetLatestEventRequest{
			Name:   e.Name,
			Labels: e.Labels,
		})
		if err != nil {
			return err
		}
		s.mu.Lock()
		cached, ok := s.latestEvents[e.EventKey]
		if !ok {
			s.latestEvents[e.EventKey] = resp.Event
			s.mu.Unlock()
			continue
		}
		if cached.Id != e.Id {
			// There is already an event newer than the given one.
			s.mu.Unlock()
			continue
		}
		s.latestEvents[e.EventKey].Status = resp.Event.Status
		s.latestEvents[e.EventKey].StatusDescription = resp.Event.StatusDescription
		s.mu.Unlock()
	}
	return nil
}
```

pkg/app/piped/apistore/eventstore/store.go
```diff
@@ -177,3 +172,37 @@ func (s *store) GetLatest(ctx context.Context, name string, labels map[string]st
...
+			continue
+		}
+		if cached.Id != e.Id {
+			// There is already an event newer than the given one.
+			s.mu.Unlock()
+			continue
+		}
+		s.latestEvents[e.EventKey].Status = e.Status
+		s.latestEvents[e.EventKey].StatusDescription = e.StatusDescription
+		s.mu.Unlock()
```
nakabonne, line 205: Or, how about overriding the cache by events fetched from control-plane again?
```suggestion
		resp, err := s.apiClient.GetLatestEvent(ctx, &pipedservice.GetLatestEventRequest{
			Name:   e.Name,
			Labels: e.Labels,
		})
		if err != nil {
			return err
		}
		s.mu.Lock()
		cached, ok := s.latestEvents[e.EventKey]
		if !ok {
			s.latestEvents[e.EventKey] = resp.Event
			s.mu.Unlock()
			continue
		}
		if cached.Id != e.Id {
			// There is already an event newer than the given one.
			s.mu.Unlock()
			continue
		}
		s.latestEvents[e.EventKey].Status = resp.Event.Status
		s.latestEvents[e.EventKey].StatusDescription = resp.Event.StatusDescription
		s.mu.Unlock()
```

nakabonne: @nghialv This solves this concern https://github.com/pipe-cd/pipecd/pull/3108#issuecomment-1021933373, right?

nghialv: I see. That is a nice idea but it will add some extra calls.
A better way is to use the returned events from `ReportEventStatuses` RPC to update the cache since they were updated by the server.

```
if _, err := s.apiClient.ReportEventStatuses(ctx, &pipedservice.ReportEventStatusesRequest{Events: es}); err != nil ```

A nice point of those ways is that we don't have to rewrite the logic of updating the event model what has already been done by the server.

But I think both of them still not be able to solve the root problem.
We will find a good solution for all local stores not just `event` store in the future.

Anyway, let's go with the current implementation for now. 🙌

Comment (nakabonne) [reply pkg/app/piped/apistore/eventstore/store.go:205]: Alright!

File: pkg/app/piped/apistore/eventstore/store.go [hunk 6]
```diff
@@ -29,22 +29,21 @@ import (
...
-
 type Store interface {
 	// Run starts syncing the event list with the control-plane.
 	Run(ctx context.Context) error
-	// Getter returns a getter for retrieving an event.
-	Getter() Getter
+	// GetLatest returns the latest event that meets the given conditions.
+	// All objects returned here must be treated as read-only.
+	GetLatest(ctx context.Context, name string, labels map[string]string) (*model.Event, bool)
+	// GetLatest updates the status of the latest events.
```
khanhtc1202, line 38:
```suggestion
	// UpdateStatuses updates the status of the latest events.
```
nits

nakabonne: Good catch!

khanhtc1202:
Here you go 🙌 
/approve

PR #3116, opened 2022-01-25
Fix to pass ctx to NewStore in livestatestore

knanao:
Fix to pass `ctx` in order to initialize client when call NewStore.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/livestatestore/cloudrun/store.go
+++ b/pkg/app/piped/livestatestore/cloudrun/store.go
@@ -34,7 +34,7 @@ type Store struct {
 type Getter interface {
 }
 
-func NewStore(cfg *config.CloudProviderCloudRunConfig, cloudProvider string, appLister applicationLister, logger *zap.Logger) *Store {
+func NewStore(_ context.Context, cfg *config.CloudProviderCloudRunConfig, cloudProvider string, appLister applicationLister, logger *zap.Logger) *Store {
 	logger = logger.Named("cloudrun").
 		With(zap.String("cloud-provider", cloudProvider))
 
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -308,7 +308,7 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 	var liveStateGetter livestatestore.Getter
 	// Start running application live state store.
 	{
-		s := livestatestore.NewStore(cfg, applicationLister, p.gracePeriod, input.Logger)
+		s := livestatestore.NewStore(ctx, cfg, applicationLister, p.gracePeriod, input.Logger)
 		group.Go(func() error {
 			return s.Run(ctx)
 		})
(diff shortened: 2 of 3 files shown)


nghialv:
Nice adding.
/lgtm

pkg/app/piped/livestatestore/cloudrun/store.go
```diff
@@ -34,7 +34,7 @@ type Store struct {
 type Getter interface {
 }
 
-func NewStore(cfg *config.CloudProviderCloudRunConfig, cloudProvider string, appLister applicationLister, logger *zap.Logger) *Store {
+func NewStore(ctx context.Context, cfg *config.CloudProviderCloudRunConfig, cloudProvider string, appLister applicationLister, logger *zap.Logger) *Store {
```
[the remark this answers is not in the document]
khanhtc1202: Wait, why does this note exist 👀  `cfg` really not being used?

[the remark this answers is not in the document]
knanao: It has been pointed out because the `NewStore` implementation has not been done yet.
[This PR](https://github.com/pipe-cd/pipecd/pull/3114) will resolve this issue.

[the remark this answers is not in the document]
khanhtc1202: I see, thanks for explaination 🙏

khanhtc1202:
Here you go 🚀 
/approve

pkg/app/piped/livestatestore/cloudrun/store.go
```diff
@@ -34,7 +34,7 @@ type Store struct {
 type Getter interface {
 }
 
-func NewStore(cfg *config.CloudProviderCloudRunConfig, cloudProvider string, appLister applicationLister, logger *zap.Logger) *Store {
+func NewStore(_ context.Context, cfg *config.CloudProviderCloudRunConfig, cloudProvider string, appLister applicationLister, logger *zap.Logger) *Store {
```
nakabonne, line 37: Tell me how to use ctx in brief?
I'm not sure your plan but basically letting struct have cotnext.Context is not a good manner.

nakabonne:
It's too late lol

pkg/app/piped/livestatestore/cloudrun/store.go
khanhtc1202: I guess it used to initialize the client of the store, not to be assigned as a property of the store. We could fix that on the caller side (on this PR: https://github.com/pipe-cd/pipecd/pull/3114#discussion_r791434181) if necessary.

nakabonne: In that case, that makes sense. Just using for initial communication, right 👍

knanao: Yep, you're right.
Therefore, it's assumed that client will be initialized in `NewStore` since I want the store to have a `client` instead of `ctx`

Issue #3109, opened 2022-01-24
Change all NOT_HANDLED events that a new event created to OUTDATED


nakabonne:
**What would you like to be added**:
If an event of the same type is issued before an event is handled, the event issued first will be `NOT_HANDLED` forever.
It can confuse you, so we need to update them periodically.

![image](https://user-images.githubusercontent.com/19730728/150749987-90a44ef6-e9b6-4119-a20f-901be840a914.png)

**Why is this needed**:

PR #3117, opened 2022-01-25
Ensure to make previous events OUTDATED whenever registering an event

nakabonne:
Whenever you register an event, it will make previous events `OUTDATED`.

Ref https://github.com/pipe-cd/pipecd/issues/3109

```release-note
NONE
```


nakabonne:
We settled on not doing this.
/close

PR #3121, opened 2022-01-26
Enable edit application config filename via edit app form

khanhtc1202:

Fixes #

```release-note
Enable to update application configuration filename via edit application form
```


nakabonne:
Nice. In the future, it would be nice if we could update it on the application detail page.
/lgtm

khanhtc1202:
> In the future, it would be nice if we could update it on the application detail page.

I guess, we're going to lock edit feature on the console, so looks like we're not going that way 😢

knanao:
GoGo!
/approve

PR #3164, opened 2022-01-30
Fix unable to fetch list Piped in case of using MySQL as datastore

khanhtc1202:

Since we're going to remove the `environment` concept from PipeCD project, the field `EnvIds` is now nullable and will be replaced with an empty array value (`[]`) due to our current MySQL index ensure. 

Since we list all available piped based on ProjectId value and we only have one index `piped_project_id_env_ids_asc` currently on Piped table, that index will be used.

> If a multi-valued key part has an empty array, no entries are added to the index, and the data record is not accessible by an index scan.

ref: https://dev.mysql.com/doc/refman/8.0/en/create-index.html#create-index-multi-valued

This PR adds another index on ProjectId so that MySQL will use this index instead of the multiple column index. The multiple column index will be removed once we completely remove the `environment` concept from PipeCD.

Thanks @amaany3 for investigation (PR #3151)

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/datastore/mysql/ensurer/indexes.sql
+++ b/pkg/datastore/mysql/ensurer/indexes.sql
@@ -111,6 +111,9 @@ CREATE INDEX name_project_id_status_updated_at_desc ON Event (Name, ProjectId, S
 -- Piped table indexes
 --
 
+-- index on `ProjectId` ASC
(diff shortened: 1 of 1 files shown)


knanao:
Nice!
/lgtm

nakabonne:
Thank you! (@amaany3 too)
/approve

PR #3175, opened 2022-02-01
Make sure to show status description even if that is too long

nakabonne:

Before

image

After

Kapture 2022-02-01 at 16 48 43

I'm aware that it's not the best, but it settles the worst case where we can't see the entire description.

Fixes https://github.com/pipe-cd/pipecd/issues/3081

```release-note
NONE
```


knanao:
Nice:)
/lgtm

khanhtc1202:
Here you go 🚀 
/approve

Issue #2865, opened 2021-11-30
Move redis out of api dependencies list


khanhtc1202:
**What would you like to be added**:

Currently, `redis` is counted as web_api and piped_api dependency but we should move it out of those API dependencies list to keep the API logic flexible.

ref: 
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/grpcapi/web_api.go#L72
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/grpcapi/piped_api.go#L65

**Why is this needed**:

PR #3312, opened 2022-02-24
Define hash cache interface

Hosshii:
Define hash cache interface.

rel #2865

```release-note
NONE
```

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -50,7 +50,7 @@ type multiGetter struct {
 	getters []Getter
 }
 
-// MultiGetter combines a lit of getters into a single getter.
+// MultiGetter combines a list of getters into a single getter.
 func MultiGetter(getters ...Getter) Getter {
 	all := make([]Getter, 0, len(getters))
 	for _, r := range getters {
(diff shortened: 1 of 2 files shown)


khanhtc1202:
Sorry, I don't really get the point of defining this new interface and how would it be used in order to resolve issue #2865 . It would be nice if you can explain it in the PR description or even better if we can address it by source code of this PR.

Hosshii:
I'm sorry. I was in too much of a hurry.
I was planning to replace the part that uses redis cache in [web_api](https://github.com/pipe-cd/pipecd/blob/0c6a1b3f464b659e83cc5f1aa280fa8f914e4c92/pkg/app/server/grpcapi/web_api.go#L124), and [piped_api](https://github.com/pipe-cd/pipecd/blob/0c6a1b3f464b659e83cc5f1aa280fa8f914e4c92/pkg/app/server/grpcapi/piped_api.go#L103) with HCache interface(made by  this PR), implement hash cache interface to [redis cache](https://github.com/pipe-cd/pipecd/blob/0c6a1b3f464b659e83cc5f1aa280fa8f914e4c92/pkg/cache/rediscache/cache.go#L27), and pass it as an argument.

PR #3328, opened 2022-03-01
Display version information to web

### LINKED ISSUE #1958: Display version information to the web frontend for easy reporting

Hosshii:
Display version information to web.

image

Fixes #1958

```release-note
Show running version and documentation link on the web console
```


pkg/app/web/src/components/header/index.tsx
```diff
@@ -206,11 +206,23 @@ export const Header: FC = memo(function Header() {
...
+          <Link
+            href="https://pipecd.dev/docs/"
+            target="_blank"
+            rel="noreferrer"
+          >
+            Docs
+          </Link>
+        </MenuItem>
+        <MenuItem disabled={true} dense={true}>
+          {process.env.STABLE_VERSION?.substring(0, 10)}
```
khanhtc1202, line 224:
```suggestion
          Version: {process.env.STABLE_VERSION?.substring(0, 10)}
```
nits

pkg/app/web/package.json
```diff
@@ -101,4 +101,4 @@
     "redux": "^4.0.5",
     "yup": "^0.32.9"
   }
-}
+}
```
khanhtc1202, line 104: missing last linebreak

pkg/app/web/src/components/header/index.tsx
khanhtc1202: btw, not sure why do we need to cut the string, why not just show the full version string? Since the last part is the commit hash and can be used to refer to source code on pipecd repo, which could be useful in some cases 👀

Hosshii: image
I think this one is easier to understand, but the display looks like this. Which one do you think is better?

khanhtc1202: I get your point 👍 Then, how about remove the `Version: ` part and show the full version string?

Hosshii: image
Sorry, my comment was not updated and I missed it.
This is what it looks like when I put them all up. Personally, I find it too long.

Hosshii: Normally, dirty is not displayed.

khanhtc1202: I guess, making the `Docs` text longer (such as `Documentations`) will make the last line less long 😄 
We can try to shift all the text to the right side as well. Happy to hear other opinions cc @nghialv @knanao

nghialv: I vote for "Documentation" to be unified with tab name on our homepage.

nghialv: And it should open a new tab when clicking on the documentation link. So maybe we also need a new tab icon for it.

khanhtc1202: The current implementation already has `target=_blank` properties for the `Docs` link, so I think it should be fine without the icon, I think only one option in a list that has a different icon is a bit odd.

nghialv: Right! I am fine without icon as well.

knanao: > I vote for "Documentation" to be unified with tab name on our homepage.
And it should open a new tab when clicking on the documentation link. So maybe we also need a new tab icon for it.

I agree with this.
And it may be better to put an icon next to "Documentations" like this.
Screen Shot 2022-03-01 at 20 18 05

nghialv: That is exactly what I was thinking about.

nghialv: About the version string, I think showing the full one is fine. Because on the production it is not that long.

pkg/app/web/src/components/header/index.tsx
```diff
@@ -206,11 +206,23 @@ export const Header: FC = memo(function Header() {
...
             Settings
           </Link>
         </MenuItem>
+        <MenuItem>
+          <Link
+            href="https://pipecd.dev/docs/"
+            target="_blank"
+            rel="noreferrer"
+          >
+            Docs
```
nghialv, line 220:
```suggestion
            Documentation
```

pkg/app/web/src/components/header/index.tsx
nghialv: Looks so beautiful!

Just a nit:

How do you think about moving the menu popup to be lower to avoid overlapping the tab bar?
We can resolve it in another PR.

khanhtc1202:
It's beautiful 😁
/approve

Hosshii:
@nghialv 
Sorry, I didn't push the code because I thought there might be more changes. I have pushed https://github.com/pipe-cd/pipecd/pull/3328/commits/986df95ca1d1686b48bac2eb2819c0c0b5a1298a .

nghialv:
> @nghialv Sorry, I didn't push the code because I thought there might be more changes. I have pushed [986df95](https://github.com/pipe-cd/pipecd/commit/986df95ca1d1686b48bac2eb2819c0c0b5a1298a) .

Got it. 👍

PR #3382, opened 2022-03-10
Show artifact versions

ffjlabo:

This PR changes to show artifact versions with links.

PipeCD

PipeCD

A part of https://github.com/pipe-cd/pipecd/issues/3303

```release-note
Be able to show artifact versions with links
```

### DIFF
--- a/web/src/components/application-detail-page/application-detail/index.tsx
+++ b/web/src/components/application-detail-page/application-detail/index.tsx
@@ -134,7 +134,7 @@ const MostRecentlySuccessfulDeployment: FC<{
                     <>
                       {i > 0 && ","}
                       <Link
-                        href={`//${v.url}`}
+                        href={v.url.includes("://") ? v.url : `//${v.url}`}
                         target="_blank"
                         rel="noreferrer"
                       >
--- a/web/src/components/application-detail-page/application-detail/index.tsx
+++ b/web/src/components/application-detail-page/application-detail/index.tsx
@@ -125,7 +125,30 @@ const MostRecentlySuccessfulDeployment: FC<{
             label="Deployed At"
             value={<span title={date.format()}>{date.fromNow()}</span>}
           />
-          <DetailTableRow label="Version" value={deployment.version} />
+          <DetailTableRow
+            label="Version"
+            value={
+              deployment.versionsList.length !== 0 ? (
+                <>
+                  {deployment.versionsList.map((v, i) => (
+                    <>
+                      {i > 0 && ","}
+                      <Link
+                        href={`//${v.url}`}
+                        target="_blank"
+                        rel="noreferrer"
+                      >
+                        {v.version} ({v.name})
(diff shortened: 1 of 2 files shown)


web/src/components/application-detail-page/application-detail/index.tsx
```diff
@@ -125,7 +125,30 @@ const MostRecentlySuccessfulDeployment: FC<{
...
+          <DetailTableRow
+            label="Version"
+            value={
+              deployment.versionsList.length !== 0 ? (
+                <>
+                  {deployment.versionsList.map((v, i) => (
+                    <>
+                      {i > 0 && ","}
+                      <Link
+                        href={`//${v.url}`}
```
khanhtc1202, line 137: should only add this `//` in case the url does not contain schema.
Can use v.url.includes(`://`) as the condition.

nghialv:
Love this!

What do you guys thing about changing the layout of the version?
For example, `v0.6.0 (helloworld)` -> `helloworld:v0.6.0`.

nghialv:
Btw, should we trim the version in case it is too long?

Like commit hash, we should show only the first 7 characters.

`helloworld:36515f2...` instead of `helloworld:sha36515f2d60ad86cc1f7b7e8d97e935741e5cf9f1`

ffjlabo:
> Love this!
> 
> What do you guys thing about changing the layout of the version? For example, `v0.6.0 (helloworld)` -> `helloworld:v0.6.0`.

I see, may be `helloworld:v0.6.0` style is natural for users because they often show that style in artifact files (e.g. k8s manifest). 



> Btw, should we trim the version in case it is too long?
> 
> Like commit hash, we should show only the first 7 characters.
> 
> `helloworld:36515f2...` instead of `helloworld:sha36515f2d60ad86cc1f7b7e8d97e935741e5cf9f1`

I agree with it!

ffjlabo:
How about the above?

nghialv:
> How about the above?

@khanhtc1202 @knanao what do you think?

khanhtc1202:
yeh, one vote from me. Please go ahead and address it 🙏

ffjlabo:
These are after fixed.

PipeCD

PipeCD

khanhtc1202:
nits, I think we should only trim the long text as it's a table cell content. On the detail page, we should just show it clearly, wdyt?

ffjlabo:
I see! Then I'll fix to truncate it in case the long text over the size of the table cell.

ffjlabo:
On the detail page, I fixed it to show full version info and split them into multiple lines.
PipeCD

nghialv:
Looks beautiful! ❤️ 
/lgmt

ffjlabo:
sorry I found some mistakes, so I'll fix

ffjlabo:
I fixed, PTAL. 🙇

nghialv:
👍 
/lgtm

khanhtc1202:
it's beautiful, thank you 😀 
/approve

PR #3383, opened 2022-03-10
Run tests in parallel for pkg/app/piped

Hosshii:
Add `t.Parallel` to execute tests in parallel. And delete unused test files.

Fixes #

```release-note
NONE
```


Hosshii:
This is more faster than others ([add t.parallel to all test](https://github.com/pipe-cd/pipecd/pull/3384) and [add t.parallel to slow test](https://github.com/pipe-cd/pipecd/pull/3386)).
So could you review this PR?

knanao (commented):
Nice catch!
Let me check one point.
If we run cases of table driven tests sequentially, the tests will be extremely slow?

Hosshii:
I think If each case of the table is slow, parallelization would speed it up. Otherwise, the cost of parallelization would be more expensive.

Hosshii:
I wrote these tests before https://github.com/pipe-cd/pipecd/pull/3370#discussion_r823411355. So these tests uses `t.Parallel` in sub tests. But it might be good to remove it.

knanao:
Thanks.
It seems good to decide the coding policy of our team and decide which one to use.

> we can use parallel on test function level or test cases level only, wdyt?

That's why, let we apply this!

Hosshii:
Thank you!  I removed sub-level t.parallel

knanao:
Awesome!
/lgtm

khanhtc1202:
Here you go 🙌 
/approve

PR #3388, opened 2022-03-11
Add index to CompletedAt

Hosshii:
Currently, completedAt does not exists on mysql column, and so order by completedAt will be failed.
To fix this, this PR create generated column of completedAt and add index to it. CompletedAt is nullable.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/datastore/mysql/ensurer/indexes.sql
+++ b/pkg/datastore/mysql/ensurer/indexes.sql
@@ -78,9 +78,9 @@ CREATE INDEX deployment_status_updated_at_desc ON Deployment (Status, UpdatedAt 
 ALTER TABLE Deployment ADD COLUMN PipedId VARCHAR(36) GENERATED ALWAYS AS (data->>"$.piped_id") VIRTUAL NOT NULL;
 CREATE INDEX deployment_piped_id ON Deployment (PipedId);
 
--- index on `CompletedAt` ASC
+-- index on `CompletedAt` DESC
 ALTER TABLE Deployment ADD COLUMN CompletedAt INT(11) GENERATED ALWAYS AS (data->>"$.completed_at") VIRTUAL NULL;
-CREATE INDEX deployment_completed_at ON Deployment (CompletedAt);
+CREATE INDEX deployment_completed_at_desc_id ON Deployment (CompletedAt DESC, Id);
 
 -- index on `DeploymentChainId` ASC and `UpdatedAt` DESC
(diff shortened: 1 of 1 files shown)


pkg/datastore/mysql/ensurer/indexes.sql
```diff
@@ -78,6 +78,10 @@ CREATE INDEX deployment_status_updated_at_desc ON Deployment (Status, UpdatedAt
 ALTER TABLE Deployment ADD COLUMN PipedId VARCHAR(36) GENERATED ALWAYS AS (data->>"$.piped_id") VIRTUAL NOT NULL;
 CREATE INDEX deployment_piped_id ON Deployment (PipedId);
 
+-- index on `CompletedAt` ASC
+ALTER TABLE Deployment ADD COLUMN CompletedAt INT(11) GENERATED ALWAYS AS (data->>"$.completed_at") VIRTUAL NULL;
+CREATE INDEX deployment_completed_at ON Deployment (CompletedAt);
```
khanhtc1202, line 83: nits, I don't think we need this index, instead of making this, maybe multiple columns index on CompletedAt and Id column is better, since this CompletedAt column is not be referred independently but in query with other columns. Should use `explain` query and check it out on MySQL cli.

Hosshii: Thank you, I'll check it.

Hosshii:
I fixed it. PTAL.

khanhtc1202:
Nice work, thanks 🙌 
/approve

pkg/datastore/mysql/ensurer/indexes.sql
```diff
@@ -78,6 +78,10 @@ CREATE INDEX deployment_status_updated_at_desc ON Deployment (Status, UpdatedAt
 ALTER TABLE Deployment ADD COLUMN PipedId VARCHAR(36) GENERATED ALWAYS AS (data->>"$.piped_id") VIRTUAL NOT NULL;
 CREATE INDEX deployment_piped_id ON Deployment (PipedId);
 
+-- index on `CompletedAt` DESC
```
khanhtc1202, line 81: nits, this should be on `CompletedAt` and `Id`

Hosshii: Thank you. I updated, PTAL

khanhtc1202:
Thanks 🙌 
/approve

PR #3424, opened 2022-03-22
Update login note

khanhtc1202:

It would look like this

Screen Shot 2022-03-23 at 9 35 27

Fixes #

```release-note
NONE
```

### DIFF
--- a/web/src/components/login-page/index.tsx
+++ b/web/src/components/login-page/index.tsx
@@ -101,8 +101,8 @@ export const LoginPage: FC = memo(function LoginPage() {
               />
               {isPlayEnvironment && (
                 <div className={classes.note}>
-                  Try project name <strong>play</strong> if you want to join the
-                  playground environment
+                  Input <strong>play</strong> if you want to join the playground
+                  environment
                 </div>
               )}
             </div>
--- a/web/src/components/login-page/index.tsx
+++ b/web/src/components/login-page/index.tsx
@@ -101,8 +101,8 @@ export const LoginPage: FC = memo(function LoginPage() {
               />
               {isPlayEnvironment && (
                 <div className={classes.note}>
-                  Try project name <strong>play</strong> if you want to access
-                  the playground environment
+                  Try project name <strong>play</strong> if you want to join the
+                  playground environment
                 </div>
               )}
             </div>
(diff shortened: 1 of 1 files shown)


web/src/components/login-page/index.tsx
```diff
@@ -101,7 +101,7 @@ export const LoginPage: FC = memo(function LoginPage() {
               />
               {isPlayEnvironment && (
                 <div className={classes.note}>
-                  Try with <strong>play</strong> name if you want to join the
+                  Try project name <strong>play</strong> if you want to join the
```
knanao, line 104: nits
```suggestion
                  Try to input <strong>play</strong> as project name if you want to join the
```

knanao (commented):
Is better off `textAlign: "left"` ?
https://github.com/pipe-cd/pipecd/pull/3424/files#diff-859b4e56fe1e23be4f03551006e60f9de7a224b2d91542372eefb519fb58cdbbR43

web/src/components/login-page/index.tsx
khanhtc1202: Nice idea, but this text is a bit too long to be shown in a single line 🙈

knanao: Oh, I got it.
Does it not also work if you set `testAlign: "left"`?

khanhtc1202: align right
Screen Shot 2022-03-23 at 9 32 10

align left
Screen Shot 2022-03-23 at 9 32 53

🙈

knanao: 😇 
allright, let's keep it that way.

nghialv, line 104: or just 😄 

```
Enter <strong>play</strong> if you want to join the playground environment
```

khanhtc1202: `Enter` is a bit odd to me, so lets me keep it as is 😄

khanhtc1202: But, maybe `Input <strong>play</strong> if you want to join the playground environment` is good. How do you think? @nghialv @knanao

knanao: Looks good!

khanhtc1202: addressed by https://github.com/pipe-cd/pipecd/pull/3424/commits/2e02521956bcdbe44baf69112c9fcc1d952969d0 🙌

nghialv:
Nice.

/approve

Issue #3426, opened 2022-03-22
Enable to fetch application description information on adding app


khanhtc1202:
**What would you like to be added**:

Currently, on adding application (via suggestion or manual) we don't fetch the application `spec.description` information in `app.pipecd.yaml` file. So, even we have prepared `spec.description` value in application configuration file, that value will not be reflected on the web console until we make other changes on `app.pipecd.yaml` to trigger `piped_api.UpdateApplicationConfigurations`.

**Why is this needed**:

niladrix719:
hey would love to give it a try

Tony-Stark20:
hey @khanhtc1202 I would like to work on this

Prachi-Gupta2808:
Hello @khanhtc1202 , can I work on this issue if not yet resolved? Also it would be great if you could provide me with a little more description about the issue.

PR #3433, opened 2022-03-23
Add description infor on adding application via suggestion form

khanhtc1202:

Add application's description to add application request so that the applications added via suggestion form will be registered with its description.

The manual adding application form still does not have labels and description input fields, so there is no way to add those values on adding right now.

Part of #3426

```release-note
NONE
```


khanhtc1202:
Need to rethink about application model's shard, so close for now
/close

PR #3470, opened 2022-03-30
Use offset instead of tzdb id

Hosshii:
Using tzdb ID may requires tzdb. It cause problem when tzDB does not esits (such as alpine linux).
So use offset instead of tzdb ID.

Fixes #3465

```release-note
NONE
```


pkg/app/server/service/webservice/service.proto
```diff
@@ -438,8 +438,8 @@ message GetInsightDataRequest {
     int64 range_from = 2 [(validate.rules).int64.gt = 0];
     int64 range_to = 3 [(validate.rules).int64.gt = 0];
     string application_id = 4;
-    // tzdb Id like `Asia/Tokyo`. null means UTC.
-    string timezone = 5;
+    // seconds east of UTC
+    int32 offset = 5;
```
khanhtc1202, line 442: should be int64 🤔

Hosshii: A day is 86400 seconds, so I think int32 is sufficient (go [uses](https://pkg.go.dev/time#FixedZone) int)

knanao:
NIce
/approve

PR #3476, opened 2022-03-31
Update decode interface

khanhtc1202:

Previously, the decode interface is
```go
type ShardDecoder interface {
	// Decode unmarshals all given raw data parts to a given entity e.
	Decode(e interface{}, parts ...[]byte) error
}
```
But in some case, we need to know which shard these parts of data belongs to in order to merge (decode) those data correctly. This PR updates the decode interface to
```go
type ShardDecoder interface {
	// Decode unmarshals all given raw data parts to a given entity e.
	Decode(e interface{}, parts map[Shard][]byte) error
}
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/datastore/commandstore.go
+++ b/pkg/datastore/commandstore.go
@@ -55,7 +55,7 @@ func (c *commandCollection) GetUpdatableShard() (Shard, error) {
 	}
 }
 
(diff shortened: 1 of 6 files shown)


knanao (commented):
Nice catch!
/lgtm

PR #3524, opened 2022-04-12
Update docs for piped configuration

knanao:

Fixes #

```release-note
NONE
```


khanhtc1202:
should we update docs in docs-dev/ and docs/ as well?

knanao:
@khanhtc1202 
docs-dev/ has already been updated by https://github.com/pipe-cd/pipecd/pull/3518.
but hasn't run `make sync-stable-docs` yet. that's why, let me sync it.

PR #3528, opened 2022-04-13
Add a dialog for skipping the analysis stage

knanao:
This enables it to skip just the analysis stage from web UI.
This means that this doesn't support the other stages.

https://user-images.githubusercontent.com/50069775/164153255-2099e1c4-31b1-4113-8512-e8f7eaad3762.mov

Fixes #

```release-note
Add a dialog for skipping the analysis stage.
```

### DIFF
--- a/web/src/components/deployments-detail-page/log-viewer/index.tsx
+++ b/web/src/components/deployments-detail-page/log-viewer/index.tsx
@@ -157,9 +157,10 @@ export const LogViewer: FC = memo(function LogViewer() {
     }
   };
 
-  const isSkipping = useAppSelector(
-    selectDeploymentStageIsSkipping(activeStage?.id)
-  );
+  const isSkipping = (): boolean => {
+      const stageId = activeStage ? activeStage.id : "";
+      return useAppSelector(selectDeploymentStageIsSkipping(stageId))
+  };
 
   if (!stageLog || !activeStage) {
     return null;
@@ -189,7 +190,7 @@ export const LogViewer: FC = memo(function LogViewer() {
                   onClick={() => setSkipTargetId(activeStage.id)}
                   variant="contained"
                   endIcon={<SkipNext />}
-                  disabled={isSkipping}
+                  disabled={isSkipping()}
                 >
                   SKIP
                 </Button>
--- a/web/src/components/deployments-detail-page/log-viewer/index.tsx
+++ b/web/src/components/deployments-detail-page/log-viewer/index.tsx
@@ -183,7 +183,7 @@ export const LogViewer: FC = memo(function LogViewer() {
                   onClick={() => setOpenSkipDialog(true)}
                   variant="contained"
                   endIcon={<SkipNext />}
-                  disabled={isSkippable}
+                  disabled={!isSkippable}
                 >
                   SKIP
                 </Button>
--- a/web/src/components/deployments-detail-page/log-viewer/index.tsx
+++ b/web/src/components/deployments-detail-page/log-viewer/index.tsx
@@ -28,15 +28,15 @@ import {
   Stage,
   StageStatus,
   skipStage,
-  selectDeploymentStageIsSkipping,
-  updateSkippingState,
+  selectDeploymentStageIsSkippable,
+  updateSkippableState,
 } from "~/modules/deployments";
 import { selectStageLogById, StageLog } from "~/modules/stage-logs";
 import { Log } from "./log";
 
 const INITIAL_HEIGHT = 400;
 const TOOLBAR_HEIGHT = 48;
-const ANALYSIS_NAME = "ANALYSIS";
+const ANALYSIS_STAGE_NAME = "ANALYSIS";
 
 function useActiveStageLog(): [Stage | null, StageLog | null] {
   return useShallowEqualSelector<[Stage | null, StageLog | null]>((state) => {
@@ -126,8 +126,8 @@ export const LogViewer: FC = memo(function LogViewer() {
   const dispatch = useAppDispatch();
   const [handlePosY, setHandlePosY] = useState(maxHandlePosY - INITIAL_HEIGHT);
   const logViewHeight = maxHandlePosY - handlePosY;
-  const [skipTargetId, setSkipTargetId] = useState<string | null>(null);
-  const isOpenSkipDialog = Boolean(skipTargetId);
+  const [isOpenSkipDialog, setOpenSkipDialog] = useState(false);
+  const stageId = activeStage ? activeStage.id : "";
 
   const handleOnClickClose = (): void => {
     dispatch(clearActiveStage());
@@ -147,20 +147,13 @@ export const LogViewer: FC = memo(function LogViewer() {
   );
 
   const handleSkip = (): void => {
-    if (skipTargetId) {
-      const deploymentId = stageLog ? stageLog.deploymentId : "";
-      dispatch(
-        skipStage({ deploymentId: deploymentId, stageId: skipTargetId })
-      );
-      dispatch(updateSkippingState({ stageId: skipTargetId }));
-      setSkipTargetId(null);
-    }
+    const deploymentId = stageLog ? stageLog.deploymentId : "";
+    dispatch(skipStage({ deploymentId: deploymentId, stageId: stageId }));
+    dispatch(updateSkippableState({ stageId: stageId }));
+    setOpenSkipDialog(false);
   };
 
-  const isSkipping = (): boolean => {
-      const stageId = activeStage ? activeStage.id : "";
-      return useAppSelector(selectDeploymentStageIsSkipping(stageId))
-  };
+  const isSkippable = useAppSelector(selectDeploymentStageIsSkippable(stageId));
 
   if (!stageLog || !activeStage) {
     return null;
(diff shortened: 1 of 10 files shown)


web/src/components/deployments-detail-page/pipeline/index.tsx
```diff
@@ -283,6 +315,21 @@ export const Pipeline: FC<PipelineProps> = memo(function Pipeline({
...
+
+        <Dialog open={isOpenSkipDialog} onClose={() => setSkipTargetId(null)}>
+          <DialogTitle>Skip stage</DialogTitle>
+          <DialogContent>
+            <DialogContentText>
+              {`To skip this stage, click "SKIP".`}
+            </DialogContentText>
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={() => setSkipTargetId(null)}>CANCEL</Button>
```
knanao, line 327: FIXME: I wanna make it to hide the dialog and show the process log, but now we can just only hide the dialog.
I wonder how to call the both of `setSkipTargetId(null)` and `handleOnClickStage` at the same onClieck.

knanao: Addressed by showing the process log when clicking the `ANALYSIS`.

nghialv:
@knanao Great!

I want to confirm one thing, was the "SKIP" popup shown after clicking the "ANALYSIS" stage?
If that, will it override the action of clicking the stage to see the log viewer?

nghialv:
@knanao Great!

I want to confirm one thing, was the "SKIP" popup shown after clicking the "ANALYSIS" stage?
If that, will it override the action of clicking the stage to see the log viewer?

knanao:
@nghialv 
Good point of view!

> I want to confirm one thing, was the "SKIP" popup shown after clicking the "ANALYSIS" stage?
Exactly, but I fixed it to show the log and open the dialog when clicking the stage by https://github.com/pipe-cd/pipecd/pull/3528/commits/629d9516180ac23c2fb37e3970bfd15d778173e4.

nghialv:
@knanao Thanks for your explanation.

I see. Now it shows both log viewer and confirmation popup once we clicked that stage, right?
Tbh, I feel it a bit annoying when showing that popup every time even when I just want to see the log viewer and don't intend to skip the stage.
So I think we should find somewhere to put a SKIP button to reveal the popup.

WDYT?

knanao:
@nghialv 
Well, I got your point.
For example, we can think about putting the SKIP button like below.
But we should get `stage.id` to stop the specific stage, that's why I feel like it's a little difficult.
Screen Shot 2022-04-18 at 17 10 40

nghialv:
@knanao Yes, I see. I am thinking of any way to add a stage action button (such as "SKIP", "APPROVE") in the stage area.
For example, add a small icon on the left half of the "ANALYSIS" stage button. 🤔

nghialv:
Or adding the SKIP, APPROVE button in the log viewer bar?

@khanhtc1202 Do you have any idea?

nghialv:
Another idea is changing the ANALYSIS stage to a button that holds a triangle on the right side to select an action. (Like the CANCEL button)

khanhtc1202:
The `ANALYSIS` stage button is quite small compared to others, so how about having text `Skip` clickable at the right bottom of that stage button 🤔

khanhtc1202:
Screen Shot 2022-04-19 at 9 49 20

Or maybe we can add an icon skip to the head of the log header line or text/button skip after stage description at that line. IMO, that skip feature should rarely be used or only be used in intention, which means not everyone wants to know about that at the first look. Making docs or letting users find out about that after time of use is a better choice. Besides, if we want to add the skip feature for other stages than ANALYSIS, changing the stage button is not a good idea, because every button would look so heavy with that kind of dropdown icon or text skip in it. wdyt?

knanao:
@nghialv @khanhtc1202 
Thank you guys for the good suggestions.

> we can add an icon skip to the head of the log header line or text/button skip after stage description at that line.

I would think this is a good idea and I would like to adapt this one like this.
How do you guys think about this?

knanao (commented):
plz, hold this temporarily in order to store the state of skipping.

knanao (commented):
/hold cancel

knanao:
@nghialv @khanhtc1202 
Finally, I fixed the UI hence PTAL when you guys have time.

nghialv:
@knanao Well done! 🙌

web/src/components/deployments-detail-page/log-viewer/index.tsx
```diff
@@ -4,20 +4,39 @@ import {
...
+  skipStage,
+  selectDeploymentStageIsSkipping,
+  updateSkippingState,
+} from "~/modules/deployments";
 import { selectStageLogById, StageLog } from "~/modules/stage-logs";
 import { Log } from "./log";
 
 const INITIAL_HEIGHT = 400;
 const TOOLBAR_HEIGHT = 48;
+const ANALYSIS_NAME = "ANALYSIS";
```
khanhtc1202, line 39:
```suggestion
const ANALYSIS_STAGE_NAME = "ANALYSIS";
```
nits, and caller place at L186

web/src/modules/deployments/index.ts
```diff
@@ -79,13 +79,15 @@ const initialState = deploymentsAdapter.getInitialState<{
   hasMore: boolean;
   cursor: string;
   minUpdatedAt: number;
+  skipping: Record<string, boolean>;
```
khanhtc1202, line 82: maybe `skippable` is a better name for this 🤔

web/src/components/deployments-detail-page/log-viewer/index.tsx
```diff
@@ -96,6 +126,8 @@ export const LogViewer: FC = memo(function LogViewer() {
   const dispatch = useAppDispatch();
   const [handlePosY, setHandlePosY] = useState(maxHandlePosY - INITIAL_HEIGHT);
   const logViewHeight = maxHandlePosY - handlePosY;
+  const [skipTargetId, setSkipTargetId] = useState<string | null>(null);
```
khanhtc1202, line 129: not sure why we need this variable, we can use `activeStage.Id` to get the current stage id directly.

knanao: I'm sorry if I'm misunderstanding this.
I use these properties for also opening and closing a dialog but please give a advice if there are better implementations.

knanao: That makes sense and is pretty good!
Let me apply it 🙌

khanhtc1202: 🙌

nghialv:
Nice work!
/lgtm

web/src/components/deployments-detail-page/log-viewer/index.tsx
```diff
@@ -135,6 +176,18 @@ export const LogViewer: FC = memo(function LogViewer() {
         <Divider />
         <Toolbar variant="dense" className={classes.toolbar}>
           <div className={classes.toolbarLeft}>
+            {activeStage.name == ANALYSIS_STAGE_NAME &&
+              activeStage.status == StageStatus.STAGE_RUNNING && (
```
khanhtc1202, line 180:
```suggestion
            {activeStage.name === ANALYSIS_STAGE_NAME &&
              activeStage.status === StageStatus.STAGE_RUNNING && (
```

web/src/modules/deployments/index.ts
```diff
@@ -258,6 +273,16 @@ export const deploymentsSlice = createSlice({
         ) {
           state.canceling[action.payload.deploymentId] = false;
         }
+        if (
+          action.payload.type === Command.Type.SKIP_STAGE &&
+          (action.payload.status === CommandStatus.COMMAND_FAILED ||
+            action.payload.status === CommandStatus.COMMAND_TIMEOUT)
+        ) {
+          state.skippable[action.payload.stageId] = false;
```
khanhtc1202, line 281: I think with the name `skippable` we should reverse the condition. Which means in case of failure or timeout, the stage should be skippable (this set should be = true) and at L285, we will set the value to false (make stage not skippable in case of click to skip button recently and in processing). WDYT?

knanao: Oops, I mistook this one. let me fix it 🙌 
Thank you.

web/src/modules/deployments/index.ts
```diff
@@ -172,6 +174,19 @@ export const approveStage = createAsyncThunk<
...
+  { deploymentId: string; stageId: string }
+>("deployments/skip", async (props, thunkAPI) => {
+  const { commandId } = await deploymentsApi.skipStage(props);
+  await thunkAPI.dispatch(fetchCommand(commandId));
+});
+
+export const updateSkippableState = createAsyncThunk<void, { stageId: string }>(
+  "deployments/skippable",
+  () => {}
+);
```
khanhtc1202, line 188: nits, `updateSkippableState` is a bit odd and hard to understand since the name is too general. How about making this action accepts `skippable: bool` besides `stageId`? Or rename this to `lockSkippableState`? 🤔

khanhtc1202:
Nice ✨ ✨ 
/approve

khanhtc1202:
@knanao web test is failed 👀

web/src/modules/deployments/index.test.ts
```diff
@@ -272,4 +274,24 @@ describe("deploymentsSlice reducer", () => {
...
   });
+
+  describe("updateSkippableState", () => {
+    it(`should handle ${updateSkippableState.fulfilled.type}`, () => {
+      expect(
+        deploymentsSlice.reducer(initialState, {
+          type: updateSkippableState.fulfilled.type,
+          meta: {
+            arg: {
+              stageId: "stage-id",
```
khanhtc1202, line 285: need to add skippable value here

PR #3573, opened 2022-04-22
Rename the grpc api for the RBAC

knanao:
Fix to rename the grpc API for the RBAC into `XXXProjectUserGroup`.

Fixes #

```release-note
NONE
```


nghialv:
@knanao Thank you.

> Add the role field in DeleteProjectUserGroupRequest.

I still don't get why we need that field.
Firstly, please let me confirm the UI, will the UI ① be deleted as soon as the UI ② is added, right?
And the built-in role is not deletable but the old user group (configured via old RBAC) is deletable, right?

knanao:
@nghialv 

> Firstly, please let me confirm the UI, will the UI ① be deleted as soon as the UI ② is added, right?

Nope.
I think we should provide some release stages not to confuse the user.
First of all, we're going to provide the view of the `UserGroup` tab and `RBAC Role` sections like this.
This means the user can just only see the group and role which have already been added.
Screen Shot 2022-04-22 at 17 22 49

Second, we're gotta provide the `Add` function and `Delete` function to edit old user groups like this.
Besides, we're gotta remove the edit button for the old RBAC config.
Screen Shot 2022-04-22 at 17 23 34

Third, we're gotta provide the `Add` function to add a new RBAC role.
Screen Shot 2022-04-22 at 17 18 11

And Last one, we'll remove the view of the old RBAC config like this.
Screen Shot 2022-04-22 at 17 33 10

knanao:
> And the built-in role is not deletable but the old user group (configured via old RBAC) is deletable, right?

Exactly, but the user who is using the old RBAC won't be able to delete the old user group via the view of the one from the second release.

knanao:
We talked about this directly, this was decided like this consequently.

First, we're gonna provide the `User Group` tab, the new RBAC role section, and remove the old RBAC role section.
Screen Shot 2022-04-25 at 14 20 42

Finally, we're gotta provide the Add and Edit function to add a new RBAC role.
Screen Shot 2022-04-22 at 17 33 10

knanao:
@nghialv @khanhtc1202 
I fixed to just only rename the grpc API and defined the release stage.
That's why PTAL when you guys have enough time.

nghialv (approved):
Thank you. Nice.

khanhtc1202 (approved):
👌🏻

PR #3578, opened 2022-04-24
Support PodDisruptionBudget

chaspy:
PodDisruptionBudget is not supported to show health status in an application view.

Fixes https://github.com/pipe-cd/pipecd/issues/3526

```release-note
Show health status of PodDisruptionBudget resource
```


knanao (commented):
Nice catch!
But now, k8s have already supported `policy/v1`, that's why let's add the app version below.
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/cloudprovider/kubernetes/resourcekey.go#L50

chaspy:
@knanao Thanks, added.

knanao (approved):
Nice!

nghialv (approved):
Looks nice. Thank you.

PR #3641, opened 2022-05-19
Fix the docs to add analysis type

knanao:

Fixes #

```release-note
NONE
```


khanhtc1202:
Just don't remember which version we added the Datadog as AnalysisProvider 😂

khanhtc1202 (approved):
Nice catch

nghialv (approved):
Nice.

knanao:
@khanhtc1202 
Me neither, but as you can know from the article, it would be after v0.25.0.
that's why this is fine🙌 
https://github.com/pipe-cd/pipecd/blob/master/docs/content/en/docs-v0.25.x/operator-manual/piped/adding-an-analysis-provider.md

PR #3647, opened 2022-05-20
Fix app config reporter skip sync configuration on adding app

khanhtc1202:

Previously, piped can only update the application configuration when the repository which contains that app has new commits after the application be registered to the pipecd database. In case of the app configuration had existed before the app be added, the new registered application configuration can not be synced. This PR updates the logic on appconfigreporter package to ensure the piped only skip syncing applications in case of application's last commit has been scanned, not in case of the application containing repository's last commit has been scanned as currently.

Fixes #

```release-note
Fix application's configuration is not synced on registering to pipecd control plane
```

### DIFF
--- a/pkg/app/piped/appconfigreporter/appconfigreporter.go
+++ b/pkg/app/piped/appconfigreporter/appconfigreporter.go
@@ -188,7 +188,7 @@ func (r *Reporter) updateRegisteredApps(ctx context.Context, headCommits map[str
 		return fmt.Errorf("failed to update application configurations: %w", err)
 	}
 
-	// Memorize registered application which be updated above.
+	// Memorize registered applications, which are updated above.
 	for _, app := range outOfSyncRegisteredApps {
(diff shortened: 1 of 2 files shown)


pkg/app/piped/appconfigreporter/appconfigreporter.go
```diff
@@ -160,51 +155,49 @@ func (r *Reporter) scanAppConfigs(ctx context.Context) error {
...
 		&pipedservice.UpdateApplicationConfigurationsRequest{
-			Applications: registeredApps,
+			Applications: outOfSyncRegisteredApps,
 		},
 	)
 	if err != nil {
 		return fmt.Errorf("failed to update application configurations: %w", err)
 	}
+
+	// Memorize registered application which be updated above.
```
knanao, line 191:
```suggestion
	// Memorize the registered application which is updated above.
```

knanao (approved):
Looks neats🎉

nghialv (approved):
Nice improvement!

PR #3687, opened 2022-05-27
Fix unable to plan deployment in case of changing app config

### LINKED ISSUE #3686: Unable to plan deployment in case of changing application config input
**What happened**:

In case the application configuration input is changed (input section) the application of kind K8s will be failed with `Unable to plan the deployment` error

**How to reproduce it**:

**Environment**:
- `piped` version: 0.32.0
- `control-plane` version:
- Others:

khanhtc1202:

Fixes #3686 

```release-note
Fix unable to plan deployment in case of changing configuration of kind Kubernetes application
```

### DIFF
--- a/pkg/app/piped/planner/kubernetes/kubernetes.go
+++ b/pkg/app/piped/planner/kubernetes/kubernetes.go
@@ -200,7 +200,12 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
 			return
 		}
 
-		loader := provider.NewLoader(in.ApplicationName, runningDs.AppDir, runningDs.RepoDir, in.GitPath.ConfigFilename, cfg.Input, in.GitClient, in.Logger)
+		runningCfg := runningDs.ApplicationConfig.KubernetesApplicationSpec
+		if runningCfg == nil {
+			err = fmt.Errorf("unable to find the running configuration (%v)", err)
+			return
+		}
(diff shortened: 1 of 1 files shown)


pkg/app/piped/planner/kubernetes/kubernetes.go
```diff
@@ -200,7 +200,12 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
 			return
 		}
 
-		loader := provider.NewLoader(in.ApplicationName, runningDs.AppDir, runningDs.RepoDir, in.GitPath.ConfigFilename, cfg.Input, in.GitClient, in.Logger)
+		runningCfg := runningDs.ApplicationConfig.KubernetesApplicationSpec
+		if runningCfg == nil {
```
knanao, line 204: Sorry, I couldn't understand this meaning.
What happens with the first plan?

khanhtc1202: cfg.Input is the application configuration of the target commit (the one which we're going to deploy), but at that place what we need is the currently running application configuration, that is why I pull runningCfg there

ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/planner/kubernetes/kubernetes.go#L56-L65

knanao: Oh, I totally get it.
I misunderstood the application manifest and application config.
Thank you.

knanao (approved):

nghialv (approved):
Nice catch!

PR #3692, opened 2022-05-30
Add error message to GetInsightData API

### LINKED ISSUE #3650: Error message shown empty on insight page load
**What happened**:

Screen Shot 2022-05-23 at 11 46 37

**How to reproduce it**:

access to insight page

**Environment**:
- `piped` version: v0.31.0
- `control-plane` version: v0.31.0
- Others:

Szarny:

Add "GetInsightData API is not yet implemented" error message to GetInsightData API.

Before
image

After
スクリーンショット 2022-05-30 12 40 07

Fixes #3650 

```release-note
NONE
```

### DIFF
--- a/pkg/app/server/grpcapi/web_api.go
+++ b/pkg/app/server/grpcapi/web_api.go
@@ -1434,7 +1434,7 @@ func (a *WebAPI) ListAPIKeys(ctx context.Context, req *webservice.ListAPIKeysReq
 
 // GetInsightData returns the accumulated insight data.
 func (a *WebAPI) GetInsightData(ctx context.Context, req *webservice.GetInsightDataRequest) (*webservice.GetInsightDataResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	return nil, status.Error(codes.Unimplemented, "GetInsightData API is not yet implemented.")
 
 	// claims, err := rpcauth.ExtractClaims(ctx)
 	// if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/app/server/grpcapi/web_api.go
```diff
@@ -1434,7 +1434,7 @@ func (a *WebAPI) ListAPIKeys(ctx context.Context, req *webservice.ListAPIKeysReq
 
 // GetInsightData returns the accumulated insight data.
 func (a *WebAPI) GetInsightData(ctx context.Context, req *webservice.GetInsightDataRequest) (*webservice.GetInsightDataResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	return nil, status.Error(codes.Unimplemented, "GetInsightData API is not yet implemented.")
```
khanhtc1202, line 1437: How about `Not yet implemented` or `Not available`?

khanhtc1202: Lets make it `Not available` 😄

Szarny: Since this message is directly displayed in the UI, I was wondering if it would be better to clarify to the user what is not yet implemented.
What do you think?

khanhtc1202: same as your reason, but I think end-users should not be annoyed by which endpoint/API is not yet implemented, so generally mentioning that the feature/UI is not available is enough to me. We - as the dev - should know about what is not yet implemented (by checking the network debugging tool if it's an HTTP request, but in this case gRPC only returns 200 even in case it's an error 😓 ).
In short, I think users don't care which API is failed, only us care about that and we should not annoy users by making those kind of message 😄

Szarny: I understand! then I would like to update it with `Not available` 👍

Szarny: now it's done!
image

nghialv (approved):
Thank you.

khanhtc1202 (approved):
Nice ✨

PR #3721, opened 2022-06-03
Fix the script of generating release

knanao:

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/gen-release.sh
+++ b/hack/gen-release.sh
@@ -24,8 +24,8 @@ done
 echo "Prepare release for version $1"
 
 # Update release file
-RELEASE_FILE=release/RELEASE
-echo -e "version: $1" > $RELEASE_FILE
+RELEASE_FILE=RELEASE
+sed -i '' "s/tag:.*/tag: $1/" $RELEASE_FILE
 
 CONTENT_DIR=docs/content/en
 # Render release note template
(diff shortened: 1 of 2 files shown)


hack/gen-release.sh
```diff
@@ -24,8 +24,8 @@ done
 echo "Prepare release for version $1"
 
 # Update release file
-RELEASE_FILE=release/RELEASE
-echo -e "version: $1" > $RELEASE_FILE
+RELEASE_FILE=RELEASE
+sed -i '' "s/tag:.*/tag: $1/" $RELEASE_FILE
```
khanhtc1202, line 28: I know we don't have a convention for this but just think it's better to unify with other places we cut this value, so
```bash
head -n 1 RELEASE | cut -d ' ' -f 2
```
(as we did in Makefile)

knanao: I see. Let me replace it with this 🙌

knanao: @khanhtc1202 
Oops, sorry let me check one point.
This script updates the tag in the RELEASE file hence I think this args will be required.

khanhtc1202: ah got that, it's my miss 🙏

khanhtc1202 (approved):
Nice catch

nghialv (approved):
Nice catch!

PR #3723, opened 2022-06-03
Auto-complete Piped form when only one Piped has been added

Szarny:

Add feature to auto-complete Piped form when only one Piped has been added.

image

If only one Piped is added, it is obvious that the user will select that Piped when adding the application, so it would be better to have this feature to save the user time interacting with the UI.

Fixes #

```release-note
Able to auto-complete Piped form when only one Piped has been added
```


nghialv (approved):
That is obviously better.

knanao (approved):
Nice improvement!

PR #3724, opened 2022-06-03
Omit empty fields in the "Show the configuration" dialog

### LINKED ISSUE #3712: UX improvement related to "Show Piped configuration" feature

- Expand the width of the dialog to display the Piped configuration.
- Display Piped configuration with YAML syntax highlight.
- If the Piped configuration has not yet been loaded, make the corresponding list item disabled.
- others?

**Why is this needed**:

To improve UX when users use this feature.

Szarny:

Omit empty fields (except required fields) on "Show the configuration" dialog.

Some fields of Piped config are `struct`, so `omitempty` is not applicable. Although `omitempty` can be applied by changing the field to a pointer to a struct, this PR leaves it as is because the scope of the change is too large.

スクリーンショット 2022-06-03 15 50 23

Fixes #3712

```release-note
Omit empty fields in the "Show the configuration" dialog
```

### DIFF
--- a/RELEASE
+++ b/RELEASE
@@ -1,4 +1,4 @@
-tag: v0.32.3
+tag: v0.32.4
 
 releaseNoteGenerator:
   showCommitter: false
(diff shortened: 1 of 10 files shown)


khanhtc1202 (approved):
✨

nghialv (approved):
Nice improvement!

PR #3732, opened 2022-06-06
Fixed bug that Cloud Provider is always Kubernetes in "Show Piped config" dialog

### LINKED ISSUE #3731: Display wrong cloud provider in Show Piped config dialog
**What happened**:

Whatever Cloud Provider is specified in piped config, it's displayed on web console as Kubernetes.
image

Configs are displayed according to the specified Cloud Provider.

**How to reproduce it**:

Specify a Cloud Provider other than Kubernetes and open the Show piped config dialog.

**Environment**:
- `piped` version: 
- `control-plane` version:
- Others:

Szarny:

Fixed bug that Cloud Provider is always Kubernetes in "Show Piped config" dialog (also add unit test to check that the bug has been resolved) .

This bug occurred because the `UnmarshalJSON` method of the `PipedCloudProvider` structure (the method called when `json.Unmarshal` is invoked) did not consider the case where the type of `type` value is int32.
(cf. https://github.com/pipe-cd/pipecd/blob/master/pkg/config/piped.go#L476)

So I have modified the `type` unmarshal logic so that it works fine when a number representing a certain Cloud Provider type is directly specified in the `type`.

After this PR merged, the Cloud Provider will be displayed correctly.

image

**Consideration**

The Cloud Provider Type displayed in the dialog shows an int32 value, but it might be more user-friendly to display a string converted from the numerical value.

Fixes #3731

```release-note
Fixed bug that Cloud Provider is always Kubernetes in "Show Piped config" dialog
```

### DIFF
--- a/pkg/config/piped.go
+++ b/pkg/config/piped.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/pipe-cd/pipecd/pkg/model"
@@ -451,7 +453,7 @@ func (r *HelmChartRegistry) Mask() {
 
 type PipedCloudProvider struct {
 	Name string                `json:"name,omitempty"`
-	Type model.ApplicationKind `json:"type,string,omitempty"`
+	Type model.ApplicationKind `json:"type,string"`
 
 	KubernetesConfig *CloudProviderKubernetesConfig `json:"kubernetesConfig,omitempty"`
 	TerraformConfig  *CloudProviderTerraformConfig  `json:"terraformConfig,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/config/piped.go
```diff
@@ -451,7 +453,7 @@ func (r *HelmChartRegistry) Mask() {
 
 type PipedCloudProvider struct {
 	Name string                `json:"name,omitempty"`
-	Type model.ApplicationKind `json:"type,string,omitempty"`
+	Type model.ApplicationKind `json:"type,string"`
```
Szarny, line 456: Remove `omitempty` so that Kubernetes-type (`0`) cloud provider will shown in dialog

khanhtc1202: I see 👍 While masking the configuration, I think it would be better if we use the enum string here instead of the enum value (instead of "0", should use KUBERNETES). wdyt?

Szarny: > I think it would be better if we use the enum string here instead of the enum value

I agree, but `PipedCloudProvider.Type` is referenced from so many files so a large modification is needed if we do this.
I made this fix because I thought I should finish fixing the bug first. 
What do you think?

khanhtc1202:
I will continue to work on this, so close the current implementation
/close

PR #3735, opened 2022-06-07
Fix bug that Cloud Provider is always Kubernetes in "Show Piped config" dialog 

### LINKED ISSUE #3731: Display wrong cloud provider in Show Piped config dialog

khanhtc1202:

Screen Shot 2022-06-07 at 17 35 48

The root cause of this issue is that we use the ApplicationKind (an enum) as the cloud provider type in the application configuration. Before submitting the piped configuration to the control plane, we do unmarshal/marshal the piped configuration, an enum value is cast to int32 and when we need to cast it back, it turns out not working (the below image shows the error in detail - thx @Szarny )

image

To resolve that, this PR added a new type CloudProviderType (define it equal to string) and use that in all places refer to pipedCfg.cloudProvider.

Fixes #3731 

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/driftdetector/detector.go
+++ b/pkg/app/piped/driftdetector/detector.go
@@ -96,7 +96,7 @@ func NewDetector(
 
 	for _, cp := range cfg.CloudProviders {
 		switch cp.Type {
-		case model.ApplicationKind_KUBERNETES:
+		case model.CloudProviderKubernetes:
 			sg, ok := stateGetter.KubernetesGetter(cp.Name)
 			if !ok {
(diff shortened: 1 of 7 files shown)


nghialv (approved):
Nice idea! 👍

PR #3745, opened 2022-06-10
Add "Restart Piped" command handler to launcher

Szarny:

According to the https://github.com/pipe-cd/pipecd/issues/3534, it's useful that there is the feature to restart piped from the web console.

To implement this feature, I have added a command handler to handle RestartPiped Command to launcher as a continuation of #3725 

Note that this PR does not implement any specific API or UI changes, so there are no user-facing changes.

Fixes partially #3534 

```release-note
NONE
```


pkg/app/launcher/cmd/launcher/launcher.go
```diff
@@ -267,6 +305,21 @@ func (l *launcher) run(ctx context.Context, input cli.Input) error {
...
+			case cmd := <-cmdCh:
+				l.handleCommand(ctx, input, cmd)
+
+			case <-ctx.Done():
+				input.Logger.Info("a worker has been stopped")
+				return
+			}
+		}
+	}
+	go commandHandler(ctx, l.commandCh)
```
knanao, line 321: How about using goroutine with errorGroup instead of creating a simple goroutine?

pkg/app/launcher/cmd/launcher/launcher.go
```diff
@@ -279,6 +332,7 @@ func (l *launcher) run(ctx context.Context, input cli.Input) error {
 			select {
 			case <-ticker.C:
 				// Don't return an error to continue piped execution.
+				l.enqueueNewCommands(ctx, input)
```
khanhtc1202, line 335: I feel it's a bit overdo here. Since the command restart piped is not that much and it should be handled immediately on given, making a queue of commands for that is unneeded 🤔 Should we just list commands and restart it instead of sending them to the commandhandler? wdyt @knanao @Szarny

pkg/app/launcher/cmd/launcher/launcher.go
```diff
@@ -302,35 +356,109 @@ func (l *launcher) run(ctx context.Context, input cli.Input) error {
...
+	}
+}
+
+func (l *launcher) handleCommand(ctx context.Context, input cli.Input, cmd model.ReportableCommand) {
+	logger := input.Logger.With(
+		zap.String("command", cmd.Id),
+	)
+	logger.Info("received a restart piped command to handle")
+
+	l.detectRestartCommand = true
```
khanhtc1202, line 418: Just feel like this one is not an effective move to implement this feature, we need on ticker (to fetch the commands) and one other ticker(util launcher self checkInterval) to restart the piped. Should we rewrite some part where we restart piped by launcher to reuse that?
ref: current rerun logic https://github.com/pipe-cd/pipecd/blob/master/pkg/app/launcher/cmd/launcher/launcher.go#L248-L262

pkg/app/launcher/cmd/launcher/launcher.go
knanao: I got your point and I feel it's better but it's up to you since there are already some implementations like this. 🙌

pkg/app/launcher/cmd/launcher/launcher.go
khanhtc1202: we can keep the use of this `l.detectRestartCommand` flag but instead of forking a new go routine to just change this flag value, we can make this in the current `enqueueNewCommands`, and in execute function, we have `shouldRelaunch()` function which check for the flag and restart the piped for us 🤔

khanhtc1202:
duplicated with https://github.com/pipe-cd/pipecd/pull/3831 so close this

PR #3753, opened 2022-06-14
Fix to add safe.directory before running git command

knanao:
We got an [error](https://github.com/pipe-cd/pipecd/runs/6872568012?check_suite_focus=true) like we could not run the git command on a docker container in actions-gh-release due to [this](https://github.com/actions/checkout/issues/760).
Hence added the workspace as a safe directory before running the git command.

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice

nghialv (approved):
👍

PR #3760, opened 2022-06-21
Fix to separete the milestones for .pipe/ and app config

knanao:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/eventwatcher/eventwatcher.go
+++ b/pkg/app/piped/eventwatcher/eventwatcher.go
@@ -81,9 +81,13 @@ type watcher struct {
 
 	// All cloned repository will be placed under this.
 	workingDir string
-	// Maximum timestamp of the last Event read.
+	// Maximum timestamp of the last Event read from .pipe/.
 	// A map from repo-id to the UNIX timestamp that has been read.
 	milestoneMap sync.Map
+
+	// Maximum timestamp of the last Event read from the application config.
+	// A map from repo-id to the UNIX timestamp that has been read.
+	executionMilestoneMap sync.Map
 	// Cache for the last scanned commit and event watcher configs for each application.
 	lastScannedConfig sync.Map
 }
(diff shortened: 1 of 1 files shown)


pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -81,9 +81,13 @@ type watcher struct {
...
 	// All cloned repository will be placed under this.
 	workingDir string
-	// Maximum timestamp of the last Event read.
+	// Maximum timestamp of the last Event read from .pipe/.
 	// A map from repo-id to the UNIX timestamp that has been read.
 	milestoneMap sync.Map
+
+	// Maximum timestamp of the last Event read from the application config.
+	// A map from repo-id to the UNIX timestamp that has been read.
+	executionMilestoneMap sync.Map
```
khanhtc1202, line 90: sorry, I can not get the point of having a separate sync.Map for events defined in-app config instead of using the same one for both event kind and in-app define. Could you provide an unexpected case that occurs if we don't create this map?

knanao: Sure.
The current event watcher process flow is as follows.
1. Refer to the event watcher config file under .pipe/ and if there is an event that is not handled, it is processed and the createdAt of the last handled event is saved as a milestone to load just only the event which has not loaded yet in the next iteration.

2. Extract the event watcher config from the target application config, and process any not handled events.

Hence, If we use the same one, it could be happened that the values of `notHandledEvents` will differ depending on the order in which the event is registered because the milestone was updated.

https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/eventwatcher/eventwatcher.go#L337

khanhtc1202: Ah, got it. `ListNotHandled` interface depends on milestone 👍 Thanks for the explanation.

khanhtc1202 (approved):
👍

nghialv (approved):
👍

PR #3764, opened 2022-06-22
Fix the format of Cloud Run

knanao:

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch

nghialv (approved):
Right. That is the official name.

PR #3769, opened 2022-06-26
Ensure to have PlatformProvider value in application model on adding new application

khanhtc1202:

By PR #3767 we only ensure that applications that already existed in the pipecd datastore will have the value PlatformProvider based on CloudProvider value, this PR ensures that all new adding applications will have that value for PlatformProvider field too.

Fixes #

```release-note
NONE
```


nghialv (approved):
👍

knanao (approved):

PR #3797, opened 2022-07-06
Make gen/release-docs command ouput consistent

khanhtc1202:

This PR makes the output of `make gen/release-docs` consistent. Previously, once the command run, if we run that command again, a new directory `docs-dev/` will be copied and stored under `docs-v0.x.x` directory which is annoying. Besides, I added logic to ensure update `docs/config.toml` file to add a route to the new version docs only if the version docs route has not existed yet.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/gen-release-docs.sh
+++ b/hack/gen-release-docs.sh
@@ -40,6 +40,7 @@ menu:
 EOT
 
 # Create new $CONTENT_DIR/docs-$1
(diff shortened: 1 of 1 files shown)


nghialv (approved):
Nice improvement! This is exactly what I also want.

knanao (approved):
Nice catch!

PR #3798, opened 2022-07-07
Add the application integration test for MySQL

knanao:
Add the integration test for MySQL by several PRs.
This PR just only tests the application model.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/datastore/mysql/mysql.go
+++ b/pkg/datastore/mysql/mysql.go
@@ -17,6 +17,7 @@ package mysql
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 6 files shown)


pkg/datastore/mysql/mysql.go
```diff
@@ -80,6 +81,10 @@ func NewMySQL(url, database string, opts ...Option) (*MySQL, error) {
 // Find implementation for MySQL
 func (m *MySQL) Find(ctx context.Context, col datastore.Collection, opts datastore.ListOptions) (datastore.Iterator, error) {
 	kind := col.Kind()
+	if opts.Cursor != "" && len(opts.Orders) == 0 {
+		return nil, errors.New("opts.Cursor also requires Orders to be set")
+	}
+
```
knanao, line 87: We should add this line because panic would occur without this.

PR #3814, opened 2022-07-21
Add platformProviders config to piped

khanhtc1202:

This PR only adds PlatformProviders as piped configuration option, all logic around usage of CloudProviders are remained and will be addressed by another PR.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/executor/terraform/terraform.go
+++ b/pkg/app/piped/executor/terraform/terraform.go
@@ -80,8 +80,8 @@ func findTerraform(ctx context.Context, version string, lp executor.LogPersister
 	return path, true
 }
 
-func findCloudProvider(in *executor.Input) (name string, cfg *config.PlatformProviderTerraformConfig, found bool) {
-	name = in.Application.CloudProvider
+func findCloudProvider(in *executor.Input) (cfg *config.PlatformProviderTerraformConfig, found bool) {
+	var name = in.Application.CloudProvider
 	if name == "" {
 		in.LogPersister.Error("Missing the CloudProvider name in the application configuration")
 		return
--- a/pkg/app/piped/executor/terraform/terraform.go
+++ b/pkg/app/piped/executor/terraform/terraform.go
@@ -80,7 +80,7 @@ func findTerraform(ctx context.Context, version string, lp executor.LogPersister
 	return path, true
 }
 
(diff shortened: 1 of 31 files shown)


pkg/app/piped/executor/terraform/terraform.go
```diff
@@ -80,8 +80,8 @@ func findTerraform(ctx context.Context, version string, lp executor.LogPersister
 	return path, true
 }
 
-func findCloudProvider(in *executor.Input) (name string, cfg *config.CloudProviderTerraformConfig, found bool) {
-	name = in.Application.CloudProvider
+func findCloudProvider(in *executor.Input) (cfg *config.PlatformProviderTerraformConfig, found bool) {
```
knanao, line 83:
```suggestion
func findPlatformProvider(in *executor.Input) (cfg *config.PlatformProviderTerraformConfig, found bool) {
```

pkg/config/piped.go
```diff
@@ -194,7 +197,7 @@ func (s *PipedSpec) HasCloudProvider(name string, t model.ApplicationKind) bool
 }
 
 // FindCloudProvider finds and returns a Cloud Provider by name and type.
-func (s *PipedSpec) FindCloudProvider(name string, t model.ApplicationKind) (PipedCloudProvider, bool) {
+func (s *PipedSpec) FindCloudProvider(name string, t model.ApplicationKind) (PipedPlatformProvider, bool) {
```
knanao, line 200:
```suggestion
func (s *PipedSpec) FindPlatformProvider(name string, t model.ApplicationKind) (PipedPlatformProvider, bool) {
```

pkg/app/piped/executor/terraform/terraform.go
khanhtc1202: This PR only adds PlatformProviders as configuration of piped.
Lets me address those suggestions in another PR since this changed the behavior of Piped and I want to address those things on one PR 🙏

pkg/config/piped.go
khanhtc1202: same as https://github.com/pipe-cd/pipecd/pull/3814#discussion_r926254623

knanao (approved):
👍

nghialv (approved):
Here you go!

PR #3827, opened 2022-08-01
Fix fail clone PipedSpec object

khanhtc1202:

The root cause of this problem is that we use a customized UnmarshalJSON function for objects of kind PlatformProvider (and AnalysisProvider, SecretManagement, etc), while in the Clone function, we implement deep cloning object by marshal the object to JSON then, re-unmarshal it to a new object.
The re-unmarshal logic based on customized UnmarshalJSON and its fields' names are changed, hence the original MarshalJSON can not marshal the object to a correct JSON => the second time we re-unmarshal from wrong JSON to object, the object changed from its original version.

To handle that issue, this PR added customized MarshalJSON logic corresponding to the customized UnmarshalJSON logic, such that we can marshal an object of those kinds correctly.

Thanks @nghialv for pointing out this 👍 

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/config/piped.go
+++ b/pkg/config/piped.go
@@ -467,22 +467,54 @@ type PipedPlatformProvider struct {
 	Name string                     `json:"name"`
 	Type model.PlatformProviderType `json:"type"`
 
-	KubernetesConfig *PlatformProviderKubernetesConfig `json:"kubernetesConfig,omitempty"`
-	TerraformConfig  *PlatformProviderTerraformConfig  `json:"terraformConfig,omitempty"`
-	CloudRunConfig   *PlatformProviderCloudRunConfig   `json:"cloudRunConfig,omitempty"`
-	LambdaConfig     *PlatformProviderLambdaConfig     `json:"lambdaConfig,omitempty"`
-	ECSConfig        *PlatformProviderECSConfig        `json:"ecsConfig,omitempty"`
+	KubernetesConfig *PlatformProviderKubernetesConfig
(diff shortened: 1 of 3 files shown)


pkg/config/piped.go
```diff
@@ -466,22 +465,54 @@ type PipedPlatformProvider struct {
 	Name string                     `json:"name"`
 	Type model.PlatformProviderType `json:"type"`
 
-	KubernetesConfig *PlatformProviderKubernetesConfig `json:"kubernetesConfig,omitempty"`
-	TerraformConfig  *PlatformProviderTerraformConfig  `json:"terraformConfig,omitempty"`
-	CloudRunConfig   *PlatformProviderCloudRunConfig   `json:"cloudRunConfig,omitempty"`
-	LambdaConfig     *PlatformProviderLambdaConfig     `json:"lambdaConfig,omitempty"`
-	ECSConfig        *PlatformProviderECSConfig        `json:"ecsConfig,omitempty"`
```
nghialv, line 473: Why did we remove these tags?

khanhtc1202: It's not being used anyway, since when we marshal the object to JSON, only one of them values is being used and displayed as `config` in the output of MarshalJSON function.
Screen Shot 2022-08-01 at 18 14 58

nghialv: Nice point. Got it.

nghialv (approved):
Great!

knanao (approved):

PR #3829, opened 2022-08-01
Update codegen container image

khanhtc1202:

This change the codegen container image by remove its entrypoint, the codegen.sh script will be refer from source code instead of from container image.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/gen.yaml
+++ b/.github/workflows/gen.yaml
@@ -12,8 +12,8 @@ jobs:
       uses: actions/checkout@v3
     - name: Generate code
       uses: docker://ghcr.io/pipe-cd/codegen@sha256:5e10b6416d507d54b8ca9ab58820f11633d88af1d9fdcce65b5e568abbbe3d9e #v0.32.0
-      run: ./tool/codegen/codegen.sh
       with:
+        entrypoint: ./tool/codegen/codegen.sh
         args: /github/workspace
     - name: Show Git status
       shell: bash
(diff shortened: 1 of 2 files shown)


nghialv (approved):
👍

knanao (commented):
Sorry, I'm not sure about the intent of this PR.
Could you tell me about it?

khanhtc1202:
> Sorry, I'm not sure about the intent of this PR. Could you tell me about it?

Currently, we use a versioned container image for the codegen validation CI step, that container image contains a specified `codegen.sh` script (ref: [Dockerfile](https://github.com/pipe-cd/pipecd/blob/master/tool/codegen/Dockerfile#L59-L60)) and is added on container image build time. The issue with this approach is that, if we want to change the content of that codegen script itself, the new script will not be referred and the CI codegen validation step will always fail (it uses the old generate script file attached to the container image on build time).

This PR makes the CI step refer to the codegen script on run so that changes on the script will be reflected, and the container image for that CI step only contains the required lib/tools to run the script, not the script itself.

knanao:
I got your point but this docker image would be just base-image, so I'm glad you to change `make gen/code` too.

khanhtc1202:
> I got your point but this docker image would be just base-image, so I'm glad you to change `make gen/code` too.

Sure, but lets me address it by a separated PR, (also need to update this version here to [GithubAction - codegen](https://github.com/pipe-cd/pipecd/blob/master/.github/workflows/gen.yaml#L14)) after I can confirm the works of this new base container image 😄

knanao (approved):

khanhtc1202:
@nghialv @knanao sorry guys, could you approve again 🙏

knanao (approved):

knanao (approved):

nghialv (approved):
👍

PR #3836, opened 2022-08-09
Fix duplicated providers listed in application form

khanhtc1202:

The root cause of this issue is that when the currently running piped reports its metadata to the controlplane, only piped's platformProviders values in controlplane database are updated, but we forget to remove the old cloudProviders value, such that it exists and will be returned later in the web console application adding form.

Screen Shot 2022-08-09 at 11 20 09

Fixes #

```release-note
Fix platform providers shows duplicated values in application form
```

### DIFF
--- a/pkg/datastore/pipedstore.go
+++ b/pkg/datastore/pipedstore.go
@@ -224,6 +224,7 @@ func (s *pipedStore) UpdateDesiredVersion(ctx context.Context, id, version strin
 
 func (s *pipedStore) UpdateMetadata(ctx context.Context, id, version, config string, pps []*model.Piped_PlatformProvider, repos []*model.ApplicationGitRepository, se *model.Piped_SecretEncryption, startedAt int64) error {
 	return s.update(ctx, id, func(piped *model.Piped) error {
+		piped.CloudProviders = nil
 		piped.PlatformProviders = pps
 		piped.Repositories = repos
 		piped.SecretEncryption = se


pkg/datastore/pipedstore.go
```diff
@@ -224,6 +224,7 @@ func (s *pipedStore) UpdateDesiredVersion(ctx context.Context, id, version strin
 
 func (s *pipedStore) UpdateMetadata(ctx context.Context, id, version, config string, pps []*model.Piped_PlatformProvider, repos []*model.ApplicationGitRepository, se *model.Piped_SecretEncryption, startedAt int64) error {
 	return s.update(ctx, id, func(piped *model.Piped) error {
+		piped.CloudProviders = nil
```
knanao, line 227: I have one point I would like to confirm.
Is this safe for the user using only `cloudProviders` settings in piped?

khanhtc1202: The piped execution only refers to its local configuration and itself knows about the CloudProvider (suppose it has only CloudProviders and the piped runs the old version since the new one will merge all CloudProviders -> PlatformProviders), so I think it's okay since they know which "cloudProvider" to use on its side.

Otherwise, though we remove CloudProviders information from the piped model here, it only be used to show on the application adding form selection, so I think it's okay 🤔

knanao: I got your point! Thanks.

knanao (approved):

nghialv (approved):
Nice catch!

PR #3857, opened 2022-08-18
Make target group configuration as not required in ECS application configuration 

khanhtc1202:

In some specified cases, users want to be able to run an ECS task without exposing it to the net through LB. So we should accept ECS application configuration without targetGroups configuration.

Fixes #

```release-note
Make target group configuration as not required in ECS application configuration
```


nghialv (approved):
👍

knanao (approved):

PR #3904, opened 2022-09-12
Fix to check whether the name of role is built-in role or not

knanao:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/model/project.go
+++ b/pkg/model/project.go
@@ -413,7 +413,7 @@ func (p *Project) AddUserGroup(sso, role string) error {
 	if p.HasUserGroup(sso) {
 		return fmt.Errorf("%s is already being used. The SSO group must be unique", sso)
 	}
-	if !p.HasRBACRole(role) {
+	if !p.HasRBACRole(role) && !isBuiltinRBACRole(role) {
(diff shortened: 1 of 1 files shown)


pkg/model/project.go
```diff
@@ -413,7 +413,7 @@ func (p *Project) AddUserGroup(sso, role string) error {
 	if p.HasUserGroup(sso) {
 		return fmt.Errorf("%s is already being used. The SSO group must be unique", sso)
 	}
-	if !p.HasRBACRole(role) {
+	if !p.HasRBACRole(role) && !isBuiltinRBACRole(role) {
```
khanhtc1202, line 416: this looks like a leak to me, does it mean that there are some cases that `role` is not yet added into project RBAC roles but is being used as a role to add usergroup?

knanao: The build-in role is not added to a project hence we have to check whether the role which is used by the user group is a built-in role or not when adding a user group.

khanhtc1202: Ah get your point 👍 Then it's lgtm, thanks for the explanation.

khanhtc1202 (approved):
Nice catch

nghialv (approved):
👍

PR #3913, opened 2022-09-14
Fix to add a configuration reference into under the managing application

knanao:

Fixes #

```release-note
NONE
```


khanhtc1202:
I don't think this go like this, because the configuration-reference page contains others feature config (eg. EventWatcher) which is not placed under `managing-application` page, maybe we need to revise that as well 🤔

knanao:
@khanhtc1202 
Oh, I misunderstood it and I got the current docs design.
I feel it's a little discomfort to place EventWatcher ref under managing-application, so let's keep this. 🙌 

Thank you!

PR #3917, opened 2022-09-15
Fix the hack command for cherry-pick

knanao:
The `git ls-remote` command returns 0 status code regardless of the results without `--exit-code` option.
But we don't want to use one because we want to use `set -o errexit`.
Hence I fixed to handle the output result instead of that.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/cherry-pick.sh
+++ b/hack/cherry-pick.sh
@@ -59,8 +59,8 @@ echo
 # Check whether remote branch exists
 echo "+++ Checking whether remote branch exists..."
 BASE_TAG=$(echo ${BRANCH} | sed -e 's/.*-//' -e 's/x/0/g')
-git ls-remote --exit-code --heads origin ${BRANCH} > /dev/null 2>&1
-if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


hack/cherry-pick.sh
```diff
@@ -59,8 +59,8 @@ echo
 # Check whether remote branch exists
 echo "+++ Checking whether remote branch exists..."
 BASE_TAG=$(echo ${BRANCH} | sed -e 's/.*-//' -e 's/x/0/g')
-git ls-remote --exit-code --heads origin ${BRANCH} > /dev/null 2>&1
-if [ $? -ne 0 ]; then
+IS_EXIST=$(git ls-remote --heads origin ${BRANCH} | wc -l)
+if [ $IS_EXIST -ne 1 ]; then
```
khanhtc1202, line 63: just want to confirm, is there any case that the IS_EXIST value here is greater than 1, looks like this count something here 🤔

knanao: I suppose that case does not exist because we can not create same name branches.

khanhtc1202 (approved):
👍

nghialv (approved):
👍

PR #3924, opened 2022-09-27
Remove sealsecret code

khanhtc1202:

Deprecated feature, use SecretManagement instead
https://pipecd.dev/docs/user-guide/managing-application/secret-management/

Fixes #

```release-note
NONE
```


knanao:
@khanhtc1202 
Sorry, plz rebase this from a master branch.

knanao (commented):
Thank you!
It's enough time passed to un-support this feature, and let's not forget to add a breaking change section to the next release note.
https://github.com/pipe-cd/pipecd/blob/master/docs/content/en/blog/releases/v0.10.0.md#deprecated-features

btw, how about changing `Does this PR introduce a user-facing change?` to announce this breaking change?

khanhtc1202:
@knanao Good catch, but as far as I know from the codebase, looks like the SealedSecret manifest kind are not being used (un-supported) for a while already means that users can not use it already, so I guess we can ignore that from the release note. This PR should be counted as removing unused code, wdyt?

knanao (approved):
@khanhtc1202 
Oh, I see. Nice catch!

PR #3926, opened 2022-09-28
Delete the deploymentChain from built-in roles

knanao:
Currently, the DeploymentChain resource is not used to call gPRC, so delete this from built-in roles.

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch 👍

PR #3932, opened 2022-09-30
Quickstart command

khanhtc1202:

This PR adds `pipectl quickstart` command.

The current usage:
```shell
$ pipectl quickstart
$ pipectl quickstart --uninstall
```

Fixes #

```release-note
Add pipectl quickstart command
```

### DIFF
--- a/cmd/pipectl/main.go
+++ b/cmd/pipectl/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/event"
 	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/piped"
 	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/planpreview"
+	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/quickstart"
 	"github.com/pipe-cd/pipecd/pkg/cli"
 )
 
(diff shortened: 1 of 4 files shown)


knanao:
Umm, I agree with removing the previous command but I don't think this is essential.
If these prerequisites were reduced, it would be a nice improvement but this looks just like a rapper. 
Is there a specific meaning for preparing this command?

khanhtc1202:
since the install controlplane is just a wrapper to helm command and not bring a well defined enough interface, to me it just add complexity to the usage of the pipectl command, that why I think it's better to remove it. Besides, we haven't officially introduced it yet, so remove it don't cause any issue to our users, I guess.

If it's really a need to have that kind of sub command, I will rethink and add it later, wdyt?

khanhtc1202:
And for the new `pipectl quickstart` I omit the fact that we need to preinstall helm so that helm not familiar users can easily make a try with PipeCD in their local, that's the point 👀

khanhtc1202:
I separated the removing `pipectl install` command to this PR, lets merge it first before this PR 🙏 
https://github.com/pipe-cd/pipecd/pull/3933

knanao:
> ...helm not familiar users can easily make a try with PipeCD in their local

I suppose many users will copy and paste the code when installing a quickstart.
Hence I think the users don't need to be familiar with helm.

> And for the new pipectl quickstart I omit the fact that we need to preinstall helm 

However, this seems enough value to add a new command because this is an effective way to reduce the user's labor and get rid of any barriers like allergic reactions.

khanhtc1202:
> However, this seems enough value to add a new command because this is an effective way to reduce the user's labor and get rid of any barriers like allergic reactions.

yes, that's my point 👍

knanao (approved):
I see, here you go!

PR #3947, opened 2022-10-13
Introduce pipectl quickstart command

khanhtc1202:

The quickstart command for pipecd try out will run like this

https://user-images.githubusercontent.com/32532742/195658913-ab18145c-3932-4222-8504-151b518460f7.mp4

Fixes #

```release-note
New pipectl quickstart command
```

### DIFF
--- a/pkg/app/pipectl/cmd/quickstart/quickstart.go
+++ b/pkg/app/pipectl/cmd/quickstart/quickstart.go
@@ -24,6 +24,7 @@ import (
 	"os/exec"
 	"path"
 
+	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
 
 	"github.com/pipe-cd/pipecd/pkg/cli"
@@ -32,8 +33,12 @@ import (
 
 const (
 	defaultHelmVersion = "3.8.2"
-	helmReleaseName    = "pipecd"
-	helmChartRepoName  = "oci://ghcr.io/pipe-cd/chart/pipecd"
+
+	helmControlPlaneReleaseName = "pipecd"
+	helmPipedReleaseName        = "piped"
+
+	helmChartControlPlaneRepoName = "oci://ghcr.io/pipe-cd/chart/pipecd"
+	helmChartPipedRepoName        = "oci://ghcr.io/pipe-cd/chart/piped"
 
 	helmQuickstartValueRemotePath = "https://raw.githubusercontent.com/pipe-cd/pipecd/%s/quickstart/control-plane-values.yaml"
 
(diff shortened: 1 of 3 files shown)


knanao (commented):
Nice work!
Will you fix this command to create a default piped at sever side automatically?
If so, I suppose it's better to include that at this time.

khanhtc1202:
Basically I tried to automate the piped registration but it turns out not work smoothly. We can:
- making the piped id and key as seeding data in datastore, in that case making filedb as quickstart datastore makes think easier but a bit risky in case users try to use quickstart as their real env (as we discussed before).
- making pipectl command itself makes requests to register piped id/key in background and users only input gitRemoteRepo (by anyway). In such case, pipectl have to bypass webservice registerPiped rpc and it risky as well.

So I think it would be better to make it straight forward, users have to register piped via the web console pop up by pipectl (just like gcloud cli from gcp). If you have any better flow, please share 😁

knanao:
@khanhtc1202 
> making the piped id and key as seeding data in datastore...

I think this is not a big deal because a quickstart env must not be used as someone's real env.

> In such case, pipectl have to bypass webservice registerPiped rpc and it risky as well.

I wonder what points you feel risky? I would say you just add the interface to `APIServiceClient`.

khanhtc1202:
For the first one, creating piped with seeding data is not a big deal as you said, but the piped id and key which be used as the seeding data need to be shared in code as a special one and in pipectl quickstart command code, we need a way to know about that id (making it as const for instance), that's not really nice IMO.

For the second one, to make requests to the control plane, the pipectl needs an API key which for now can only be registered via the web console. In that case, it turns out that we need to do the same thing I do in this PR (open up the browser and make users register for the API key to use pipectl, and so on), which makes the issue the same with registering the piped. If we allow only RegisterPiped rpc without an API key, that is kind of trying to bypass our API auth logic ourselves, which is a bit risky and adds unneeded complexity to the API side.

Consider the two methods, I guess copy the gcloud command flow is better 🤔

knanao:
> but the piped id and key which be used as the seeding data need to be shared in code as a special one and in pipectl quickstart command code

Right, but it's okay to deal with a default piped because it's also possible to create a piped and keys as you know.

> to make requests to the control plane, the pipectl needs an API key which for now can only be registered via the web console.

What you mention is `webservice` which should be called by literally web service like control plane.
But I think we should focus on `apiservice` in this case like this because we're discussing pipeclt.
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/service/apiservice/service.proto

However, it's up to you to make the final decision. 👍

khanhtc1202:
@knanao 
> What you mention is webservice which should be called by literally web service like control plane.
But I think we should focus on apiservice in this case like below because we're discussing pipeclt.

Yes, I'm talking about the pipectl thus I mention the API key for pipectl calling API service because we need that API key as mentioned here
https://pipecd.dev/docs/user-guide/command-line-tool/#authentication
Are we on the same page? 👀

knanao:
> Even if we make RegisterPiped rpc available in API service, we still need to register an API key to enable pipectl API client to request to the control plane server for that registration.

Oh, I see and that makes sense.
Let's keep this improvement 🙌

pkg/app/pipectl/cmd/quickstart/quickstart.go
```diff
@@ -87,36 +100,195 @@ func (c *command) run(ctx context.Context, input cli.Input) error {
...
+		"-n",
+		c.namespace,
+	}
+	var stderr bytes.Buffer
+	cmd := exec.CommandContext(ctx, kubectl, args...)
+	cmd.Stderr = &stderr
+
+	// Wait the control plane service to be ready.
+	// TODO: Find a better way to handle this instead of time sleep
+	time.Sleep(time.Minute)
```
knanao, line 225: How about using `kubectl rollout status` command or a `--timeout` option for helm?

knanao (approved):
Nice work!

PR #3952, opened 2022-10-18
Fix typos on notification event

kurochan:
Fix notification event name in docs and test from `DEPLOYMENT_STARTED` to `DEPLOYMENT_TRIGGERED`.


khanhtc1202 (approved):
Nice catch 👍

PR #3963, opened 2022-10-21
Fix the menu design for APIKey

knanao:
In order to match with another UI, fixed the size of width.

# before
Screen Shot 2022-10-21 at 16 59 45

# afeter
Screen Shot 2022-10-21 at 16 59 34

Fixes #

```release-note
NONE
```

### DIFF
--- a/web/src/components/settings-page/api-key/index.tsx
+++ b/web/src/components/settings-page/api-key/index.tsx
@@ -66,10 +66,8 @@ const EmptyTableContent = memo(function EmptyTableContent() {
   );
 });
 
-const ITEM_HEIGHT = 48;
 const menuStyle = {
   style: {
-    maxHeight: ITEM_HEIGHT * 4.5,
     width: "25ch",
   },
 };
--- a/web/src/components/settings-page/api-key/index.tsx
+++ b/web/src/components/settings-page/api-key/index.tsx
@@ -66,6 +66,14 @@ const EmptyTableContent = memo(function EmptyTableContent() {
   );
 });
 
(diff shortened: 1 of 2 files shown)


web/src/components/settings-page/api-key/index.tsx
```diff
@@ -166,6 +174,7 @@ export const APIKeyPage: FC = memo(function APIKeyPage() {
         open={Boolean(anchorEl)}
         anchorEl={anchorEl}
         onClose={handleCloseMenu}
+        PaperProps={menuStyle}
```
khanhtc1202, line 177: could you attach a screenshot to show what this change makes to the UI?  👀

knanao: fixed the description. 🙏

khanhtc1202: I see, to set the width of the menu, should we use `className` instead? and besides, looks like it's redundant to set the maxHeight as well, so I guess it's better to remove unneeded code 🤔

knanao: In this case, `className`doesn't work because we need to override the underlying paper class.
If you don't wanna use `PaperProps`, can use `classes={{paper: classes.menu}}` instead of that/

khanhtc1202: got your point, then lets use PaperProps and remove unneeded maxHeight, wdyt? 👀

khanhtc1202 (approved):
🙌

PR #3964, opened 2022-10-24
Feat the new feature for RBAC

knanao:

https://user-images.githubusercontent.com/50069775/197912760-038b51d4-abdf-4d4a-a6c9-9ef394ae4d10.mov

Fixes #

```release-note
Enable to manage a custom RBAC role through control plane
```

### DIFF
--- a/web/src/components/settings-page/project/components/rbac-form/components/role/index.tsx
+++ b/web/src/components/settings-page/project/components/rbac-form/components/role/index.tsx
@@ -1,9 +1,7 @@
 import {
-  Menu,
-  MenuItem,
+  Button,
   Paper,
   Table,
-  IconButton,
   TableBody,
   TableCell,
   makeStyles,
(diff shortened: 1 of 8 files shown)


web/src/components/settings-page/project/components/rbac-form/components/role/index.tsx
```diff
@@ -34,58 +43,84 @@ const useStyles = makeStyles((theme) => ({
...
-      policy.resourcesList.map((resource) => {
-        resources.push(RBAC_RESOURCE_TYPE_TEXT[resource.type]);
+  const handleSubmit = useCallback(
+    (values: { name: string; policies: string }) => {
+      const params = {
+        name: values.name,
+        policies: parseRBACPolicies({ policies: values.policies }),
+      };
+      dispatch(addRBACRole(params)).then((result) => {
+        if (addRBACRole.fulfilled.match(result)) {
```
khanhtc1202, line 63: I'm still a bit worried about this one. Let's say, this `then` will run after the dispatch function finish, then if the API handling is still not finished, it means that if the check failed, and even if the API runs successfully, the toast and so on handlers will not be triggered. This way to handle only ensures that there will be no error toast and success toast showing at the same time, but in other cases problem existed, wdyt?

khanhtc1202: lets check and handle that later since we don't have plan for this currently 🤔

khanhtc1202 (approved):
Nice work 👍

web/src/components/settings-page/project/components/rbac-form/components/role/index.tsx
knanao: I see, thanks for reviewing.

PR #3967, opened 2022-10-28
Fix the docs of eventwatcher

knanao:

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/content/en/docs-dev/user-guide/event-watcher.md
+++ b/docs/content/en/docs-dev/user-guide/event-watcher.md
@@ -18,8 +18,7 @@ While it empowers you to build pretty versatile workflows, the canonical use cas
 This guide walks you through configuring Event watcher and how to push an Event.
 
 ## Prerequisites
-If you want to use EventWatcher configuration files under the `.pipe/` directory, before we get into configuring EventWatcher, be sure to configure Piped. See [here](/docs/user-guide/managing-piped/configuring-event-watcher/) for more details.
-In other cases, there is no prerequisite. Let's move on to Usage.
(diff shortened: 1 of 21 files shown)


docs/content/en/docs-dev/user-guide/event-watcher.md
```diff
@@ -18,8 +18,7 @@ While it empowers you to build pretty versatile workflows, the canonical use cas
 This guide walks you through configuring Event watcher and how to push an Event.
 
 ## Prerequisites
-If you want to use EventWatcher configuration files under the `.pipe/` directory, before we get into configuring EventWatcher, be sure to configure Piped. See [here](/docs/user-guide/managing-piped/configuring-event-watcher/) for more details.
-In other cases, there is no prerequisite. Let's move on to Usage.
+Before we get into configuring EventWatcher, be sure to configure Piped. See [here](/docs/user-guide/managing-piped/configuring-event-watcher/) for more details.
```
khanhtc1202, line 21: This is only true in case of using `.pipe/` pattern, isn't it?

knanao: Sorry, we need to configure `repoID` hence fixed like this.

docs/content/en/docs-dev/user-guide/managing-piped/configuring-event-watcher.md
```diff
@@ -27,6 +27,8 @@ spec:
 ```
 
 ### [optional] Specify Eventwatcher files Piped will use
+>NOTE: This way is valid only for defining events using `.pipe/`.
```
khanhtc1202, line 30: should we add a link to the docs to this `.pipe/` pattern. I feel the docs are a bit complicated here.

khanhtc1202 (approved):
Nice improvement 👍

PR #3974, opened 2022-11-03
pipectl delete command

### LINKED ISSUE #3809: pipectl delete command
I would like to remove an application with the pipectl command
If you already have the command, please let me know.
If not, I would appreciate your consideration to add it.

funera1:
Add the ability to remove applications from the command line tool.

Fixes #3809 

```release-note
Add pipectl delete application subcommand.
```


pkg/app/pipectl/cmd/application/delete.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020 The PipeCD Authors.
```
khanhtc1202, line 1:
```suggestion
// Copyright 2022 The PipeCD Authors.
```

pkg/app/pipectl/cmd/application/delete.go
```diff
@@ -0,0 +1,67 @@
...
+	appID string
+}
+
+func newDeleteCommand(root *command) *cobra.Command {
+	c := &delete{
+		root: root,
+	}
+	cmd := &cobra.Command{
+		Use:   "delete",
+		Short: "Delete application.",
```
khanhtc1202, line 39:
```suggestion
		Short: "Delete an application.",
```

pkg/app/pipectl/cmd/application/delete.go
```diff
@@ -0,0 +1,67 @@
...
+		RunE:  cli.WithContext(c.run),
+	}
+
+	cmd.Flags().StringVar(&c.appID, "app-id", c.appID, "The application ID.")
```
khanhtc1202, line 43: nits, the flag name can be `id`, but it's up to you 👍

funera1: Thanks for the code review!
I think ```app-id``` flag is better than ```id``` flag, because we can see at a glance what the id is and the flag of ```pipectl application get``` is also ```app-id```, so it's consistent.

khanhtc1202: Nice catch, then lets keep it as is 👍

khanhtc1202 (approved):
Nice improvement! Thank you 🙌

funera1:
I'm sorry.
I hit the update branch button by mistake.

PR #3991, opened 2022-11-18
Fix to override namespace if set for k8s manifest

knanao:
Currently, `pipecd.dev/resource-key` annotation is wrong because of this, which means every key will have been `apiVersion:ResourceKind:default:appName` if not having it in the manifest directly although the namespace is set in the `app.config.yaml`.

This will resolve the problem that k8s resources aren't pruned even if running Quick Sync

Fixes #

```release-note
Fix the problem that k8s resources are not pruned even if running Quick Sync
```


khanhtc1202 (approved):
Nice catch

PR #4004, opened 2022-11-29
Migrate from legacy loaders to asset module

knanao:
`file-loader` is already deprecated for webpack v5 hence migrate to asset module.
https://webpack.js.org/guides/asset-modules/#custom-data-uri-generator

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch 🍠

PR #4006, opened 2022-11-29
Fix to escape the user-provided value for ops

knanao:
This is not necessary but escapes special characters like <, >, &, ' and " to avoid unexpected behavior.

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
🍠

PR #4033, opened 2022-11-30
Keep new line at end of file in yamlprocessor

kurochan:
Currently, yamlprocessor doesn't respect new line at end of file if the input file contains it or not.
This behavior causes unnecessary differences and should be corrected.

For example, EventWatcher generates these changes.
image

Fixes #

```release-note
Keep new line at end of file in yamlprocessor
```

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/manifest.go
+++ b/pkg/app/piped/platformprovider/kubernetes/manifest.go
@@ -224,11 +224,16 @@ func ParseManifests(data string) ([]Manifest, error) {
 		manifests = make([]Manifest, 0, len(parts))
 	)
 
-	for _, part := range parts {
+	newLineAtEOF := strings.HasSuffix(data, "\n")
+	for i, part := range parts {
 		// Ignore all the cases where no content between separator.
 		part = strings.TrimSpace(part)
 		if len(part) == 0 {
 			continue
+		}
+		// Append new line which trim by document separator.
+		if i != len(parts)-1 || newLineAtEOF {
+			part += "\n"
 		}
 		var obj unstructured.Unstructured
 		if err := yaml.Unmarshal([]byte(part), &obj); err != nil {
--- a/pkg/app/piped/platformprovider/kubernetes/manifest.go
+++ b/pkg/app/piped/platformprovider/kubernetes/manifest.go
@@ -224,15 +224,13 @@ func ParseManifests(data string) ([]Manifest, error) {
 		manifests = make([]Manifest, 0, len(parts))
 	)
 
-	newLineAtEOF := strings.HasSuffix(data, "\n")
 	for i, part := range parts {
 		// Ignore all the cases where no content between separator.
-		part = strings.TrimSpace(part)
-		if len(part) == 0 {
+		if len(strings.TrimSpace(part)) == 0 {
 			continue
 		}
 		// Append new line which trim by document separator.
-		if i != len(parts)-1 || newLineAtEOF {
+		if i != len(parts)-1 {
 			part += "\n"
 		}
 		var obj unstructured.Unstructured
--- a/pkg/yamlprocessor/yamlprocessor.go
+++ b/pkg/yamlprocessor/yamlprocessor.go
@@ -25,6 +25,7 @@ import (
 
 type Processor struct {
 	file *ast.File
+	newLineAtEOF bool
 }
 
 func NewProcessor(data []byte) (*Processor, error) {
@@ -33,8 +34,14 @@ func NewProcessor(data []byte) (*Processor, error) {
 		return nil, err
 	}
 
+	nl := false
+	if len(data) > 0 && data[len(data)-1] == '\n' {
+		nl = true
+	}
+
 	return &Processor{
 		file: f,
+		newLineAtEOF: nl,
 	}, nil
 }
 
@@ -104,5 +111,10 @@ func (p *Processor) ReplaceString(path, value string) error {
 }
 
 func (p *Processor) Bytes() []byte {
-	return []byte(p.file.String())
+	str := p.file.String()
+	if p.newLineAtEOF {
+		str += "\n"
+	}
+
+	return []byte(str)
 }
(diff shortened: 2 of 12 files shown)


pkg/yamlprocessor/yamlprocessor_test.go
```diff
@@ -253,6 +253,16 @@ foo:
   - baz`),
 			wantErr: false,
 		},
+		{
+			name:    "new line at end of yml given",
+			yml:     `foo: bar
```
khanhtc1202, line 258: what if raw yaml has more than one `\n` at the end of file 🤔

kurochan: @khanhtc1202 It's just as you said...

How about simply always adding a new line at the EOF?

According to the POSIX Stndard, The definition of a LINE is
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

khanhtc1202: In that case, I guess diff will be shown if the input file does not contain a new line at EOF, thus the idea of making no diff caused by a new line at EOF is not guaranteed.

I think we should update the logic to count new lines at EOF and add (if necessary) new line(s) to the end of parsed file, wdyt? 👀

kurochan: My suggestion is to always add single newline to the end of the output file, whether or not there are some newlines at the end of the source file.

khanhtc1202: I understand your suggestion 👍 How about changing it to: add a new line to the EOF in case of there is no `\n` at EOF? If we just add a new line anyway (regardless there is a new line at the EOF or not), in case there is already one `\n` at EOF, we added an unnecessary new line and could cause a diff shown, wdyt?

kurochan: I think there is no `\n` at EOF in generated yaml. Because yamlprocessor calls here (https://github.com/goccy/go-yaml/blob/master/ast/ast.go#L610 ).

```
return strings.Join(doc, "\n")
```

Based on this, which of the following should I do?

1. modify yamlprocessor to add single newline to the end of the output file
2. add a new line to the EOF in case of there is no \n at EOF
3. Make Pull Request to `goccy/go-yaml` to append `\n` at EOF
4. Do 2 and 3

khanhtc1202: I think (2) is good enough for pipecd, about (3) may need discuss on `goccy/go-yaml` repo

kurochan: OK, I'll fix it in the way of (2).

And (3) will discuss on https://github.com/goccy/go-yaml/pull/329 .

pkg/yamlprocessor/yamlprocessor.go
```diff
@@ -104,5 +104,11 @@ func (p *Processor) ReplaceString(path, value string) error {
 }
 
 func (p *Processor) Bytes() []byte {
-	return []byte(p.file.String())
+	str := p.file.String()
+
+	if len(str) > 0 && str[len(str)-1] != '\n' {
```
khanhtc1202, line 109: Could you add a reminder here (TODO or NOTE) to remove this once github.com/goccy/go-yaml/pull/329 is merged and released 😄

kurochan: updated https://github.com/pipe-cd/pipecd/pull/4033/commits/fabf67ee34c041c22e72f36a660f2dcbc1d90931

khanhtc1202:
@kurochan there is a failed test, pls re-check it 😄

kurochan:
@khanhtc1202 
I'm struggling on test fail on `pkg/app/piped/executor/kubernetes/kubernetes_test.go` .
The following differences appear and I haven't come up with a good fix yet 🤔 
```
Diff:                                
--- Expected                                
+++ Actual                                
@@ -2,3 +2,3 @@                                
 data:                                
-  envoy-config: |-                                
+  envoy-config: |                                
     admin:                                
TestPatchManifest/multi_ops_with_a_given_field  
```
Do you have any ideas to fix this differences?

khanhtc1202:
@kurochan Probably it's caused by this convention
https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines#:~:text=Block%20styles%20with%20block%20chomping%20indicator
So I think your change should support all YAML convention includes: `>`, `|` and its mix with `-`, `+` etc

pkg/yamlprocessor/yamlprocessor.go
kurochan: @khanhtc1202 I updated go-yaml and removed TODO.

pkg/app/piped/platformprovider/kubernetes/manifest.go
```diff
@@ -224,12 +224,17 @@ func ParseManifests(data string) ([]Manifest, error) {
 		manifests = make([]Manifest, 0, len(parts))
 	)
 
-	for _, part := range parts {
+	newLineAtEOF := strings.HasSuffix(data, "\n")
+	for i, part := range parts {
 		// Ignore all the cases where no content between separator.
 		part = strings.TrimSpace(part)
```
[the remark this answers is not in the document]
khanhtc1202: Looks like in case of the yaml file contains more than one empty line at the end of file, only one empty line is remained thus the diff will be shown, isn't it? In that case, lets add test to demonstrate that and I will merge this PR 🙌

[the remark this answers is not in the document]
kurochan: Thanks for pointing that out. I have fixed it and added a test case. https://github.com/pipe-cd/pipecd/pull/4033/commits/0e250b194b6ead5508f41e4a9eb65ef445bfe3c1

khanhtc1202 (approved):
Nice work! Thanks for the contribution 🙌

knanao (approved):

PR #4047, opened 2022-12-04
Replace deprecated actions commands with env files

knanao:
save-state and set-output commands are already deprecated hence replace it with env files.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch 🍠

PR #4049, opened 2022-12-04
Add release commands to make

khanhtc1202:

Rename commands to `release/*` so that we can remember that we need those commands for release flow 😄 

Fixes #

```release-note
NONE
```


knanao (approved):
Great!

PR #4052, opened 2022-12-05
Update actions-gh-release version to v2.4.0

knanao:

Fixes #

```release-note
NONE
```


.github/workflows/release.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      # - uses: pipe-cd/actions-gh-release@v2.3.6
+      # - uses: pipe-cd/actions-gh-release@v2.4.0
```
khanhtc1202, line 17: this is not being used anyway, so how about removing this file?

knanao: Exactly, let's remove this.

khanhtc1202 (approved):
👍

PR #4053, opened 2022-12-05
Ensure grouping version docs

khanhtc1202:

Previously, we had to remember to self-change version `vX.Y.Z` to `vX.Y.x` whenever we run `make release/docs`, with this change, version docs will always be set to `vX.Y.x` pattern. 

Fixes #

```release-note
NONE
```


knanao (approved):
Nice work.

PR #4054, opened 2022-12-06
use helper "serviceAccountName" on create role binding

golemiso:
When you override service account name to use existent one, role binding should be toward the service account you set.

Fixes #

```release-note

```


knanao (approved):
Cool, nice catch

PR #4055, opened 2022-12-06
Update gen release docs command

khanhtc1202:

Change logic to update `docs/config.yaml` so that we can reduce manual steps while generating release docs.

Fixes #

```release-note
NONE
```


nghialv (approved):
Great!

PR #4060, opened 2022-12-08
Fix bug on parsing empty kubernetes manifest

kurochan:
Fixes the bug on parsing leading document separator.

For example:
```
---
apiVersion: v1
kind: ConfigMap
```
```
---
---
---
apiVersion: v1
```

Fixes #

```release-note
Fix bug on parsing empty kubernetes manifest
```

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/manifest_test.go
+++ b/pkg/app/piped/platformprovider/kubernetes/manifest_test.go
@@ -1,0 +1,121 @@
+// Copyright 2022 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


pkg/app/piped/platformprovider/kubernetes/manifest_test.go
```diff
@@ -0,0 +1,121 @@
...
+		manifests string
+		want      []Manifest
+	}{
+		{
+			name: "empty1",
+		},
+		{
+			name:      "empty2",
+			manifests: "",
+		},
```
knanao, line 55: Nice improvements.
These seem to be the same cases, so I think one or the other is sufficient. wdyt?

kurochan: corrected my mistake 🙏

knanao (approved):
There we go. 🚀

PR #4061, opened 2022-12-09
Update web dependencies

khanhtc1202:

Resolve dependencies conflict between react-thunk, redux, and redux-toolkit so that we can update immer to v9.0.7. Also update the source code reflect the change required on new versions of dependecies.

Fixes #

```release-note
NONE
```


knanao (approved):
Nice improvements, thanks!

PR #4068, opened 2022-12-12
Add the log for detecting pruned manifest

knanao:

Fixes #

```release-note
NONE
```


pkg/app/piped/executor/kubernetes/primary.go
```diff
@@ -161,6 +161,7 @@ func (e *deployExecutor) ensurePrimaryRollout(ctx context.Context) model.StageSt
 		e.LogPersister.Errorf("Failed while loading running manifests (%v)", err)
 		return model.StageStatus_STAGE_FAILURE
 	}
+	e.LogPersister.Successf("Successfully loaded %d live resources", len(runningManifests))
```
khanhtc1202, line 164: how about list up running manifests instead of just print out the len 🤔

knanao: Sounds good.
I added the log of loaded resources.

nghialv (approved):
👍

PR #4072, opened 2022-12-13
Fix app live resource lister wrong attribute

khanhtc1202:

Fix the appLiveResourceLister attribute, previously this bug could not be found because we keep both CloudProvider and PlatformProvider duplicated in the Application model.

Fixes #

```release-note
NONE
```


nghialv (approved):
Nice catch!

PR #4077, opened 2022-12-15
Add --include-crds option to include CRDS for helm

knanao:

Fixes #

```release-note
Fix the problem to not deploy crd resources through helm
```


khanhtc1202 (approved):
Nice catch

PR #4079, opened 2022-12-16
Support Cloud Run application for planpreview

knanao:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/planpreview/cloudrundiff.go
+++ b/pkg/app/piped/planpreview/cloudrundiff.go
@@ -1,3 +1,17 @@
+// Copyright 2022 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


knanao:
If this seems to be good, let me check if it will work correctly after merging.

pkg/app/piped/planpreview/cloudrundiff.go
```diff
@@ -0,0 +1,109 @@
+package planpreview
```
khanhtc1202, line 1: Missing file header

pkg/app/piped/planpreview/cloudrundiff.go
```diff
@@ -0,0 +1,123 @@
...
+	if lastCommit != "" {
+		runningDSP := deploysource.NewProvider(
+			b.workingDir,
+			deploysource.NewGitSourceCloner(b.gitClient, b.repoCfg, "running", lastCommit),
+			*app.GitPath,
+			b.secretDecrypter,
+		)
+		oldManifest, err = b.loadCloudRunManifests(ctx, *app, runningDSP)
+		if err != nil {
+			fmt.Fprintf(buf, "failed to load cloud run manifest at the running commit (%v)\n", err)
```
khanhtc1202, line 56: Should we print the running commit infor as well? same with error log in L43

Issue #3484, opened 2022-04-04
Enable to configure kubectl version via piped configuration


knanao:
**What would you like to be added**:

**Why is this needed**:

PR #4081, opened 2022-12-16
Support setting kubectlVersion via piped config

khanhtc1202:

The priority to determine kubectl version would be:
- Version setting via application config
- Version setting via piped config
- Default version

Fixes #

```release-note
Enable to specify default kubectl version via piped config
```


knanao:
I do think we have discussed supporting such a feature before, but I forgot what makes us happy.

khanhtc1202:
Well can say that at some places where users don't want config kubectl version on their application configuration and at the same time, they want to use kubectl at a newer version than the piped default version, this change could help, wdyt?

knanao:
Right, as you mentions, it's a hassle to specify per the application configuration file.

knanao (approved):

knanao:
https://github.com/pipe-cd/pipecd/issues/3484

PR #4084, opened 2022-12-16
support standalone ECS task

### LINKED ISSUE #2734: Support deploy standalone ECS task

Support deploy standalone ECS task

**Why is this needed**:

Currently, PipeCD only supports running ECS tasks handled by ECS service. In some cases, users want to start standalone ECS tasks for one-time tasks such as DB migration.

TonkyH:
 Support deploy standalone ECS task. but only `Quick Sync`.
PipeCD deployed only ecs tasks managed by the ecs service, but now it is possible to deploy standalone tasks.

Fixes #2734 

```release-note
NONE
```


examples/ecs/standalone-task/app.pipecd.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: pipecd.dev/v1beta1
```
khanhtc1202, line 1: Should remove those `examples` files from this PR, we can add those files later.

khanhtc1202:
Some names changed after the PR https://github.com/pipe-cd/pipecd/pull/4080, please rebase your branch with latest master 👀

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -150,6 +150,25 @@ func (c *client) RegisterTaskDefinition(ctx context.Context, taskDefinition type
...
+		Cluster:        cluster.ClusterArn,
+		Group:          taskDefinition.Family,
+		LaunchType:     types.LaunchTypeFargate,
+		NetworkConfiguration: &types.NetworkConfiguration{
+			AwsvpcConfiguration: &awsVpcConfiguration,
+		},
+	}
+
+	output, err := c.ecsClient.RunTask(ctx, input)
+
```
knanao, line 165: nits
```suggestion
```

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -157,6 +184,48 @@ func applyServiceDefinition(ctx context.Context, cli provider.Client, serviceDef
...
+	taskDefinition types.TaskDefinition,
+) (*ecs.RunTaskOutput, error) {
+	output, err := cli.RunTask(ctx, clusterDefinition, awsVpcConfiguration, taskDefinition)
+	if err != nil {
+		return nil, fmt.Errorf("unable to run ECS task %s: %v", *taskDefinition.TaskDefinitionArn, err)
+	}
+	return output, nil
+}
+
+func isStandaloneTask(in *config.ECSDeploymentInput) bool {
```
knanao, line 201: I suppose it's better to implement this as a config.ECSDeploymentInput's method.

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -83,6 +84,32 @@ func findPlatformProvider(in *executor.Input) (name string, cfg *config.Platform
...
+		in.LogPersister.Errorf("Failed to load VPC configuration (%v)", err)
+		return types.AwsVpcConfiguration{}, false
+	}
+
+	in.LogPersister.Infof("Successfully loaded the VPC configuration at commit %s", ds.Revision)
+	return awsVpcConfiguration, true
+}
+
+func loadClusterDefinition(in *executor.Input, clusterDefinitionFile string, ds *deploysource.DeploySource) (types.Cluster, bool) {
+	in.LogPersister.Infof("Loading Cluster manifest at commit %s", ds.Revision)
```
knanao, line 101:
```suggestion
	in.LogPersister.Infof("Loading cluster manifest at commit %s", ds.Revision)
```

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -83,6 +84,32 @@ func findPlatformProvider(in *executor.Input) (name string, cfg *config.Platform
 	return
 }
 
+func LoadVpcConfiguration(in *executor.Input, vpcConfiguration string, ds *deploysource.DeploySource) (types.AwsVpcConfiguration, bool) {
+	in.LogPersister.Infof("Loading vpc manifest at commit %s", ds.Revision)
+
+	awsVpcConfiguration, err := provider.LoadVpcConfiguration(ds.AppDir, vpcConfiguration)
+	if err != nil {
+		in.LogPersister.Errorf("Failed to load VPC configuration (%v)", err)
```
knanao, line 92: I'm not so concerned about it, but I want to unify whether `vpc` is lower or upper case.

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -83,6 +84,32 @@ func findPlatformProvider(in *executor.Input) (name string, cfg *config.Platform
...
+	in.LogPersister.Infof("Successfully loaded the VPC configuration at commit %s", ds.Revision)
+	return awsVpcConfiguration, true
+}
+
+func loadClusterDefinition(in *executor.Input, clusterDefinitionFile string, ds *deploysource.DeploySource) (types.Cluster, bool) {
+	in.LogPersister.Infof("Loading Cluster manifest at commit %s", ds.Revision)
+
+	clusterDefinition, err := provider.LoadClusterDefinition(ds.AppDir, clusterDefinitionFile)
+	if err != nil {
+		in.LogPersister.Errorf("Failed to load ECS Cluster definition (%v)", err)
```
knanao, line 105:
```suggestion
		in.LogPersister.Errorf("Failed to load ECS cluster definition (%v)", err)
```

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -83,6 +84,32 @@ func findPlatformProvider(in *executor.Input) (name string, cfg *config.Platform
 	return
 }
 
+func LoadVpcConfiguration(in *executor.Input, vpcConfiguration string, ds *deploysource.DeploySource) (types.AwsVpcConfiguration, bool) {
```
knanao, line 87: Is this fine to define as a private function?
```suggestion
func loadVpcConfiguration(in *executor.Input, vpcConfiguration string, ds *deploysource.DeploySource) (types.AwsVpcConfiguration, bool) {
```

pkg/app/piped/executor/ecs/ecs.go
TonkyH: Since it's a proper noun, I'd like to capitalize it.

pkg/app/piped/executor/ecs/ecs.go
TonkyH: thx, That's better 🙇🚀

pkg/app/piped/executor/ecs/deploy.go
```diff
@@ -82,6 +82,23 @@ func (e *deployExecutor) ensureSync(ctx context.Context) model.StageStatus {
 	if !ok {
 		return model.StageStatus_STAGE_FAILURE
 	}
+
+	if e.appCfg.Input.IsStandaloneTask() {
```
knanao, line 86: How about defining `escInput := e.appCfg.Input` at the beginning of this method?

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -157,6 +184,42 @@ func applyServiceDefinition(ctx context.Context, cli provider.Client, serviceDef
...
+}
+
+func runStandaloneTask(
+	ctx context.Context,
+	in *executor.Input,
+	cloudProviderName string,
+	cloudProviderCfg *config.PlatformProviderECSConfig,
+	clusterDefinition types.Cluster,
+	awsVpcConfiguration types.AwsVpcConfiguration,
+	taskDefinition types.TaskDefinition,
```
knanao, line 208: we're customary to pass logs at the end of args, so modify them as this.
```suggestion
func runStandaloneTask(
	ctx context.Context,
	cloudProviderName string,
	cloudProviderCfg *config.PlatformProviderECSConfig,
	clusterDefinition types.Cluster,
	awsVpcConfiguration types.AwsVpcConfiguration,
	taskDefinition types.TaskDefinition,
	lp executor.LogPersister,
```

pkg/app/piped/executor/ecs/deploy.go
TonkyH: That sounds great🚀
It's easy to understand.

pkg/app/piped/executor/ecs/ecs.go
TonkyH: I am not very good at English yet, so I am sorry if I did not understand your intentions.

Since `in *executor.Input` is used in `client, err := provider.DefaultRegistry().Client(cloudProviderName, cloudProviderCfg, in.Logger)`, we need to pass `in *executor.Input`.
Also, `in *executor.Input` has `LogPersister` and `*zap.Logger`.

https://github.com/pipe-cd/pipecd/pull/4084/files#diff-0ced6421324def8505cb6dadaa49f469d0855ef22437ba6d8ad7f5279a61b2e4R210

knanao: oh, right. thanks.

pkg/config/application_ecs.go
```diff
@@ -34,6 +34,10 @@ func (s *ECSApplicationSpec) Validate() error {
 }
 
 type ECSDeploymentInput struct {
+	// The name of network configuration file placing in application directory.
+	VpcConfigurationFile string `json:"vpcConfigurationFile"`
+	// The name of cluster definition file placing in application directory.
+	ClusterDefinitionFile string `json:"clusterDefinitionFile"`
```
khanhtc1202, line 40: Looks like this file will only contain the clusterArn, and we will not create the cluster using piped so I think we should pass the clusterArn directly instead. WDYT?

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -157,6 +184,42 @@ func applyServiceDefinition(ctx context.Context, cli provider.Client, serviceDef
 	return service, nil
 }
 
+func applyStandaloneTask(
+	ctx context.Context,
+	cli provider.Client,
+	clusterDefinition types.Cluster,
+	awsVpcConfiguration types.AwsVpcConfiguration,
+	taskDefinition types.TaskDefinition,
+) (*ecs.RunTaskOutput, error) {
```
khanhtc1202, line 193: We don't use this `ecs.RunTaskOuput` anyway so I think it's better to just return the error.
ref: https://github.com/pipe-cd/pipecd/pull/4084/files#diff-0ced6421324def8505cb6dadaa49f469d0855ef22437ba6d8ad7f5279a61b2e4R215

pkg/app/piped/platformprovider/ecs/ecs.go
```diff
@@ -37,6 +38,7 @@ type ECS interface {
 	CreateService(ctx context.Context, service types.Service) (*types.Service, error)
 	UpdateService(ctx context.Context, service types.Service) (*types.Service, error)
 	RegisterTaskDefinition(ctx context.Context, taskDefinition types.TaskDefinition) (*types.TaskDefinition, error)
+	RunTask(ctx context.Context, cluster types.Cluster, awsVpcConfiguration types.AwsVpcConfiguration, taskDefinition types.TaskDefinition) (*ecs.RunTaskOutput, error)
```
khanhtc1202, line 41: Should remove the `ecs.RunTaskOutput` from this function signature then, we can remove the `service/ecs` dependency from this file because we don't use that anyway.

pkg/app/piped/executor/ecs/ecs.go
TonkyH: I think you are right. I'll fix it🚀

pkg/app/piped/platformprovider/ecs/ecs.go
TonkyH: I'll fix this file to remove the `service/ecs` dependency🙇

pkg/config/application_ecs.go
TonkyH: I couldn't think of a way to pass the clusterArn directly, so I had to pass it via file...
Do you have any good ideas?😭

khanhtc1202: I mean, treating it as a string and making this field like `clusterArn` is enough, isn't it?
I saw the RunTaskInput only requires ClusterArn 👀 
ref: https://github.com/pipe-cd/pipecd/pull/4084/files/600b7403062cf90e5a806cbb9a92e72981f02594#diff-d72df5132a62e1bc9918e8313b516291c54b87a84acd3686bf5233adf4775cccR156

TonkyH: I apologize if I did not understand your intent🙇

`clusterArn` has to be received from outside. because we want to execute the task within the specified cluster.
The only way I could think of to receive `clusterArn` is via file.
Also, The RunTaskInput requires TaskDefinition field.
ref: https://github.com/aws/aws-sdk-go-v2/blob/main/service/ecs/api_op_RunTask.go#L77

khanhtc1202: I mean current implementation of client.RunTask passes whole types.Cluster object but in the function body when you prepare RunTaskInput, only cluster.ClusterArn be used. It means we don't need the whole types.Cluster object anyway, so we should let users pass only input.clusterArn from start instead of focusing them to make a new cluster.yaml which contains almost nothing. 👀

TonkyH: Thank you for your detailed explanation😄
I understand. I'll give it a try🚀

pkg/config/application_ecs.go
```diff
@@ -47,6 +51,12 @@ type ECSDeploymentInput struct {
 	AutoRollback *bool `json:"autoRollback,omitempty" default:"true"`
 }
 
+func (in *ECSDeploymentInput) IsStandaloneTask() bool {
+	existClusterFile := in.ClusterDefinitionFile != ""
+	existVpcConfFile := in.VpcConfigurationFile != ""
+	return existClusterFile && existVpcConfFile
+}
+
```
khanhtc1202, line 59: I think this check should depend on whether the input contains serviceDef or not.

pkg/app/piped/executor/ecs/deploy.go
```diff
@@ -78,11 +78,21 @@ func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
...
-	servicedefinition, ok := loadServiceDefinition(&e.Input, e.appCfg.Input.ServiceDefinitionFile, e.deploySource)
+
+	if ecsInput.IsStandaloneTask() {
+		if !runStandaloneTask(ctx, &e.Input, e.platformProviderName, e.platformProviderCfg, taskDefinition, &ecsInput) {
+			return model.StageStatus_STAGE_FAILURE
+		}
+		return model.StageStatus_STAGE_SUCCESS
+	}
+
+	servicedefinition, ok := loadServiceDefinition(&e.Input, ecsInput.ServiceDefinitionFile, e.deploySource)
```
khanhtc1202, line 95: Nice 😋

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -150,6 +150,32 @@ func (c *client) RegisterTaskDefinition(ctx context.Context, taskDefinition type
...
 }
 
+func (c *client) RunTask(
+	ctx context.Context,
+	taskDefinition types.TaskDefinition,
+	capacityProviderStrategy []types.CapacityProviderStrategyItem,
+	clusterArn string,
+	launchType string,
+	awsVpcConfiguration *types.AwsVpcConfiguration,
+) ([]types.Task, []types.Failure, error) {
```
khanhtc1202, line 160: Why don't we just return the error? The outputs aren't used anyway, isn't it?

pkg/config/application_ecs.go
```diff
@@ -34,9 +38,21 @@ func (s *ECSApplicationSpec) Validate() error {
...
+	// The launch type on which to run your task.
+	// If a launchType is specified, the capacityProviderStrategy parameter must be omitted.
+	LaunchType string `json:"launchType"`
+	// VpcConfiguration ECSVpcConfiguration `json:"awsvpcConfiguration"`
+	AwsVpcConfiguration types.AwsVpcConfiguration `json:"awsvpcConfiguration"`
+	// The capacity provider strategy to use for the task.
+	// If a capacityProviderStrategy is specified, the launchType parameter must be omitted.
+	CapacityProviderStrategy []types.CapacityProviderStrategyItem `json:"capacityProviderStrategy"`
 	// The name of service definition file placing in application directory.
 	// Default is service.json
```
khanhtc1202, line 54: Remove this note as well if we remove the default value.

pkg/config/application_ecs.go
```diff
@@ -34,9 +38,21 @@ func (s *ECSApplicationSpec) Validate() error {
...
+	// The Amazon Resource Name (ARN) that identifies the cluster.
+	ClusterArn string `json:"clusterArn"`
+	// The launch type on which to run your task.
+	// If a launchType is specified, the capacityProviderStrategy parameter must be omitted.
+	LaunchType string `json:"launchType"`
+	// VpcConfiguration ECSVpcConfiguration `json:"awsvpcConfiguration"`
+	AwsVpcConfiguration types.AwsVpcConfiguration `json:"awsvpcConfiguration"`
+	// The capacity provider strategy to use for the task.
+	// If a capacityProviderStrategy is specified, the launchType parameter must be omitted.
+	CapacityProviderStrategy []types.CapacityProviderStrategyItem `json:"capacityProviderStrategy"`
```
khanhtc1202, line 52: Is this duplicated with the LaunchType? Should we keep only LaunchType available to simplify the configuration?

pkg/config/application_ecs.go
```diff
@@ -34,9 +38,21 @@ func (s *ECSApplicationSpec) Validate() error {
 }
 
 type ECSDeploymentInput struct {
+	// The name of network configuration file placing in application directory.
+	VpcConfigurationFile string `json:"vpcConfigurationFile"`
+	// The Amazon Resource Name (ARN) that identifies the cluster.
+	ClusterArn string `json:"clusterArn"`
+	// The launch type on which to run your task.
+	// If a launchType is specified, the capacityProviderStrategy parameter must be omitted.
```
khanhtc1202, line 46: nits, should add a link to AWS docs
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_types.html

pkg/config/application_ecs.go
```diff
@@ -34,9 +38,21 @@ func (s *ECSApplicationSpec) Validate() error {
...
+	LaunchType string `json:"launchType"`
```
khanhtc1202, line 47: Should set a default value for this field, `FARGATE` is good in this case, I guess.

pkg/config/application_ecs.go
```diff
@@ -34,9 +38,21 @@ func (s *ECSApplicationSpec) Validate() error {
 }
 
 type ECSDeploymentInput struct {
+	// The name of network configuration file placing in application directory.
+	VpcConfigurationFile string `json:"vpcConfigurationFile"`
```
khanhtc1202, line 42: Is this duplicated with L49 (AwsVpcConfiguration field), or do you want to support both on purpose?

pkg/app/piped/platformprovider/ecs/client.go
TonkyH: I think you are right. I'll fix it!

pkg/config/application_ecs.go
TonkyH: I forgot to delete this line🙇

pkg/config/application_ecs.go
TonkyH: The difference is whether to use `CapacityProviderStrategy`

>Should we keep only LaunchType available to simplify the configuration?

I agree. I found the current configuration a bit complicated!

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -157,6 +157,48 @@ func applyServiceDefinition(ctx context.Context, cli provider.Client, serviceDef
...
+		taskDefinition,
+		ecsInput.ClusterArn,
+		ecsInput.LaunchType,
+		&ecsInput.AwsVpcConfiguration,
+	)
+
+	if err != nil {
+		return fmt.Errorf("unable to run ECS task %s: %v", *taskDefinition.TaskDefinitionArn, err)
+	}
+	return nil
```
khanhtc1202, line 177:
```suggestion
	return cli.RunTask(
		ctx,
		taskDefinition,
		ecsInput.ClusterArn,
		ecsInput.LaunchType,
		&ecsInput.AwsVpcConfiguration,
	)
```

pkg/config/application_ecs.go
```diff
@@ -14,7 +14,11 @@
 
 package config
 
-import "encoding/json"
+import (
+	"encoding/json"
+
+	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
```
khanhtc1202, line 20: Should convert this types.AwsVpcConfiguration to platformprovider ecs types so that we can remove the depending on 3rd package from this configuration package.

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -157,6 +157,43 @@ func applyServiceDefinition(ctx context.Context, cli provider.Client, serviceDef
...
+	taskDefinition types.TaskDefinition,
+	ecsInput *config.ECSDeploymentInput,
+) bool {
+	client, err := provider.DefaultRegistry().Client(cloudProviderName, cloudProviderCfg, in.Logger)
+	if err != nil {
+		in.LogPersister.Errorf("Unable to create ECS client for the provider %s: %v", cloudProviderName, err)
+		return false
+	}
+
+	err = applyStandaloneTask(ctx, client, taskDefinition, ecsInput)
```
khanhtc1202, line 189: In this case, directly call to cli.RunTask and remove applyStandaloneTask could be a better choice 🤔

TonkyH: I think you are right.
I was creating unnecessary functions...🙇

khanhtc1202 (approved):
Here you go 🚀

PR #4086, opened 2022-12-17
Implement lint/go command for local

knanao:
In fact, we have to fix all lint errors before using this command.
So let me fix all of them one by one in another PR.
https://github.com/pipe-cd/pipecd/runs/10154286748

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,8 +138,15 @@ run/site:
 # Lint commands
 
 .PHONY: lint/go
+lint/go: FIX ?= false
+lint/go: VERSION ?= sha256:78d1bbd01a9886a395dc8374218a6c0b7b67694e725dd76f0c8ac1de411b85e8 #v1.46.2
+lint/go: FLAGS ?= --rm -e GOLANGCI_LINT_CACHE=/repo/.cache/golangci-lint -v ${PWD}:/repo -w /repo -it
 lint/go:
-	@echo "Unimplemented"
+ifeq ($(FIX),true)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -138,8 +138,15 @@ run/site:
 # Lint commands
 
 .PHONY: lint/go
+lint/go: FIX ?= false
+lint/go: VERSION ?= sha256:78d1bbd01a9886a395dc8374218a6c0b7b67694e725dd76f0c8ac1de411b85e8 #v1.46.2
+lint/go: FLAGS ?= --rm -e GOLANGCI_LINT_CACHE=/repo/.cache/golangci-lint -v ${PWD}:/repo -w /repo -it
 lint/go:
-	@echo "Unimplemented"
+ifeq ($(FIX),true)
+	docker run ${FLAGS} golangci/golangci-lint@${VERSION} golangci-lint run --fix
```
khanhtc1202, line 146: Should we add `--rm` to remove the container after the job is done? 🤔

knanao: this one?
https://github.com/pipe-cd/pipecd/pull/4086/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R143

khanhtc1202: Ah true, I missed reading that part 👍

khanhtc1202 (approved):
Nice improvement 👍

PR #4102, opened 2023-01-05
fix null pointer

TonkyH:
The null pointer was causing the error propagation to not work.
So I've corrected it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/platformprovider/ecs/client.go
+++ b/pkg/app/piped/platformprovider/ecs/client.go
@@ -170,7 +170,7 @@ func (c *client) RunTask(ctx context.Context, taskDefinition types.TaskDefinitio
 
 	_, err := c.ecsClient.RunTask(ctx, input)
 	if err != nil {
-		return fmt.Errorf("failed to run ECS task %s: %w", *taskDefinition.TaskDefinitionArn, err)
+		return fmt.Errorf("failed to run ECS task %s: %w", *taskDefinition.Family, err)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


khanhtc1202:
Not related to this PR changes but you should update this
https://github.com/pipe-cd/pipecd/pull/4102/files#diff-d72df5132a62e1bc9918e8313b516291c54b87a84acd3686bf5233adf4775cccR156
pass `*taskDefinition.TaskDefinitionArn` instead of Family there.

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -170,7 +170,7 @@ func (c *client) RunTask(ctx context.Context, taskDefinition types.TaskDefinitio
 
 	_, err := c.ecsClient.RunTask(ctx, input)
 	if err != nil {
-		return fmt.Errorf("failed to run ECS task %s: %w", *taskDefinition.TaskDefinitionArn, err)
+		return fmt.Errorf("failed to run ECS task %s: %w", *taskDefinition.Family, err)
```
khanhtc1202, line 173: Should keep it as is and add a check for TaskDefinicationArn before running task since it doesn't make sense to run a task that does not exist. Please refer to 
https://github.com/pipe-cd/pipecd/blob/21936efa898995e4cdaefb91710f956ae1fe0d52/pkg/app/piped/platformprovider/ecs/client.go#L179-L181

khanhtc1202 (approved):
Here you go 👍

PR #4120, opened 2023-01-11
Templating commit message used by Event watcher

### LINKED ISSUE #1334: Templating commit message used by Event watcher
Currently, we can set the commit message used by event watcher.
It would be nice if the commit message can be templating.

```yaml
kind: Piped
spec:
  eventWatcher:
    repos:
      - repoId: foo
        commitMessage: update values {.Value} by {.EventName}
```
**Why is this needed**:

funera1:
Template the eventWatcher's commitMessage using environment variables.  

Fixes #1334 

```release-note
NONE
```


pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -149,12 +151,20 @@ func (w *watcher) run(ctx context.Context, repo git.Repo, repoCfg config.PipedRe
 		commitMsg                  string
 		includedCfgs, excludedCfgs []string
 	)
+
+	// get rule to parse template
```
khanhtc1202, line 155:
```suggestion
	// Get rule to parse template.
```
as our team convention 😄

pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -747,3 +757,18 @@ func modifyText(path, regexText, newValue string) ([]byte, bool, error) {
 
 	return newText, false, nil
 }
+
+// parse msg according to rules
```
khanhtc1202, line 761:
```suggestion
// parseTemplate parses msg according to rules.
```
nits

pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -149,12 +151,20 @@ func (w *watcher) run(ctx context.Context, repo git.Repo, repoCfg config.PipedRe
 		commitMsg                  string
 		includedCfgs, excludedCfgs []string
 	)
+
+	// Get rule to parse template.
+	rules := map[string]string{}
+	for _, env := range os.Environ() {
```
knanao, line 157: Sorry, I couldn't catch your intention.
You intended that templating variables such as `{.Value}` and `{.EventName}` are passed through env?
IMO, it's possible to use these variables without setting them as an env var because it's passed when registering events.

funera1: I see.
I was thinking I should set variables other than value and eventName, but maybe only value and eventName would be better.

knanao: In this context, I think it's enough that we support only what the user input.

pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -633,6 +636,9 @@ func (w *watcher) commitFiles(ctx context.Context, latestData, eventName, commit
 
 	if commitMsg == "" {
 		commitMsg = fmt.Sprintf(defaultCommitMessageFormat, latestData, eventName)
+	} else {
+		rules := map[string]string{"Value": latestData, "EventName": eventName}
+		commitMsg = parseTemplate(commitMsg, rules)
 	}
```
knanao, line 642: How about like this?

```suggestion
	commitMsg = parseCommitMsg(commitMsg, latestData, eventName)
```

pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -747,3 +753,18 @@ func modifyText(path, regexText, newValue string) ([]byte, bool, error) {
...
+		return msg
+	}
+
+	buf := new(strings.Builder)
+	if err = t.Execute(buf, rules); err != nil {
+		return msg
+	}
+
+	return buf.String()
+}
```
knanao, line 770: This is an example of parseCommitMsg implementation.

```suggestion
// argsTemplate represents a collection of available template arguments.
type argsTemplate struct {
   Value  string 
   EventName string
}


// parseCommitMsg parses event watcher's commit message.
// Currently, only { .Vaule } and { .EventName } are supported.
func parseCommitMsg(msg, latestDate, eventName string) string {
        if msg == "" {
		return fmt.Sprintf(defaultCommitMessageFormat, latestData, eventName)
	}
	
	args := argsTemplate{"Value": latestData, "EventName": eventName}
	t, err := template.New("EventWatcherCommitMsgTemplate").Parse(msg)
	if err != nil {
		return msg
	}

	buf := new(strings.Builder)
	if err := t.Execute(buf, args); err != nil {
		return msg
	}
	return buf.String()
}
```

pkg/app/piped/eventwatcher/eventwatcher.go
funera1: OK. Thank you for your review!

knanao (approved):
Thank you.

pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -747,3 +752,28 @@ func modifyText(path, regexText, newValue string) ([]byte, bool, error) {
...
 }
+
+// argsTemplate represents a collection of available template arguments.
+type argsTemplate struct {
+	Value     string
+	EventName string
+}
+
+// parseCommitMsg parses event watcher's commit message.
+// Currently, only { .Vaule } and { .EventName } are supported.
```
khanhtc1202, line 763:
```suggestion
// Currently, only { .Value } and { .EventName } are supported.
```

khanhtc1202 (approved):
Here you go 👍

PR #4124, opened 2023-01-12
Add last use of api key on webui

### LINKED ISSUE #4118: Show last use of API key on API key setting page

Show time of last use for a specified API key on the API key setting page

**Why is this needed**:

kevin55156:
Enabling us to know when api keys are created and used at last time.

Fixes #4118 

```release-note
add created-time and last-used-time of api keys on web-ui
```


knanao:
@kevin55156 
Great, it's better to be able to check what gRPC was called too by a API key next.

Anyway, Thank you for your nice work. 🚀

web/src/components/settings-page/api-key/index.tsx
```diff
@@ -82,11 +82,17 @@ export const APIKeyPage: FC = memo(function APIKeyPage() {
   const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(
     null
   );
+  console.log(keys);
```
khanhtc1202, line 85: Please remove this

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -67,6 +68,11 @@ func (v *Verifier) Verify(ctx context.Context, key string) (*model.APIKey, error
 		return nil, fmt.Errorf("unable to find API key %s from datastore, %w", keyID, err)
 	}
 
+	// update the time API key was last used
+	if err := v.apiKeyStore.UpdateLastUsedAt(ctx, keyID, apiKey.ProjectId); err != nil {
```
khanhtc1202, line 72: if only check it here, the UpdateLastUsedAt only be triggered on case of apiKeyCache missed, isn't it?

kevin55156:
> @kevin55156 Great, it's better to be able to check what gRPC was called too by a API key next.

Thank you, I think so, too.
I will create the new issue about this feature.

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -28,6 +28,7 @@ import (
 
 type apiKeyGetter interface {
 	Get(ctx context.Context, id string) (*model.APIKey, error)
+	UpdateLastUsedAt(ctx context.Context, id, projectID string) error
```
khanhtc1202, line 31: the name of the interface is XXXGetter, so it should contain only getter methods, not update methods. We should update the name of this interface to apiKeyStore.

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -54,37 +55,41 @@ func (v *Verifier) Verify(ctx context.Context, key string) (*model.APIKey, error
...
-	if err := checkAPIKey(apiKey, keyID, key); err != nil {
+	if err := checkAPIKey(ctx, v, apiKey, keyID, key); err != nil {
 		return nil, err
 	}
 
 	return apiKey, nil
 }
 
-func checkAPIKey(apiKey *model.APIKey, id, key string) error {
+func checkAPIKey(ctx context.Context, v *Verifier, apiKey *model.APIKey, id, key string) error {
```
khanhtc1202, line 82: If we need to pass the Verifier as a parameter here, then better to make this checkAPIKey as a private method of the Verifier struct.

pkg/datastore/apikey.go
```diff
@@ -148,3 +149,16 @@ func (s *apiKeyStore) Disable(ctx context.Context, id, projectID string) error {
...
+	return s.ds.Update(ctx, s.col, id, func(e interface{}) error {
+		k := e.(*model.APIKey)
+		if k.ProjectId != projectID {
+			return fmt.Errorf("invalid project id, expected %s, got %s", k.ProjectId, projectID)
+		}
+
+		k.LastUsedAt = now
+		return k.Validate()
+	})
+}
```
khanhtc1202, line 164: this is never used, isn't it?

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -23,24 +23,30 @@ import (
...
-	apiKeyStore apiKeyGetter
-	logger      *zap.Logger
+	apiKeyCache         cache.Cache
+	apiKeyStore         apiKeyGetter
+	apiKeyLastUsedCache cache.Cache
+	logger              *zap.Logger
 }
 
-func NewVerifier(ctx context.Context, getter apiKeyGetter, logger *zap.Logger) *Verifier {
+const apiKeyLastUsedCacheHashKey = "HASHKEY:PIPED:API_KEYS" //nolint:gosec
```
knanao, line 42: [nits] I feel it's better to make it sharding per project.
Because it's supposed to be called per a project.

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -54,37 +60,42 @@ func (v *Verifier) Verify(ctx context.Context, key string) (*model.APIKey, error
...
-func checkAPIKey(apiKey *model.APIKey, id, key string) error {
+func (v *Verifier) checkAPIKey(ctx context.Context, apiKey *model.APIKey, id, key string) error {
 	if apiKey.Disabled {
 		return fmt.Errorf("the api key %s was already disabled", id)
 	}
 
 	if err := apiKey.CompareKey(key); err != nil {
 		return fmt.Errorf("invalid api key %s: %w", id, err)
 	}
+	now := time.Now().Unix()
```
knanao, line 95: This is not a testable method, so how about moving this function to the receiver?

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -23,24 +23,30 @@ import (
...
+	apiKeyCache         cache.Cache
+	apiKeyStore         apiKeyGetter
+	apiKeyLastUsedCache cache.Cache
+	logger              *zap.Logger
 }
 
-func NewVerifier(ctx context.Context, getter apiKeyGetter, logger *zap.Logger) *Verifier {
+const apiKeyLastUsedCacheHashKey = "HASHKEY:PIPED:API_KEYS" //nolint:gosec
+
+func NewVerifier(ctx context.Context, getter apiKeyGetter, rd redis.Redis, logger *zap.Logger) *Verifier {
```
khanhtc1202, line 44: instead of passing redis directly here and creating HashCache based redis in the constructor, you can create HashCache for lastUsedCache in server.go and pass it to this constructor.

https://github.com/pipe-cd/pipecd/blob/master/cmd/pipecd/server.go#L198-L213
https://github.com/pipe-cd/pipecd/blob/master/cmd/pipecd/server.go#L72

That also helps reduce the definition for `apiKeyLastUsedCacheHashKey` to one (only have to define it in server.go)

khanhtc1202: Or maybe if we only need Putter, let's define interface Putter here just like we did with 
https://github.com/pipe-cd/pipecd/pull/4124/files#diff-b8762b28ce9988e536f2d9961930f127631c2e8a55e489963913dfbda93e6567R31

pkg/app/server/apikeyverifier/verifier.go
kevin55156: Server must decide hash key name before NewVerifier is called, but which project is not decided at that time.
Piped client is associated with one project, but defaultPipedStatHashKey does not include project id because of the same reason.

cmd/pipecd/server.go
```diff
@@ -69,7 +69,8 @@ var (
 )
 
 const (
-	defaultPipedStatHashKey = "HASHKEY:PIPED:STATS"
+	defaultPipedStatHashKey    = "HASHKEY:PIPED:STATS"
+	apiKeyLastUsedCacheHashKey = "HASHKEY:PIPED:API_KEYS" //nolint:gosec
```
khanhtc1202, line 73: Just want to confirm, do we still need this `nolint` annotation?

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -54,37 +56,46 @@ func (v *Verifier) Verify(ctx context.Context, key string) (*model.APIKey, error
...
 	}
-
 	// If the cache data was not found,
 	// we have to retrieve from datastore and save it to the cache.
 	apiKey, err = v.apiKeyStore.Get(ctx, keyID)
 	if err != nil {
 		return nil, fmt.Errorf("unable to find API key %s from datastore, %w", keyID, err)
 	}
 
+	// update the time API key was last used
```
khanhtc1202, line 71:
```suggestion
	// Update last time the API key was used.
```

pkg/app/server/apikeyverifier/verifier.go
```diff
@@ -54,37 +56,46 @@ func (v *Verifier) Verify(ctx context.Context, key string) (*model.APIKey, error
...
+	if err := v.apiKeyLastUsedCache.Put(id, now); err != nil {
+		return fmt.Errorf("unable to update the time API key %s was last used, %w", id, err)
+	}
 
 	return nil
 }
+
+func (v *Verifier) now() int64 {
+	return time.Now().Unix()
+}
```
khanhtc1202, line 101: Should make nowFunc a field in Verifier struct, like this.

https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/controller/planner.go#L66

cmd/pipecd/server.go
kevin55156: Yes, CI golang-cli fails without this nolint

pkg/app/server/apikeyverifier/verifier_test.go
```diff
@@ -42,6 +42,11 @@ func (g *fakeAPIKeyGetter) Get(_ context.Context, id string) (*model.APIKey, err
 	return nil, fmt.Errorf("not found")
 }
 
+type fakeRedisHashCache struct{}
+
+func (f *fakeRedisHashCache) Put(k string, v interface{}) error {
+	return nil
+}
```
khanhtc1202, line 49:
```suggestion
}

```

khanhtc1202 (approved):
Nice work 🙌

PR #4127, opened 2023-01-13
Add environment variable expire web session

### LINKED ISSUE #4107: Shorten expiration of session on WebUI
Shorten the expiration of session on WebUI.

**Why is this needed**:
Long expiration of session induces security risks.

kevin55156:
Enabling developers change session ttl with config file

Fixes #4107 

```release-note
Add field "sessionTtl" to sharedSSOConfig
```


docs/content/en/docs-dev/user-guide/managing-controlplane/configuration-reference.md
```diff
@@ -147,6 +147,7 @@ Must be one of the following objects:
 |-|-|-|-|
 | name | string | The unique name of the configuration. | Yes |
 | provider | string | The SSO service provider. Can be one of the following values<br>`GITHUB`, `GOOGLE`... | Yes |
+| sessionTtl | int | The time to live of session for sso login | Yes |
```
khanhtc1202, line 150:
```suggestion
| sessionTtl | int | The time to live of session for SSO login. Unit is `minute`. | Yes |
```

quickstart/control-plane-values.yaml
```diff
@@ -6,6 +6,8 @@ config:
     apiVersion: "pipecd.dev/v1beta1"
     kind: ControlPlane
     spec:
+      stateKey: gnvraeiegrojblskgnlhej
+      address: http://127.0.0.1:8080
```
khanhtc1202, line 10: Could you remove this 👀

pkg/model/project.proto
```diff
@@ -97,6 +97,8 @@ message ProjectSSOConfig {
     }
 
     Provider provider = 1 [(validate.rules).enum.defined_only = true];
+    // session ttl for users (munites)
+    int64 session_ttl = 2 [(validate.rules).int64.gt = 0];
```
khanhtc1202, line 101: With the rule set like this, I guess project object which is stored in datastore currently without this value will cause error, right?

kevin55156: When server take configuration file in, sever does not use getter generated by grpc, so validation does not works.
If the field of sessionTtl does not exist in config file, the zero is set at that time.

docs/content/en/docs-dev/user-guide/managing-controlplane/configuration-reference.md
```diff
@@ -147,6 +147,7 @@ Must be one of the following objects:
 |-|-|-|-|
 | name | string | The unique name of the configuration. | Yes |
 | provider | string | The SSO service provider. Can be one of the following values<br>`GITHUB`, `GOOGLE`... | Yes |
+| sessionTtl | int | The time to live of session for SSO login. Unit is `minute`. | Yes |
```
khanhtc1202, line 150: Do we have any reason to mark this field as required? We have a default value for this setting anyway.

docs/content/en/docs-dev/user-guide/managing-controlplane/configuration-reference.md
```diff
@@ -147,6 +147,7 @@ Must be one of the following objects:
 |-|-|-|-|
 | name | string | The unique name of the configuration. | Yes |
 | provider | string | The SSO service provider. Can be one of the following values<br>`GITHUB`, `GOOGLE`... | Yes |
+| sessionTtl | int | The time to live of session for SSO login. Unit is `minute`. Default is 7 hours. | No |
```
khanhtc1202, line 150: I think the current default TTL is 7 days
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/httpapi/auth_handler.go#L51

khanhtc1202: And looks like the current default time is in units of hours, so could you recheck around that part?

kevin55156: Thank you, I have changed "minutes" to "hours"

khanhtc1202 (approved):
Nice work, thanks 🙌

PR #4132, opened 2023-01-15
Fix docs wrong params name

khanhtc1202:

Based on this place, it should be `appArgs` instead of `args`.

Fixes #

```release-note
NONE
```


knanao (approved):
Nice catch!

PR #4137, opened 2023-01-19
Add kubectl version to stage log

### LINKED ISSUE #978: What version of kubectl is used?
Even if the kubectl version specified, the stage log doesn't tell us what version is used.

**Why is this needed**:

funera1:

Fixes #978

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/executor/kubernetes/kubernetes.go
+++ b/pkg/app/piped/executor/kubernetes/kubernetes.go
@@ -94,6 +94,10 @@ func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
 		}
 	}
 
+	if e.appCfg.Input.KubectlVersion != "" {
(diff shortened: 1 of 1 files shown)


pkg/app/piped/executor/kubernetes/kubernetes.go
```diff
@@ -94,6 +94,10 @@ func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
 		}
 	}
 
+	if e.appCfg.Input.KubectlVersion != "" {
+		e.LogPersister.Infof("Kubectl version is %s.", e.appCfg.Input.KubectlVersion)
```
knanao, line 98:
```suggestion
		e.LogPersister.Infof("kubectl version %s will be used.", e.appCfg.Input.KubectlVersion)
```

knanao (approved):

PR #4140, opened 2023-01-19
Add pipecd builtin tags to ECS resources

TonkyH:

Add PipeCD builtin tags for ECS resources.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/executor/ecs/ecs.go
+++ b/pkg/app/piped/executor/ecs/ecs.go
@@ -141,11 +141,12 @@ func applyTaskDefinition(ctx context.Context, cli provider.Client, taskDefinitio
 }
 
 func applyServiceDefinition(ctx context.Context, cli provider.Client, serviceDefinition types.Service) (*types.Service, error) {
-	service, found, err := cli.ServiceExists(ctx, *serviceDefinition.ClusterArn, *serviceDefinition.ServiceName)
+	found, err := cli.ServiceExists(ctx, *serviceDefinition.ClusterArn, *serviceDefinition.ServiceName)
 	if err != nil {
 		return nil, fmt.Errorf("unable to validate service name %s: %v", *serviceDefinition.ServiceName, err)
 	}
 
+	var service *types.Service
 	if found {
 		service, err = cli.UpdateService(ctx, serviceDefinition)
 		if err != nil {
--- a/pkg/app/piped/executor/ecs/ecs.go
+++ b/pkg/app/piped/executor/ecs/ecs.go
@@ -98,7 +98,7 @@ func loadServiceDefinition(in *executor.Input, serviceDefinitionFile string, ds 
 
 	serviceDefinition.Tags = append(
 		serviceDefinition.Tags,
-		provider.CreateTags(map[string]string{
+		provider.MakeTags(map[string]string{
 			provider.LabelManagedBy:   provider.ManagedByPiped,
 			provider.LabelPiped:       in.PipedConfig.PipedID,
 			provider.LabelApplication: in.Deployment.ApplicationId,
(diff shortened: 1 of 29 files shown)


pkg/app/piped/executor/ecs/rollback.go
```diff
@@ -105,7 +105,7 @@ func rollback(ctx context.Context, in *executor.Input, platformProviderName stri
 	// Re-register TaskDef to get TaskDefArn.
 	// Consider using DescribeServices and get services[0].taskSets[0].taskDefinition (taskDefinition of PRIMARY taskSet)
 	// then store it in metadata store and use for rollback instead.
-	td, err := client.RegisterTaskDefinition(ctx, taskDefinition)
+	td, err := client.RegisterTaskDefinition(ctx, taskDefinition, serviceDefinition.Tags)
```
khanhtc1202, line 108: I think we only care about the cluster/service/task thus we don't need to tag the taskDefinition, since it isn't a part of the application anyway. wdyt?

TonkyH: I can also get TaskDef from Task, and it certainly seems unnecessary...
I'll give it some more thought and fix it!🙇

khanhtc1202: Any besides, we should check whether tags which are added to the service is automatically added to the tasks which are controlled by service or not as well 👀

TonkyH: The `--propagateTags` parameter can be used to copy tags from a task definition or service to a task in the service🚀

pkg/app/piped/platformprovider/ecs/ecs.go
```diff
@@ -19,13 +19,18 @@ import (
...
+	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 	"go.uber.org/zap"
 	"golang.org/x/sync/singleflight"
 
 	"github.com/pipe-cd/pipecd/pkg/config"
 )
 
+const (
+	LabelApplication string = "pipecd-dev-application" // The application this resource belongs to.
```
khanhtc1202, line 31: Could you add more builtin tags for ECS resources. For instant, in case of CloudRun, we have as below
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/cloudrun/cloudrun.go#L86-L93

I think we should also have: TagManagedBy (value will be piped in all cases), LabelPiped, LabelCommitHash as the PipeCD builtin tags for ECS resources.

pkg/app/piped/platformprovider/ecs/ecs.go
```diff
@@ -109,3 +119,11 @@ var defaultRegistry = &registry{
 func DefaultRegistry() Registry {
 	return defaultRegistry
 }
+
+func CreateTags(keyValue map[string]string) []types.Tag {
```
khanhtc1202, line 123:
```suggestion
func MakeTags(tags map[string]string) []types.Tag {
```
nits

khanhtc1202 (dismissed):
Nice work 👍

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -92,6 +92,20 @@ func loadServiceDefinition(in *executor.Input, serviceDefinitionFile string, ds
 		return types.Service{}, false
 	}
 
+	if serviceDefinition.PropagateTags == "" {
+		serviceDefinition.PropagateTags = types.PropagateTagsService
+	}
```
khanhtc1202, line 97: Sorry, I missed this part. I think we should always propagate the tags from the service to its tasks due to tags contains out builtin tags as well. So it's better to set the value to CreateServiceInput.PropagateTags directly in CreateService method. WDYT?

TonkyH: I agree.
I think we should always propagate tags from the service to its tasks.
So, set the value directly to CreateServiceInput.PropagateTags in the CreateService method.

khanhtc1202 (approved):
Here you go 🙌

PR #4142, opened 2023-01-22
Disable provenance feature on docker build for unsupported platforms

kurochan:

Explicitly set `provenance=false` on docker build. Because of PipeCD supports CloudRun and related to issue https://github.com/docker/buildx/issues/1533 .

However, the provenance is a part of security feature, I am wondering whether it should be uniformly disabled.

Related: 

```release-note
NONE
```

### DIFF
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -74,6 +74,7 @@ jobs:
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
           push: true
+          provenance: false
           context: .
           file: cmd/pipecd/Dockerfile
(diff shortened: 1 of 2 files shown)


knanao (commented):
The image used by Cloud Run is pushed to GCR.
Hence how about just fixing the image used by Cloud Run?

knanao:
Thank you for sharing the issue.

From this comment, I've actually checked whether Cloud Run has already supported multi-platform built, and it works as expected in both GCR and GAR.
https://github.com/docker/buildx/issues/1533#issuecomment-1416814739

But currently, the actions' version which we're using is not affected by this.
Hence let us update the necessary config when using this feature.
https://github.com/docker/build-push-action#:~:text=Subdirectory%20for%20Git,at%20the%20moment.

That's why let me close this temporarily.

PR #4144, opened 2023-01-23
Add terraform drift detection

### LINKED ISSUE #2542: Automated configuration drift detection for terraform
Epic description
---

funera1:

Fixes #2542

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/driftdetector/terraform/detector.go
+++ b/pkg/app/piped/driftdetector/terraform/detector.go
@@ -13,3 +13,309 @@
 // limitations under the License.
 
 package terraform
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/pipe-cd/pipecd/pkg/app/piped/livestatestore/terraform"
+	provider "github.com/pipe-cd/pipecd/pkg/app/piped/platformprovider/terraform"
+	"github.com/pipe-cd/pipecd/pkg/app/piped/toolregistry"
+	"github.com/pipe-cd/pipecd/pkg/cache"
+	"github.com/pipe-cd/pipecd/pkg/config"
+	"github.com/pipe-cd/pipecd/pkg/git"
+	"github.com/pipe-cd/pipecd/pkg/model"
+	"go.uber.org/zap"
+)
+
+type applicationLister interface {
+	ListByPlatformProvider(name string) []*model.Application
+}
+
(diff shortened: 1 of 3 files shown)


funera1:
There are two points of concern.
First, it does not use Cache, which would likely result in poor performance. However, terraform's drift detection is different from other kubernetes and cloud runs in the way drift detection is done, so implementing cache is also difficult.
Second, the other two output diff details in makeSyncState, but terraform does not. This is because PlanResult did not have a method called Render. However, it may be possible to prepare the details if we can get the output format of Render and the raw output of terraform plan

pkg/app/piped/driftdetector/terraform/detector.go
```diff
@@ -13,3 +13,318 @@
...
+		m    = make(map[string][]*model.Application)
+	)
+	for _, app := range apps {
+		repoID := app.GitPath.Repo.Id
+		m[repoID] = append(m[repoID], app)
+	}
+	return m
+}
+
+func (d *detector) loadApplicationConfiguration(repoPath string, app *model.Application) (*config.Config, error) {
```
knanao, line 316:
```suggestion
func loadApplicationConfiguration(repoPath string, app *model.Application) (*config.Config, error) {
```

pkg/app/piped/driftdetector/terraform/detector.go
```diff
@@ -13,3 +13,318 @@
...
+		appDir  = filepath.Join(repoDir, app.GitPath.Path)
+	)
+
+	// Load config
+	cpCfg := d.provider.TerraformConfig
+	cfg, err := d.loadApplicationConfiguration(repoDir, app)
+	appCfg := cfg.TerraformApplicationSpec
+	if err != nil {
+		return fmt.Errorf("failed to load application configuration: %w", err)
+	}
```
knanao, line 186:
```suggestion
	if err != nil {
		return fmt.Errorf("failed to load application configuration: %w", err)
	}
	appCfg := cfg.TerraformApplicationSpec
```

knanao (commented):
Nice work.
I left some nits comments, so please check it out.
And let's call this feature to check whether working correctly.
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/driftdetector/detector.go#L97

funera1:
Thank you for your nice review

knanao (approved):

PR #4150, opened 2023-01-27
Add --create-namespace flag when uses helm version greater than v3

### LINKED ISSUE #1202: Create namespace if not exist the namespace with helm3

When I install a new application in another namespace with helm, I'm pleased that it will be created if that namespace doesn't exist. helm3 has `--create-namespace` option!

**Why is this needed**:

funera1:

Fixes #1202

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/google/go-github/v29 v29.0.3
 	github.com/google/uuid v1.2.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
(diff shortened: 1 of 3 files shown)


knanao (commented):
Currently, PipeCD has not used `helm install`, which means this is not just for helm.
IMO, I feel it's better to add an optional field to `KubernetesDeploymentInput` in order to create namespace if it doesn't exist.

funera1:
Thank you for nice review. I will try to change it.

PR #4154, opened 2023-02-01
Add Install interface for the k8s applier

knanao:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/applier.go
+++ b/pkg/app/piped/platformprovider/kubernetes/applier.go
@@ -35,21 +35,36 @@ type Applier interface {
 	ReplaceManifest(ctx context.Context, manifest Manifest) error
 	// Delete deletes the given resource from Kubernetes cluster.
 	Delete(ctx context.Context, key ResourceKey) error
+	// Install does installing a release or upgrading it if it already exists.
+	Install(ctx context.Context) error
 }
 
 type applier struct {
+	appName          string
+	appDir           string
 	input            config.KubernetesDeploymentInput
 	platformProvider config.PlatformProviderKubernetesConfig
+	gc               gitClient
 	logger           *zap.Logger
 
 	kubectl  *Kubectl
+	helm     *Helm
 	initOnce sync.Once
 	initErr  error
 }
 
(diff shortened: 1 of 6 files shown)


pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -35,21 +35,36 @@ type Applier interface {
 	ReplaceManifest(ctx context.Context, manifest Manifest) error
 	// Delete deletes the given resource from Kubernetes cluster.
 	Delete(ctx context.Context, key ResourceKey) error
+	// Install does installing a release or upgrading it if it already exists.
+	Install(ctx context.Context) error
```
khanhtc1202, line 39: To be honest, I feel a bit weird about the use of this interface. Most of the functions this interface contains are used for applying/creating resources based on loaded/manipulated manifests loader, this Install function does not really relate to applying tasks. Since we haven't know how we're using this interface anyway, I suggest separating this Intall interface instead of making it part of Applier. wdyt?

knanao: @khanhtc1202 
I don't think it is too strange to have the install interface in the applier in terms of applying resources, but I understand the feeling.

Instead of that, how about creating the new interface, Installer?
```
type Installer interface {
	// Install the chart to Kubernetes.
	Install(ctx context.Context) error
}
```

khanhtc1202: > To have the install interface in the applier in terms of applying resources.

I understand this feeling as well 😂 I think the hard part to make the decision for this is we don't know how it's used on the caller side, so how about to implement that part first. TBH, instead of adding not yet used interface, I think reviewing from how it would be used and define a NOT YET implemented interface is easier and we can confirm the follow works before actually implement the underlying execution. wdyt? 👀

knanao: @khanhtc1202 
I've just realized that the way installing release with helm cannot add built-in annotations to manifest because not using manifests in it.
Hence we have to think about this again before implementing this.

khanhtc1202: Oops, okay 🙆‍♂️

PR #4174, opened 2023-02-10
Select the container registry to use with run/pipecd

### LINKED ISSUE #4172: Select the container registry to use with run/pipecd

**Why is this needed**:

funera1:
Maintain compatibility with previous versions

It would be nice to have a free choice of docker registry because fetching from the local registry will be ImagePullBackOff in my environment. I have compatibility with previous versions, so it shouldn't be a problem

Fixes #4172

```release-note
NONE
```


Makefile
```diff
@@ -102,15 +103,15 @@ run/pipecd:
 	yarn --cwd web build
 
 	@echo "Building docker image and pushing it to local registry..."
-	docker build -f cmd/pipecd/Dockerfile -t localhost:5001/pipecd:$(BUILD_VERSION) .
-	docker push localhost:5001/pipecd:$(BUILD_VERSION)
+	docker build -f cmd/pipecd/Dockerfile -t $(DOCKER_REGISTRY)/pipecd:$(BUILD_VERSION) .
```
khanhtc1202, line 106: I don't think this fix the requirement since run/pipecd always build a new image container  on run, so I guess we still can't use previous image for debugging 🤔

funera1: I see...
Thank you.

PR #4177, opened 2023-02-10
Make the deploying symbol rotate in reverse

khanhtc1202:

Fixes #

```release-note
NONE
```


knanao (commented):
Thanks.
Just in case, I'm glad if you share the demo to check whether working as expected.

khanhtc1202:
Sure, it would look like this

Kapture 2023-02-10 at 14 21 45

knanao (approved):
Thank you.

PR #4181, opened 2023-02-13
Update the interval for Terraform's drift detection

knanao:
For the performance, this changes the interval of Terraform's drift detection.
This param should be mutable through piped config in the end.

Fixes #

```release-note
NONE
```


khanhtc1202:
> This param should be mutable through piped config in the end.

Just to confirm, this change will be addressed by a separated PR, right?

knanao:
@khanhtc1202 
Right, this is for the perfomance check!

khanhtc1202 (approved):
Here you go!

PR #4202, opened 2023-02-20
Enable to ignore manifests' fields from kubernetes' drift detection

### LINKED ISSUE #4190: Enable to ignore manifests' fields from drift detection

**Why is this needed**:

Some autogenerated fields in k8s manifests changed on each rerun, thus users want to ignore them from the drift detection feature. Would be best if we can make it available via application config.

funera1:
Added functionality to prevent drift detection from detecting fields in kubernetes manifests that are set in driftDetection.ignoreFields in application config.

Fixes part of #4190

```release-note
NONE
```


pkg/app/piped/platformprovider/kubernetes/diff.go
```diff
@@ -168,6 +169,9 @@ func (r *DiffListResult) Render(opt DiffRenderOptions) string {
 		} else if opt.MaskConfigMap && key.IsConfigMap() {
 			opts = append(opts, diff.WithMaskPath("data"))
 			needMaskValue = true
+		} else if len(opt.IgnoreFields) > 0 {
```
knanao, line 172: Why `else if` is used here?

pkg/app/piped/driftdetector/kubernetes/detector.go
```diff
@@ -370,3 +375,16 @@ func makeSyncState(r *provider.DiffListResult, commit string) model.ApplicationS
 		Timestamp:   time.Now().Unix(),
 	}
 }
+
+func (d *detector) getDrifteDetectionConfig(repoDir string, app *model.Application) (*config.DriftDetection, error) {
```
knanao, line 379:
```suggestion
func (d *detector) getDriftDetectionConfig(repoDir string, app *model.Application) (*config.DriftDetection, error) {
```

pkg/diff/renderer.go
```diff
@@ -97,6 +104,16 @@ func (r *Renderer) Render(ns Nodes) string {
 	}
 
 	for _, n := range ns {
+		isSkip := false
+		for _, ignorePathPrefix := range r.ignorePathPrefixs {
+			if strings.HasPrefix(n.PathString, ignorePathPrefix) {
+				isSkip = true
+			}
+		}
+		if isSkip {
```
knanao, line 113: Could you make this simpler?

funera1:
@knanao
Please review my code when you have time.

pkg/diff/diff.go
```diff
@@ -77,6 +86,13 @@ func DiffUnstructureds(x, y unstructured.Unstructured, opts ...Option) (*Result,
 }
 
 func (d *differ) diff(path []PathStep, vx, vy reflect.Value) error {
+	for _, ignorePathPrefix := range d.ignorePathPrefixs {
```
knanao, line 89: Please add new test cases to ensure that this works as we expected.

funera1: Ok!

funera1: I thought that ignore in result.addNode is the simplest and most solid way.
I think that if we add such an implementation in the future, there will be an ignore leakage, because there are cases like diffSlice where addNode is done without going through the diff function.
However, if I have the ignore option in the differ struct, I cannot ignore in the result.addNode. What do you think?

pkg/diff/diff.go
```diff
@@ -119,6 +132,7 @@ func (d *differ) diff(path []PathStep, vx, vy reflect.Value) error {
 	}
 
 	switch vx.Kind() {
+	// diffSliceは中でaddNodeしてるから中でもしすり抜けてるならここで省きたい
```
funera1, line 135: Remove the commnet

pkg/diff/diff.go
```diff
@@ -161,12 +175,18 @@ func (d *differ) diffSlice(path []PathStep, vx, vy reflect.Value) error {
 
 	for i := minLen; i < vx.Len(); i++ {
 		nextPath := newSlicePath(path, i)
+		if d.isContainIgnorePathPrefixs(nextPath) {
```
funera1, line 178: Hmmm

pkg/diff/diff.go
```diff
@@ -161,12 +175,18 @@ func (d *differ) diffSlice(path []PathStep, vx, vy reflect.Value) error {
...
+			return nil
+		}
 		nextValueX := vx.Index(i)
 		d.result.addNode(nextPath, nextValueX.Type(), nextValueX.Type(), nextValueX, reflect.Value{})
 	}
 
 	for i := minLen; i < vy.Len(); i++ {
 		nextPath := newSlicePath(path, i)
+		if d.isContainIgnorePathPrefixs(nextPath) {
```
funera1, line 187: Hmmm

pkg/diff/diff.go
funera1: Implementation is possible by providing a wrapper function.

funera1: I implemented it and it went PANIC, so I'm going to proceed the way I've been doing it.

pkg/diff/diff.go
```diff
@@ -26,6 +27,7 @@ type differ struct {
 	ignoreAddingMapKeys           bool
 	equateEmpty                   bool
 	compareNumberAndNumericString bool
+	ignorePathPrefixs             []string
```
knanao, line 30: nits
```suggestion
	ignoredPaths             []string
```

pkg/diff/diff.go
```diff
@@ -56,6 +58,13 @@ func WithCompareNumberAndNumericString() Option {
 	}
 }
 
+// WithIgnorePathPrefixs configures fields to ignore specified in application config.
+func WithIgnorePathPrefixs(ignorePathPrefixs []string) Option {
+	return func(d *differ) {
+		d.ignorePathPrefixs = ignorePathPrefixs
+	}
+}
+
```
knanao, line 67:
```suggestion
// WithIgnoredPaths configures ignored fields.
func WithIgnoredPaths(paths []string) Option {
	return func(d *differ) {
		d.ignoredPaths = paths
	}
}

```

pkg/diff/diff.go
```diff
@@ -322,3 +341,13 @@ func newMapPath(path []PathStep, index string) []PathStep {
 	})
 	return next
 }
+
+func (d *differ) isContainIgnorePathPrefixs(path []PathStep) bool {
```
knanao, line 345:
```suggestion
func (d *differ) isIgnoredPaths(path []PathStep) bool {
```

knanao (commented):
I've left some nits comments, please check it out when you have time.

funera1:
Thanks for the NICE comment. That one is simpler and better.

pkg/diff/diff.go
```diff
@@ -322,3 +341,101 @@ func newMapPath(path []PathStep, index string) []PathStep {
...
+	pathString := makePathString(path)
+	for _, ignoredPath := range d.ignoredPaths {
+		if strings.HasPrefix(pathString, ignoredPath) {
+			return true
+		}
+	}
+	return false
+}
+
+func (d *differ) isIgnoredPathsForString(pathString string) bool {
```
knanao, line 434: This could be replaced by `isIgnoredPaths`.

pkg/diff/diff.go
```diff
@@ -322,3 +341,101 @@ func newMapPath(path []PathStep, index string) []PathStep {
...
 	return next
 }
+
+func (d *differ) addNode(path []PathStep, tx, ty reflect.Type, vx, vy reflect.Value) {
+	if len(d.ignoredPaths) > 0 {
+		if d.isIgnoredPaths(path) {
+			return
+		}
+
+		pathString := makePathString(path)
```
knanao, line 351: How about removing `isIgnoredPathsForString` by calling `makePathStiring` or `path.String()` before that.
```suggestion
		pathString := makePathString(path)
		if d.isIgnoredPaths(pathString) {
			return
		}
```

pkg/diff/diff.go
```diff
@@ -161,22 +174,28 @@ func (d *differ) diffSlice(path []PathStep, vx, vy reflect.Value) error {
 
 	for i := minLen; i < vx.Len(); i++ {
 		nextPath := newSlicePath(path, i)
+		if d.isIgnoredPaths(nextPath) {
+			continue
+		}
```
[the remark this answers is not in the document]
funera1: Oh, it is not necessary. Thank you.

pkg/diff/diff.go
```diff
@@ -161,22 +174,28 @@ func (d *differ) diffSlice(path []PathStep, vx, vy reflect.Value) error {
...
 		nextValueX := vx.Index(i)
-		d.result.addNode(nextPath, nextValueX.Type(), nextValueX.Type(), nextValueX, reflect.Value{})
+		d.addNode(nextPath, nextValueX.Type(), nextValueX.Type(), nextValueX, reflect.Value{})
 	}
 
 	for i := minLen; i < vy.Len(); i++ {
 		nextPath := newSlicePath(path, i)
+		if d.isIgnoredPaths(nextPath) {
+			continue
+		}
```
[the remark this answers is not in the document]
funera1: Oh, it is not necessary. Thank you.

pkg/diff/diff.go
funera1: Your implementation is better. Thank you.

knanao (approved):
Thank you.

Issue #4149, opened 2023-01-27
Add the status of failing to load application config file


knanao:
**What would you like to be added**:
Currently, there is no way to check the error failing to load app config on the console.
It's useful to make it possible to debug its error without checking the log every time. 

https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/trigger/trigger.go#L226

**Why is this needed**:

funera1:
I would like to add the following three functions for convenience
- Get application-id from application-name
- Get a list of deployment-ids from application-id
- Get deployment's stage log from deployment-id

PR #4206, opened 2023-02-21
Add status of failing load application config

funera1:
Add status for load Application failure and added icon for that status.

Fixes part of #4149 

Icon view (As an example, I applied an invalid config status icon instead by the synced status icon.)
image
```release-note
NONE
```


funera1:
@knanao 
Please review my code when you have time.

khanhtc1202:
Would like to use this mark ⚠️ since the current icon in the picture has quite same color with the out of sync icon 👀

funera1:
@khanhtc1202 
Thank you for nice comment. Would you want to use this icon if you add for example WARNING status.
If that problem is ok, I agree to use this icon:smiley:.

khanhtc1202:
@funera1 Got your point. That warning icon maybe mean to some generic warning for sure. But from the users point of view, `warning: application configuration loaded failed` should be treated as a warning, isn't it? 🤔

funera1:
@khanhtc1202 
I understand. That's certainly true. I will apply that icon. Thanks.

funera1:
Changed icon view:smiley:
image

khanhtc1202:
@funera1 how about the text next to it, should not be `Synced`, right? 🤔

khanhtc1202:
@funera1 how about the text next to it, should not be `Synced`, right? 🤔

khanhtc1202:
@funera1 how about the text next to it, should not be `Synced`, right? 🤔

funera1:
@khanhtc1202 
I'm trying with Synced status instead because I can't get INVALID_CONFIG until this PR version of control plane is released.

khanhtc1202:
Oops, got your point, in that case, it's okay to keep it in single PR like this. Thanks 🙌

knanao (approved):

PR #4207, opened 2023-02-21
Fix unable to delete resource that removed from git for k8s apps

khanhtc1202:

Previously due to this wrong iteration make changed via `setNamespace` does not returned after loader.LoadManifests run, thus annotation and logic which based on manifest.ResourceKey.Namespace go wrong way

From
```
ResourceKey: apps/v1:Deployment:default:simple
```
To
```
ResourceKey: apps/v1:Deployment:test:simple
```

Fixes #

```release-note
Fix unable to delete k8s resources which those manifests removed from Git
```

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/loader.go
+++ b/pkg/app/piped/platformprovider/kubernetes/loader.go
@@ -173,8 +173,8 @@ func setNamespace(manifests []Manifest, namespace string) {
 	if namespace == "" {
 		return
 	}
-	for _, m := range manifests {
-		m.Key.Namespace = namespace
+	for i := range manifests {
+		manifests[i].Key.Namespace = namespace
 	}
 }
 


pkg/app/piped/platformprovider/kubernetes/loader.go
```diff
@@ -173,8 +173,8 @@ func setNamespace(manifests []Manifest, namespace string) {
 	if namespace == "" {
 		return
 	}
-	for _, m := range manifests {
-		m.Key.Namespace = namespace
+	for i := range manifests {
+		manifests[i].Key.Namespace = namespace
```
knanao, line 177: I could understand what this wants to do, but originally the manifest is not a pointer, right?

khanhtc1202: basically, this function accepts `manifests` which of type slice as it input. Go makes all function's input pass by value, thus it copy this manifests slide and manipulate thing inside the function. The point to note here is that even it copy the slide, the underlying array of that slide  is remained to be the same, so basically change "that" array will change after function returns. Previously, instead of change the underlying array, `m` is a temporary var that ref to element of slide and change make on it does not reflect on the underlying array thus the changes are not remained after function end.

https://go.dev/play/p/Osr6eY4eh0A

knanao: Oh, I totally understand. Thank you.

knanao (approved):

PR #4223, opened 2023-02-24
Add custom sync stage

kevin-namba:
Add custom sync executor to support deploy tools other than terraform.
```yaml
kind: LambdaApp
spec:
  name: simple
  labels:
    env: example
    team: abc
  planner:
    alwaysUsePipeline: true
  pipeline:
    stages:
      - name: CUSTOM_SYNC
        desc: "test"
        with:
          timeout: 10m
          envs:
            AWS_PROFILE: "piped-test"
          run: |
            sam build
            echo y | sam deploy --profile $AWS_PROFILE
```

Fixes part of https://github.com/pipe-cd/pipecd/issues/1089

```release-note
Add custom stage that users can define how the stage works
```


pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,127 @@
...
+	)
+	ctx := sig.Context()
+	ds, err := e.TargetDSP.Get(ctx, e.LogPersister)
+	if err != nil {
+		e.LogPersister.Errorf("Failed to prepare target deploy source data (%v)", err)
+		return model.StageStatus_STAGE_FAILURE
+	}
+	e.repoDir = ds.RepoDir
+	e.appDir = ds.AppDir
+	if e.StageConfig.CustomStageOptions.Timeout.Duration() != 0*time.Second {
```
knanao, line 67: nits
```suggestion
	if e.StageConfig.CustomStageOptions.Timeout != 0 {
```

pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,127 @@
...
+		timeout = e.StageConfig.CustomStageOptions.Timeout.Duration()
+	}
+
+	c := make(chan model.StageStatus, 1)
+	go func() {
+		result := e.executeCommand(e.StageConfig.CustomStageOptions)
```
knanao, line 73: The arg seems not necessary because the receiver already has it.

pkg/model/stage.go
```diff
@@ -92,6 +92,8 @@ const (
 	// the CANARY variant resources has been cleaned.
 	StageECSCanaryClean Stage = "ECS_CANARY_CLEAN"
 
+	StageCustomStage Stage = "CUSTOM_STAGE"
```
khanhtc1202, line 95:
```suggestion
	// StageCustomStage represents the stage where users can
	// run their defined scripts as part of the deployment.
	StageCustomStage Stage = "CUSTOM_STAGE"
```

khanhtc1202:
Guys, I started to think that should we rename this to `BIN_EXEC` or `SCRIPT_RUNNING`, since the stage named `CUSTOM_STAGE` is a bit too common and may be hard for us later when we want to add new kinds of stages. wdyt?

kevin-namba:
> Guys, I started to think that should we rename this to `BIN_EXEC` or `SCRIPT_RUNNING`, since the stage named `CUSTOM_STAGE` is a bit too common and may be hard for us later when we want to add new kinds of stages. wdyt?

I agree with your opinion. This stage includes only running scripts but users can not customize the stage.

pkg/model/stage.go
```diff
@@ -91,6 +91,9 @@ const (
 	// StageECSCanaryClean represents the stage where
 	// the CANARY variant resources has been cleaned.
 	StageECSCanaryClean Stage = "ECS_CANARY_CLEAN"
+	// StageCustomSync represents the stage where users can
+	// run their defined scripts as part of the deployment.
```
khanhtc1202, line 95:
```suggestion
	// StageCustomSync represents the stage where users can use their
	// defined scripts to sync the application's state instead of the KIND_SYNC stage.
```

pkg/config/testdata/application/custom-stage.yaml
```diff
@@ -0,0 +1,12 @@
...
+spec:
+  pipeline:
+    stages:
+      - name: CUSTOM_SYNC
+        desc: "deploy by sam"
+        with:
+          env: 
+            AWS_PROFILE: default
+          runs: 
+            - "sam deploy -g {{ .AWS_PROFILE }}"
```
khanhtc1202, line 12:
```suggestion
            - "sam deploy -g {{ .AWS_PROFILE }}"

```

pkg/config/application.go
```diff
@@ -396,6 +402,12 @@ type WaitApprovalStageOptions struct {
 	MinApproverNum int      `json:"minApproverNum" default:"1"`
 }
 
+type CustomSyncOptions struct {
+	Timeout Duration          `json:"timeout"`
```
khanhtc1202, line 406: should use default to set default value for this option
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/config/piped.go#L56

pkg/config/application.go
```diff
@@ -396,6 +402,12 @@ type WaitApprovalStageOptions struct {
...
+	Env     map[string]string `json:"env"`
```
khanhtc1202, line 407: Should be `envs`? 🤔

kevin-namba: I name it the same name as Github Actions.
https://docs.github.com/ja/actions/learn-github-actions/variables
and CircleCI uses `environment`
https://circleci.com/docs/ja/env-vars/

pkg/config/application_test.go
```diff
@@ -516,3 +516,66 @@ func TestGenericPostSyncConfiguration(t *testing.T) {
...
+			expectedSpec: &LambdaApplicationSpec{
+				GenericApplicationSpec: GenericApplicationSpec{
+					Timeout: Duration(6 * time.Hour),
+					Pipeline: &DeploymentPipeline{
+						Stages: []PipelineStage{
+							{
+								Name: model.StageCustomSync,
+								Desc: "deploy by sam",
+								CustomSyncOptions: &CustomSyncOptions{
+									Timeout: 600000000000,
```
khanhtc1202, line 541: Why does this value and the one set in yaml does not fit 🤔

pkg/config/application.go
khanhtc1202: I guess we should keep the same convention for all stubs, and we have `runs` as a list of commands (plural), wdyt?

pkg/config/application_test.go
```diff
@@ -516,3 +516,66 @@ func TestGenericPostSyncConfiguration(t *testing.T) {
...
+
+func TestCustomSyncConfig(t *testing.T) {
+	testcases := []struct {
+		fileName           string
+		expectedKind       Kind
+		expectedAPIVersion string
+		expectedSpec       interface{}
+		expectedError      error
+	}{
+		{
```
khanhtc1202, line 528: This stage should require `runs` as it's configuration since defining this stage without runs make no sense. You can refer to other stage validation to know how to add validate step for config.

pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,119 @@
...
+func Register(r registerer) {
+	f := func(in executor.Input) executor.Executor {
+		return &deployExecutor{
+			Input: in,
+		}
+	}
+	r.Register(model.StageCustomSync, f)
+}
+
+// Execute starts waiting for the specified duration.
```
khanhtc1202, line 49:
```suggestion
// Execute exec the user-defined scripts in timeout duration.
```

pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,119 @@
...
+func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
+	var (
+		originalStatus = e.Stage.Status
+	)
```
khanhtc1202, line 53:
```suggestion
	var originalStatus = e.Stage.Status
```

pkg/config/application.go
```diff
@@ -253,6 +255,14 @@ func (s *PipelineStage) UnmarshalJSON(data []byte) error {
...
 
 	switch s.Name {
+	case model.StageCustomSync:
+		s.CustomSyncOptions = &CustomSyncOptions{}
+		if len(gs.With) > 0 {
+			err = json.Unmarshal(gs.With, s.CustomSyncOptions)
+		}
+		if s.CustomSyncOptions.Timeout <= 0 {
+			s.CustomSyncOptions.Timeout = defaultCustomSyncTimeout
+		}
```
khanhtc1202, line 265: No need to do this, we set all default values here

https://github.com/pipe-cd/pipecd/blob/master/pkg/config/config.go#L215-L217

using https://github.com/creasty/defaults

pkg/config/application.go
```diff
@@ -396,6 +406,12 @@ type WaitApprovalStageOptions struct {
 	MinApproverNum int      `json:"minApproverNum" default:"1"`
 }
 
+type CustomSyncOptions struct {
```
khanhtc1202, line 409: Should define validate function for this StageCustomSync and add it here
https://github.com/pipe-cd/pipecd/blob/master/pkg/config/application.go#L135

pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+}
+
+func (e *deployExecutor) executeCommand(opts *config.CustomSyncOptions) model.StageStatus {
+
+	binDir := toolregistry.DefaultRegistry().GetBinDir()
+	pathFromOS := os.Getenv("PATH")
+
+	path := binDir + ":" + pathFromOS
+	envs := make([]string, len(opts.Env))
```
khanhtc1202, line 99:
```suggestion
	envs := make([]string, 0, len(opts.Env))
```
should be like this or `append` at L101 will add new values to len+1, len+2, ...

pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,119 @@
...
+	for key, value := range opts.Env {
+		envs = append(envs, key+"="+value)
+	}
+	for _, v := range opts.Runs {
+		cmd := exec.Command("/bin/sh", "-c", v)
+		e.LogPersister.Infof("RUN %s (env: %v)", v, envs)
+		cmd.Dir = e.appDir
+		cmd.Env = append(os.Environ(), append(envs, "PATH="+path)...)
```
khanhtc1202, line 107: Any way to make this run only once for all `run`? looks like this is triggered for each command in runs.

khanhtc1202: Even if we don't have any way to this `cmd` shared between `opts.Runs`, make those appends to build os.Environ() outside the for loop could be good since the `append` is a costly op.

kevin-namba: I have checked Github Actions and CircleCI and I noticed that run field should not be array.
```yaml
run: |
  command1
  command2
```
To do that, we do not have to run more than once.

pkg/config/application.go
kevin-namba: I thought `env` is not list but map, so it does not have to be plural, but I saw other map field uses plural https://pipecd.dev/docs/user-guide/configuration-reference/
I think `envs` is better in this case, too.

pkg/app/piped/executor/custom/custom.go
khanhtc1202: Nice point, lets adopt that 👍

pkg/config/application.go
khanhtc1202: OKay, lets use `envs` to keep unify with others configurations in pipecd.

pkg/config/application.go
kevin-namba: https://github.com/pipe-cd/pipecd/blob/master/pkg/config/application.go/#L396
WaitApproval timeout define default time  not in the structure but in the unmarshaljson.
https://github.com/pipe-cd/pipecd/blob/master/pkg/config/application.go/#L268

pkg/config/application.go
kevin-namba: https://github.com/pipe-cd/pipecd/pull/4223#discussion_r1129136931

pkg/config/application.go
khanhtc1202: Which do you think is better? Set in executor side or wrap all in configuration stub 🤔

kevin-namba: I think the latter one (set it in the stub) is better in terms of the visibility.

kevin-namba: Thank you for this PR.
https://github.com/pipe-cd/pipecd/pull/4258/files

khanhtc1202: Please add validate function for this stage as I mentioned in https://github.com/pipe-cd/pipecd/pull/4223#discussion_r1128940960

pkg/app/piped/executor/custom/custom.go
```diff
@@ -0,0 +1,123 @@
...
+	path := binDir + ":" + pathFromOS
+	envs := make([]string, 0, len(opts.Envs))
+	for key, value := range opts.Envs {
+		envs = append(envs, key+"="+value)
+	}
+	cmd := exec.Command("/bin/sh", "-c", opts.Run)
+	e.LogPersister.Infof("Runnnig commands...")
+	for _, v := range strings.Split(opts.Run, "\n") {
+		if v != "" {
+			e.LogPersister.Infof("   %s (env: %v)", v, envs)
```
khanhtc1202, line 106: Do we need to print `(env: %v)` here? I mean, we may in case envs is empty and always print this doesn't make sense.

pkg/config/application.go
```diff
@@ -267,6 +278,7 @@ func (s *PipelineStage) UnmarshalJSON(data []byte) error {
 		}
 		if s.WaitApprovalStageOptions.Timeout <= 0 {
 			s.WaitApprovalStageOptions.Timeout = defaultWaitApprovalTimeout
+			fmt.Println(s.WaitApprovalStageOptions.Timeout)
```
khanhtc1202, line 281: Remove unneeded print.

pkg/app/piped/executor/registry/registry.go
```diff
@@ -104,4 +105,5 @@ func init() {
 	ecs.Register(defaultRegistry)
 	wait.Register(defaultRegistry)
 	waitapproval.Register(defaultRegistry)
+	custom.Register(defaultRegistry)
```
khanhtc1202, line 108:
```suggestion
	customsync.Register(defaultRegistry)
```
should rename it like this

pkg/app/piped/executor/custom/custom.go
kevin-namba: I think to show env is bad for visibility, too. And in case users want to see environment variables, users can use echo command to see it.

khanhtc1202: Then lets remove it
```suggestion
			e.LogPersister.Infof("   %s", v)
```

khanhtc1202 (approved):
Nice work! Thank you 🙌

PR #4232, opened 2023-02-28
Add ability to see invalid config errors from application-detail-page

### LINKED ISSUE #4149: Add the status of failing to load application config file

funera1:
You are able to see invalid config errors from application-detail-page when application status is INVALID_CONFIG.

Fixes #4149

```release-note
NONE
```


web/src/components/application-detail-page/application-detail/index.tsx
```diff
@@ -302,7 +302,11 @@ export const ApplicationDetail: FC<ApplicationDetailProps> = memo(
 
               {app.syncState && (
                 <SyncStateReason
-                  summary={app.syncState.shortReason}
+                  summary={
+                    app.syncState.shortReason
+                      ? app.syncState.shortReason
```
khanhtc1202, line 307: does this mean there is a case in which shortReason exists? In such case, please ensure that the shortReason exists in all cases, we want that reason to be shown at this place.

funera1:
I'm trying to check the operation now, but it's going to take some time because pipecd is behaving in a way that I don't expect. Please wait:bow:

funera1:
After checking the behavior, it seemed that this implementation would probably not be able to open application-detail-page.
Therefore, I would like to use the policy of sending shortReason on the piped side.
(In the meantime, the actual image cannot be seen until this policy is merged.)

khanhtc1202:
@funera1 Sorry, I don't fully understand what do you mean, but it looks like you can just push `ShortReason` here just do the same thing with `Reason`, isn't it?

https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/trigger/trigger.go#L239

pkg/app/piped/trigger/trigger.go
```diff
@@ -232,12 +232,17 @@ func (t *Trigger) checkRepoCandidates(ctx context.Context, repoID string, cs []c
...
+				shortReasonLimit = len(err.Error())
+			}
 			req := &pipedservice.ReportApplicationSyncStateRequest{
 				ApplicationId: app.Id,
 				State: &model.ApplicationSyncState{
-					Status:    model.ApplicationSyncStatus_INVALID_CONFIG,
-					Reason:    err.Error(),
-					Timestamp: time.Now().Unix(),
+					Status:      model.ApplicationSyncStatus_INVALID_CONFIG,
+					ShortReason: err.Error()[:shortReasonLimit],
```
khanhtc1202, line 243:
```suggestion
					ShortReason: "failed to load application configuration",
```

pkg/app/piped/trigger/trigger.go
```diff
@@ -232,12 +232,17 @@ func (t *Trigger) checkRepoCandidates(ctx context.Context, repoID string, cs []c
 			)
 
 			// Set ApplicationSyncState to INVALID_CONFIG when LoadApplication fails.
+			shortReasonLimit := 100
+			if len(err.Error()) < shortReasonLimit {
+				shortReasonLimit = len(err.Error())
+			}
```
khanhtc1202, line 238:
```suggestion
```

pkg/app/piped/trigger/trigger.go
funera1: I think this statement is the same meaning as the status INVALID_CONFIG and is not necessary. Instead, if we display part of the actual error (in most cases, the full text, since error messages are not long) in shortReason, the user can easily see the error. What do you think?

khanhtc1202: I think this is a UI problem, since users can see the "actual" error message in detail page; showing it here is redundant is most case. The same idea with `Some thing went wrong` text message on other services; users are supposed to check the error message (along with other details information) in detail page rather than showing everything in the listing page. 

And besides, INVALID_CONFIG is the status and it's displayed by icon, not everyone have sense of what it is by seeing the icon, but text like this would help. WDYT? 🤔

khanhtc1202: And beisdes, it's safer for the UI to be not broken (by long text) without adding unneeded if condition to cut the text if it's longer than limit or thing.

funera1: I see, I understand. Thanks for the good review.

khanhtc1202:
Could you provide picture to show what does it looks like with in use? 😄

funera1:
@khanhtc1202 
Since it's a piped change, you won't see shortReason until it's merged, but Reason will look like this😄
imageimage

khanhtc1202:
> you won't see shortReason until it's merged

Oops, I didn't notice we have this 👀

knanao:
@funera1 
Umm, as I said before, I feel that it's better like the below instead of the show details. wydt?
https://dev.pipecd.dev/applications/d7e5d60d-b321-4aa8-91bc-449bdb61b3e1?project=pipecd

funera1:
@knanao 
So ```ShortReason: err.Error()[:shortReasonLimit]``` is better than ```ShortReason: "failed to load application configuration"```?

funera1:
@knanao 
I see.
I was thinking that we should consider how to deal with long errors, but it doesn't seem it is not so long, and if we are trouble, we can deal with it.
So, I think I will implement the policy of displaying errors without SHOW DETAIL.

funera1:
@knanao 
It looks like I will need to upgrade the version of material-ui to version 5 to implement your policy. I would like to discuss this.

reference: https://zenn.dev/taizo_pro/articles/4aeacbb7c1e104

funera1:
HIDE
image

MORE
image
Now, it looks like this images.

knanao:
@funera1 
Sorry, I could not get your point.
If you wanna use `line-clamp` like the ref at our version, how about using `-webkit-line-clamp` as a css property?

funera1:
I see. I will try it in the way. Thank you.

funera1:
@knanao 
Please review this PR when you have time:bow:

knanao:
@funera1 
Sound good.
Just in case, Could you show me the demo?

web/src/components/application-detail-page/application-detail/sync-state-reason/index.tsx
```diff
@@ -65,3 +65,51 @@ export const SyncStateReason: FC<SyncStateReasonProps> = ({
...
+
+  return (
+    <>
+      <div className={classes.summary}>
+        {showReason && (
+          <Typography variant="body2" color="textSecondary">
+            {detail}
+          </Typography>
+        )}
+        {!showReason && (
```
knanao, line 94: Ternary operator could be used here.

web/src/components/application-detail-page/application-detail/sync-state-reason/index.tsx
```diff
@@ -65,3 +65,51 @@ export const SyncStateReason: FC<SyncStateReasonProps> = ({
...
+          <Typography variant="body2" color="textSecondary">
+            {detail.slice(0, MAX_DISPLAY_LENGTH) + "..."}
+          </Typography>
+        )}
+
+        {detail && (
+          <>
```
knanao, line 101: This is not necessary.

funera1:
@knanao It's demo. How about it?
 For demonstration purposes, MAX_DISPLAY_LENGTH=10. For production, MAX_DISPLAY_LENGTH=100.

https://user-images.githubusercontent.com/60760935/222652820-dd52c276-5daa-4de9-af11-0341c82be692.mov

funera1:
View of MAX_DISPLAY_LENGTH=100
image

web/src/components/application-detail-page/application-detail/sync-state-reason/index.tsx
knanao: No, I mean, `<>` is not necessary.

knanao (approved):
Thank you!

PR #4234, opened 2023-02-28
Fix unable to plan deployment on failed to determine app version

khanhtc1202:

Previously, these functions re-used `err` variable, which defined in function signature. That make the caller side always return failed in case we can't determine the application version (which only be used in showing stub). The right behaviour should be only show version `unknown` and keep deploying as usual.

ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/controller/planner.go#L211-L214

Thanks @kevin55156 for pointing out this 👍 

Fixes #

```release-note
Fix unable to plan deployment if failed to determine application version(s)
```

### DIFF
--- a/pkg/app/piped/planner/cloudrun/cloudrun.go
+++ b/pkg/app/piped/planner/cloudrun/cloudrun.go
@@ -62,8 +62,7 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
 		in.Logger.Warn("unable to determine target version", zap.Error(e))
 	}
 
-	out.Versions, err = p.determineVersions(ds.AppDir, cfg.Input.ServiceManifestFile)
-	if err != nil {
+	if versions, e := p.determineVersions(ds.AppDir, cfg.Input.ServiceManifestFile); e != nil || len(versions) == 0 {
 		in.Logger.Warn("unable to determine target versions", zap.Error(err))
 		out.Versions = []*model.ArtifactVersion{
 			{
@@ -71,6 +70,8 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
 				Version: "unknown",
 			},
 		}
+	} else {
+		out.Versions = versions
 	}
 
 	autoRollback := *cfg.Input.AutoRollback
--- a/pkg/app/piped/planner/cloudrun/cloudrun.go
+++ b/pkg/app/piped/planner/cloudrun/cloudrun.go
@@ -63,7 +63,7 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
 	}
 
 	if versions, e := p.determineVersions(ds.AppDir, cfg.Input.ServiceManifestFile); e != nil || len(versions) == 0 {
-		in.Logger.Warn("unable to determine target versions", zap.Error(err))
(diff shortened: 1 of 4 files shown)


pkg/app/piped/planner/cloudrun/cloudrun.go
```diff
@@ -55,22 +55,23 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
 	}
 
 	// Determine application version from the manifest.
-	if version, e := p.determineVersion(ds.AppDir, cfg.Input.ServiceManifestFile); e == nil {
-		out.Version = version
-	} else {
+	if version, e := p.determineVersion(ds.AppDir, cfg.Input.ServiceManifestFile); e != nil {
```
knanao, line 58: Let`s follow the go practice and this change is just refactoring, right?
```suggestion
	if version, err := p.determineVersion(ds.AppDir, cfg.Input.ServiceManifestFile); err != nil {
```

pkg/app/piped/planner/cloudrun/cloudrun.go
```diff
@@ -55,22 +55,23 @@ func (p *Planner) Plan(ctx context.Context, in planner.Input) (out planner.Outpu
...
 		out.Version = "unknown"
 		in.Logger.Warn("unable to determine target version", zap.Error(e))
+	} else {
+		out.Version = version
 	}
 
-	out.Versions, err = p.determineVersions(ds.AppDir, cfg.Input.ServiceManifestFile)
-	if err != nil {
+	if versions, e := p.determineVersions(ds.AppDir, cfg.Input.ServiceManifestFile); e != nil || len(versions) == 0 {
```
knanao, line 65: 👍

pkg/app/piped/planner/cloudrun/cloudrun.go
khanhtc1202: Sorry for not make it clear, but I want this error to be e instead of err, to avoid shadowing with err defined in function signature 👀

knanao: Sorry, I could not catch your point.
In this way, the err scope is restricted to if statement, right?

khanhtc1202: I guess that `:` is for "version" variable, and err still the outer one. Don't want to take risk there so better to make it clear that it's a different one and be sure we don't return it by return statement.

khanhtc1202: fyi, we did the same in case of kubernetes planner

ref https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/planner/kubernetes/kubernetes.go#L93

knanao: Umm, I guess it's common if there is no specific thing.

> fyi, we did the same in case of kubernetes planner

The same could be said here.

knanao: Okay, I got your point, and let's keep it.
Thanks.

khanhtc1202: Thank you 🙏

knanao (approved):
Here you go!

PR #4239, opened 2023-03-01
Revert "Revert "Change diff detail's header""

khanhtc1202:
Reverts pipe-cd/pipecd#4236

### DIFF
--- a/pkg/app/piped/driftdetector/cloudrun/detector.go
+++ b/pkg/app/piped/driftdetector/cloudrun/detector.go
@@ -207,8 +207,8 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 	d.logger.Info(fmt.Sprintf("application %s has a live service manifest", app.Id))
 
 	result, err := provider.Diff(
-		headManifest,
 		liveManifest,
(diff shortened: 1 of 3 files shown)


knanao:
I feel this should be updated by @funera1 with another PR instead of this revert PR.

khanhtc1202:
> I feel this should be updated by @funera1 with another PR instead of this revert PR.

I don't think so since this fix panic on master, `Out of sync` issue should be carefully fix later 👀

khanhtc1202:
Or another option to resolve panic on master is to revert this https://github.com/pipe-cd/pipecd/pull/4202
wdyt?

knanao:
@khanhtc1202 
> We should resolve panic on master branch to not block other works. wdyt?
Right, but I wonder what the cause of the panic?

khanhtc1202:
okay let's revert #4202 👍
@knanao @funera1

knanao (commented):
@khanhtc1202 
Okay, that's why let's close this.

funera1:
@knanao @khanhtc1202 
I reverted #4202, so please you check it.

PR #4244, opened 2023-03-03
Enable to filter application by its piped

### LINKED ISSUE #4212: Enable to filter applications by its piped

**Why is this needed**:

funera1:
Support for piped filtering of applications list with Web UI and pipectl

Fixes #4212

View example
Each application name is the same as piped name.
image

image

image
```release-note
NONE
```


funera1:
image
image
imageimage

knanao:
Cool, how about removing "All" as well as Application Name because it's not the kind of application.

knanao (commented):
If we filter applications with piped, how do the matched applications line up?
If it's lined up following the updated time, I suppose you need to create a custom index.

funera1:
Do you mean Piped name, not Application name?
The list of Piped names is implemented in Select, not Autocomplete like Application name, so it seemed difficult to specify all Piped without All.

funera1:
>If we filter applications with piped, how do the matched applications line up?
If it's lined up following the updated time, I suppose you need to create a custom index.

I will investigate it. Thank you for nice review.

knanao:
> Do you mean Piped name, not Application name?
Sorry, I mean I thought it's better not to use the "All" property, as well as the labels filter.

funera1:
@knanao 
Please review this PR when you have time:bow:

funera1:
https://user-images.githubusercontent.com/60760935/224869782-d044ef9f-b61b-4c05-be68-54d940b1b54c.mov

knanao (approved):

PR #4246, opened 2023-03-03
Add custom sync rollback(just support lambda)

kevin-namba:
After a pipeline including custom sync fails, the the pipeline should run the same script in the running commit 
as a rollback.

Fixes part of https://github.com/pipe-cd/pipecd/issues/1089

```release-note
Enable us to use rollback when a pipeline including one custom sync stage fails
```


pkg/app/piped/executor/customsync/rollback.go
```diff
@@ -0,0 +1,108 @@
...
+		e.LogPersister.Errorf("Unable to determine the last deployed commit to rollback. It seems this is the first deployment.")
+		return model.StageStatus_STAGE_FAILURE
+	}
+
+	runningDS, err := e.RunningDSP.Get(ctx, e.LogPersister)
+	if err != nil {
+		e.LogPersister.Errorf("Failed to prepare running deploy source data (%v)", err)
+		return model.StageStatus_STAGE_FAILURE
+	}
+	e.repoDir = runningDS.RepoDir
```
khanhtc1202, line 66: Looks like this never be used? 🤔

pkg/model/stage.go
```diff
@@ -101,6 +101,8 @@ const (
 	// This stage is AUTOMATICALLY GENERATED and can not be used
 	// to specify in configuration file.
 	StageRollback Stage = "ROLLBACK"
+
+	StageCustomSyncRollback Stage = "CUSTOM_SYNC_ROLLBACK"
```
khanhtc1202, line 105:
```suggestion
	// StageCustomSyncRollback represents a state where
	// all changes made by the CUSTOM_SYNC stage will be reverted to
	// bring back the pre-deploy stage.
	StageCustomSyncRollback Stage = "CUSTOM_SYNC_ROLLBACK"
```

pkg/app/piped/executor/customsync/rollback.go
```diff
@@ -0,0 +1,107 @@
...
+		return model.StageStatus_STAGE_FAILURE
+	}
+	e.appDir = runningDS.AppDir
+
+	runningCustomSyncConfigs, ok := runningDS.GenericApplicationConfig.GetStagesFromName(model.StageCustomSync)
+	if !ok || runningCustomSyncConfigs == nil {
+		e.LogPersister.Errorf("There are no custom sync in the running commit")
+	}
+	if len(runningCustomSyncConfigs) > 1 {
+		e.LogPersister.Errorf("There are custom sync stages more than one stage.")
```
khanhtc1202, line 73:
```suggestion
		e.LogPersister.Errorf("There are more than one custom sync stages.")
```

khanhtc1202: Also, should we return here? I mean, we decided only to handle rollback in case only one CUSTOM_SYNC stage is used, so processing further, in this case, makes no sense.

pkg/app/piped/executor/customsync/rollback.go
```diff
@@ -0,0 +1,108 @@
...
+
+	runningDS, err := e.RunningDSP.Get(ctx, e.LogPersister)
+	if err != nil {
+		e.LogPersister.Errorf("Failed to prepare running deploy source data (%v)", err)
+		return model.StageStatus_STAGE_FAILURE
+	}
+	e.appDir = runningDS.AppDir
+
+	if len(runningDS.GenericApplicationConfig.Pipeline.Stages) > 1 {
+		e.LogPersister.Errorf("There are more than one custom sync stages.")
```
khanhtc1202, line 69: should return model.StageStatus_STAGE_FAILURE here as well?

khanhtc1202 (approved):
Nice work! Let's 🚀

PR #4199, opened 2023-02-17
Change diff detail's header

funera1:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4248, opened 2023-03-06
Fixed a bug in #4199 that resulted in wrong Out of Sync.

funera1:
Fixed a bug in #4199 that resulted in wrong Out of Sync.
Also, during this modification, the diff option WithIgnoreAddingMapKeys was made commutative. That is, diff a b = diff b a.
I would like to discuss the validity of this modification.

Fixes #

```release-note
NONE
```


funera1:
@knanao 
Please review this PR, when you have time:bow:

knanao (approved):

pkg/app/piped/platformprovider/kubernetes/diff.go
```diff
@@ -53,19 +53,19 @@ type DiffListChange struct {
 func Diff(old, new Manifest, logger *zap.Logger, opts ...diff.Option) (*diff.Result, error) {
 	if old.Key.IsSecret() && new.Key.IsSecret() {
 		var err error
-		new.u, err = normalizeNewSecret(old.u, new.u)
+		old.u, err = normalizeNewSecret(old.u, new.u)
```
khanhtc1202, line 56: sorry but why does `normalizeNewSecret` return `old` stub 🤔 This does not make sense to me 👀

funera1: In a nutshell, it is because we swapped the order of headManifest and liveManifest in Diff.
ignoreAddingMapKeys in diff/diff.go is to ignore LiveManifestw and normalizeNewSecret depends on to ignore LiveManifest with ignoreAddingMapKeys!v

khanhtc1202: Thanks for clarifying, then in that case, why not just mention directly here that it's Live/Head manifest or, even better Base/Compare instead of Old/New one here? 🤔 I'm not talking about the logic but the readability of the code, return value from `normalizeNewSecret` is set to `old` one doesn't make sense to me, that's my point 👀

funera1: I see what you mean. I too have a problem with the fact that old/new is not explicitly tied to live/head.
However, since the relationship between old/new and live/head is not only in this section, but is tied across the board, I also think that a separate Issue or PR should be created to rename it.

Or is your point that normalizeNewSecret should be normalizeOldSecret? If so, I will create an improved PR immediately.

khanhtc1202: >  I also think that a separate Issue or PR should be created to rename it.

Yes, It's what I mean too 👍 The logic around this part is tricky, and if the flow & naming are hard to follow, it's hard for us to debug next time problems occur around here.

funera1: Ok, I will create this issue. Thank you for nice comment:smile:

PR #4258, opened 2023-03-08
Refactor config package to remove unnecessary default values setter

khanhtc1202:

Remove unnecessary default value setters from `GenericApplicationSpec` struct and use the `creasty/defaults` package to set the defaults instead.

Fixes #

```release-note
NONE
```


knanao (commented):
Nice catch.

knanao (approved):

PR #4268, opened 2023-03-15
Fix Missing RBAC configuration bug

knanao:

Fixes #

```release-note
Fix the Missing RBAC configuration bug when login with SSO
```


khanhtc1202 (approved):
Nice catch 👍

PR #4275, opened 2023-03-16
Support update application

sivchari:
Support update application operation through grpc api.

Fixes #

```release-note
Add UpdateApplication rpc for grpc api service
```


khanhtc1202:
Hi @sivchari , thanks for the contribution 😄 
Could you share the background of this change which makes by this PR? Basically, PipeCD's idea is to avoid functions that commonly update model objects done on this PR. The reason is mostly that we want to separate which clients/callers can update which part of the model object clearly so that we can implement our filedb without lock (no concurrent update, thus no need to implement lock object on update)

In the case of the Application model, you can see which part of the model can be updated by which client here:
https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/applicationstore.go#L112

ref: https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0008-simplify-datastore.md for the idea behind PipeCD database design

sivchari:
Hi @khanhtc1202 ,  thanks for the comment.
We're creating the provider of PipeCD, so we want a new API to update application. 
Is it right that `PipedId`, `PlatformProvider`, `GitPath (only GitPath.ConfigFilename)`,  `Disabled` and `Deleted` can be updated ?
If you have a better idea, please let me know. Thanks.

https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/applicationstore.go#L127-L147
```go
	clientShardStruct := model.Application{
		// Fields which required in all shard for validation on update.
		Id:        me.Id,
		ProjectId: me.ProjectId,
		CreatedAt: me.CreatedAt,
		UpdatedAt: me.UpdatedAt,
		// Fields which exist in both AgentShard and ClientShard but AgentShard has
		// a higher priority since those fields can only be updated by PipedCommander.
		Kind: me.Kind,
		Name: me.Name,
		// Fields which exist in both AgentShard and ClientShard but ClientShard has
		// a higher priority since those fields can only be updated by WebCommander.
		PipedId:          me.PipedId,
		PlatformProvider: me.PlatformProvider,
		// Note: Only GitPath.ConfigFilename is changeable.
		GitPath: me.GitPath,
		// Fields which exist only in ClientShard.
		Disabled:  me.Disabled,
		Deleted:   me.Deleted,
		DeletedAt: me.DeletedAt,
	}
```

khanhtc1202:
> Is it right that `PipedId`, `PlatformProvider`, `GitPath (only GitPath.ConfigFilename)`, `Disabled` and `Deleted` can be updated ?

@sivchari Yes, as you mentioned, those fields are currently allowed to be updated via "client" (pipectl and web). And I think the new terraform provider would act as a new client (just as other current clients to pipecd, which are pipectl and web), so we should keep this convention to not break the pipecd filedb model.

pkg/datastore/applicationstore.go
```diff
@@ -268,6 +269,14 @@ func (s *applicationStore) List(ctx context.Context, opts ListOptions) ([]*model
 	return apps, cursor, nil
 }
 
+func (s *applicationStore) Update(ctx context.Context, app *model.Application) error {
```
khanhtc1202, line 272: @sivchari Also, my point is if the function name is called UpdateXXX (UpdateApplication in this case), others may assume that this method updates the whole Application model, but in fact, we only update some fields and have to check which is updated and which is not. That is why instead of building a common update whole model, I created an Update function that fits the usecase (as you called in this function body: UpdateBasicInfo, UpdateConfiguration, UpdateSyncState)

sivchari: OK, I created a new API called UpdateApplication to communicate via gRPC, I think the naming is correct since it only scopes PipeCD applications, the API calls the Update function that fits the usecase. What do you think?

sivchari:
@khanhtc1202 
This PR is ready for review.

pkg/app/server/service/apiservice/service.proto
```diff
@@ -150,6 +151,18 @@ message DisableApplicationResponse {
     string application_id = 1 [(validate.rules).string.min_len = 1];
 }
 
+message UpdateApplicationRequest {
+    string application_id = 1 [(validate.rules).string.min_len = 1];
+    string piped_id = 2 [(validate.rules).string.min_len = 1];
+    string platform_provider = 3 [(validate.rules).string.min_len = 1];
+    model.ApplicationGitPath git_path = 4 [(validate.rules).message.required = true];
+    bool disabled = 5;
```
khanhtc1202, line 159: Looks like this is redundant since we have Disable/Enable Application rpc already 👀

khanhtc1202:
Sorry for the late. Could you re-run `make gen/code` for this change?

pkg/app/server/service/apiservice/service.proto
sivchari: Surely, I delete disabled field.

sivchari:
@khanhtc1202 
I fixed it.

pkg/app/server/service/apiservice/service.proto
```diff
@@ -150,6 +151,17 @@ message DisableApplicationResponse {
     string application_id = 1 [(validate.rules).string.min_len = 1];
 }
 
+message UpdateApplicationRequest {
+    string application_id = 1 [(validate.rules).string.min_len = 1];
+    string piped_id = 2 [(validate.rules).string.min_len = 1];
```
khanhtc1202, line 156: This field is not being used in the current implementation of UpdateApplication rpc, right?
If you want to update that field as well, it's okay to make it. Also, I read through the current implementation of UpdateApplication rpc, and it looks like it does exactly the same as we do in web_api.go (ref [code](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/grpcapi/web_api.go#L530-L551)). So how about updating it like that? (Of course we have to keep the API key check stub for the client api.go version)

sivchari: Thanks for good review. `piped_id` can be referenced from application, so this field doesn't need and I improve the implementation of UpdateApplication rpc.

pkg/app/server/grpcapi/api.go
```diff
@@ -46,6 +46,8 @@ type apiApplicationStore interface {
 	Enable(ctx context.Context, id string) error
 	Disable(ctx context.Context, id string) error
 	UpdateConfigFilename(ctx context.Context, id, filename string) error
+	UpdateConfiguration(ctx context.Context, id, pipedID, platformProvider, configFilename string) error
+	UpdatePlatformProvider(ctx context.Context, id string, platFormProvider string) error
```
khanhtc1202, line 50: this can be removed due to unused.

pkg/app/server/grpcapi/api.go
```diff
@@ -343,6 +345,30 @@ func (a *API) ListApplications(ctx context.Context, req *apiservice.ListApplicat
...
+	app, err := getApplication(ctx, a.applicationStore, req.ApplicationId, a.logger)
+	if err != nil {
+		return nil, err
+	}
+
+	if app.ProjectId != key.ProjectId {
+		return nil, status.Error(codes.InvalidArgument, "Requested application does not belong to your project")
+	}
+
+	if err := a.applicationStore.UpdateConfiguration(ctx, req.ApplicationId, app.PipedId, req.PlatformProvider, req.GitPath.ConfigFilename); err != nil {
```
khanhtc1202, line 363: Instead of `app.PipedId` should be `req.PipedId`. I mean, you can even change the piped which handles this application via this grpc, so why not?
Just as the same we do in web_api.go (ref: [code](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/grpcapi/web_api.go#L530-L551))

pkg/app/server/grpcapi/api.go
```diff
@@ -343,6 +344,30 @@ func (a *API) ListApplications(ctx context.Context, req *apiservice.ListApplicat
...
+		return nil, err
+	}
+
+	piped, err := getPiped(ctx, a.pipedStore, req.PipedId, a.logger)
+	if err != nil {
+		return nil, gRPCStoreError(err, fmt.Sprintf("failed to get piped %s", req.PipedId))
+	}
+
+	if piped.ProjectId != key.ProjectId {
+		return nil, status.Error(codes.InvalidArgument, "Requested application does not belong to your project")
```
khanhtc1202, line 359:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "Requested piped does not belong to your project")
```

khanhtc1202:
@sivchari Nice work, thanks 🙏 
Only a small change is required. Please update these comments

https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/applicationstore.go#L138
https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/applicationstore.go#L164

to `a higher priority since those fields can only be updated by WebCommander/PipectlCommander.` due to the change of this PR.

sivchari:
Ready for review.

pkg/app/server/grpcapi/api.go
khanhtc1202, line 359:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "Requested piped does not belong to your project")
```

khanhtc1202 (approved):
Here you go 🙌

khanhtc1202:
@sivchari Sorry, could you rebase this branch by pulling the master branch? I just removed the CLA sign action on master, so maybe the actions triggered by this PR can't be done due to CLA not found issue.
https://github.com/pipe-cd/pipecd/pull/4335

sivchari:
@khanhtc1202 
Hmm... I know this changes, so I've already done rebased.

khanhtc1202:
@sivchari Could you try close this PR and open a new one.

PR #4284, opened 2023-03-23
Fix to unable to plan deployment if failed to load manifest files

kevin-namba:
https://github.com/pipe-cd/pipecd/pull/4234 fixed to enable  to plan deployment if determineVersion fails, but terraform and kubernetes even fails if loadManifest fails.

Fixes #

```release-note
NONE
```


khanhtc1202:
Sorry but I don't get your point on this change. It doesn't make sense to keep planning deployment if we cannot load all application manifests; that's reasonable, right?

kevin-namba:
> @kevin-namba Sorry but I don't get your point on this change. It doesn't make sense to keep planning deployment if we cannot load all application manifests; that's reasonable, right?
Using custom sync, developers do not prepare terraform module or kubernetes manifests, but currently pipecd can not plan pipeline for terraform or kubernetes.

PR #4289, opened 2023-03-26
Add attachment feature that supports attaching files to application manifests

khanhtc1202:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -535,7 +535,7 @@ func (e *SecretEncryption) Validate() error {
 }
 
 type Attachment struct {
-	// List of files which contain embedding source data.
+	// Map of name to refer with the file path which contain embedding source data.
 	Sources map[string]string `json:"sources"`
 	// List of files to be embedded before using.
 	Targets []string `json:"targets"`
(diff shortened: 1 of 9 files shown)


pkg/config/application.go
```diff
@@ -519,6 +527,25 @@ func (e *SecretEncryption) Validate() error {
 	return nil
 }
 
+type Attachment struct {
+	// List of files which contain embedding source data.
+	Sources map[string]string `json:"sources"`
```
knanao, line 532: It's possible to understand necessary to specify the source path here, but it might be hard to get why this field is map instead of slice just from this comment.

knanao (commented):
Can we merge this PR before the next release?
If not, let's make this hold temporarily.

khanhtc1202:
> Can we merge this PR before the next release?

I plan to attach this change to the next release but not mention it in docs, so we can make it available as an alpha test version for some teams.

pkg/config/application.go
khanhtc1202: Updated, thank you 👍

khanhtc1202:
Changed my mind, and will make this available as a `pre-release` version. Thank you 👍

knanao (approved):
Here you go!

PR #4291, opened 2023-03-27
Fix docs Piped Cloud Run service manifest

hosht:

Move `run.googleapis.com/ingress` and `run.googleapis.com/ingress-status` to the correct location.
ref. https://cloud.google.com/run/docs/reference/yaml/v1

`run.googleapis.com/ingress-status` is an output-only annotation.
It may not be necessary.

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch, thank you 👍

PR #4292, opened 2023-03-27
Add support for setting ignore fields for each resource key

funera1:

It didn’t distinguish fields by resource key to ignore drift detection. So, this PR resolves the problem.

Input form is `apiVersion:kind:namespace:name#filedPath`.

For example,
```
…
driftDetection: 
    ignoreFields: 
        - apps/v1:Deployment:default:simple#spec.replicas 
        - apps/v1:Deployment:default:simple#spec.template.spec.containers.0.args
```

You may find it annoying to have to specify the Key, but we plan to add a future feature that will allow you to easily copy the path of the field you want to ignore from the web ui

Fixes #

```release-note
NONE
```


pkg/app/piped/driftdetector/kubernetes/detector.go
```diff
@@ -193,9 +193,17 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 		return err
 	}
 
-	ignoreFields := make([]string, 0)
+	ignoreConfig := make(map[string][]string, 0)
```
khanhtc1202, line 196: nits, can we rename it ad ignoreConfig 👀

pkg/diff/diff.go
```diff
@@ -59,14 +60,14 @@ func WithCompareNumberAndNumericString() Option {
 }
 
 // WithIgnoredPaths configures ignored fields.
-func WithIgnoredPaths(paths []string) Option {
+func WithIgnoreConfig(config map[string][]string) Option {
```
khanhtc1202, line 63: change the comment on above this function header, the name is not matched

pkg/app/piped/driftdetector/kubernetes/detector.go
```diff
@@ -193,9 +193,17 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
...
 	if ddCfg != nil {
-		ignoreFields = ddCfg.IgnoreFields
+		for _, ignoreField := range ddCfg.IgnoreFields {
+			// ignoreField is 'apiVersion:kind:namespace:name#fieldPath'
```
khanhtc1202, line 199: Should add this line to the config/application as well to mention what we expect as driftDetection.ignoreFields

funera1: I changed it. How about that?

pkg/diff/diff.go
funera1: :+1:

pkg/app/piped/driftdetector/kubernetes/detector.go
khanhtc1202: 👍

khanhtc1202 (approved):
Nice work, thank you 🙌

PR #4295, opened 2023-03-28
Fix error handling in to parse ignoreFields

funera1:
Since it is returning when the input is not in key#path, it seems that drift detection is only working when the input is correct.

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
👍

PR #4301, opened 2023-03-29
Rearrange configuration reference docs sections and add missing docs

khanhtc1202:

This pull rearrange configuration reference docs sections and add missing config sections like WaitApprovalStageOptions, WaitStageOptions

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/content/en/docs-dev/user-guide/configuration-reference.md
+++ b/docs/content/en/docs-dev/user-guide/configuration-reference.md
@@ -108,6 +108,7 @@ spec:
 | name | string | The application name. | Yes if you set the application through the application configuration file |
 | labels | map[string]string | Additional attributes to identify applications. | No |
 | description | string | Notes on the Application. | No |
+| input | [LambdaDeploymentInput](#lambdadeploymentinput) | Input for Lambda deployment such as path to function manifest file... | No |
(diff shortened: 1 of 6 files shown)


docs/content/en/docs-dev/user-guide/configuration-reference.md
```diff
@@ -183,7 +184,7 @@ spec:
 | labels | map[string]string | Additional attributes of event. This can make an event definition unique even if the one with the same name exists. | No |
 | replacements | [][EventWatcherReplacement](#eventwatcherreplacement) | List of places where will be replaced when the new event matches. | Yes |
 
-## EventWatcherReplacement
+### EventWatcherReplacement
```
knanao, line 187: What are the differences between `##` and `###`?
If everything nested in `##` are `###`, `KubernetesDeploymentInput` and more should be under `###`?

khanhtc1202: You are right! All `###` will be nested under `##`, but the reason why I don't make `KubernetesDeploymentInput` as `###` is because that section and its parent are too far from each other, so that makes it nested there makes no sense. If a section stands right after its parent, then I make it nested, if it is far from its parent, keep it the same as the level of the surrounding section. That's the logic of the current structure.

You can see the difference below.

Before
Screen Shot 2023-03-30 at 10 01 07

After
Screen Shot 2023-03-30 at 10 01 18

knanao (approved):

PR #4303, opened 2023-03-29
Change application invalid config detail UI

khanhtc1202:

Change application invalid config detail UI from using button to using end text

Before

https://user-images.githubusercontent.com/32532742/228563611-0be70d30-f407-4e05-b863-d81ccc807ab4.mp4

After

https://user-images.githubusercontent.com/32532742/228563651-ab0f0f06-72e1-4193-ab1b-ed1f4851c705.mp4

Fixes #

```release-note
NONE
```

### DIFF
--- a/web/src/components/application-detail-page/application-detail/sync-state-reason/index.tsx
+++ b/web/src/components/application-detail-page/application-detail/sync-state-reason/index.tsx
@@ -76,7 +76,7 @@ export const InvalidConfigReason: FC<SyncStateReasonProps> = ({ detail }) => {
   const [showReason, setShowReason] = useState(false);
 
   const msgHeader = "Failed to load aplication config: ";
-  const MAX_DISPLAY_LENGTH = 70;
+  const MAX_DISPLAY_LENGTH = 200;
   if (detail.length < MAX_DISPLAY_LENGTH) {
     return (
       <>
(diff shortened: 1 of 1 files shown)


web/src/components/application-detail-page/application-detail/sync-state-reason/index.tsx
```diff
@@ -100,14 +105,12 @@ export const InvalidConfigReason: FC<SyncStateReasonProps> = ({ detail }) => {
...
-            variant="text"
-            size="small"
-            className={classes.showButton}
+          <span
+            className={classes.showText}
             onClick={() => setShowReason(!showReason)}
           >
-            {showReason ? "HIDE" : "MORE"}
-          </Button>
+            {showReason ? "show less" : "show more"}
```
knanao, line 112:
```suggestion
            {showReason ? "SHOW LESS" : "SHOW MORE"}
```

knanao (commented):
By default, it might be better not to display the button, and set the `more details` only if the text is more than the max length.

khanhtc1202:
> By default, it might be better not to display the button, and set the `more details` only if the text is more than the max length.

Yes, it is already like that. In the sample video, I changed the length to show how it is when the size of the text is over the max showing value.
https://github.com/pipe-cd/pipecd/pull/4303/files#diff-f52d077868b409e7cfd23c0ada42583c3323f153f6e4af8fe1a490d51a148f1eL75

knanao (approved):

PR #4314, opened 2023-04-05
Add RegisterPiped api for grpcapi

arabian9ts:
Support register piped operation through grpc api.

I'm developing terraform provider for pipecd.
I'd like to manage pipecd application with piped id created by terraform as following.
I added RegisterPiped rpc for grpc api for that reason.

```hcl
resource "pipecd_piped" "something" {
  name = "example-piped"
  desc = "This is example."
}

resource "pipecd_application" "something" {
  name     = "example-application"
  piped_id = pipecd_piped.something.id
  kind     = "CLOUDRUN"
  ...
}
```

In the process of that, I found there are no function to enable applications.

Fixes #

```release-note
Add RegisterPiped api for grpcapi
```


knanao (commented):
Thank you for your contribution, and this looks good.
But be careful that this gPRC is not installing piped but just registering piped to the control plane.
Hence it is necessary to install piped using helm or something in addition to this.

PR #4317, opened 2023-04-10
Bump google.golang.org/api

knanao:
Updates google.golang.org/api module to use the latest google.golang.org/api/run/v1.
The old version doesn't support the fields, livenessProbe and startupProbe for Cloud Run, hence we need to update it.

Fixes #

```release-note
NONE
```


test/integration/filestore/gcs/gcs_test.go
```diff
@@ -195,23 +195,3 @@ func TestList(t *testing.T) {
 		})
 	}
 }
-
-// Check if it panics when connect with closed client.
-func TestClose(t *testing.T) {
```
khanhtc1202, line 200: Why do we need to remove this test? Did the library change its behavior? 🤔

knanao: Yes, it was changed to not cause panic when the client already was closed, but it's not big deal for us.

khanhtc1202 (approved):
Here you go 👍

PR #4355, opened 2023-05-29
Support slack api

### LINKED ISSUE #4352: Support Slack API to send notification

Now, PipeCD only supports Webhook URL to send notification to Slack. Slack provides the API to post a message. I think it's good to support this way, too.

**Why is this needed**:

Slack announces Incoming Webhooks is deperecated.
ref. https://api.slack.com/legacy/custom-integrations/messaging/webhooks

sivchari:

Fixes #4352 

I support a new way to send notification to slack channel.

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/notifier/slack.go
+++ b/pkg/app/piped/notifier/slack.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	slackgo "github.com/slack-go/slack"
 	"go.uber.org/zap"
 
 	"github.com/pipe-cd/pipecd/pkg/config"
@@ -40,12 +41,13 @@ const (
 )
 
 type slack struct {
-	name       string
-	config     config.NotificationReceiverSlack
-	webURL     string
-	httpClient *http.Client
-	eventCh    chan model.NotificationEvent
-	logger     *zap.Logger
+	name        string
+	config      config.NotificationReceiverSlack
+	webURL      string
+	httpClient  *http.Client
+	slackClient *slackgo.Client
+	eventCh     chan model.NotificationEvent
+	logger      *zap.Logger
 }
 
 func newSlackSender(name string, cfg config.NotificationReceiverSlack, webURL string, logger *zap.Logger) *slack {
@@ -56,8 +58,9 @@ func newSlackSender(name string, cfg config.NotificationReceiverSlack, webURL st
 		httpClient: &http.Client{
 			Timeout: 5 * time.Second,
 		},
-		eventCh: make(chan model.NotificationEvent, 100),
-		logger:  logger.Named("slack"),
+		slackClient: slackgo.New(cfg.OAuthToken),
+		eventCh:     make(chan model.NotificationEvent, 100),
+		logger:      logger.Named("slack"),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/config/piped.go
```diff
@@ -919,13 +919,18 @@ func (n *NotificationReceiver) Mask() {
 }
 
 type NotificationReceiverSlack struct {
-	HookURL string `json:"hookURL"`
+	HookURL    string `json:"hookURL"`
+	OAuthToken string `json:"oauthToken"`
+	ChannelID  string `json:"channelID"`
```
khanhtc1202, line 924: Could you add validation for this struct? I think we need something like
- Ensure that if HookURL already available, no OAuthToken / ChannelID be set (since we only use one of them), and the revert condition as well.
- If the OAuthToken is available ChannelID is required as well since we need that to push message in case of send message via API

pkg/app/piped/notifier/slack.go
```diff
@@ -101,12 +104,19 @@ func (s *slack) sendEvent(ctx context.Context, event model.NotificationEvent) {
 		s.logger.Info(fmt.Sprintf("ignore event %s", event.Type.String()))
 		return
 	}
-	if err := s.sendMessage(ctx, msg); err != nil {
-		s.logger.Error(fmt.Sprintf("unable to send notification to slack: %v", err))
+	if len(s.config.HookURL) != 0 {
+		if err := s.sendMessageViaHookURL(ctx, msg); err != nil {
+			s.logger.Error(fmt.Sprintf("unable to send notification to slack: %v", err))
+		}
```
khanhtc1202, line 110: Only one of `sendMessageViaHookURL` and `sendMessageViaAPI` should be called, so maybe we should add a return statement here to ensure we don't double send messages.

sivchari:
@khanhtc1202 
I fixed it. Please review 🙇

pkg/config/piped.go
```diff
@@ -919,13 +926,35 @@ func (n *NotificationReceiver) Mask() {
...
+	}
+	if n.OAuthToken != "" {
+		if n.HookURL != "" {
+			return errors.New("slack notification receiver: oauthToken is set, so hookURL should be empty")
+		}
+		if n.ChannelID == "" {
+			return errors.New("slack notification receiver: oauthToken is set, so channelID should not be empty")
+		}
+	}
+	return nil
```
khanhtc1202, line 957:
```suggestion
	if n.HookURL != "" && n.OAuthToken != "" {
		return errors.New("only one of hookURL or oauthToken and channelID should be used")
	}
	
	if n.HookURL != "" {
		return nil
	}
	
	if n.OAuthToken == "" || n.ChannelID == "" {
		return errors.New("missing oauthToken or channelID configuration")
	}
	return nil
```
nits, how about this?

sivchari:
```go
	if n.HookURL != "" && n.OAuthToken != "" {
		return errors.New("only one of hookURL or oauthToken and channelID should be used")
	}
```
If HookURL and ChannelID is set, is this configuration valid ?

```go
	if n.OAuthToken == "" || n.ChannelID == "" {
		return errors.New("missing oauthToken or channelID configuration")
	}
```
👍

sivchari: @khanhtc1202 
wdyt?

khanhtc1202: I see, then, it should be

```golang
	if n.HookURL != "" && (n.OAuthToken != "" || n.ChannelID == "") {
```

khanhtc1202:
Thanks @sivchari, this PR is LGTM 🙏 
Could you update your commit messages and add `Sign-off-by` information? We're following the DCO for now, and requires signing off every commits 🙏 
https://pipecd.dev/docs/contribution-guidelines/contributing/#contributor-license-agreement

sivchari:
Improve UT 😸

pkg/config/piped.go
```diff
@@ -947,7 +947,7 @@ func (n *NotificationReceiverSlack) Validate() error {
 	if n.HookURL != "" {
 		return nil
 	}
-	if n.OAuthToken == "" || n.ChannelID == "" {
+	if (n.OAuthToken != "" && n.ChannelID == "") || (n.OAuthToken == "" && n.ChannelID != "") {
```
khanhtc1202, line 950: Not sure why we should adopt this over the previous version 😄 Previous one fit the error message more than the current condition and it's easier to read 🤔

sivchari: The previous version returns error, if HookURL and OauthToken, ChannelID doesn't set.

sivchari: wdyt?

khanhtc1202: > The previous version returns error, if HookURL and OauthToken, ChannelID doesn't set.

I don't think so, after L947 ( If `n.HookURL != ""` check ), the HookURL must be empty. Then after that, `if n.OAuthToken == "" || n.ChannelID == ""` if one of oauthToken or channelID is empty, it returns error. That's what I mean, isn't it?

sivchari: In the first place, NotificationReceiverSlack requires HookURL or OAuthToken and ChannelID. So it's no way that both parameters is empty. I completely misunderstand this point. Sorry 🐱

khanhtc1202 (approved):
Nice work! Thank you 👏

sivchari:
🚀

khanhtc1202:
@sivchari Please also update `/docs-dev` reflects this change on notification configuration 🙏

PR #4356, opened 2023-05-29
Add versioning and release cycle to docs

kentakozuka:

Defining detail of our release strategy.


RELEASES.md
```diff
@@ -1,8 +1,22 @@
...
+## Release Cycle
+- Major releases: arbitrary timing
+- Minor releases: roughly every 2 months
+- Patch releases: roughly every 2-3 weeks
+
+Release candidates(vX.X.X-alpha) for minor versions are available for 2-4 weeks before the release.
 
 ## Major release
-This refers to the release of new features.
+This refers to the release of breaking changes.
```
khanhtc1202, line 19: I don't think we're following this right? 🤔 I mean, we have not released any major yet, and minor does contain breaking changes. Directly mentioning that we're not follow the Semantic version maybe better.

RELEASES.md
```diff
@@ -1,8 +1,22 @@
-# Release Process
-This document explains the process to release new version and to address patch release.
+# Releases
+This document explains the process to release new versions.
+
+## Versioning
+Versions are expressed as `vX.Y.Z` as well as Semantic Versioning;
```
khanhtc1202, line 5: I don't think we're following this right? 🤔 I mean, we have not released any major yet, and minor does contain breaking changes. Directly mentioning that we're not follow the Semantic version maybe better.

RELEASES.md
```diff
@@ -1,8 +1,22 @@
...
+- `X` is the major version
+- `Y` is the minor version
+- `Z` is the patch version
+
+## Release Cycle
+- Major releases: arbitrary timing
+- Minor releases: roughly every 2 months
+- Patch releases: roughly every 2-3 weeks
+
+Release candidates(vX.X.X-alpha) for minor versions are available for 2-4 weeks before the release.
```
khanhtc1202, line 16:
```suggestion
Release candidates(vX.Y.Z-alpha) for minor versions are available for 2-4 weeks before the release.
```

RELEASES.md
```diff
@@ -1,8 +1,22 @@
...
-This document explains the process to release new version and to address patch release.
+# Releases
+This document explains the process to release new versions.
+
+## Versioning
+Versions are expressed as `vX.Y.Z` as well as Semantic Versioning;
+
+- `X` is the major version
+- `Y` is the minor version
+- `Z` is the patch version
```
khanhtc1202, line 9:
```suggestion
- `X` is the major version
- `Y` is the minor version <- current major release
- `Z` is the patch version <- current minor / patch release
```

RELEASES.md
khanhtc1202, line 19:
```suggestion
This refers to the release of new features or breaking changes.
```
If only breaking changes, then this would rarely happen since we don't expect breaking change.

sivchari: I think PipeCD should only once release major version at some point. That way,  we can explain a major version is `This refers to the release of breaking changes (PipeCD committed to backward compatibility, so a next breaking changes will not happen)`.

khanhtc1202: In such case, all new features also only be released as patch, that sounds not really nice 🤔

sivchari: I focus on major release.  I think all new features is released as minor, if PipeCD secure backward compatibility or do these feature have no backward compatibility you mention to ?

RELEASES.md
kentakozuka: I meant that versions are "expressed" as well as SemVer, but it does not mean that we follow the rules of it. It may confuse readers. I will fix it 👍

RELEASES.md
kentakozuka: IMHO, we need to use the term "minor versions" for the current major versions (`Y`), so that everyone cannot be confused. Also, we need to use the term "patch versions" for the current minor / patch versions (`Z`). WDYT?

RELEASES.md
kentakozuka: If we increment major versions(`X`), minor versions(`Y`) always change, too. So I guess the explanation of major releases can simply refer to "breaking changes".

khanhtc1202 (approved):
Thank you 🙏

PR #4369, opened 2023-06-05
Platform provider deployment detail page

### LINKED ISSUE #3418: Show platformProvider in the deployment details page

**Why is this needed**:

anhpnv:
- Display Platform Provider in Deployment detail page

image

Fixes #3418 

```release-note
NONE
```

### DIFF
--- a/web/src/components/deployments-detail-page/deployment-detail/index.tsx
+++ b/web/src/components/deployments-detail-page/deployment-detail/index.tsx
@@ -166,7 +166,11 @@ export const DeploymentDetail: FC<DeploymentDetailProps> = memo(
                       </Link>
                     }
                   />
-                  <DetailTableRow label="Piped" value={piped.name + " (provider:" + " " + deployment.platformProvider + ")" } />
+                  <DetailTableRow label="Piped" value={piped.name} />
+                  <DetailTableRow
+                    label="Platform Provider"
+                    value={deployment.platformProvider}
+                  />
                   <DetailTableRow label="Summary" value={deployment.summary} />
                 </tbody>
               </table>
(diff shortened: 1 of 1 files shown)


khanhtc1202:
Hi @anhpnv, thanks for your contribution! Since the `Platform Provider` information belongs (and sticks) to the piped, how about to show the platform provider like
```
Piped test-piped (provider: kubernetes-default)
```
instead of making it a new separated label in the deployment detail block 🤔

anhpnv:
> Hi @anhpnv, thanks for your contribution! Since the `Platform Provider` information belongs (and sticks) to the piped, how about to show the platform provider like
> 
> ```
> Piped test-piped (provider: kubernetes-default)
> ```
> 
> instead of making it a new separated label in the deployment detail block 🤔

Hi @khanhtc1202 , i understand, i 'll check it 😃

anhpnv:
image
I have changed the position of Platform Provider.
And i have an ideal to make it more highlight, like this:
image
What do you think @khanhtc1202 😄😄

khanhtc1202:
Could you provider that image which "labeled" platform provider is shown with other field as well, just worried it looks not fit with the surrounding 😄

khanhtc1202:
Btw, this is how it looks like in the case of showing the platform provider in the application detail page. I think we should keep the same way for simplicity
Screen Shot 2023-06-06 at 15 37 16

anhpnv:
> Btw, this is how it looks like in the case of showing the platform provider in the application detail page. I think we should keep the same way for simplicity Screen Shot 2023-06-06 at 15 37 16

I agree, you mean like this, right ?

image

khanhtc1202 (approved):
LGTM, thanks 🙏

Issue #4349, opened 2023-05-25
Support Terraform import block


golemiso:
**What would you like to be added**:

[Terraform v1.5.0-beta2](https://github.com/hashicorp/terraform/releases/tag/v1.5.0-beta2) has new feature `import block`.
I tried that with PipeCD and faced an error.

error message:
```
reason: failed while calculating diff, unable to parse plan output
```

plan result with import block (this occurs parse-error):
```
Plan: 1 to import, 11 to add, 0 to change, 0 to destroy.
```

**Why is this needed**:

khanhtc1202:
Hi @golemiso, thanks for reporting. Just want to confirm, did you try that with this configuration set to `v1.5.0-beta2` 👀 
https://pipecd.dev/docs/user-guide/configuration-reference/#terraformdeploymentinput:~:text=No-,terraformVersion,-string

golemiso:
I did.

`unable to parse plan output` occurred because of regexp.
https://github.com/pipe-cd/pipecd/blob/534e95756010012d6561e2f5ba5d945cf7c9bd5a/pkg/app/piped/platformprovider/terraform/terraform.go#L316

PR #4396, opened 2023-06-14
Fix import block cause failure of parsing regex

kentakozuka:

Fixes [#](https://github.com/pipe-cd/pipecd/issues/4349)

```release-note
NONE
```

### DIFF
--- a/pkg/app/piped/platformprovider/terraform/terraform.go
+++ b/pkg/app/piped/platformprovider/terraform/terraform.go
@@ -314,7 +314,7 @@ func (t *Terraform) makeCommonCommandArgs() (args []string) {
 
 var (
 	// Import block was introduced from Terraform v1.5.0.
-	// TODO: Show import in output.
+	// TODO: Show import in output and add it to diff calculation.
 	planHasChangeRegex = regexp.MustCompile(`(?m)^Plan:(?: \d+ to import,)?? (\d+) to add, (\d+) to change, (\d+) to destroy.$`)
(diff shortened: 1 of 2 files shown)


pkg/app/piped/platformprovider/terraform/terraform.go
```diff
@@ -313,7 +313,9 @@ func (t *Terraform) makeCommonCommandArgs() (args []string) {
 }
 
 var (
-	planHasChangeRegex = regexp.MustCompile(`(?m)^Plan: (\d+) to add, (\d+) to change, (\d+) to destroy.$`)
+	// Import block was introduced from Terraform v1.5.0.
+	// TODO: Show import in output.
+	planHasChangeRegex = regexp.MustCompile(`(?m)^Plan:(?: \d+ to import,)?? (\d+) to add, (\d+) to change, (\d+) to destroy.$`)
```
khanhtc1202, line 318: Could you add a test to ensure this does not break piped with Terraform version before v1.5.0 🙏

pkg/app/piped/platformprovider/terraform/terraform_test.go
```diff
@@ -0,0 +1,36 @@
+package terraform
```
khanhtc1202, line 1: Please add the file header (the copyright text)

kentakozuka: oh yea. forgot it. Thanks!

PR #4400, opened 2023-06-15
Remove pre-installation tools from piped-base image

kentakozuka:

Fixes #

```release-note
NONE
```


khanhtc1202:
@kentakozuka Maybe we should separate this file removal to another PR so that if we need to revert this, just only revert the installer script back since we completely removed asdf from our plan already.
ref: pkg/app/piped/toolregistry/tool_shared.go

kentakozuka:
> @kentakozuka Maybe we should separate this file (ref: pkg/app/piped/toolregistry/tool_shared.go) removal to another PR so that if we need to revert this, just only revert the installer script back since we completely removed asdf from our plan already.

@khanhtc1202 Fixed 👍

khanhtc1202 (approved):
Let's go 🚀

PR #4405, opened 2023-06-19
Update piped-base without preinstall tools

kentakozuka:

Fixes #

```release-note
NONE
```


cmd/piped/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # https://github.com/pipe-cd/pipecd/pkgs/container/piped-base/22991493?tag=v0.31.0-30-g13f4bf9
```
khanhtc1202, line 1: Please update the comment as well

kentakozuka: @khanhtc1202 Thanks. Fixed 👍

khanhtc1202 (approved):
Let's go 🚀

Issue #4413, opened 2023-06-21
Managed certificates and backend config are missing with helm upgrade


kentakozuka:
**What happened**:

When running `helm upgrade pipecd` with both gateway and grafana managed certificate and backend config, pipecd's objects cannot be applied.

This is because `{{-` remove new line on the left side and the templating result will be;

```
# Source: pipecd/templates/backend-config.yaml
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: pipecd
spec:
  iap:
    enabled: true
    oauthclientCredentials:
      secretName: pipecd-iap---
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: pipecd-grafana
spec:
  iap:
    enabled: true
    oauthclientCredentials:
      secretName: pipecd-monitoring-iap
```

**What you expected to happen**:

```
# Source: pipecd/templates/backend-config.yaml
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: pipecd
spec:
  iap:
    enabled: true
    oauthclientCredentials:
      secretName: pipecd-iap
---
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: pipecd-grafana
spec:
  iap:
    enabled: true
    oauthclientCredentials:
      secretName: pipecd-monitoring-iap
```

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

PR #4414, opened 2023-06-21
Fix managed certificates and backend config are missing with helm upg…

kentakozuka:

Fixes https://github.com/pipe-cd/pipecd/issues/4413

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch 👍

Issue #4422, opened 2023-06-21
Default settings of grafana ingresss is incorrect


kentakozuka:
**What happened**:
Now, the default settings for grafana ingress is the following;
```yaml
  ingress:
    enabled: false
    hosts: []
    path: /* # set in pipecd chart values.yaml
    pathType: Prefix # set in grafana chart values.yaml
```

If the default settings are applied, ingress encounters an error;

```
> kubectl describe ingresses.networking.k8s.io pipecd-grafana
...
Warning  Translate  7m26s (x174 over 16h)  loadbalancer-controller  Translation failed: invalid ingress spec: failed to validate prefix path /* due to invalid wildcard
```

The error says that `pathType: Prefix` does not support wildcard `/*`.
I we want to allow all routes, the settings should be either;
1. `pathType: Prefix`, `/`
2. `pathType: ImplementationSpecific`, `/*`

Reference:
- https://kubernetes.io/ja/docs/concepts/services-networking/ingress/#%E3%83%91%E3%82%B9%E3%81%AE%E3%82%BF%E3%82%A4%E3%83%97
- https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2066

**What you expected to happen**:

The `pathType: ImplementationSpecific` is literally "implemented specifically" by providers so that `pathType: Prefix`, `/` is better and this is [grafana's default settings](https://github.com/grafana/helm-charts/blob/55e562b2dae9a0cb9b3180f399ca1ab014c34b53/charts/grafana/values.yaml#L234-L237).

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

PR #4423, opened 2023-06-21
Fix path settings of grafana ingresss is incorrect

kentakozuka:

Fixes https://github.com/pipe-cd/pipecd/issues/4422

```release-note
NONE
```


khanhtc1202:
Looks like Grafana chart changed its behavior since our current stg works with the previous configuration 🤔 Lets update and see how it works on dev 👍

khanhtc1202 (approved):

PR #4435, opened 2023-06-25
Refactor fmt.Errorf to log.Fatal

seipan:
Refactored because it was notated with fmt.Errorf and os.Exit instead of log.Fatal

Fixes #

```release-note
NONE
```


khanhtc1202 (approved):
Nice catch, thanks 👍

Issue #4452, opened 2023-06-28
Documentation search shows results of each versions


kentakozuka:
**What happened**:

**What you expected to happen**:

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

PR #4453, opened 2023-06-28
Fix document search shows pages that are not selected

kentakozuka:

Fixes https://github.com/pipe-cd/pipecd/issues/4452


khanhtc1202 (approved):
Looks neat 💯

Issue #4116, opened 2023-01-10
Quickstart support Arm architecture


funera1:
**What would you like to be added**:
I want to get it to work without ``docker pull mysql:8.0.23 -platform linux/amd64`` on Arm architecture.

**Why is this needed**:

funera1:
reference: [Multi-architecture images with image indexes](https://kubernetes.io/docs/concepts/containers/images/#multi-architecture-images-with-image-indexes)

PR #4457, opened 2023-06-29
Update quickstart to support arm hardware(ex: M1)

gitbluf:
**What this PR does / why we need it**: Currently quickstart doesn't work on ARM machines unless manual steps are taken.

**Which issue(s) this PR fixes**: #4116

Fixes # Quckstart for arm architecture CPU's (tested on M1 Pro)
Update Helm chart values to support different mysql image.

- **How are users affected by this change**: Enables arm users to easily use quickstart

### DIFF
--- a/pkg/app/pipectl/cmd/quickstart/quickstart.go
+++ b/pkg/app/pipectl/cmd/quickstart/quickstart.go
@@ -150,7 +150,7 @@ func (c *command) installControlPlane(ctx context.Context, helm string, input cl
 		"--values",
 		fmt.Sprintf(helmQuickstartValueRemotePath, c.version),
 		"--set",
-		fmt.Sprintf("mysql.image=%s", selectImage()),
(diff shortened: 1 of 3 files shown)


pkg/app/pipectl/cmd/quickstart/quickstart.go
```diff
@@ -312,6 +314,19 @@ func openbrowser(url string) error {
 	return err
 }
 
+func selectMySqlImage() string {
```
khanhtc1202, line 317: should be `selectMySQLImage()`

khanhtc1202:
Thanks @gitbluf , overall looks great, just a small nits 🙌

pkg/app/pipectl/cmd/quickstart/quickstart.go
gitbluf: just pushed the update

khanhtc1202 (approved):
Great! Thank you 🙌

PR #4463, opened 2023-07-03
Specify Version For Templating Tools

### LINKED ISSUE #4382: Better way to specify version for templating tools

```yaml
kind: KubernetesApp
spec:
  input:
    namespace: sample
    helmVersion: 3.8.2
    kustomizeVersion: 5.0.3
...
```

This configuration should install helm version v3.8.2 and kustomize version v5.0.3, both in the piped container environment, but it's not work like that now.

**Why is this needed**:

Kustomize can be configured to be used with helm via `enable-helm` & `helm-command` flags. The PipeCD currently supports [kustomizeOptions](https://pipecd.dev/docs/user-guide/configuration-reference/#kubernetesdeploymentinput:~:text=No-,kustomizeOptions,-map%5Bstring%5Dstring), which is a map of options for kustomize build command, so basically users can use `enable-helm` if they want.

The point to note here is that even if they use `helm-command` to specify helm version they want to use with kustomize build, there is no way to have that version of helm command via PipeCD configuration currently.

anhpnv:

1. Install both specify version Helm and Kustomize

Fixes #4382 

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/loader.go
+++ b/pkg/app/piped/platformprovider/kubernetes/loader.go
@@ -91,11 +91,8 @@ func (l *loader) LoadManifests(ctx context.Context) (manifests []Manifest, err e
 	}()
 	l.initOnce.Do(func() {
 		l.templatingMethod = determineTemplatingMethod(l.input, l.appDir)
-		switch l.templatingMethod {
-		case TemplatingMethodHelm:
+		if l.templatingMethod != TemplatingMethodNone {
 			l.helm, l.initErr = l.findHelm(ctx, l.input.HelmVersion)
-
-		case TemplatingMethodKustomize:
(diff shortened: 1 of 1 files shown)


anhpnv:
I have updated my code, my test case is i have installed wrong version of helm and kustomize:
```
  helmVersion: 3.27.9
  kustomizeVersion: 5.88.7
```
The point is if i write code like this
```go
l.helm, l.initErr = l.findHelm(ctx, l.input.HelmVersion)
l.kustomize, l.initErr = l.findKustomize(ctx, l.input.KustomizeVersion)
```
It may not have error for installing helm, and in web console like:
image

With my updated code, it display both :
image

What do you think ? 😃

khanhtc1202:
The error is not really user friendly but it works for me, I think it's clear enough 👍

khanhtc1202 (approved):
Thank you 🤝

PR #4468, opened 2023-07-07
Fix GA4 data is not sent

kentakozuka:

According to [docs](https://gohugo.io/templates/internal/#use-the-google-analytics-template), async script is no suitable for GA4.

Fixes 


khanhtc1202 (approved):
look neat to me 👍

PR #4489, opened 2023-07-12
Fix mockgen does not support errors.Join

kentakozuka:

Use https://github.com/uber/mock because https://github.com/golang/mock is no longer maintained.

Fixes 


khanhtc1202 (approved):

khanhtc1202:
Fixed by https://github.com/pipe-cd/pipecd/pull/4499 so we can close this 👍

PR #4493, opened 2023-07-13
Modify severity when unable to determine versions in Kubernetes

kanata2:
This error is not critical error and there are cases that users have no way to fix.
In addition to this, the severity of same error in other providers(cloudrun, lambda, ecs, terraform) is already set WARN.

None

No


khanhtc1202 (approved):
Nice catch, thank you 🙏

PR #4504, opened 2023-07-15
Add spec autoCreateNamespace configuration for creating a ns when doesn't exist

hungran:
- hey team, I'm new here so really need your help and correct me if I'm wrong on smt
this PR will try to use `kubectl create namespace <ns_name>` when we put flag `AutoCreateNamespace` to avoid issue when the ns doesn't exit

- basis I created field `AutoCreateNamespace` under `KubernetesDeploymentInput`  following this comment https://github.com/pipe-cd/pipecd/pull/4150#pullrequestreview-1272107822

**Which issue(s) this PR fixes**: https://github.com/pipe-cd/pipecd/issues/1202

Fixes #

- **How are users affected by this change**: ? not sure


docs/content/en/docs-dev/user-guide/configuration-reference.md
```diff
@@ -304,6 +304,7 @@ One of `yamlField` or `regex` is required.
 | helmOptions | [HelmOptions](#helmoptions) | Configurable parameters for helm commands. | No |
 | namespace | string | The namespace where manifests will be applied. | No |
 | autoRollback | bool | Automatically reverts all deployment changes on failure. Default is `true`. | No |
+| AutoCreateNamespace | bool | Automatically create a new namespace if it does not exist. Default is `false`. | No |
```
khanhtc1202, line 307:
```suggestion
| autoCreateNamespace | bool | Automatically create a new namespace if it does not exist. Default is `false`. | No |
```

pkg/config/application_kubernetes.go
```diff
@@ -94,6 +94,10 @@ type KubernetesDeploymentInput struct {
 	// Automatically reverts all deployment changes on failure.
 	// Default is true.
 	AutoRollback *bool `json:"autoRollback,omitempty" default:"true"`
+
+	// Automatically create a new namespace if it does not exist.
+	// Default is false.
+	AutoCreateNamespace *bool `json:"autoCreateNamespace,omitempty" default:"false"`
```
khanhtc1202, line 100:
```suggestion
	AutoCreateNamespace bool `json:"autoCreateNamespace,omitempty"`
```
We only need to use the type `*bool` in case of the default value is set to true, or else the bool type on golang has the default value is set to false 😄 
ref: https://github.com/creasty/defaults/issues/30#issuecomment-942435138 (we're using this package for defaults values in configuration object)

pkg/config/application_kubernetes_test.go
```diff
@@ -97,7 +97,8 @@ func TestKubernetesApplicationConfig(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: newBoolPointer(false),
```
khanhtc1202, line 101:
```suggestion
					AutoCreateNamespace: false,
```
or just remove it, due to the default value is set to false for the bool type in Golang.

pkg/config/application_test.go
```diff
@@ -424,7 +427,8 @@ func TestTrueByDefaultBoolConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: newBoolPointer(true),
```
khanhtc1202, line 431:
```suggestion
					AutoCreateNamespace: true,
```

pkg/config/application_kubernetes.go
hungran: oh cool thanks, I couldn't imagined that from the time being!

pkg/config/application_kubernetes_test.go
hungran: yeah, got it as mentioned above! thanks

pkg/config/application_kubernetes_test.go
```diff
@@ -135,7 +136,8 @@ func TestKubernetesApplicationConfig(t *testing.T) {
...
```
hungran, line 140:
```suggestion
					AutoCreateNamespace: false,
```

pkg/config/application_test.go
```diff
@@ -315,7 +315,8 @@ func TestGenericTriggerConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: newBoolPointer(false),
```
hungran, line 319:
```suggestion
					AutoCreateNamespace: false,
```

pkg/config/application_test.go
```diff
@@ -366,7 +367,8 @@ func TestTrueByDefaultBoolConfiguration(t *testing.T) {
...
```
hungran, line 371:
```suggestion
					AutoCreateNamespace: false,
```

pkg/config/application_test.go
```diff
@@ -395,7 +397,8 @@ func TestTrueByDefaultBoolConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(false),
+					AutoRollback:        newBoolPointer(false),
+					AutoCreateNamespace: newBoolPointer(false),
```
hungran, line 401:
```suggestion
					AutoCreateNamespace: false,
```

pkg/config/application_test.go
```diff
@@ -493,7 +497,8 @@ func TestGenericPostSyncConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: newBoolPointer(false),
```
hungran, line 501:
```suggestion
					AutoCreateNamespace: false,
```

pkg/config/application_test.go
```diff
@@ -613,7 +618,8 @@ func TestGenericAnalysisConfiguration(t *testing.T) {
...
```
hungran, line 622:
```suggestion
					AutoCreateNamespace: false,
```

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -63,6 +63,15 @@ func (a *applier) ApplyManifest(ctx context.Context, manifest Manifest) error {
...
 		return a.initErr
 	}
+	if a.input.AutoCreateNamespace == true {
+		if err := a.kubectl.CreateNamespace(
+			ctx,
+			a.platformProvider.KubeConfigPath,
+			a.getNamespaceToRun(manifest.Key),
+		); err != nil {
+			return err
+		}
```
khanhtc1202, line 73: This will be called at each manifest applying (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/executor/kubernetes/kubernetes.go#L236) means that the previously applied manifest creates NS may cause the latter manifest apply to return an error in case of creatingNamespace run. I think we should treat the error here more carefully, like:
- error.Is(NamespaceAlreadyExistErr, err) -> ignore
- other error -> return err
WDYT? 🤔

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -80,6 +89,15 @@ func (a *applier) CreateManifest(ctx context.Context, manifest Manifest) error {
...
```
khanhtc1202, line 99: Same with https://github.com/pipe-cd/pipecd/pull/4504/files#r1270312251

khanhtc1202, line 99: Same as https://github.com/pipe-cd/pipecd/pull/4504#discussion_r1270312251

pkg/app/piped/platformprovider/kubernetes/applier.go
hungran: thanks for the feedback ❤️ 
I aware that so rather than call CreateNamespace, for the time being, I was though can have a func CheckNamespaceExist but sound be same because we run it every time manifest applying
so in the CreateNamespace func, it won't throw error if namespace already exist 

https://github.com/pipe-cd/pipecd/pull/4504/files#diff-0f62d61e20f316f805df0111036eaec110ebcbb6792639dda6bb483581832df3R237-R242
just wonder if I can have other better output rather than using `fmt.Printf`

khanhtc1202: Yes, you can check out this implementation 😄 I mean, we can create a new error named `errorNamespaceAlreadyExists` like this one. And at this place, we can check whether the error here is that error or not, if that error is already exist, just ignore it, or else return err.
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/kubernetes/kubectl.go#L31-L32

hungran: @khanhtc1202 thanks, I don't mind to implement the handle the error if ns already exist which also came in mind when I started 😅, just though if we need it or not because as I mentioned above, in the func CreateNamespace will not throw error if it exist. Please correct me if I missing something

hungran: Well, nevermind, I think got your point, will Implement that. Thank you 😀 @khanhtc1202

hungran: @khanhtc1202 again, thank you, I implemented handle err and ignore if match the case

pkg/app/piped/platformprovider/kubernetes/kubectl.go
```diff
@@ -28,8 +28,10 @@ import (
 )
 
 var (
-	errorReplaceNotFound = errors.New("specified resource is not found")
-	errorNotFoundLiteral = "Error from server (NotFound)"
+	errorReplaceNotFound    = errors.New("specified resource is not found")
+	errorNotFoundLiteral    = "Error from server (NotFound)"
+	errReplaceAlreadyExists = errors.New("resource already exists")
```
khanhtc1202, line 33:
```suggestion
	errResourceAlreadyExists = errors.New("resource already exists")
```
should be like this?

pkg/app/piped/platformprovider/kubernetes/kubectl.go
```diff
@@ -223,3 +225,22 @@ func (c *Kubectl) Get(ctx context.Context, kubeconfig, namespace string, r Resou
...
+	if kubeconfig != "" {
+		args = append(args, "--kubeconfig", kubeconfig)
+	}
+	args = append(args, "create", "namespace", namespace)
+
+	cmd := exec.CommandContext(ctx, c.execPath, args...)
+	out, err := cmd.CombinedOutput()
+
+	if strings.Contains(string(out), errAlreadyExistsLiteral) {
+		return errReplaceAlreadyExists
```
khanhtc1202, line 240: Rename due to https://github.com/pipe-cd/pipecd/pull/4504#discussion_r1271260320

pkg/config/testdata/application/truebydefaultbool-false-explicitly.yaml
```diff
@@ -6,3 +6,4 @@ spec:
       disabled: false
   input:
     autoRollback: false
+    autoCreateNamespace: false
```
khanhtc1202, line 9: Default is false so no need to add this

pkg/config/testdata/application/k8s-app-helm.yaml
```diff
@@ -36,3 +37,4 @@ spec:
     helmValueFiles:
     - values.yaml
     helmVersion: 3.1.1
+    autoCreateNamespace: false
```
khanhtc1202, line 40: Default is false so no need to add this

pkg/config/application_test.go
```diff
@@ -613,7 +618,8 @@ func TestGenericAnalysisConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: false,
```
khanhtc1202, line 622: Default is false so no need to add this

pkg/config/application_test.go
```diff
@@ -493,7 +497,8 @@ func TestGenericPostSyncConfiguration(t *testing.T) {
...
```
khanhtc1202, line 501: Default is false so no need to add this

pkg/config/application_test.go
```diff
@@ -395,7 +397,8 @@ func TestTrueByDefaultBoolConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(false),
+					AutoRollback:        newBoolPointer(false),
+					AutoCreateNamespace: false,
```
khanhtc1202, line 401: Default is false so no need to add this

pkg/config/application_test.go
```diff
@@ -315,7 +315,8 @@ func TestGenericTriggerConfiguration(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: false,
```
khanhtc1202, line 319: Default is false so no need to add this

pkg/config/application_test.go
```diff
@@ -366,7 +367,8 @@ func TestTrueByDefaultBoolConfiguration(t *testing.T) {
...
```
khanhtc1202, line 371: Default is false so no need to add this

pkg/config/application_kubernetes_test.go
```diff
@@ -97,7 +97,8 @@ func TestKubernetesApplicationConfig(t *testing.T) {
 					},
 				},
 				Input: KubernetesDeploymentInput{
-					AutoRollback: newBoolPointer(true),
+					AutoRollback:        newBoolPointer(true),
+					AutoCreateNamespace: false,
```
khanhtc1202, line 101: Default is false so no need to add this

pkg/config/application_kubernetes_test.go
```diff
@@ -135,7 +136,8 @@ func TestKubernetesApplicationConfig(t *testing.T) {
...
```
khanhtc1202, line 140: Default is false so no need to add this

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -64,6 +64,17 @@ func (a *applier) ApplyManifest(ctx context.Context, manifest Manifest) error {
 		return a.initErr
 	}
 
+	if a.input.AutoCreateNamespace == true {
+		err := a.kubectl.CreateNamespace(
+			ctx,
+			a.platformProvider.KubeConfigPath,
+			a.getNamespaceToRun(manifest.Key),
+		)
+		if err != nil && !errors.Is(err, errReplaceAlreadyExists) {
```
khanhtc1202, line 73: Rename due to https://github.com/pipe-cd/pipecd/pull/4504#discussion_r1271260320

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -81,6 +92,17 @@ func (a *applier) CreateManifest(ctx context.Context, manifest Manifest) error {
...
```
khanhtc1202, line 101: Rename due to https://github.com/pipe-cd/pipecd/pull/4504#discussion_r1271260320

khanhtc1202 (commented):
Looks neat, just left a few nits. Thank you ❤️

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -81,6 +92,17 @@ func (a *applier) CreateManifest(ctx context.Context, manifest Manifest) error {
 		return a.initErr
 	}
 
+	if a.input.AutoCreateNamespace == true {
```
khanhtc1202, line 95:
```suggestion
	if a.input.AutoCreateNamespace {
```

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -64,6 +64,17 @@ func (a *applier) ApplyManifest(ctx context.Context, manifest Manifest) error {
...
```
khanhtc1202, line 67:
```suggestion
	if a.input.AutoCreateNamespace {
```

khanhtc1202 (dismissed):
Great work, thank you ❤️

khanhtc1202:
@hungran Could you check the false test 🙏 
https://github.com/pipe-cd/pipecd/actions/runs/5634502628/job/15264954333?pr=4504
(On local, you can run make test/go to run the test as on the CI)

hungran:
> @hungran Could you check the false test 🙏 https://github.com/pipe-cd/pipecd/actions/runs/5634502628/job/15264954333?pr=4504 (On local, you can run make test/go to run the test as on the CI)

@khanhtc1202  yeah, I might missed run the test after changed the test file, fixed

pkg/config/testdata/application/k8s-app-helm.yaml
```diff
@@ -22,6 +22,7 @@ spec:
     helmValueFiles:
     - values.yaml
     helmVersion: 3.1.1
+    autoCreateNamespace: true
```
khanhtc1202, line 25: Do we need this?

hungran: I think no, will remove it :) thanks

khanhtc1202 (approved):
Thank you 👍

PR #4512, opened 2023-07-19
Add caution note to RBAC documentation

seitarof:
I stumbled upon this part when setting up the User Group, so I thought it would be better to include it in the documentation.

Fixes #


kentakozuka (approved):
Amazing!
Thank you so much :)

PR #4518, opened 2023-07-20
Add ability to set apiKey and applicationKey for AnalysisProviderDatadogConfig directly instead of reading files

### LINKED ISSUE #4492: Support setting apiKey and applicationKey for AnalysisProviderDatadogConfig directly instead of just via file

Currently, the AnalysisProviderDatadogConfig configuration only supports auth information to be set via file (mount to the piped container). It would be great if we could support setting those values directly via fields like `apiKeyData` and `applicationKeyData` as for other configurations.

ref: https://pipecd.dev/docs-v0.41.x/user-guide/managing-piped/configuration-reference/#analysisproviderdatadogconfig

**Why is this needed**:

hungran:

add an option to set apiKey + applicationKey AnalysisProviderDatadogConfig directly in yaml file

Fixes #4492

### DIFF
--- a/docs/content/en/docs-dev/user-guide/managing-piped/configuration-reference.md
+++ b/docs/content/en/docs-dev/user-guide/managing-piped/configuration-reference.md
@@ -180,8 +180,8 @@ Must be one of the following structs:
 | address | string | The address of Datadog API server. Only "datadoghq.com", "us3.datadoghq.com", "datadoghq.eu", "ddog-gov.com" are available. Defaults to "datadoghq.com" | No |
 | apiKeyFile | string | The path to the api key file. | Yes |
 | applicationKeyFile | string | The path to the application key file. | Yes |
-| apiKey | string | Optional: API Key for Datadog API server. | No |
-| applicationKey | string | Optional: Application Key for Datadog API server | No |
+| apiKeyData | string | Base64 API Key for Datadog API server. Either apiKeyData or apiKeyFile must be set | No |
+| applicationKeyData | string | Base64 Application Key for Datadog API server. Either applicationKeyFile or applicationKeyData must be set | No |
 
 ## EventWatcher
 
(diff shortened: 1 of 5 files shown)


khanhtc1202:
@hungran Yes, it's. Thanks so much for contributions 😄

docs/content/en/docs-dev/user-guide/managing-piped/configuration-reference.md
```diff
@@ -179,6 +179,8 @@ Must be one of the following structs:
 | address | string | The address of Datadog API server. Only "datadoghq.com", "us3.datadoghq.com", "datadoghq.eu", "ddog-gov.com" are available. Defaults to "datadoghq.com" | No |
 | apiKeyFile | string | The path to the api key file. | Yes |
 | applicationKeyFile | string | The path to the application key file. | Yes |
+| apiKey | string | Optional: API Key for Datadog API server. | No |
+| applicationKey | string | Optional: Application Key for Datadog API server | No |
```
khanhtc1202, line 183:
```suggestion
| apiKeyData | string | Optional: API Key for Datadog API server. | No |
| applicationKeyData | string | Optional: Application Key for Datadog API server | No |
```
should add suffix `Data` to unify with other options like [this](https://github.com/pipe-cd/pipecd/blob/master/docs/content/en/docs-dev/user-guide/managing-piped/configuration-reference.md?plain=1#L218)

khanhtc1202: Also as the piped convention, the field which holds key/token data should hold the base64 encode version of the key/token, so I think we should do the same here as well.

ref: https://pipecd.dev/docs-v0.44.x/user-guide/managing-piped/configuration-reference/#:~:text=Yes-,pipedKeyData,-string

hungran:
> @hungran Thanks so much for contributions 😄
> 
> We should allow only one of those fields (if it provides the same data) to avoid misleading. Ref to other fields if you want reference 👍 https://github.com/pipe-cd/pipecd/blob/master/pkg/config/piped.go#L1101-L1115

Thanks for feedback, got your point, I added it

docs/content/en/docs-dev/user-guide/managing-piped/configuration-reference.md
hungran: yes, better if we hold key/token as base64 when direct on yaml file. Implemented that! 🙏

PR #4520, opened 2023-07-21
Add application related labels to deployment_status metric

kanata2:
**What this PR does / why we need it**:
Added application_id, application_name labels to deployment_status metric.

**Which issue(s) this PR fixes**:

Fixes: #4475

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**: No
- **Is this breaking change**: N/A
- **How to migrate (if breaking change)**: N/A

PR #4521, opened 2023-07-21
Fix deployment status use an incorrect status

kentakozuka:

Fixes https://github.com/pipe-cd/pipecd/pull/4520/files#r1270234381


khanhtc1202 (approved):
👍

PR #4522, opened 2023-07-21
Disable terraform drift detection

kentakozuka:

Part of https://github.com/pipe-cd/pipecd/issues/4508

### DIFF
--- a/docs/content/en/docs-v0.44.x/user-guide/managing-piped/configuration-reference.md
+++ b/docs/content/en/docs-v0.44.x/user-guide/managing-piped/configuration-reference.md
@@ -110,7 +110,6 @@ Must be one of the following structs:
 | Field | Type | Description | Required |
 |-|-|-|-|
 | vars | []string | List of variables that will be set directly on terraform commands with `-var` flag. The variable must be formatted by `key=value`. | No |
-| driftDetectionEnabled | bool | Enable drift detection. Default is `true` | No |
 
 ### PlatformProviderCloudRunConfig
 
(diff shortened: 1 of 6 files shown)


docs/content/en/docs-v0.44.x/user-guide/managing-piped/configuration-reference.md
```diff
@@ -110,6 +110,7 @@ Must be one of the following structs:
 | Field | Type | Description | Required |
 |-|-|-|-|
 | vars | []string | List of variables that will be set directly on terraform commands with `-var` flag. The variable must be formatted by `key=value`. | No |
+| driftDetectionEnabled | bool | Enable drift detection. Default is `true` | No |
```
khanhtc1202, line 113: Not yet release, so we don't need this.

pkg/config/piped.go
```diff
@@ -642,6 +642,7 @@ type PlatformProviderTerraformConfig struct {
 	// 'image_id_list=["ami-abc123","ami-def456"]'
 	// 'image_id_map={"us-east-1":"ami-abc123","us-east-2":"ami-def456"}'
 	Vars []string `json:"vars,omitempty"`
+	DriftDetectionEnabled *bool `json:"driftDetectionEnabled" default:"true"`
```
khanhtc1202, line 645: I got your point on making this flag here, but still think it's a bit complicated to make this here. Users must know which provider controls the current application if they want to disable the application drift detection feature 🤔 Maybe we should find a better place for this. Still, yes, it solves the current problem. How about adding TODO to find a better place for this or mentioning that this is a temp instead of making this a fix for https://github.com/pipe-cd/pipecd/issues/4508 and closing the issue 🤔

kentakozuka: Agree. I added comments.

kentakozuka:
@khanhtc1202 Fixed 👍

khanhtc1202 (approved):
Thank you 🙏

Issue #4530, opened 2023-07-27
Support enable selection of listener rules


moko-poi:
**What would you like to be added**:
Allow PipeCD to choose which listener rules to rewrite

**Why is this needed**:
Currently, PipeCD updates the listener rules for the default setting. 

When multiple listener rules exist, the default one is not used, preventing the use of canary release.

khanhtc1202:
This issue changed to ensure the ECS_TRAFFIC_ROUTING stage does not change other rules than Forwarding type rule

PR #4540, opened 2023-08-07
feat(ECS): enable selection of listener rules

moko-poi:

Fixes https://github.com/pipe-cd/pipecd/issues/4530

```
kind: ECSApp
input:
  taskDefinitionFile: taskdef.yaml
  serviceDefinitionFile: servicedef.yaml
  targetGroups:
    primary:
    canary:
```


pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -415,9 +433,16 @@ func routing(ctx context.Context, in *executor.Input, platformProviderName strin
 		return false
 	}
 
-	if err := client.ModifyListeners(ctx, currListenerArns, routingTrafficCfg); err != nil {
-		in.LogPersister.Errorf("Failed to routing traffic to PRIMARY/CANARY variants: %v", err)
-		return false
+	if len(listenerRules) == 0 {
+		if err := client.ModifyListeners(ctx, currListenerArns, routingTrafficCfg); err != nil {
```
kentakozuka, line 437: Move [these lines](https://github.com/pipe-cd/pipecd/blob/ce26e5b6ae53ad95d6f8b4a3998a2a9a21d743e3/pkg/app/piped/executor/ecs/ecs.go#L412-L416) inside this block because only this line references `currListenerArns`.

kentakozuka (commented):
Thank you for the PR!
Can you update [docs](https://github.com/pipe-cd/pipecd/blob/master/docs/content/en/docs-dev/user-guide/configuration-reference.md#ecsdeploymentinput) as well?

pkg/app/piped/executor/ecs/ecs.go
moko-poi: Thank you for the review, @kentakozuka!

I've made the suggested changes and moved the lines inside the block as per your feedback. 
[bb1b2f3](https://github.com/pipe-cd/pipecd/pull/4540/commits/bb1b2f34d29a6f4b6a4e2cc011b1775569fec1c7)

 I've also updated the documentation to reflect the changes made in this PR.
 [9a2a9c0](https://github.com/pipe-cd/pipecd/pull/4540/commits/9a2a9c09c763930abfefe7983c84b189a43718b9)

Please let me know if there are any other points I should address.

kentakozuka (commented):
Thanks! LGTM 🎉 
@khanhtc1202 Can you review for approval?

moko-poi:
@kentakozuka @khanhtc1202 
The CI has failed due to codecov/patch.

The coverage drop is due to missing test files. Adding tests might be beyond this PR's scope. 
How should we address this?

Best regards,

khanhtc1202:
@moko-poi Don't worry about that. We can approve and ignore that 👌

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -409,15 +427,21 @@ func routing(ctx context.Context, in *executor.Input, platformProviderName strin
...
-	currListenerArns, err := client.GetListenerArns(ctx, primaryTargetGroup)
-	if err != nil {
-		in.LogPersister.Errorf("Failed to get current active listeners: %v", err)
-		return false
-	}
-
-	if err := client.ModifyListeners(ctx, currListenerArns, routingTrafficCfg); err != nil {
-		in.LogPersister.Errorf("Failed to routing traffic to PRIMARY/CANARY variants: %v", err)
-		return false
+	if len(listenerRules) == 0 {
```
khanhtc1202, line 430: @moko-poi let's make this easier to follow by removing the if/else condition here and return early (as pipecd code convention)
I think we can make it
```go
if len(listenerRules) != 0 {
   // Do modify listener rules
   return true
}
  // Do modify listener
  return true
```

moko-poi: @khanhtc1202 
Thank you for the feedback.

I agree that following the pipecd code convention and using early returns will make the code clearer.
I refactored the if/else condition as you suggested and updated the PR.
[29554b1](https://github.com/pipe-cd/pipecd/pull/4540/commits/29554b1725290064014bcafd28be2a5de344d770)

khanhtc1202:
@moko-poi Thanks so much for your contribution 🙏 There are 2 points that concern me in this current implementation.

First, not sure why we need to nest the configuration as `listernerRules.rules`, `listernerRules` directly contains rules make more sense to me.

```yaml
kind: ECSApp
input:
  taskDefinitionFile: taskdef.yaml
  serviceDefinitionFile: servicedef.yaml
  targetGroups:
    primary:
    canary:
  listenerRules:
      - rule1
      - rule2
```

Second, as far as I understand, the current implementation (ref: [code](https://github.com/pipe-cd/pipecd/pull/4540/files#diff-0ced6421324def8505cb6dadaa49f469d0855ef22437ba6d8ad7f5279a61b2e4R430)), piped can perform routing via
- Updating rule (which forwarding canary/primary target group)
- Updating listener (currently, clear all rules of the listener and only forwarding rule remained)

We currently don't have any way to validate the rules (set by users) is the rules which should be used to perform routing or not. That could lead to unexpected behavior if users miss setting the wrong rules to be used. Across the issue requirement, how about to keep finding current listeners of the given primary workload, then describe the listeners and find the forwarding rule to update, other rules will be remained and be included in the ModifyListener input?

We can use this API to describe specified listeners: https://docs.aws.amazon.com/cli/latest/reference/elbv2/describe-listeners.html

moko-poi:
@khanhtc1202

Thank you for the feedback! I truly appreciate your insights.

Regarding the nesting of listenerRules.rules:You're absolutely right; having listenerRules directly contain the rules seems more intuitive.

I have now made the necessary modifications based on your suggestions. Please take a look at the updated implementation and let me know if there are any further changes or adjustments needed.
24adec0

In addition to the previous changes, I've also addressed the validation of provided listener rules against the existing rules. I'd appreciate it if you could review this new implementation as well. Thank you!
394adbf

moko-poi:
@kentakozuka @khanhtc1202 
Thank you for the Public MTG the other day. 

Based on the feedback received, I've decided not to add a new feature but to upgrade the existing functionality. 

When you have a moment, I'd appreciate it if you could review the changes.

khanhtc1202:
@moko-poi Could you update the code to pass tests 👀

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -417,6 +417,30 @@ func (c *client) GetListenerArns(ctx context.Context, targetGroup types.LoadBala
 	return arns, nil
 }
 
+func (c *client) GetListenerRuleArns(ctx context.Context, listenerArns []string) ([]string, error) {
+	var ruleArns []string
+
+	// 各リスナーのルールを取得
```
khanhtc1202, line 423: Please use comment in English

moko-poi: @khanhtc1202
Thank you for the feedback.
I'll make sure to use English in the comments moving forward.
[d799ab0](https://github.com/pipe-cd/pipecd/pull/4540/commits/d799ab0af2bb3920493324bcd3e6cc4c48972933)

moko-poi:
@khanhtc1202 
I've made the necessary updates to the code. Please take a look and let me know if there are any further changes needed. Thanks for the feedback!

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -432,11 +456,45 @@ func (c *client) getLoadBalancerArn(ctx context.Context, targetGroupArn string)
...
+						ForwardConfig: &elbtypes.ForwardActionConfig{
+							TargetGroups: []elbtypes.TargetGroupTuple{
+								{
+									TargetGroupArn: aws.String(routingTrafficCfg[0].TargetGroupArn),
+									Weight:         aws.Int32(int32(routingTrafficCfg[0].Weight)),
+								},
+								{
+									TargetGroupArn: aws.String(routingTrafficCfg[1].TargetGroupArn),
+									Weight:         aws.Int32(int32(routingTrafficCfg[1].Weight)),
+								},
```
khanhtc1202, line 480: Will this always change a rule to `ForwardAction` type, which forwards traffic to primary/canary workload? if I recall correctly, we want to only modify the forward traffic actions, and keep "default forward 404" like action, isn't it?

moko-poi: Thank you for your review. 

The code is designed to modify only the rules with 'forward' action types; all other rules will remain unchanged.

[e91ab8a](https://github.com/pipe-cd/pipecd/pull/4540/commits/e91ab8ac0f357a502c269a74d64097faf38add8b)

moko-poi: @khanhtc1202

moko-poi:
@kentakozuka @khanhtc1202 
Closing this PR due to conflicts and delays. A new, updated PR has been created here: https://github.com/pipe-cd/pipecd/pull/4668. Please refer to the new PR for further discussion.

I apologize for any inconvenience and invite you to join the discussion on the new PR. Thank you for your understanding and collaboration.

PR #4551, opened 2023-08-18
Support vpc configuration to lambda

### LINKED ISSUE #4550: Support VPC Configuration for lambda

I use PipeCD to manage to many lambda applications. And a lot of lambda functions connect to a network. A lambda function which PipeCD manages doesn't support to vpc configuration. So I improve lambda function to support it. I added VPC configuration and I refactored [CreateFunction](https://docs.aws.amazon.com/ja_jp/lambda/latest/dg/API_CreateFunction.html) and [UpdateFunction](https://docs.aws.amazon.com/ja_jp/lambda/latest/dg/API_UpdateFunctionConfiguration.html). If vpc configuration is non nil, CreateFunction passes this configuration to client input and UpdateFunction calls UpdateConfiguration in additional. 

**Why is this needed**:

sivchari:

Fixes #4550 

Users can configure a vpc configuration for lambda function.

No

### DIFF
--- a/docs/content/en/docs-dev/user-guide/configuration-reference.md
+++ b/docs/content/en/docs-dev/user-guide/configuration-reference.md
@@ -9,7 +9,7 @@ description: >
 ## Kubernetes Application
 
 ``` yaml
-apiVersion: pipecd.dev/v1beta1
+apiVersion: pipecd.devI mean instead of adding it here, only make the vpc configuration available via Function manifest file. I think we should keep this configuration file contains generic configurations only./v1beta1
 kind: KubernetesApp
 spec:
   input:
@@ -121,7 +121,6 @@ spec:
 | timeout | duration | The maximum length of time to execute deployment before giving up. Default is 6h. | No |
 | notification | [DeploymentNotification](#deploymentnotification) | Additional configuration used while sending notification to external services. | No |
 | postSync | [PostSync](#postsync) | Additional configuration used as extra actions once the deployment is triggered. | No |
-| vpcConfig | [VPCConfig](#vpcconfig) | Additional configuration used to link to specific vpc. | No |
 | eventWatcher | [][EventWatcher](#eventwatcher) | List of configurations for event watcher. | No |
 
 ## ECS application
@@ -437,13 +436,6 @@ One of `yamlField` or `regex` is required.
 
 | Field | Type | Description | Required |
 |-|-|-|-|
-
-## VPCConfig
-
-| Field | Type | Description | Required |
-|-|-|-|-|
-| securityGroupIds | []string | List of security group ids. | Yes |
-| subnetIds | []string | List of subnet ids. | Yes |
 
 ## ECSDeploymentInput
 
--- a/docs/content/en/docs-dev/user-guide/configuration-reference.md
+++ b/docs/content/en/docs-dev/user-guide/configuration-reference.md
@@ -9,7 +9,7 @@ description: >
 ## Kubernetes Application
 
 ``` yaml
-apiVersion: pipecd.devI mean instead of adding it here, only make the vpc configuration available via Function manifest file. I think we should keep this configuration file contains generic configurations only./v1beta1
+apiVersion: pipecd.dev/v1beta1
 kind: KubernetesApp
 spec:
   input:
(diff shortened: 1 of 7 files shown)


khanhtc1202:
@sivchari How about adding that vpc configuration here instead of the application configuration? Since it quite specific for Lambda (or AWS service only)
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/lambda/function.go#L58-L73

sivchari:
@khanhtc1202 Thanks, oh, I've already added it, doesn't it ?

https://github.com/pipe-cd/pipecd/pull/4551/files#diff-7b8805cb45eb45990964cb95f8e8afdd1619eb58eeed7b190462d91374d015f6R73

docs/content/en/docs-dev/user-guide/configuration-reference.md
```diff
@@ -121,6 +121,7 @@ spec:
 | timeout | duration | The maximum length of time to execute deployment before giving up. Default is 6h. | No |
 | notification | [DeploymentNotification](#deploymentnotification) | Additional configuration used while sending notification to external services. | No |
 | postSync | [PostSync](#postsync) | Additional configuration used as extra actions once the deployment is triggered. | No |
+| vpcConfig | [VPCConfig](#vpcconfig) | Additional configuration used to link to specific vpc. | No |
```
khanhtc1202, line 124: @sivchari I mean instead of adding it here, only make the vpc configuration available via Function manifest file. I think we should keep this configuration file contains generic configurations only.

sivchari: @khanhtc1202 
Ah, sorry. I fixed it. Like this ?

docs/content/en/docs-dev/user-guide/configuration-reference.md
```diff
@@ -9,7 +9,7 @@ description: >
 ## Kubernetes Application
 
 ``` yaml
-apiVersion: pipecd.dev/v1beta1
+apiVersion: pipecd.devI mean instead of adding it here, only make the vpc configuration available via Function manifest file. I think we should keep this configuration file contains generic configurations only./v1beta1
```
khanhtc1202, line 12: what was this? 😆

pkg/app/piped/platformprovider/lambda/function_test.go
```diff
@@ -99,6 +99,40 @@ func TestParseFunctionManifest(t *testing.T) {
...
+  "spec": {
+	  "name": "SimpleFunction",
+	  "role": "arn:aws:iam::xxxxx:role/lambda-role",
+	  "memory": 128,
+	  "timeout": 5,
+	  "image": "ecr.region.amazonaws.com/lambda-simple-function:v0.0.1",
+      "vpcConfig": {
+          securityGroupIds: ["sg-1234567890", "sg-0987654321"],
+          subnetIds: ["subnet-1234567890", "subnet-0987654321"]
+      }
```
khanhtc1202, line 116: ident

sivchari: fix

docs/content/en/docs-dev/user-guide/configuration-reference.md
sivchari: fix

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -133,6 +133,12 @@ func (c *client) CreateFunction(ctx context.Context, fm FunctionManifest) error
 			ImageUri: aws.String(fm.Spec.ImageURI),
 		}
 	}
+	if fm.Spec.VPCConfig != nil {
+		input.VpcConfig = &types.VpcConfig{
+			SecurityGroupIds: fm.Spec.VPCConfig.SecurityGroupIDs,
+			SubnetIds:        fm.Spec.VPCConfig.SubnetIDs,
+		}
+	}
```
khanhtc1202, line 141: Could you move this block to L128 or after L152, I want to keep "packing using container image" and "Zip on S3" block near by since they mean to same thing (code of the function).

khanhtc1202 (commented):
Almost LGTM. Left a simple nits.

docs/content/en/docs-dev/user-guide/configuration-reference.md
khanhtc1202: 🐈

pkg/app/piped/platformprovider/lambda/client.go
sivchari: You are right. I fix.

khanhtc1202 (dismissed):
Great, thank you 🙌

khanhtc1202:
@sivchari one testcase is failed, could you check?

sivchari:
@khanhtc1202 
All test is passed.

khanhtc1202 (approved):
LGreatTM 👍

PR #4553, opened 2023-08-20
fix  the link of terraform applications autorollback in docs

nnnkkk7:
 The link of terraform applications autorollback is wrong.


khanhtc1202 (approved):
Nice catch, thank you 🙌

PR #4554, opened 2023-08-21
Improve notification

### LINKED ISSUE #4547: Make it possible to webhook notification for Rolling Back event

After a deployment failed and is rolling back, a functionality is needed to notify the event status related to `DEPLOYMENT_ROLLING_BACK`.

**Why is this needed**:

By adding events for PipeCD [Configuring notifications](https://pipecd.dev/docs-dev/user-guide/managing-piped/configuring-notifications/) to the `spec.notifications.routes.events[]` field in piped config, any service can receive deployment events.

For example, configure it as follows.

```yaml
spec:
  notifications:
    routes:
      - name: sample
        events:
          - DEPLOYMENT_ROLLING_BACK
          - DEPLOYMENT_SUCCEEDED
          - DEPLOYMENT_FAILED
          - DEPLOYMENT_CANCELLED
        receiver: sample-receiver
    receivers:
      - name: sample-receiver
        ...
        ...
```

However, in the current implementation, even if you set `DEPLOYMENT_ROLLING_BACK`, any services cannot receive the rolling back event.
This is because the current event handler is only checked by `DEPLOYMENT_SUCCEEDED`, `DEPLOYMENT_FAILED`, and `DEPLOYMENT_CANCELLED`.

FYI:
- https://github.com/pipe-cd/pipecd/blob/master/pkg/model/notificationevent.pb.go#L44-L47
- https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/controller/scheduler.go#L614-L655

sivchari:

I improve reportDeploymentCompleted.

Fixes #4547 


pkg/app/piped/controller/scheduler.go
```diff
@@ -612,6 +612,20 @@ func (s *scheduler) reportDeploymentCompleted(ctx context.Context, status model.
 
 	defer func() {
 		switch status {
+
+		case model.DeploymentStatus_DEPLOYMENT_ROLLING_BACK:
```
kentakozuka, line 616: `status` cannot be `DeploymentStatus_DEPLOYMENT_ROLLING_BACK` in the current implementation. See [here](https://github.com/pipe-cd/pipecd/blob/46fd914f2852da0a3d9d5c7ee711402207b3c621/pkg/app/piped/controller/scheduler.go#L404)

kentakozuka (commented):
@sivchari 
Thanks for creating PR!
It seems that this PR contains two issues;
- Add notification of rollback
- Add mentions on notifications

It would be better to split it into two different PRs 👍 
Also, about https://github.com/pipe-cd/pipecd/issues/4547, please join the conversation on the issue. There are things that we need to decide.

pkg/app/piped/controller/scheduler.go
GotoRen: I wrote my opinion [here](https://github.com/pipe-cd/pipecd/issues/4547#issuecomment-1694820677).
Thank you @sivchari !

PR #4560, opened 2023-08-24
fix: update function configuration

sivchari:

@khanhtc1202 
I'm so sorry, update function configuration has been already defined and called. So I deleted my changes.
And I fixed order to call UpdateFunctionCode and UpdateFunctionConfiguration. Because UpdateFunctionConfiguration is failed, if Lambda state is pending. 

Fixes #

### DIFF
--- a/pkg/app/piped/platformprovider/lambda/client.go
+++ b/pkg/app/piped/platformprovider/lambda/client.go
@@ -188,6 +188,14 @@ func (c *client) CreateFunctionFromSource(ctx context.Context, fm FunctionManife
 }
 
 func (c *client) UpdateFunction(ctx context.Context, fm FunctionManifest) error {
+	// UpdateFunctionConfiguration must be called before UpdateFunctionCode.
+	// Lambda has named by state.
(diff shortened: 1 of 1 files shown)


pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -287,6 +272,10 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 			Environment: &types.Environment{
 				Variables: fm.Spec.Environments,
 			},
+			VpcConfig: &types.VpcConfig{
+				SecurityGroupIds: fm.Spec.VPCConfig.SecurityGroupIDs,
+				SubnetIds:        fm.Spec.VPCConfig.SubnetIDs,
+			},
```
khanhtc1202, line 278: Should check nullity for `fm.Spec.VPCConfig` before call to its fields

sivchari: I fixed it.

khanhtc1202 (approved):
Great catch, thank you 👏

PR #4565, opened 2023-08-24
update: support to wait to update function

sivchari:

Last patch. I support to wait to update lambda function.
If lambda function state is pending, UpdateFunctionCode is failed. So I check the lambda satate.

Fixes #

### DIFF
--- a/pkg/app/piped/platformprovider/lambda/client.go
+++ b/pkg/app/piped/platformprovider/lambda/client.go
@@ -227,6 +227,11 @@ func (c *client) UpdateFunction(ctx context.Context, fm FunctionManifest) error 
 }
 
 func (c *client) UpdateFunctionFromSource(ctx context.Context, fm FunctionManifest, zip io.Reader) error {
+	// Update function configuration.
+	if err := c.updateFunctionConfiguration(ctx, fm); err != nil {
+		return err
+	}
+
 	data, err := io.ReadAll(zip)
 	if err != nil {
 		return err
@@ -247,11 +252,6 @@ func (c *client) UpdateFunctionFromSource(ctx context.Context, fm FunctionManife
 	_, err = c.client.UpdateFunctionCode(ctx, codeInput)
 	if err != nil {
 		return fmt.Errorf("failed to update function code for Lambda function %s: %w", fm.Spec.Name, err)
-	}
-
-	// Update function configuration.
-	if err = c.updateFunctionConfiguration(ctx, fm); err != nil {
-		return err
 	}
 
 	// Tag/Untag function if necessary.
@@ -295,32 +295,24 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 	if !updateFunctionConfigurationSucceed {
 		return fmt.Errorf("failed to update configuration for Lambda function %s: %w", fm.Spec.Name, err)
 	}
-	if err := c.waitFunctionUpdated(ctx, fm.Spec.Name); err != nil {
-		return fmt.Errorf("failed to wait for Lambda function %s updated: %w", fm.Spec.Name, err)
-	}
-	return nil
-}
 
-func (c *client) waitFunctionUpdated(ctx context.Context, functionName string) error {
-	retry := backoff.NewRetry(RequestRetryTime, backoff.NewConstant(RetryIntervalDuration))
+	// Wait until function updated successfully.
+	retry = backoff.NewRetry(RequestRetryTime, backoff.NewConstant(RetryIntervalDuration))
 	input := &lambda.GetFunctionInput{
-		FunctionName: aws.String(functionName),
+		FunctionName: aws.String(fm.Spec.Name),
 	}
-	_, err := retry.Do(ctx, func() (any, error) {
+	_, err = retry.Do(ctx, func() (any, error) {
 		output, err := c.client.GetFunction(ctx, input)
 		if err != nil {
 			return nil, err
 		}
 		if output.Configuration.LastUpdateStatus != types.LastUpdateStatusSuccessful {
 			return nil, fmt.Errorf("failed to update Lambda function %s, status code %v, error reason %s",
-				functionName, output.Configuration.LastUpdateStatus, *output.Configuration.LastUpdateStatusReason)
+				fm.Spec.Name, output.Configuration.LastUpdateStatus, *output.Configuration.LastUpdateStatusReason)
 		}
 		return nil, nil
 	})
-	if err != nil {
-		return err
-	}
-	return nil
+	return err
 }
 
 func (c *client) PublishFunction(ctx context.Context, fm FunctionManifest) (string, error) {
--- a/pkg/app/piped/platformprovider/lambda/client.go
+++ b/pkg/app/piped/platformprovider/lambda/client.go
@@ -295,6 +295,31 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 	if !updateFunctionConfigurationSucceed {
 		return fmt.Errorf("failed to update configuration for Lambda function %s: %w", fm.Spec.Name, err)
 	}
+	if err := c.waitFunctionUpdated(ctx, fm.Spec.Name); err != nil {
+		return fmt.Errorf("failed to wait for Lambda function %s updated: %w", fm.Spec.Name, err)
+	}
+	return nil
+}
+
+func (c *client) waitFunctionUpdated(ctx context.Context, functionName string) error {
+	retry := backoff.NewRetry(RequestRetryTime, backoff.NewConstant(RetryIntervalDuration))
+	input := &lambda.GetFunctionInput{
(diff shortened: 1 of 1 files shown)


pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -27,9 +27,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/service/lambda"
 	"github.com/aws/aws-sdk-go-v2/service/lambda/types"
-	"go.uber.org/zap"
-
+	helper "github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
```
khanhtc1202, line 30: Let's not use hashicorp retry but pipecd package 👀 
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/backoff/backoff.go

sivchari: I use retry package to get lambda function state.

khanhtc1202:
@sivchari Could you remove grpc update and go packages update from this PR 👀

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -295,6 +295,30 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
...
+			return nil, err
+		}
+		if output.Configuration.LastUpdateStatus != types.LastUpdateStatusSuccessful {
+			return nil, fmt.Errorf("failed to update Lambda function %s, status code %v, error reason %s",
+				functionName, output.Configuration.LastUpdateStatus, *output.Configuration.LastUpdateStatusReason)
+		}
+		return nil, nil
+	}); err != nil {
+		return err
+	}
```
khanhtc1202, line 321:
```suggestion
	input := &lambda.GetFunctionInput{
		FunctionName: aws.String(functionName),
	}
	
	retry := backoff.NewRetry(RequestRetryTime, backoff.NewConstant(RetryIntervalDuration))
	_, err := retry.Do(ctx, func() (interface{}, error) {
		output, err := c.client.GetFunction(ctx, input)
		if err != nil {
			return nil, err
		}
		if output.Configuration.LastUpdateStatus != types.LastUpdateStatusSuccessful {
			return nil, fmt.Errorf("failed to update Lambda function %s, status code %v, error reason %s",
				functionName, output.Configuration.LastUpdateStatus, *output.Configuration.LastUpdateStatusReason)
		}
		return nil, nil
	})
	
	return err
	}
```
How about this?

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -295,6 +295,30 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 	if !updateFunctionConfigurationSucceed {
 		return fmt.Errorf("failed to update configuration for Lambda function %s: %w", fm.Spec.Name, err)
 	}
+	if err := c.waitFunctionUpdated(ctx, fm.Spec.Name, 10*time.Second); err != nil {
+		return fmt.Errorf("failed to wait for Lambda function %s updated: %w", fm.Spec.Name, err)
+	}
+	return nil
+}
+
+func (c *client) waitFunctionUpdated(ctx context.Context, functionName string, timeout time.Duration) error {
```
khanhtc1202, line 304: this `timeout` is not being used

khanhtc1202:
btw, @sivchari  due to the comment [here](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/lambda/client.go#L191-L195)
Should we rearrange this function as well (to be called before the UpdateFunctionCode)
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/lambda/client.go#L252-L255

sivchari:
@khanhtc1202 
This answer is very strong Yes ! I rearranged it.

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -232,6 +232,11 @@ func (c *client) UpdateFunctionFromSource(ctx context.Context, fm FunctionManife
 		return err
 	}
 
+	// Update function configuration.
+	if err := c.updateFunctionConfiguration(ctx, fm); err != nil {
+		return err
+	}
+
```
khanhtc1202, line 239: Probably should be at the top of this function (before L230 reading zip) to avoid unnecessary disk read if the update function configuration return error.

sivchari: You're right. Thnaks.

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -295,6 +295,31 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 	if !updateFunctionConfigurationSucceed {
 		return fmt.Errorf("failed to update configuration for Lambda function %s: %w", fm.Spec.Name, err)
 	}
+	if err := c.waitFunctionUpdated(ctx, fm.Spec.Name); err != nil {
+		return fmt.Errorf("failed to wait for Lambda function %s updated: %w", fm.Spec.Name, err)
+	}
+	return nil
+}
+
+func (c *client) waitFunctionUpdated(ctx context.Context, functionName string) error {
```
khanhtc1202, line 304: How about moving this function body to L298? Looks like this function is specified to be used only by `updateFunctionConfiguration`, so I don't think we have enough reason to separate it as an independent function 🤔 Also, the current function name is quite common, `waitFunctionUpdated` while we have many `UpdateFunctionXXX`, which could cause misleading. So removing this function and moving all of its body to be part of `updateFunctionConfiguration` could be a better choice if you only want to wait after updating function configuration. wdyt?

sivchari: Okay.

> Also, the current function name is quite common, waitFunctionUpdated while we have many UpdateFunctionXXX, which could cause misleading.

I thinks so too.

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -295,6 +295,24 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 	if !updateFunctionConfigurationSucceed {
 		return fmt.Errorf("failed to update configuration for Lambda function %s: %w", fm.Spec.Name, err)
 	}
+
+	input := &lambda.GetFunctionInput{
```
khanhtc1202, line 299:
```suggestion
	// Wait until function updated successfully.
	input := &lambda.GetFunctionInput{
```

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -295,6 +295,24 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
...
+		if output.Configuration.LastUpdateStatus != types.LastUpdateStatusSuccessful {
+			return nil, fmt.Errorf("failed to update Lambda function %s, status code %v, error reason %s",
+				fm.Spec.Name, output.Configuration.LastUpdateStatus, *output.Configuration.LastUpdateStatusReason)
+		}
+		return nil, nil
+	})
+	if err != nil {
+		return err
+	}
 	return nil
```
khanhtc1202, line 316:
```go
	return err
```
nits

khanhtc1202 (commented):
Look neat, just left nits comments 😀

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -295,7 +295,23 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 	if !updateFunctionConfigurationSucceed {
 		return fmt.Errorf("failed to update configuration for Lambda function %s: %w", fm.Spec.Name, err)
 	}
-	return nil
+
+	// Wait until function updated successfully.
+	input := &lambda.GetFunctionInput{
+		FunctionName: aws.String(fm.Spec.Name),
+	}
+	_, err = retry.Do(ctx, func() (any, error) {
```
khanhtc1202, line 303: Sorry, I just noticed that we reused the previous retry object here. I think we should recreate a new one instead since the same object may contain previously updated backoff internal, thus the number of retry times is not as expected.

```suggestion
	retry = backoff.NewRetry(RequestRetryTime, backoff.NewConstant(RetryIntervalDuration))
	_, err = retry.Do(ctx, func() (any, error) {
```

khanhtc1202:
@sivchari After this got merge, I will release v0.45.3-rc1 then you can test it before we make release v0.45.3 😄

pkg/app/piped/platformprovider/lambda/client.go
sivchari: @khanhtc1202 
I fixed it.

khanhtc1202 (approved):
LGTM, thanks 🍏

PR #4568, opened 2023-08-25
fix: ECS WaitService finish before task running

tokku5552:
Fixed an issue where ECS tasks were marked as successful before they were all started

Follow PR #4548

### DIFF
--- a/pkg/app/piped/platformprovider/ecs/client.go
+++ b/pkg/app/piped/platformprovider/ecs/client.go
@@ -252,7 +252,8 @@ func (c *client) WaitServiceStable(ctx context.Context, service types.Service) e
 		Cluster:  service.ClusterArn,
 		Services: []string{*service.ServiceArn},
 	}
-
+	// TODO: Wait until a new task is started instead of sleeping.
+	time.Sleep(30 * time.Second)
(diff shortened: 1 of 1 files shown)


khanhtc1202:
@tokku5552 Thanks for contribution, please use `git commit -s` to sign-off the commit 😄 
You can find how to do it here: https://github.com/pipe-cd/pipecd/blob/master/CONTRIBUTING.md#contributor-license-agreement

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -252,7 +252,8 @@ func (c *client) WaitServiceStable(ctx context.Context, service types.Service) e
 		Cluster:  service.ClusterArn,
 		Services: []string{*service.ServiceArn},
 	}
-
+	// TODO: Wait until a new task is started instead of sleeping.
```
khanhtc1202, line 255:
```suggestion
	// Wait before first checking the service state due to the logic checking service
	// stable currently is based on `pendingCount`, which could always be `0` when
	// the service deployment has started running.
	// TODO: Wait until a new task is started instead of sleeping.
```
nits, add a note to clarify why we have to use sleep here.

khanhtc1202 (approved):
Great catch, thank you ❤️

PR #4580, opened 2023-08-30
Fix ECS not remove taskset on quick sync rollback

khanhtc1202:

In this PR I did
- Update rollback traffic routing condition, we only need to revert all traffic to primary target group in case of primary & canary target group config are available (ie. progressive sync)
- Instead of using `clean` function in rollback implementation to remove unused taskset, which can only remove CANARY taskset, for general use of rollback function, I use new defined `GetServiceTaskSets` to get all PRIMARY/ACTIVE tasksets of the current service before rolling back and remove them unused taskset later.
- Other changes are rearrange function code blocks to make the flow more like ECS deployment.

Fixes #

### DIFF
--- a/pkg/app/piped/executor/ecs/ecs.go
+++ b/pkg/app/piped/executor/ecs/ecs.go
@@ -33,9 +33,9 @@ import (
 )
 
 const (
+	activeServiceKeyName = "active-service-object"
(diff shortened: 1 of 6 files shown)


nghialv (approved):
🐬

PR #4583, opened 2023-09-01
add: dynamic mention

### LINKED ISSUE #4417: Dynamic mentions to slack
I think it is possible to send Slack notifications of Deployment WAIT APPROVALs with the current functionality. It may also be possible to add pre-defined mentions.
I would like to be able to dynamically add mentions to a specific person who triggered the Deployment.

For example, if I have set up slack id's for two people, A and B, when person A triggers a deployment, I would like to be able to send mentions only to person A, and not to person B.

**Why is this needed**:
Our team has a trunk-based development policy.
Since multiple releases are made per day, we have introduced a WAIT APPROVAL stage in the pipeline as a final check to see if a release is ready to be rolled out.
However, since it takes some time to reach the WAIT APPROVAL stage, we feel that we can use our time more effectively if we can send notifications only to those who triggered the release when the release reaches the WAIT APPROVAL stage.

seitarof:
Dynamically add mentions based on the GitHub ID of the specific person who triggered the deployment.

Fixes #4417 

Notify dynamic mention by setting as follows.
```
kind: TerraformApp
spec:
  pipeline:
    stages:
      - name: WAIT_APPROVAL
  notification:
    mentions:
      - event: DEPLOYMENT_WAIT_APPROVAL
        dynamics:
          - git: {github-id}
            slack: {slack-uid}
```

### DIFF
--- a/docs/content/en/docs-dev/user-guide/configuration-reference.md
+++ b/docs/content/en/docs-dev/user-guide/configuration-reference.md
@@ -291,6 +291,13 @@ One of `yamlField` or `regex` is required.
 |-|-|-|-|
 | event | string | The event to be notified to users. | Yes |
 | slack | []string | List of user IDs for mentioning in Slack. See [here](https://api.slack.com/reference/surfaces/formatting#mentioning-users) for more information on how to check them. | No |
+| dynamics | [] [DynamicMention](#dynamicmention) | List of dynamic mentions. | No |
+
+#### DynamicMention
+| Field | Type | Description | Required |
+|-|-|-|-|
+| git | string | GitHub ID. | No |
+| slack | string | User ID for mentioning in Slack. See [here](https://api.slack.com/reference/surfaces/formatting#mentioning-users) for more information on how to check them. | No |
 
 ## KubernetesDeploymentInput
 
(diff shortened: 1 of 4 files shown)


kentakozuka:
@seitarof
Thank you for your contribution ❤️ 
The idea of this PR is great!
I want to address some concerns about this feature.

Firstly, the name "dynamic" is a bit too confusing for those who see this configuration at the first time. "Dynamic" does not always mean mentioning the account that triggers the deployment. There may be a different way to set configurations to achieve this.

Secondly, [DeploymentNotification](https://pipecd.dev/docs-v0.45.x/user-guide/configuration-reference/#deploymentnotification) is not only for wait-approval but also other deployment stages. So, if we add configurations to DeploymentNotification, it must be used for other stages. Or this feature should be only for wait-approval 🤔 

wdyt?

kentakozuka:
I just thinking if there is a better way to implement this feature because of some reasons;
- Users have to input slack IDs into two places, `NotificationMention.slack` and `NotificationMention.dynamics`.
- Adding both git and slack IDs is a bit bothersome for most of users.

But, I do not have any better ideas yet 😢 
@khanhtc1202 wdyt?

kentakozuka:
@seitarof 
Could you write the design of implementation on the issue?
I think we need more to discuss 😄

khanhtc1202:
> I think it would be better to use this feature not only in the wait-approval stage, but also in other deployment stages.
I'll fix it to reflect other deployment stages as well.

@seitarof That's a good point, and I think we should clearly define what should be done before we start. The mentioned problem in the issue is about `Only notify users who trigger the deployment which contains WAIT_APPROVAL stage`, so if it's just like that, then it's better only to make that as part of WAIT_APPROVAL stage configuration. But, I really like your point, can we discuss that before reviewing this PR? As mentioned above, we should clearly understand what will be done before making changes to master 😄

Issue #4506, opened 2023-07-16
PipeCD Dockerfile for local building PipeCD docker images


jz543fm:
**What would you like to be added**:

I would like that will be added to PipeCD how to locally build PipeCD for custom usage, f.e,  if I want to change in docker image for my proper use case, it will be really helpful, because I tried to run in repository `local/` when I start it, check the release with `cat /etc/*release*`, builded from `local/` the OS is `Ubuntu 16.04`, but in pod f.e, pipecd-server, it's running on `Alpine Linux 3.13.2`, I do not know how to create custom Helm Chart or Docker image from the [Pipecd Chart Releases](https://github.com/pipe-cd/pipecd/pkgs/container/chart%2Fpipecd) and the doc will be also really helpful or some template

**Why is this needed**:

It will be helpful for devs, if they want to add some additional features to custom PipeCD Helm Chart / Docker image

kentakozuka:
It is necessary to fix [here](https://github.com/pipe-cd/pipecd/blob/331b3239b159071c49fdd8b96ecee608b3454fe8/manifests/pipecd/templates/deployment.yaml#L125) to change the PipeCD server container image.

PRs are always welcome :)
It would be great if you create a PR for that!

jz543fm:
Yea, I've fixed that later that image, thank's, ok.

hungran:
yeah I've same issue when trying to make image arm64, not only `pipecd` but also `piped` here
https://github.com/pipe-cd/pipecd/blob/master/manifests/piped/templates/deployment.yaml#L31-L36
what I did for local build:

- `make build`
- some babysitting command to build the image and push image into my repository with a tag
- change the image repository name and `Chart.AppVersion` to my tag

I would like we can have option to use image tag from values. wdyt?

PR #4586, opened 2023-09-03
Enable to specific image tag for container images in Helm template

hungran:
I would like we have a way to specific an image tag whenever I or somebody else want to test with new OCI image from local environment or difference registry

Fixes https://github.com/pipe-cd/pipecd/issues/4506

### DIFF
--- a/manifests/pipecd/values.yaml
+++ b/manifests/pipecd/values.yaml
@@ -30,8 +30,8 @@ gateway:
 server:
   image:
     repository: ghcr.io/pipe-cd/pipecd
-  # Overrides the image tag whose default is the chart appVersion.
-  tag: ""
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: ""
   replicasCount: 1
   args:
     cacheAddress: ""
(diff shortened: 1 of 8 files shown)


manifests/pipecd/values.yaml
```diff
@@ -30,6 +30,8 @@ gateway:
 server:
   image:
     repository: ghcr.io/pipe-cd/pipecd
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
```
khanhtc1202, line 34: As mentioned in template file, this config should be under `image` block (same as `repository`)

manifests/pipecd/values.yaml
```diff
@@ -51,6 +53,8 @@ cache:
 ops:
   image:
     repository: ghcr.io/pipe-cd/pipecd
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
```
khanhtc1202, line 57: Same here

khanhtc1202 (commented):
Great improvement, left some nits 🙏

hungran:
@khanhtc1202  thanks for your feedback, solved

khanhtc1202 (approved):
Great, thank you 🙌

Issue #4589, opened 2023-09-04
Display user defined error that occurred in MySQL


kentakozuka:
**What would you like to be added**:

**Why is this needed**:

kentakozuka:
Fixed by #4590 and #4597

PR #4590, opened 2023-09-04
Display MySQL user defined error in API Key UI

kentakozuka:

Screenshot 2023-09-04 at 12 22 49

Fixes partof https://github.com/pipe-cd/pipecd/issues/4589

### DIFF
--- a/pkg/app/server/grpcapi/grpcapi.go
+++ b/pkg/app/server/grpcapi/grpcapi.go
@@ -166,18 +166,22 @@ func getEncriptionKey(se *model.Piped_SecretEncryption) ([]byte, error) {
 }
 
 func gRPCStoreError(err error, msg string) error {
-	switch err {
-	case nil:
+	if err == nil {
 		return nil
-	case datastore.ErrNotFound, filestore.ErrNotFound, stagelogstore.ErrNotFound:
-		return status.Error(codes.NotFound, fmt.Sprintf("Entity was not found to %s", msg))
-	case datastore.ErrInvalidArgument:
-		return status.Error(codes.InvalidArgument, fmt.Sprintf("Invalid argument to %s", msg))
-	case datastore.ErrAlreadyExists:
-		return status.Error(codes.AlreadyExists, fmt.Sprintf("Entity already exists to %s", msg))
-	default:
-		return status.Error(codes.Internal, fmt.Sprintf("Failed to %s", msg))
 	}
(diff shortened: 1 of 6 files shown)


web/src/components/settings-page/api-key/index.tsx
```diff
@@ -96,10 +96,16 @@ export const APIKeyPage: FC = memo(function APIKeyPage() {
 
   const handleSubmit = useCallback(
     (values: { name: string; role: APIKey.Role }) => {
-      dispatch(generateAPIKey(values)).then(() => {
-        dispatch(fetchAPIKeys({ enabled: true }));
-        dispatch(addToast({ message: GENERATE_API_KEY_SUCCESS }));
-      });
+      dispatch(generateAPIKey(values))
+        .then(unwrapResult)
```
kentakozuka, line 100: Fixes an existing bug that success toast is displayed even if a request fails.

pkg/app/server/grpcapi/grpcapi.go
```diff
@@ -166,18 +166,22 @@ func getEncriptionKey(se *model.Piped_SecretEncryption) ([]byte, error) {
...
+	if errors.Is(err, datastore.ErrInvalidArgument) {
 		return status.Error(codes.InvalidArgument, fmt.Sprintf("Invalid argument to %s", msg))
-	case datastore.ErrAlreadyExists:
+	}
+	if errors.Is(err, datastore.ErrAlreadyExists) {
 		return status.Error(codes.AlreadyExists, fmt.Sprintf("Entity already exists to %s", msg))
-	default:
-		return status.Error(codes.Internal, fmt.Sprintf("Failed to %s", msg))
 	}
+	if errors.Is(err, datastore.ErrUserDefined) {
```
khanhtc1202, line 181: Why do we have to change from switch case to multiple if 🤔

kentakozuka: If we generate an error with a message from MySQL, handling errors with a switch cannot work properly. That's why I had to change it to multi-if blocks.

khanhtc1202: Which means the switch block can not check the type of datastore.ErrUserDefined error 🤔 It's a bit weird to me. Btw, how about adding test for this function. Looks like this function is testable without mocking anything.

kentakozuka: When an error message from MySQL is added to datastore.ErrUserDefined, checking with the switch block does not work because the new error is no longer the same as datastore.ErrUserDefined.

I agree with you to add tests for it 👍

khanhtc1202 (approved):
Great 🚀

Issue #4584, opened 2023-09-01
Including more precise deployment timestamps in webhook


GotoRen:
**What would you like to be added**:

Including in PipeCD webhook the timestamp when an existing resource was changed by piped.

Specifically, PipeCD gets the deployment status by [reportDeploymentCompleted](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/controller/scheduler.go#L597-L667). At this time, a timestamp ([CompletedAt](CompletedAt)) of when the status was obtained is included in PipeCD webhook.

We can know precisely the timing of the deployment by obtaining the timestamp when receiving the apply result of the manifests.

**Why is this needed**:

PipeCD includes the deployment start time in `CreatedAt`, overwrites the status (triggered:0 -> planned:1 -> success:4 / failed:5), updates `UpdatedAt`, and includes it in PipeCD webhook.
However, the `UpdatedAt` is not an actual deploy completed timestamp because it includes the time when   PipeCD C-Plane updated the database status.

I confirmed a time difference of about 20-40s between the timestamp notified by PipeCD webhook and the timestamp displayed in the WebUI logs.

e.g.

```json
{
  "Type": 4,
  "Metadata": {
    "deployment":
      ...
      ...
      "status_reason": "The deployment has been planned",
      "created_at": 1691115188,// 2023-08-04 11:13:08 // deploy started
      "updated_at": 1691115209 // 2023-08-04 11:13:29 // received a success status
    }
  }
}
```

![webui-log](https://github.com/pipe-cd/pipecd/assets/63791288/f7b642f9-5e96-4c78-a1cf-725baa586e45)

- `deployment_created_at`: 2023-08-04 11:13:08
- `deployment.updated_at`:  2023-08-04 11:13:29 <- 20s ~ 40s faster than successfully log

The actual time of successful applied the manifest was 11:13:54.
It should be after 11:13:54 that the success status can be included in PipeCD webhook.

khanhtc1202:
@GotoRen I'm Not sure if I missed something, but I think there is a miss here. Across the issue description, the `updated_at` timestamp is for status reason `"status_reason": "The deployment has been planned"`, while the stage log highlighted in the picture shows a message that it's `manifests applied`.

I think those timestamps are different because they're totally different events at first, aren't they?

GotoRen:
@khanhtc1202 
PipeCD Webhook seems to include the message "The deployment has been planned" in the `status_reason`,
[...]

PR #4591, opened 2023-09-04
Improve: add deployment timestamp in webhook

GotoRen:

Including in PipeCD webhook the timestamp when an existing resource was changed by piped.

---
PipeCD gets the deployment status by [reportDeploymentCompleted](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/controller/scheduler.go#L597-L667). At this time, a timestamp ([CompletedAt](CompletedAt)) of when the status was obtained is included in PipeCD webhook.

We can know precisely the timing of the deployment by obtaining the timestamp when receiving the apply result of the manifests.

Fixes https://github.com/pipe-cd/pipecd/issues/4584

### DIFF
--- a/pkg/app/piped/trigger/deployment.go
+++ b/pkg/app/piped/trigger/deployment.go
@@ -99,6 +99,7 @@ func buildDeployment(
 		Status:                    model.DeploymentStatus_DEPLOYMENT_PENDING,
 		StatusReason:              "The deployment is waiting to be planned",
 		Metadata:                  metadata,
+		CompletedAt:               app.MostRecentlySuccessfulDeployment.CompletedAt,
 		CreatedAt:                 now.Unix(),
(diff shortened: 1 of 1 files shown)


kentakozuka:
@GotoRen Sorry for being so late to review!
It is not clear to me why this change solves the issue. In order to know the exact time of deployment completion, should we change [here](https://github.com/pipe-cd/pipecd/blob/743bd9ec63df148d2c1d8df5bc217820150faa59/pkg/app/piped/controller/scheduler.go#L597) instead?
When `s.notifier.Notify` is called, CompletedAt is not updated as [this value](https://github.com/pipe-cd/pipecd/blob/743bd9ec63df148d2c1d8df5bc217820150faa59/pkg/app/piped/controller/scheduler.go#L608).

PR #4600, opened 2023-09-11
[ECS] Fix remove all previous active tasksets on QuickSync

khanhtc1202:

In case there are ACTIVE tasksets that are missed and still exist under service before running QuickSync on ECS application, this PR ensures to remove all previous Primary/Active tasksets, so that after QuickSync is applied, the service only contains the latest taskset as its primary.

Fixes #


kentakozuka (approved):
Fabulous!

nghialv (approved):

PR #4603, opened 2023-10-05
Add rfc for script run stage

ffjlabo:

Part of https://github.com/pipe-cd/pipecd/issues/4643

### DIFF
--- a/docs/rfcs/0011-script-run-stage.md
+++ b/docs/rfcs/0011-script-run-stage.md
@@ -24,7 +24,7 @@ So more simply, some users want to execute commands.
 
 1. execute any commands in their pipeline.
 
-```
+```yaml
 apiVersion: pipecd.dev/v1beta1
 kind: LambdaApp
 spec:
@@ -33,12 +33,13 @@ spec:
   pipeline:
     stages:
       - name: SCRIPT_RUN
-        env:
-          AWS_PROFILE: default
-        runs:
-          - "echo {{ .encryptedSecrets.password }} | sudo -S su"
-          - "sam build"
-          - "sam deploy -g --profile $AWS_PROFILE"
+        with:
+            env:
+              AWS_PROFILE: default
+            runs:
+              - "echo {{ .encryptedSecrets.password }} | sudo -S su"
+              - "sam build"
+              - "sam deploy -g --profile $AWS_PROFILE"
 ```
 
 2. combine with other stage
@@ -46,7 +47,7 @@ spec:
 Compared to CUSTOM_SYNC, this stage can be combined with other stage.
 For example, 
 
-```
+```yaml
 apiVersion: pipecd.dev/v1beta1
 kind: KubernetesApp
 spec:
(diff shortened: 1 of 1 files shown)


kentakozuka:
@ffjlabo Will the custom sync be removed after this feature is released?
It seems like this script run stage has all the features that the custom sync has.

ffjlabo:
@kentakozuka Yes, I think it will be deprecated and removed in the future.
Custom sync is a great way to execute any command. I think it is more useful to execute any command with other stages. But now we can't use custom sync with other stages such as K8S_SYNC because it is a kind of application.
So I proposed the script run stage to execute any command with other stages.

kentakozuka:
@ffjlabo 👌 
> But now we can't use custom sync with other stages such as K8S_SYNC because it is a kind of application.

What do you mean by this? Maybe I miss something 🤔

ffjlabo:
@kentakozuka 
> But now we can't use custom sync with other stages such as K8S_SYNC because it is a kind of application.

Oh, I'm sorry. I don't understand about custom sync correctly. custom sync can be used at any other stage. 🙏 
https://pipecd.dev/docs-v0.45.x/user-guide/managing-application/customizing-deployment/custom-sync/

But this stage is for **sync**.  In custom sync, the defined command will execute the same command when rollback.

By the script run stage, I want to realize the demand to just execute commands before or after other stages.
If we want to execute when rollback, we can define any command.

kentakozuka:
@ffjlabo Thanks! It is clear to me now.
Can you write that the custom sync would be deprecated when the script run stage is implemented?

ffjlabo:
@kentakozuka Thank you for your advise! I added it 🙏  [5f74cfd](https://github.com/pipe-cd/pipecd/pull/4603/commits/5f74cfdf6816fdf3051b43775e3c15975d55f225)

docs/rfcs/0011-script-run-stage.md
```diff
@@ -0,0 +1,208 @@
...
+
+**cons**
+- users need to prepare api for the webhooks.
+
+
+# Unresolved questions
+
+- It might be better to change from alpine to debian and so on to provide users popular commands (e.g. curl) by default.
+- It might be better to add `runsOnRollback` on each sync stage if users want to control when rollback.
+This is a just idea.
```
khanhtc1202, line 208: I don't get this part, do you mean in case of multiple script_run stages are combined, we should add `runsOnRollback` to each? And is this `runsOnRollback` is same as `onRollback` (the one used in the example above)?

docs/rfcs/0011-script-run-stage.md
```diff
@@ -0,0 +1,208 @@
...
+      - name: SCRIPT_RUN
+        with:
+          env:
+            SLACK_WEBHOOK_URL: ""
+          runs:
+            - "curl -X POST -H 'Content-type: application/json' --data '{"text":"successfully deployed!!"}' $SLACK_WEBHOOK_URL"
+          onRollback:
+            - "curl -X POST -H 'Content-type: application/json' --data '{"text":"failed to deploy: rollback"}' $SLACK_WEBHOOK_URL"
+```
+
```
khanhtc1202, line 121: Some other questions I think we should define before starting:
- Do we support multi script_run stages combine
- How the user-defined `onRollback` be run in case of multi script_run stages combine
- How the user-defined `onRollback` be run in combination with the `XXX_ROLLBACK` stage? I mean, if k8s app is rollbacked and the pipeline contains script_run, how do we perform the rollback process? Should it be `XXX_ROLLBACK` first then `SCRIPT_RUN_ROLLBACK`?

docs/rfcs/0011-script-run-stage.md
ffjlabo: @khanhtc1202 
This means It might be a way to add `onRollback` to each stage, not to create a new stage.

e.g. K8S_CANARY_ROLLOUT
```
  pipeline:
    stages:
      - name: K8S_CANARY_ROLLOUT
        with:
          replicas: 10%
          onRollback: "curl -X POST -H 'Content-type: application/json' --data '{"text":"failed to deploy: rollback"}' $SLACK_WEBHOOK_URL"
```

docs/rfcs/0011-script-run-stage.md
ffjlabo: Thank you for your questions!
I will write about these things on rfc, but in advance, I will write my opinions on this .

> Do we support multi script_run stages combine
Yes, I think so. 

> How the user-defined onRollback be run in case of multi script_run stages combine
Multi script_run stages can be executed in the order these are set.

> How the user-defined onRollback be run in combination with the XXX_ROLLBACK stage? I mean, if k8s app is rollbacked and the pipeline contains script_run, how do we perform the rollback process? Should it be XXX_ROLLBACK first then SCRIPT_RUN_ROLLBACK?

I don't think about it. I will try to check the implementation and propose it 🙇

docs/rfcs/0011-script-run-stage.md
khanhtc1202: Why does it come to this place 🤔 And also, how do we combine your idea with the current rollback logic? Do we have any plans for this?

docs/rfcs/0011-script-run-stage.md
khanhtc1202: > Multi script_run stages can be executed in the order these are set.

Either with their rollback script?

ffjlabo: I think it will be 2. (But I don't know yet it can be realized technically 🙇 )
xxx_canary -> script_run -> script_run -> rollback -> script_rollback -> script_rollback

docs/rfcs/0011-script-run-stage.md
ffjlabo: > Why does it come to this place 🤔 And also, how do we combine your idea with the current rollback logic? Do we have any plans for this?

It is a just idea, so I have no idea to realize it 🙇 
But I thought it might be more useful to be able to set `onRollback` on each stages than set a stage to execute any command.

ffjlabo:
@khanhtc1202 
I implemented the stage as an alpha version. So, I want to merge this PR and work on the rollback implementation. 🙏 
https://github.com/pipe-cd/pipecd/pull/4720

I will describe about the rollback on the original issue and fix the RFC on another PR :)

t-kikuc (approved):
Great

khanhtc1202 (approved):
Here you go 🚀

PR #4605, opened 2023-10-06
Fix unable to fetch ECS taskset tags

khanhtc1202:

The current implementation of fetching ECS taskset is wrong because ECS describe-services API does not return tasksets' TAGS. We have to use the ECS describe-task-sets with `--includes TAGS` option to get those tags for filtering logic.

refer: https://docs.aws.amazon.com/cli/latest/reference/ecs/describe-task-sets.html

Follow #4600 


ffjlabo (approved):
👍

nghialv (approved):

PR #4614, opened 2023-10-12
Fix unable to rollout ECS first taskset

khanhtc1202:

Follow PR #4600 

### DIFF
--- a/pkg/app/piped/platformprovider/ecs/client.go
+++ b/pkg/app/piped/platformprovider/ecs/client.go
@@ -275,8 +275,9 @@ func (c *client) GetServiceTaskSets(ctx context.Context, service types.Service) 
 		activeTaskSetArns = append(activeTaskSetArns, *svc.TaskSets[i].TaskSetArn)
 	}
 
+	var taskSets []*types.TaskSet
 	if len(activeTaskSetArns) == 0 {
-		return nil, fmt.Errorf("failed to get task sets of service %s: services empty", *service.ServiceName)
+		return taskSets, nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -275,8 +275,9 @@ func (c *client) GetServiceTaskSets(ctx context.Context, service types.Service)
 		activeTaskSetArns = append(activeTaskSetArns, *svc.TaskSets[i].TaskSetArn)
 	}
 
+	var taskSets []*types.TaskSet
 	if len(activeTaskSetArns) == 0 {
-		return nil, fmt.Errorf("failed to get task sets of service %s: services empty", *service.ServiceName)
+		return taskSets, nil
```
ffjlabo, line 280: imo: just return `[]*types.TaskSet{}` and define `taskSets` variable to minify its scope. 👀

ffjlabo: And it should be better to add comment for the reason.

khanhtc1202: Okey, I don't have strong opinion on this 👍 Let's me adopt your suggestion

khanhtc1202: @ffjlabo addressed by https://github.com/pipe-cd/pipecd/pull/4614/commits/8c6060cb8fe70574509af60da2ea5e2ebed40648

nghialv (approved):
🐬

PR #4648, opened 2023-11-02
Make actions plan preview handle timeout option

mi11km:
The [actions-plan-preview](https://github.com/pipe-cd/actions-plan-preview/tree/main) supports [timeout option](https://github.com/pipe-cd/actions-plan-preview/blob/main/action.yaml#L15-L18). 
However, the actions-plan-preview finished in 5m (default timeout minutes) although setting 15m in timeout option. 
Here is the actual actions logs:
スクリーンショット 2023-11-02 午後12 14 14
スクリーンショット 2023-11-02 午後2 07 55

To make the timeout option work, use -piped-handle-timeout when executing pipectl.

**Which issue(s) this PR fixes**: No issue

### DIFF
--- a/pkg/model/notificationevent_test.go
+++ b/pkg/model/notificationevent_test.go
@@ -1,0 +1,66 @@
+// Copyright 2023 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGroup(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name string
+		typ  NotificationEventType
+		want NotificationEventGroup
+	}{
+		{
+			name: "returns EVENT_DEPLOYMENT for type < 100",
+			typ:  99,
+			want: NotificationEventGroup_EVENT_DEPLOYMENT,
+		},
+		{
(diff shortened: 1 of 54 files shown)


khanhtc1202:
@mi11km Nice catch 👍 
Could you sign-off your commit using `git commit -s` 😄

mi11km:
@khanhtc1202 
Should the -timeout and -piped-handle-timeout options be able to be set to different values?

tool/actions-plan-preview/github.go
```diff
@@ -55,10 +55,10 @@ type GraphQLClient interface {
...
 // to parse and build githubEvent struct.
 // Currently, we support 2 kinds of event as below:
-// - PullRequestEvent
-//   https://pkg.go.dev/github.com/google/go-github/v36/github#PullRequestEvent
-// - IssueCommentEvent
-//   https://pkg.go.dev/github.com/google/go-github/v36/github#IssueCommentEvent
+//   - PullRequestEvent
+//     https://pkg.go.dev/github.com/google/go-github/v36/github#PullRequestEvent
+//   - IssueCommentEvent
+//     https://pkg.go.dev/github.com/google/go-github/v36/github#IssueCommentEvent
```
khanhtc1202, line 61: Should keep this as is due to go convention 👀

khanhtc1202:
> @khanhtc1202 Should the -timeout and -piped-handle-timeout options be set to different values?

👍 I think we should keep that different and have a way to enable users to set both.

tool/actions-plan-preview/github.go
mi11km: fix it in [this PR](https://github.com/pipe-cd/pipecd/pull/4648/commits/d0cb1a5deffb760baab87c916f6561958f4553e5)

pkg/model/notificationevent_test.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			e := NotificationEvent{Type: tt.typ}
+			got := e.Group()
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
```
khanhtc1202, line 66: this file is not relevant, please remove from the PR 👀

khanhtc1202 (approved):
LGTM, thanks 🥕

PR #4649, opened 2023-11-02
Add ECSTargetGroupObject to the configuration reference documentation

ouchi2501:

I added ECSTargetGroupObject as it was not documented.

I didn't know to what extent I should add the document, so I added it only to the latest version first.
Please let me know if you need to add it to `docs-dev` or any other ver.

Fixes #


khanhtc1202:
@ouchi2501 Nice catch 🙏 
Please also update the docs file under `/docs-dev` with the same content 🙏

ouchi2501:
@khanhtc1202 
Thanks for the comment 👍🏼 
Also updated `/docs-dev`.

khanhtc1202 (approved):
Thank you 🙌

PR #4653, opened 2023-11-06
Add actions/stale workflow

kentakozuka:

Fixes https://github.com/pipe-cd/pipecd/discussions/4647

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -14,7 +14,7 @@ jobs:
           close-issue-message: 'This issue was closed because it has been stalled for 7 days with no activity. Feel free to reopen if still applicable.'
           days-before-issue-stale: 180
           days-before-issue-close: 7
-          exempt-issue-labels: 'kind/bug,good first issue'
+          exempt-issue-labels: 'not-auto-close,kind/bug,good first issue'
           # PRs
           stale-pr-message: 'This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
           close-pr-message: 'This PR was closed because it has been stalled for 7 days with no activity. Feel free to reopen if still applicable.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yaml
```diff
@@ -0,0 +1,22 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          # Issues
+          stale-issue-message: 'This issue is stale because it has been open 180 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
+          close-issue-message: 'This issue was closed because it has been stalled for 7 days with no activity. Feel free to reopen if still applicable.'
+          days-before-issue-stale: 180
+          days-before-issue-close: 7
```
khanhtc1202, line 16: I want to keep good-first-issue labeled issue out of this rule, wdyt?

kentakozuka: Exempted `good first issue` and `kind/bug`.

kentakozuka:
@khanhtc1202 Fixed 👍

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,23 @@
...
+          exempt-issue-labels: 'kind/bug,good first issue'
```
khanhtc1202, line 17: Should we add one more label that we can use to prevent the CI from deleting that issue manually? I mean, I think we should have a way to bypass the logic of removing issues automatically on purpose, wdyt?

khanhtc1202: aka. I propose to add `help wanted` labels to exempt labels list.

kentakozuka: sounds good idea 👍

kentakozuka:
@khanhtc1202 Fixed 👍

.github/workflows/stale.yaml
khanhtc1202: that's better 🫡

khanhtc1202 (approved):
🫡

Issue #4657, opened 2023-11-08
Send error comment before plan preview exit


TakumaKurosawa:
**What would you like to be added**:
Send comment before exiting when error occurred.

**Why is this needed**:
It's too hard to know why plan preview exited.

TakumaKurosawa:
I'd like to work on this.
I've already make PR: https://github.com/pipe-cd/pipecd/pull/4658

khanhtc1202:
Nice, thank you 🙌

PR #4658, opened 2023-11-08
Send error comment before plan preview exit

TakumaKurosawa:
It's hard to know why plan-preview exited.
Sending error comment before exiting makes us easier to find the problem.

Fixes https://github.com/pipe-cd/pipecd/issues/4657

  - plan-preview error comment is going to be added a line break
  
Screenshot 2023-11-08 at 15 37 46

  - error comment will displayed before plan-preview exiting

### DIFF
--- a/tool/actions-plan-preview/main.go
+++ b/tool/actions-plan-preview/main.go
@@ -154,7 +154,7 @@ func main() {
 		return
 	}
 
-	if comment.IsMinimized {
+	if bool(comment.IsMinimized) {
 		log.Printf("Previous plan-preview comment has already minimized. So don't minimize anything\n")
 		return
 	}
--- a/tool/actions-plan-preview/main.go
+++ b/tool/actions-plan-preview/main.go
@@ -101,14 +101,14 @@ func main() {
 	}
 
 	if event.PRClosed {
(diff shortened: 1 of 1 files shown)


tool/actions-plan-preview/main.go
```diff
@@ -152,7 +154,7 @@ func main() {
 		return
 	}
 
-	if bool(comment.IsMinimized) {
+	if comment.IsMinimized {
```
khanhtc1202, line 157: I think this cast is to ensure we can notice if the lib type Boolean changed, so let's keep it.

khanhtc1202:
Overall looks good, just left a small nit 😄

tool/actions-plan-preview/main.go
TakumaKurosawa: I see your point👍

TakumaKurosawa: @khanhtc1202 
Applied!!
https://github.com/pipe-cd/pipecd/pull/4658/commits/449aa1b24cd33a627a0fd8909e20463294e9d8d2

khanhtc1202 (approved):
Great catch, thank you 🙌

Issue #4665, opened 2023-11-09
Prevent running standalone ECS tasks during deployment


kentakozuka:
**What would you like to be added**:

Prevent standalone ECS tasks from running at the deployment.
Adding a flag `ECSApp.runAtDeployment` to disable running at the deployment...?

**Why is this needed**:

There is the case;
- Run a task at a certain time triggered by EventBridge.
- The task should not be triggered at the deployment.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/scheduled_tasks-eventbridge-scheduler.html

kentakozuka:
Current implementation;
1. Check if the service definition exists.
2. If not, Run the task.
code is [here](https://github.com/pipe-cd/pipecd/blob/89fb20354a9968092c0df1c00ad7538a346d236e/pkg/app/piped/executor/ecs/deploy.go#L88)

kentakozuka:
Created a draft PR https://github.com/pipe-cd/pipecd/pull/4666. WDYT?

PR #4666, opened 2023-11-09
Add a flag to skip running standalone task during deployment

kentakozuka:

Fixes https://github.com/pipe-cd/pipecd/issues/4665

### DIFF
--- a/pkg/app/piped/executor/ecs/ecs.go
+++ b/pkg/app/piped/executor/ecs/ecs.go
@@ -199,6 +199,11 @@ func runStandaloneTask(
 		return false
 	}
 
+	if !*ecsInput.RunStandaloneTask {
+		in.LogPersister.Infof("Skipped running task")
+		return true
+	}
+
 	err = client.RunTask(
 		ctx,
 		*td,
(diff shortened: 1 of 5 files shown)


pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -199,6 +199,11 @@ func runStandaloneTask(
 		return false
 	}
 
+	if !*ecsInput.RunStandaloneTask {
+		in.LogPersister.Infof("Skipped to the run task")
```
t-kikuc, line 203:
```suggestion
		in.LogPersister.Infof("Skipped running task")
```
I think it would be clearer
because this if-clause means skipping RunTask, not skipping to RunTask. 
WDYT?

t-kikuc:
Successfully the codes @kentakozuka suggested worked in my dev environment!

kentakozuka:
If it looks good to you folks, I will add the docs, too.

pkg/app/piped/executor/ecs/ecs.go
```diff
@@ -199,6 +199,11 @@ func runStandaloneTask(
 		return false
 	}
 
+	if !*ecsInput.RunStandaloneTask {
+		in.LogPersister.Infof("Skipped running task")
+		return true
+	}
```
khanhtc1202, line 205: Just to confirm, with this change, the task would not be "even" created at the time piped "sync" the ECS app. We can expect other triggers to start the tasks for us, is that right? In such case, how can that trigger specify the exact taskDefinition revision to use to create the ECS task 🤔 WDYT @kentakozuka @nnnkkk7

kentakozuka: > We can expect other triggers to start the tasks for us, is that right? 

Yes. We need other triggers to start the task.

> In such case, how can that trigger specify the exact taskDefinition revision to use to create the ECS task

AFAIK, EventBridge Scheduler allows us to select the latest revision or specific one to run.

Untitled (1)

khanhtc1202: yep, I understand that they allow us to do that, but the point is how actually dev can understand that which revision of taskdef is what they want (if they create it not of the time they commit the code change). Anw, it's just my concern on how actually this function could be used in usecase, look like it not really pipecd problem (if both of you agree with that way to select revision). Besides, maybe print out registered taskdef revision on stage log could help.

khanhtc1202 (approved):
lgtm

PR #4672, opened 2023-11-14
Add rfc for pipectl init

ouchi2501:

Add init to the pipectl command.
See below for Discussion.
https://github.com/pipe-cd/pipecd/discussions/4635

Fixes #

### DIFF
--- a/docs/rfcs/0011-pipectrl-init.md
+++ b/docs/rfcs/0011-pipectrl-init.md
@@ -1,9 +1,9 @@
-- Start Date: 2023-11-24
+- Start Date: 2023-11-14
 - Target Version: 1.45.1
 
 # Summary
 
-To simplify the creation of a configuration file using the pipectrl init command.
+To simplify the creation of a configuration file using the pipectl init command.
 
 # Motivation
 
@@ -12,8 +12,8 @@ Also, build in the ability to easily migrate from the configuration files of use
 
 # Detailed design
 
-pipectrl init command will be added to the pipecdctl command.
-The pipectrl init command will have 3 features.
+pipectl init command will be added to the pipectl command.
+The pipectl init command will have 3 features.
 
 1. Create a configuration file from scratch
 2. Create a configuration file from an existing configuration file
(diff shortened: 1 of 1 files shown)


kentakozuka:
Is it only app.pipecd.yaml that the first step(1. Create a configuration file from scratch) creates?

ouchi2501:
https://github.com/pipe-cd/pipecd/pull/4672#issuecomment-1813670644
@kentakozuka 
Yes, Only app.pipecd.yaml

kentakozuka:
As you mentions [here](https://github.com/pipe-cd/pipecd/discussions/4635#discussioncomment-7507636), would it be better to generate app.pipecd.yaml from applications's configs?
I input configs in [example](https://github.com/pipe-cd/pipecd/tree/master/examples/kubernetes/simple) and PipeCD's GPT output is the following. Looks not bad. WDYT?

```yaml
kind: KubernetesApp
spec:
  input:
    # Specify the namespace if needed; otherwise, PipeCD deploys to the default namespace
    namespace: default
    kubernetes:
      manifests:
        - deployment.yaml
        - service.yaml
  pipeline:
    stages:
      # The primary deployment rollout
      - name: K8S_PRIMARY_ROLLOUT
      # The stage for manual approval (if needed)
      - name: WAIT_APPROVAL
      # The canary deployment rollout
      - name: K8S_CANARY_ROLLOUT
        with:
          replicas: 50% # Adjust the percentage as required
      # The cleanup stage for the canary deployment
      - name: K8S_CANARY_CLEAN
  trafficRouting:
    method: pod
    kubernetes:
      primary:
        labels:
          pipecd.dev/variant: primary
      canary:
        labels:
          pipecd.dev/variant: canary

```

ouchi2501:
I think it should have both features.
As for pipectl init, I think it would be less confusing for users if it was created as a general interactive init command rather than relying solely on GPT from the start!
However, I also think the ability to create a config from the application config through GPT would be very nice!
I think it would be nice to have the ability to create a config from the application config via GPT as an option for pipectl init!

kentakozuka:
@ouchi2501 Sounds very cool! Let's stick with your plan :)

khanhtc1202 (approved):

PR #4673, opened 2023-11-14
Add docs for k8s annotations

ffjlabo:

Some annotations for k8s resources are not described in official docs. So I added docs.

Fixes #


khanhtc1202 (approved):
Good catch

PR #4680, opened 2023-11-19
add ephemeralStorage option to lambda config

### LINKED ISSUE #4679: support ephemeralStorage optional filed for lambda
I would like to add an ephemeralStorage optional value to the lambda configuration.

**Why is this needed**:
Because currently it's not possible to configure ephemeralStorage for lambda.

nnnkkk7:
I added an ephemeralStorage optional value to the lambda configuration.

Fixes #4679 

No

### DIFF
--- a/pkg/app/piped/platformprovider/lambda/client.go
+++ b/pkg/app/piped/platformprovider/lambda/client.go
@@ -126,7 +126,7 @@ func (c *client) CreateFunction(ctx context.Context, fm FunctionManifest) error 
 		}
 		input.Architectures = architectures
 	}
-	if fm.Spec.EphemeralStorage.Size != 0 {
+	if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
 		input.EphemeralStorage.Size = aws.Int32(fm.Spec.EphemeralStorage.Size)
 	}
 	if fm.Spec.VPCConfig != nil {
@@ -281,7 +281,7 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 		if fm.Spec.Handler != "" {
 			configInput.Handler = aws.String(fm.Spec.Handler)
 		}
-		if fm.Spec.EphemeralStorage.Size != 0 {
+		if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
 			configInput.EphemeralStorage.Size = aws.Int32(fm.Spec.EphemeralStorage.Size)
 		}
 		if fm.Spec.VPCConfig != nil {
--- a/pkg/app/piped/platformprovider/lambda/client.go
+++ b/pkg/app/piped/platformprovider/lambda/client.go
@@ -126,7 +126,7 @@ func (c *client) CreateFunction(ctx context.Context, fm FunctionManifest) error 
 		}
 		input.Architectures = architectures
 	}
-	if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
+	if fm.Spec.EphemeralStorage != nil {
 		input.EphemeralStorage.Size = aws.Int32(fm.Spec.EphemeralStorage.Size)
 	}
 	if fm.Spec.VPCConfig != nil {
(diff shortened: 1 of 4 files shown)


pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -278,6 +281,9 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 		if fm.Spec.Handler != "" {
 			configInput.Handler = aws.String(fm.Spec.Handler)
 		}
+		if fm.Spec.EphemeralStorage.Size != 0 {
```
khanhtc1202, line 284:
```suggestion
		if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
```
avoid nil pointer error due to the omitempty json tag

khanhtc1202:
@nnnkkk7 Thanks for the contribution. For this configuration, should we also add this to `CreateFunctionXXX` step as well? I mean the following 2 places
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/lambda/client.go#L111
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/lambda/client.go#L160

pkg/app/piped/platformprovider/lambda/client.go
nnnkkk7: Thank you! I added it.
[add nil pointer check](https://github.com/pipe-cd/pipecd/pull/4680/commits/caf132172af73103066822359604429ef3840f74)

nnnkkk7:
@khanhtc1202 
Thanks! I had forgot to add  this configuration to CreateFunctionFromSource!
[add EphemeralStorage input to CreateFunctionFromSource](https://github.com/pipe-cd/pipecd/pull/4680/commits/639d0f448b035b82ea22525f618da90fc40a52ec)

pkg/app/piped/platformprovider/lambda/function.go
```diff
@@ -100,6 +103,11 @@ func (fmp FunctionManifestSpec) validate() error {
 			return fmt.Errorf("architecture is invalid: %w", err)
 		}
 	}
+	if fmp.EphemeralStorage != nil && fmp.EphemeralStorage.Size != 0 {
```
khanhtc1202, line 106: This condition is wrong 🤔 If the size value is set to `0` the validate function will be bypass.
```suggestion
	if fmp.EphemeralStorage != nil {
```

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -126,6 +126,9 @@ func (c *client) CreateFunction(ctx context.Context, fm FunctionManifest) error
 		}
 		input.Architectures = architectures
 	}
+	if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
```
khanhtc1202, line 129:
```suggestion
	if fm.Spec.EphemeralStorage != nil {
```
This condition should be updated because the value `0` will be yielded on the validation step.

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -179,7 +182,9 @@ func (c *client) CreateFunctionFromSource(ctx context.Context, fm FunctionManife
 			Variables: fm.Spec.Environments,
 		},
 	}
-
+	if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
```
khanhtc1202, line 185:
```suggestion
	if fm.Spec.EphemeralStorage != nil {
```
same as above

pkg/app/piped/platformprovider/lambda/client.go
```diff
@@ -278,6 +283,9 @@ func (c *client) updateFunctionConfiguration(ctx context.Context, fm FunctionMan
 		if fm.Spec.Handler != "" {
 			configInput.Handler = aws.String(fm.Spec.Handler)
 		}
+		if fm.Spec.EphemeralStorage != nil && fm.Spec.EphemeralStorage.Size != 0 {
```
khanhtc1202, line 286:
```suggestion
		if fm.Spec.EphemeralStorage != nil {
```
same as above

pkg/app/piped/platformprovider/lambda/function.go
nnnkkk7: Thank you!
I fixed these conditions! [fix condition](https://github.com/pipe-cd/pipecd/pull/4680/commits/8af45bb9ffde7b091a96363a0c8c30d2f2f99c61)

khanhtc1202 (approved):
LGTM, thank you 🎸

PR #4682, opened 2023-11-21
Add v2 to versionWhitelist

ffjlabo:

Added "v2" to versionWhitelist to recognize "autoscaling/v2" which is built-in resource

Fixes #


ffjlabo:
note:
Some versions also exists, but now added `v2` for the resources only supported by PipeCD like HorizontalPodAutoscaler
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#-strong-api-groups-strong

khanhtc1202 (approved):
Good catch 👍

PR #4683, opened 2023-11-22
Minimize latest plan preview comment before comment error message

TakumaKurosawa:

1. The status badge appear twice

Screenshot 2023-11-22 at 14 52 14

2. Minimize previous comment when plan-preview result has error

screencapture-github-CyberAgentAI-aoc-daifuku-app-config-pull-1323-2023-11-22-14_56_14

Fixes #

### DIFF
--- a/tool/actions-plan-preview/main.go
+++ b/tool/actions-plan-preview/main.go
@@ -141,8 +141,33 @@ func main() {
 			doComment(failureBadgeURL + "\nUnable to run plan-preview for a closed pull request.")
 			return
 		}
+
+		// Find comments we sent before
+		comment, err := findLatestPlanPreviewComment(ctx, ghGraphQLClient, event.Owner, event.Repo, event.PRNumber)
+		if err != nil {
+			log.Printf("Unable to find the previous comment to minimize (%v)\n", err)
+		}
+
 		body := makeCommentBody(event, result)
-		doComment(failureBadgeURL + "\n" + body)
+		doComment(body)
+
+		if comment == nil {
+			log.Println("plan-preview result has error")
+			os.Exit(1)
+			return
+		}
+
+		if bool(comment.IsMinimized) {
+			log.Printf("Previous plan-preview comment has already minimized. So don't minimize anything\n")
+			return
+		}
+
+		if err := minimizeComment(ctx, ghGraphQLClient, comment.ID, "OUTDATED"); err != nil {
+			log.Printf("warning: cannot minimize comment: %s\n", err.Error())
+			return
+		}
+
+		log.Printf("Successfully minimized last plan-preview result on pull request\n")
 		log.Println("plan-preview result has error")
 		os.Exit(1)
 	}
--- a/tool/actions-plan-preview/main.go
+++ b/tool/actions-plan-preview/main.go
@@ -142,60 +142,20 @@ func main() {
 			return
 		}
 
-		// Find comments we sent before
-		comment, err := findLatestPlanPreviewComment(ctx, ghGraphQLClient, event.Owner, event.Repo, event.PRNumber)
-		if err != nil {
-			log.Printf("Unable to find the previous comment to minimize (%v)\n", err)
-		}
+		minimizePreviousComment(ctx, ghGraphQLClient, event)
 
 		body := makeCommentBody(event, result)
 		doComment(body)
-
-		if comment == nil {
-			log.Println("plan-preview result has error")
-			os.Exit(1)
-			return
-		}
-
-		if bool(comment.IsMinimized) {
-			log.Printf("Previous plan-preview comment has already minimized. So don't minimize anything\n")
-			return
-		}
-
-		if err := minimizeComment(ctx, ghGraphQLClient, comment.ID, "OUTDATED"); err != nil {
-			log.Printf("warning: cannot minimize comment: %s\n", err.Error())
-			return
-		}
 
 		log.Printf("Successfully minimized last plan-preview result on pull request\n")
 		log.Println("plan-preview result has error")
 		os.Exit(1)
 	}
 
-	// Find comments we sent before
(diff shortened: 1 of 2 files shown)


ffjlabo:
@TakumaKurosawa 
Thank you for the PR!
I would like to recheck the purpose!

You want to do below, right?
- fix not to show the status badge twice
- minimize the previous comment when the plan-preview result has an error (because not minimized before)

TakumaKurosawa:
@ffjlabo 
Thanks for checking my pull request. 
Your thoughts are completely in line with mine.

tool/actions-plan-preview/main.go
```diff
@@ -141,8 +141,33 @@ func main() {
...
+
 		body := makeCommentBody(event, result)
-		doComment(failureBadgeURL + "\n" + body)
+		doComment(body)
+
+		if comment == nil {
+			log.Println("plan-preview result has error")
+			os.Exit(1)
+			return
+		}
```
khanhtc1202, line 158: What if there is no previous comment 🤔

TakumaKurosawa: @khanhtc1202 
I'm sorry for replying later.
Nothing to minimize any message if there is no previous comment.
I'm not sure what you want to know, sorry.

tool/actions-plan-preview/main.go
```diff
@@ -141,8 +141,33 @@ func main() {
...
+
+		if bool(comment.IsMinimized) {
+			log.Printf("Previous plan-preview comment has already minimized. So don't minimize anything\n")
+			return
+		}
+
+		if err := minimizeComment(ctx, ghGraphQLClient, comment.ID, "OUTDATED"); err != nil {
+			log.Printf("warning: cannot minimize comment: %s\n", err.Error())
+			return
+		}
```
ffjlabo, line 168: [IMO]
> minimize the previous comment not only when the plan-preview succeed but also the plan-preview result has an error (because not minimized before)

To realize this, It would be nice to gather these process into one method 👀 
- find latest comment
- makeCommentBody
- do comment
- minimizeComment

Then we only call the method regardless of failure or success. WDYT?

tool/actions-plan-preview/main.go
khanhtc1202: I mean in that case, this comment would be nil and I guess we can return normally, no need to exit with error here, isn't it? 🤔

tool/actions-plan-preview/main.go
TakumaKurosawa: @ffjlabo 
Yes! Exactly!
I will gather these process.

tool/actions-plan-preview/main.go
TakumaKurosawa: @khanhtc1202 
This `comment` is latest previous comment that must close.
But `if result.HasError() {}` is still `true`, so we have to exit with error.

tool/actions-plan-preview/main.go
TakumaKurosawa: @ffjlabo 
Applied!!
https://github.com/pipe-cd/pipecd/pull/4683/commits/74a9d8e231a532e2c86ab8e137119dafe86c0988

ffjlabo: @TakumaKurosawa Sorry for late reply 🙏 It looks good 👍

tool/actions-plan-preview/main.go
ffjlabo: @khanhtc1202 cc @TakumaKurosawa 
It should stop as exit 1 because the fix is in the error case in which plan-preview failed.

So if the previous comment doesn't exist, the code wants to finish as exit 1 with any errors when the plan-preview result has some errors.
This is like an early return.

ffjlabo: This is confusing, so he created a function for minimizing a previous comment.
This function does not affect whether to finish as exit 1.
https://github.com/pipe-cd/pipecd/commit/74a9d8e231a532e2c86ab8e137119dafe86c0988

khanhtc1202:
@ffjlabo please re-check 👀

tool/actions-plan-preview/main.go
khanhtc1202: Got your points, thank both 🙏

tool/actions-plan-preview/main.go
```diff
@@ -141,36 +141,21 @@ func main() {
...
 			return
 		}
+
+		minimizePreviousComment(ctx, ghGraphQLClient, event)
+
 		body := makeCommentBody(event, result)
-		doComment(failureBadgeURL + "\n" + body)
+		doComment(body)
+
+		log.Printf("Successfully minimized last plan-preview result on pull request\n")
```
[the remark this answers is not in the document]
TakumaKurosawa: @ffjlabo 
Applied!
https://github.com/pipe-cd/pipecd/pull/4683/commits/1cb42fac60cf1bbe3f7801ab639653a5d4e244f2

tool/actions-plan-preview/planpreview.go
```diff
@@ -331,3 +332,27 @@ func generateTerraformShortPlanDetails(details string) (string, error) {
...
+		log.Printf("Previous plan-preview comment has already minimized. So don't minimize anything\n")
+		return
+	}
+
+	if err := minimizeComment(ctx, ghGraphQLClient, comment.ID, "OUTDATED"); err != nil {
+		log.Printf("warning: cannot minimize comment: %s\n", err.Error())
+		return
+	}
+
+	log.Printf("Successfully minimized last plan-preview result on pull request\n")
```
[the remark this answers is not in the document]
TakumaKurosawa: https://github.com/pipe-cd/pipecd/pull/4683/commits/1cb42fac60cf1bbe3f7801ab639653a5d4e244f2

khanhtc1202:
@TakumaKurosawa The new version of action-plan-preview had been released, thank you 🙏 
https://github.com/pipe-cd/actions-plan-preview/releases/tag/v1.7.6

Issue #4616, opened 2023-10-14
Support progressive delivery for ECS apps using Service Discovery


khanhtc1202:
**What would you like to be added**:

Support ECS apps progressive delivery using only ECS components (without ALB configuration)

**Why is this needed**:

Currently, to perform progressive delivery for ECS applications, we need users to prepare ALB targetgroups. For some usecases, users' tasks may not directly accept/handle traffic from the ALB but via other tasks/components which leads to the inability to perform progressive delivery strategies.

khanhtc1202:
An idea for this is using 2 services that are configured using the same service registry id, one for the primary tasks and one for the canary tasks. Then tasks of those services will be treated as the same service to other components of users system. 💡

t-kikuc:
@khanhtc1202 

Hi, I have a question.

> For some usecases, users' tasks may not directly accept/handle traffic from the ALB but via other tasks/components

Could you give me some examples of such cases? 
(Directly calling ECS APIs or IP addresses or something? From where?)

I understand the current flow of canary deployment for ECS with ALB and target groups,
but I'd like to clarify what problem to solve.

t-kikuc:
> I understand the current flow of canary deployment for ECS with ALB and target groups,
but I'd like to clarify what problem to solve.

Thank you for the explanation.
I'll investigate the Service Discovery feature of ECS.
https://docs.aws.amazon.com/ja_jp/AmazonECS/latest/developerguide/service-discovery.html

t-kikuc:
I share my current situation of consideration.

I think this case assumes currently using ECS Service Discovery, not ECS Service Connect or App Mesh.

There can be three ways which solve the problem, though I'm not sure all of them are possible yet.

1. Enhance ECS Service Discovery - manipulate Route53??
2. Use ECS Service Connect - a young feature.
3. Use App Mesh  - Possible but complicated.

Besides, I'm wondering which is simpler, option-1 or option-2, if both are possible.

khanhtc1202:
@t-kikuc If (2) and (3) are not promising, let's focus on (1). But I don't really understand why we need to manipulate Route53 for that 🤔 

AFAIK we can create a copy of the current primary service A (call A') with `desiredCount` value set to the number of the canary variant's workloads/tasks and use the same `serviceRegistryID` with primary service A.

In that case, the CloudMap service would ensure that traffic to serviceRegistryID will be routed to both A and A' and we can use that to perform progressive delivery approach, wdyt?

[...]

Issue #4630, opened 2023-10-25
A unit test is flaky: TestGracefulStopCommand


kentakozuka:
**What happened**:

test passes

**What you expected to happen**:

not passes

**How to reproduce it**:

```
--- FAIL: TestGracefulStopCommand (0.00s)
    --- FAIL: TestGracefulStopCommand/graceful_stop_after_very_short_time (0.00s)
        binary_test.go:48: 
            	Error Trace:	/home/runner/work/pipecd/pipecd/pkg/app/launcher/cmd/launcher/binary_test.go:48
            	Error:      	Should be false
            	Test:       	TestGracefulStopCommand/graceful_stop_after_very_short_time
```

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

t-kikuc:
Sometimes, I face the same problem in the GitHub Actions test phase...

PR #4685, opened 2023-11-22
Support QuickSync for ECS without ELB or target groups

t-kikuc:
This PR supports QuickSync for ECS using Service Discovery, not ELB and target groups.
I will support canary and blue/green for Service Discovery later.

Part of #4616 

- **How are users affected by this change**: not affected including configs.
- **How to migrate (if breaking change)**:  no

### DIFF
--- a/pkg/app/piped/executor/ecs/deploy.go
+++ b/pkg/app/piped/executor/ecs/deploy.go
@@ -17,6 +17,7 @@ package ecs
 import (
 	"context"
 
+	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 	"github.com/pipe-cd/pipecd/pkg/app/piped/deploysource"
 	"github.com/pipe-cd/pipecd/pkg/app/piped/executor"
 	"github.com/pipe-cd/pipecd/pkg/config"
@@ -97,9 +98,13 @@ func (e *deployExecutor) ensureSync(ctx context.Context) model.StageStatus {
 		return model.StageStatus_STAGE_FAILURE
 	}
 
-	primary, _, ok := loadTargetGroups(&e.Input, e.appCfg, e.deploySource)
-	if !ok {
-		return model.StageStatus_STAGE_FAILURE
+	var primary *types.LoadBalancer
+	// When the service is accessed via ELB, the target group is not used.
+	if ecsInput.IsAccessedViaELB() {
+		primary, _, ok = loadTargetGroups(&e.Input, e.appCfg, e.deploySource)
+		if !ok {
+			return model.StageStatus_STAGE_FAILURE
+		}
 	}
 
 	recreate := e.appCfg.QuickSync.Recreate
--- a/pkg/app/piped/executor/ecs/deploy.go
+++ b/pkg/app/piped/executor/ecs/deploy.go
@@ -99,7 +99,7 @@ func (e *deployExecutor) ensureSync(ctx context.Context) model.StageStatus {
 	}
 
 	var primary *types.LoadBalancer
-	// When the service is accessed via ELB, the target group is not used.
+	// When the service is not accessed via ELB, the target group is not used.
 	if ecsInput.IsAccessedViaELB() {
 		primary, _, ok = loadTargetGroups(&e.Input, e.appCfg, e.deploySource)
 		if !ok {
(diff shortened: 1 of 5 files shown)


pkg/config/application_ecs.go
```diff
@@ -57,12 +57,19 @@ type ECSDeploymentInput struct {
 	// Run standalone task during deployment.
 	// Default is true.
 	RunStandaloneTask *bool `json:"runStandaloneTask" default:"true"`
+	// How the ECS service is accessed.
```
khanhtc1202, line 60: Could you list up possible values here?

pkg/config/testdata/application/ecs-app-service-discovery.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: pipecd.dev/v1beta1
+kind: ECSApp
+spec:
+  input:
+    serviceDefinitionFile: /path/to/servicedef.yaml
+    taskDefinitionFile: /path/to/taskdef.yaml
+    accessType: SERVICE_DISCOVERY
```
khanhtc1202, line 7: This constant is defined by us or by AWS lib? 🤔

pkg/config/application_ecs.go
t-kikuc: Thanks, I did it.

pkg/config/testdata/application/ecs-app-service-discovery.yaml
t-kikuc: "This constant" means 'SERVICE_DISCOVERY'?

If so, it's defined by us, not by AWS.

And other names would be added in the future, like 'APP_MESH'

khanhtc1202: Then we should define these values as constant of ECS platform provider and have validate in config package

t-kikuc: thanks, I got it.  I'll fix it.

t-kikuc: @khanhtc1202 
I did it. Could you check again?
919cc3cf26b37b119563dc30614ca224f8a1d16e

pkg/config/application_ecs.go
```diff
@@ -121,3 +142,12 @@ func (opts ECSTrafficRoutingStageOptions) Percentage() (primary, canary int) {
 	canary = 0
 	return
 }
+
+func (in *ECSDeploymentInput) validateAccessType() error {
```
khanhtc1202, line 146:
```suggestion
func (in *ECSDeploymentInput) validate() error {
```

pkg/config/application_ecs.go
```diff
@@ -121,3 +142,12 @@ func (opts ECSTrafficRoutingStageOptions) Percentage() (primary, canary int) {
...
+	switch in.AccessType {
+	case AccessTypeELB, AccessTypeServiveDiscovery:
+		return nil
```
khanhtc1202, line 149:
```suggestion
		break
```

pkg/config/application_ecs.go
```diff
@@ -121,3 +142,12 @@ func (opts ECSTrafficRoutingStageOptions) Percentage() (primary, canary int) {
...
+	default:
+		return fmt.Errorf("invalid accessType: %s", in.AccessType)
+	}
```
khanhtc1202, line 152:
```suggestion
	}
	return nil
```

t-kikuc:
@khanhtc1202 
I removed unrelated commits.

One test/go failed, but it's due to flakiness https://github.com/pipe-cd/pipecd/issues/4630
And it's passed by a re-run.

pkg/app/piped/executor/ecs/deploy.go
```diff
@@ -17,6 +17,7 @@ package ecs
 import (
 	"context"
 
+	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
```
khanhtc1202, line 20:
```suggestion
	"github.com/aws/aws-sdk-go-v2/service/ecs/types"

```
go import convention

t-kikuc: thanks, I fixed it.

khanhtc1202 (approved):
LGTM 🚀

PR #4668, opened 2023-11-11
feat(ECS): enable selection of listener rules

moko-poi:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes https://github.com/pipe-cd/pipecd/issues/4530

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
```
apiVersion: pipecd.dev/v1beta1
kind: ECSApp
input:
  taskDefinitionFile: taskdef.yaml
  serviceDefinitionFile: servicedef.yaml
  targetGroups:
    primary:
    canary:
```
- **Is this breaking change**: no
- **How to migrate (if breaking change)**:

PR #4688, opened 2023-11-28
Remove ModifyRules interface

khanhtc1202:

Since the logic ModifyListeners had been updated to ensure only modifying the actions of kind `ActionTypeEnumForward`, there is no need to separate ModifyRules and ModifyListeners.
Also, no rule exists without associating with listeners in the context of ELB.

Follow #4668 


khanhtc1202:
@moko-poi Please take a look 👀

ffjlabo (approved):
🚀

PR #4693, opened 2023-12-01
[ECS] Delete old tasksets before QuickSync except PRIMARY

### LINKED ISSUE #4692: [ECS] Unable to create new TaskSet since this exceeds the maximum limit.
**What happened**:

The below error happened and the deployment failed several times.  (ECS QuickSync)

> 5 deployments exist on the service. Unable to create new TaskSet since this exceeds the maximum limit.

Besides, that happened again even after deleting unnecessary deployments. 

related: https://github.com/pipe-cd/pipecd/discussions/4686

**How to reproduce it**:

Cancel deployments (not sure)

**Environment**:
- `piped` version:  v0.45.4-rc2
- `control-plane` version:
- Others:

t-kikuc:

If old deployments remain, the new deployment will sometimes fail.
So this PR removes old deployments before the new deployment.

Workaround for #4692

### DIFF
--- a/pkg/app/piped/executor/ecs/ecs.go
+++ b/pkg/app/piped/executor/ecs/ecs.go
@@ -258,7 +258,7 @@ func deleteOldTaskSets(ctx context.Context, client provider.Client, service type
 
 	// Remove old taskSets except PRIMARY if exist.
 	for _, prevTaskSet := range prevTaskSets {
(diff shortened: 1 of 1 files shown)


t-kikuc:
This is closed because this is just a workaround in hurry, and I'll fix it in another PR.

PR #4704, opened 2023-12-05
avoid nil pointer in ephemeral storage  settings

nnnkkk7:
If the size input is accessed, a nil pointer could be triggered, so I have fixed it.

Fixes #


khanhtc1202 (approved):
Nice catch 👍

PR #4705, opened 2023-12-05
Add documentation of pprof

### LINKED ISSUE #4687: [piped] Add documentation of pprof

as Tite
But we should discuss where to add pprof documentation

**Why is this needed**:

piped is able to profile with pprof.
Sometimes I want to measure piped's memory and CPU usage.
It's a shame that the wonderful thing about piped's support for pprof is not mentioned in the documentation.

https://github.com/pipe-cd/pipecd/blob/f10a8e770ab8821fa402f9dde15cfa1cd11638f6/pkg/app/piped/cmd/piped/piped.go#L255-L258

ouchi2501:

- Add documentation of pprof

Fixes #4687


khanhtc1202 (approved):
Great, thank you 🐎

PR #3763, opened 2022-06-22
Fix to publish the image of piped and launcher to GCR

knanao:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4712, opened 2023-12-10
feat: github action matrix for images build

hungran:
**What this PR does / why we need it**: better DRY on images build CI
this also prepare for upgrade image base, will have flexible option to put image TAG as ARG
https://github.com/pipe-cd/pipecd/discussions/4702#discussioncomment-7810428

another point, though that this could be better view
image

Part of #4713 
- [x] add matrix image build

### DIFF
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -114,7 +114,7 @@ jobs:
           context: .
           file: ${{ matrix.dockerfile }}
           platforms: linux/amd64,linux/arm64
-          tags: ${{ env.GHCR }}/pipe-cd/${{ matrix.image }}:${{ env.PIPECD_VERSION }}
+          tags: ${{ matrix.container_registry }}/pipe-cd/${{ matrix.image }}:${{ env.PIPECD_VERSION }}
       
       # Building and pushing Helm charts.
       - name: Install helm
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -1,4 +1,4 @@
-name: publish
+name: publish_image_chart
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


khanhtc1202 (dismissed):
Great, thank you ❤️

.github/workflows/publish_image_chart.yaml
```diff
@@ -76,69 +102,15 @@ jobs:
...
-      - name: Build and push helloworld image
-        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
-        with:
-          push: true
-          context: .
-          file: cmd/helloworld/Dockerfile
-          platforms: linux/amd64,linux/arm64
-          tags: |
-            ${{ env.GHCR }}/pipe-cd/helloworld:${{ env.PIPECD_VERSION }}
-            ${{ env.GCR }}/pipecd/helloworld:${{ env.PIPECD_VERSION }}
```
ffjlabo, line 140: It looks that some container images are hosted by GCR 👀 
This may be because Cloud Run can use only with Artifact Registry, GCR, Docker Hub.
So it would be nice to consider for GCR 🤔 
WDYT @khanhtc1202 ?

📝 memo
Cloud Run docs: https://cloud.google.com/run/docs/deploying
PRs that maybe related
- https://github.com/pipe-cd/pipecd/pull/3763
- https://github.com/pipe-cd/pipecd/pull/3772

hungran:
Guys, I know that it's hard to test the workflow publish build and push image by matrix from this PR
Please give me sometime on a test then show you how with the current change
@khanhtc1202 @ffjlabo

.github/workflows/publish_image_chart.yaml
hungran: @ffjlabo good catch!
if we GCR only when Cloud Run, so ideally we could bring all images to GCR
however, not sure if PipeCD want to rely on a specific vendor?
@khanhtc1202 

btw, even we won't bring all the images to GCR, we could add more configuration for the matrix by this way 
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations

already push the change base on that, please check below comment

ffjlabo (approved):
It looks good! Here we go 🚀

ffjlabo:
@khanhtc1202 plz re-check 🙏

khanhtc1202 (approved):
Great, thank you 🙏

PR #4719, opened 2023-12-14
cover gcr.io/pipecd and ghcr.io/pipe-cd

hungran:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**: 
fix PR https://github.com/pipe-cd/pipecd/pull/4712 and https://github.com/pipe-cd/pipecd/pull/4718
Test from my repo
10 jobs
7 job ghcr.io/pipe-cd
3 job gcr.io/pipecd for `helloworld`, `launcher` and `piped`
image

Fixes #

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #4718, opened 2023-12-14
fix: matrix build recorrect and use docker build and push action on v5.0.0

hungran:
**What this PR does / why we need it**: fix last PR #https://github.com/pipe-cd/pipecd/pull/4712
apologized for my last PR did not deep check test
This PR will introduce `exclude`  and update docker build and push action to v5.0.0 on all flows

image

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**: n/a

- **How are users affected by this change**: n/a
- **Is this breaking change**: n/a
- **How to migrate (if breaking change)**: n/a

PR #4721, opened 2023-12-14
fix dockefile path

hungran:
**What this PR does / why we need it**: fix the last PR 
https://github.com/pipe-cd/pipecd/pull/4719 
https://github.com/pipe-cd/pipecd/pull/4718
https://github.com/pipe-cd/pipecd/pull/4712

seem like hard to maintain the image because of fragment on base tool image and the 2nd iterator.
even better by multistage build tho?

Fixes #

**Does this PR introduce a user-facing change?**: n/a

- **How are users affected by this change**: n/a
- **Is this breaking change**: n/a

### DIFF
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -35,14 +35,19 @@ jobs:
           - pipectl
         include:
           - image: helloworld
+            dockerfile: cmd/helloworld/Dockerfile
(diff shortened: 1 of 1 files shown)


khanhtc1202 (approved):
👍👍

ffjlabo (approved):
Thank you !🚀 🚀

PR #4722, opened 2023-12-15
fix: exclude pipecd image to gcr.io as we won't need

hungran:
basis only helloworld, launcher, piped need gcr.io

Fixes #

**Does this PR introduce a user-facing change?**:n/a

- **How are users affected by this change**: n/a
- **Is this breaking change**: n/a


khanhtc1202 (approved):
Nice catch, thanks 🙏

ffjlabo (approved):
Thx👍

PR #4726, opened 2023-12-17
[ECS] Enable selection of listener rules to modify

### LINKED ISSUE #4725: [ECS] Support modifying listener rules other than defaults

Select which listener rules to modify.

**Why is this needed**:

We need to specify the rules to modify other than default one.
For example, one listener has multiple listener rules and they have different path conditions and target services.

Now PipeCD updates only the `Default` rule.
So even if we want to modify the `rule-web` below, only the `Default` rule is modified.

image

t-kikuc:

Fixes #4725

**Does this PR introduce a user-facing change?**:  Add an option in ECSApp's input.

### DIFF
--- a/pkg/app/piped/executor/ecs/deploy.go
+++ b/pkg/app/piped/executor/ecs/deploy.go
@@ -206,7 +206,7 @@ func (e *deployExecutor) ensureTrafficRouting(ctx context.Context) model.StageSt
 		return model.StageStatus_STAGE_FAILURE
 	}
 
-	if !routing(ctx, &e.Input, e.platformProviderName, e.platformProviderCfg, *primary, *canary, e.appCfg.Input.ListenerRuleArn) {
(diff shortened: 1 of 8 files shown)


t-kikuc:
This PR is similar to https://github.com/pipe-cd/pipecd/pull/4668/ , but adds `ListenerRuleSelector` to `ECSDeploymentInput`

khanhtc1202:
@moko-poi Could you review this change to ensure this fit your usecase 👀

t-kikuc:
I created another PR (https://github.com/pipe-cd/pipecd/pull/4733) so that we don't need to specify in app.pipecd.yaml.

This would be simpler for users.

khanhtc1202:
Nice, I prefer to go with the approach provided by PR #4733 . Let's hold this PR!

t-kikuc:
I close this PR because #4733 is better and merged.

PR #4733, opened 2023-12-21
[ECS] Modify ELB listener rules other than defaults without adding config

### LINKED ISSUE #4725: [ECS] Support modifying listener rules other than defaults

t-kikuc:

Modifying listener rules which is not the default rule without specifying in app.pipecd.config, unlike #4726. 

Fixes #4725 

** A short explanation ** 

If the ELB and app.pipecd.yaml are as below, the rules that will be modified are `rule-A1`,`rule-B1` and `rule-B2` 
since they have the same target groups as app.pipecd.yaml.

```
ELB-X
|-- Listener-A
|    |-- rule-A1 (default)
|        |-- targetGroup-p
|        |-- targetGroup-c
|-- Listener-B
    |-- rule-B1 (default)
    |    |-- targetGroup-p
    |    |-- targetGroup-c
    |-- rule-B2   [path-pattern: /xxx/*]
    |    |-- targetGroup-p
    |    |-- targetGroup-c
    |-- rule-B3   [path-pattern: /yyy/*]
        |-- targetGroup-x
        |-- targetGroup-y
```

app.pipecd.yaml: 

```yaml
kind: ECSApp
spec:
  input:
    targetGroups:
      primary:
        targetGroupArn: arn:aws:elasticloadbalancing:ap-northeast-1:<account-id>:targetgroup/xxx/yyy1 # targetGroup-p
      canary:
        targetGroupArn: arn:aws:elasticloadbalancing:ap-northeast-1:<account-id>:targetgroup/xxx/yyy2 # targetGroup-c
```

### DIFF
--- a/pkg/app/piped/platformprovider/ecs/client.go
+++ b/pkg/app/piped/platformprovider/ecs/client.go
@@ -438,48 +438,47 @@ func (c *client) ModifyListeners(ctx context.Context, listenerArns []string, rou
 	}
 
 	for _, listenerArn := range listenerArns {
-		// Describe the listener to get the current actions
-		describeListenersOutput, err := c.elbClient.DescribeListeners(ctx, &elasticloadbalancingv2.DescribeListenersInput{
-			ListenerArns: []string{listenerArn},
+		describeRulesOutput, err := c.elbClient.DescribeRules(ctx, &elasticloadbalancingv2.DescribeRulesInput{
+			ListenerArn: aws.String(listenerArn),
 		})
 		if err != nil {
-			return fmt.Errorf("error describing listener %s: %w", listenerArn, err)
+			return fmt.Errorf("error describing rules %s: %w", listenerArn, err)
 		}
 
-		// Prepare the actions to be modified
-		var modifiedActions []elbtypes.Action
-		for _, action := range describeListenersOutput.Listeners[0].DefaultActions {
-			if action.Type == elbtypes.ActionTypeEnumForward {
-				// Modify only the forward action
-				modifiedAction := elbtypes.Action{
-					Type: elbtypes.ActionTypeEnumForward,
-					ForwardConfig: &elbtypes.ForwardActionConfig{
-						TargetGroups: []elbtypes.TargetGroupTuple{
-							{
-								TargetGroupArn: aws.String(routingTrafficCfg[0].TargetGroupArn),
-								Weight:         aws.Int32(int32(routingTrafficCfg[0].Weight)),
-							},
-							{
-								TargetGroupArn: aws.String(routingTrafficCfg[1].TargetGroupArn),
-								Weight:         aws.Int32(int32(routingTrafficCfg[1].Weight)),
+		for _, rule := range describeRulesOutput.Rules {
+			var modifiedActions []elbtypes.Action
+			for _, action := range rule.Actions {
+				if action.Type == elbtypes.ActionTypeEnumForward && routingTrafficCfg.hasSameTargets(action.ForwardConfig.TargetGroups) {
+					// Modify only the forward action which has the same target groups.
+					modifiedAction := elbtypes.Action{
+						Type:  elbtypes.ActionTypeEnumForward,
+						Order: action.Order,
+						ForwardConfig: &elbtypes.ForwardActionConfig{
+							TargetGroups: []elbtypes.TargetGroupTuple{
+								{
+									TargetGroupArn: aws.String(routingTrafficCfg[0].TargetGroupArn),
+									Weight:         aws.Int32(int32(routingTrafficCfg[0].Weight)),
+								},
+								{
+									TargetGroupArn: aws.String(routingTrafficCfg[1].TargetGroupArn),
+									Weight:         aws.Int32(int32(routingTrafficCfg[1].Weight)),
+								},
 							},
 						},
-					},
+					}
+					modifiedActions = append(modifiedActions, modifiedAction)
+				} else {
+					modifiedActions = append(modifiedActions, action)
 				}
-				modifiedActions = append(modifiedActions, modifiedAction)
-			} else {
-				// Keep other actions unchanged
-				modifiedActions = append(modifiedActions, action)
 			}
-		}
 
-		// Modify the listener
-		_, err = c.elbClient.ModifyListener(ctx, &elasticloadbalancingv2.ModifyListenerInput{
-			ListenerArn:    aws.String(listenerArn),
-			DefaultActions: modifiedActions,
-		})
-		if err != nil {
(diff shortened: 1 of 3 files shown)


t-kikuc:
@moko-poi @nnnkkk7 
Does this PR satisfy your usecases?

pkg/app/piped/platformprovider/ecs/routing_traffic.go
```diff
@@ -14,9 +14,37 @@
...
+	})
+	sort.Slice(forwardActionTargets, func(i, j int) bool {
+		return *forwardActionTargets[i].TargetGroupArn < *forwardActionTargets[j].TargetGroupArn
+	})
+
+	for i := range c {
+		if c[i].TargetGroupArn != *forwardActionTargets[i].TargetGroupArn {
+			return false
+		}
+	}
```
khanhtc1202, line 47: Instead of sorting both, then iterating through, using an extra map with TargetGroupArn as key could provide a better performance. It's a more straightforward approach.

t-kikuc: @khanhtc1202 Thanks! I'll try

t-kikuc: @khanhtc1202 
I fixed on https://github.com/pipe-cd/pipecd/pull/4733/commits/727e13c6e407793ee28a8bff8d5ec5e11728c482

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -438,48 +438,58 @@ func (c *client) ModifyListeners(ctx context.Context, listenerArns []string, rou
...
 	for _, listenerArn := range listenerArns {
-		// Describe the listener to get the current actions
-		describeListenersOutput, err := c.elbClient.DescribeListeners(ctx, &elasticloadbalancingv2.DescribeListenersInput{
-			ListenerArns: []string{listenerArn},
+		describeRulesOutput, err := c.elbClient.DescribeRules(ctx, &elasticloadbalancingv2.DescribeRulesInput{
+			ListenerArn: aws.String(listenerArn),
 		})
 		if err != nil {
-			return fmt.Errorf("error describing listener %s: %w", listenerArn, err)
+			return fmt.Errorf("failed to describe rules %s: %w", listenerArn, err)
```
khanhtc1202, line 445:
```suggestion
			return fmt.Errorf("failed to describe rules of listener %s: %w", listenerArn, err)
```

pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -438,48 +438,58 @@ func (c *client) ModifyListeners(ctx context.Context, listenerArns []string, rou
...
-						TargetGroups: []elbtypes.TargetGroupTuple{
-							{
-								TargetGroupArn: aws.String(routingTrafficCfg[0].TargetGroupArn),
-								Weight:         aws.Int32(int32(routingTrafficCfg[0].Weight)),
-							},
-							{
-								TargetGroupArn: aws.String(routingTrafficCfg[1].TargetGroupArn),
-								Weight:         aws.Int32(int32(routingTrafficCfg[1].Weight)),
+		for _, rule := range describeRulesOutput.Rules {
+			var modifiedActions []elbtypes.Action
```
khanhtc1202, line 449:
```suggestion
			modifiedActions := make([]elbtypes.Action, 0, len(rule.Actions))
```

khanhtc1202:
LGTM with some nits 👍

t-kikuc:
@khanhtc1202 
Thank you for reviewing, I fixed two

khanhtc1202 (dismissed):
Great improvement 👍

khanhtc1202:
@ffjlabo plz check 👀👀

pkg/app/piped/platformprovider/ecs/routing_traffic.go
```diff
@@ -14,9 +14,32 @@
...
+	cMap := make(map[string]bool)
+	for _, item := range c {
+		cMap[item.TargetGroupArn] = true
+	}
+
+	for _, target := range forwardActionTargets {
+		if !cMap[*target.TargetGroupArn] {
+			return false
+		}
+	}
```
ffjlabo, line 42: [nit] It would be nice to use `struct{}` empty struct :)
This is more efficient because it does not allocate memory.

```suggestion
	cMap := make(map[string]struct{})
	for _, item := range c {
		cMap[item.TargetGroupArn] = struct{}{}
	}

	for _, target := range forwardActionTargets {
		if _, ok := cMap[*target.TargetGroupArn]; !ok {
			return false
		}
	}
```

pkg/app/piped/platformprovider/ecs/routing_traffic_test.go
```diff
@@ -0,0 +1,134 @@
...
+					TargetGroupArn: aws.String("arn:aws:elasticloadbalancing:<region>:<account-id>:targetgroup/xxx/yyy3"),
+					Weight:         aws.Int32(100),
+				},
+			},
+			expected: false,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
t-kikuc: @ffjlabo Thank you so much! I fixed

pkg/app/piped/platformprovider/ecs/routing_traffic.go
t-kikuc: @ffjlabo Thank you so much! I fixed

ffjlabo (dismissed):
LGTM 🚀

pkg/app/piped/platformprovider/ecs/routing_traffic.go
t-kikuc: @ffjlabo 
I fixed again to use `tc:=tc` as you suggested instead of `tc:=testcase`
because I found an article that recommends to use the same name.

https://go.dev/wiki/CommonMistakes

ffjlabo (approved):

khanhtc1202 (approved):
🚀

t-kikuc:
I'll add the below diagrams to the doc later.

1.  Previously, only the default rule was controlled during deployments:

    case-1

2. Now piped can also control the rules **other than the default**: 

    case-2

3. Besides, piped controls all rules that have the common target groups: 

    case-3

PR #4737, opened 2023-12-26
[ECS] Refactored from json.RawMessage to ECSTargetGroup struct

t-kikuc:

Internal refactoring.

For `pipectl init` command and ECS redesigning, it's more convenient to have struct than json.RawMessage.

no

**Does this PR introduce a user-facing change?**:  no

### DIFF
--- a/pkg/app/piped/platformprovider/ecs/target_groups.go
+++ b/pkg/app/piped/platformprovider/ecs/target_groups.go
@@ -15,11 +15,9 @@
 package ecs
 
 import (
-	"bytes"
-	"encoding/json"
 	"errors"
-	"fmt"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 
 	"github.com/pipe-cd/pipecd/pkg/config"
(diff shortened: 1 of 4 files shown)


khanhtc1202:
I don't have a strong opinion about this change. The current implementation uses json.RawMessage is due to the fact that we would keep it changeable. But since we're redesigning the whole ECS configuration, maybe changing this doesn't make too much sense 🤔 wdyt? @t-kikuc cc @ffjlabo

ffjlabo:
@khanhtc1202 
IMO, I agree to refactor it.
It looks like it is for the implementation of `pipectl init`. Maybe a concrete type is needed when creating config file, right? @t-kikuc

t-kikuc:
@khanhtc1202 

> It looks like it is for the implementation of pipectl init. Maybe a concrete type is needed when creating config file, right?

Yes!  Thanks @ffjlabo 

We can decide the `ECSTargetGroupObject`'s fields as the doc below, and we can modify if needed in the future.
https://pipecd.dev/docs-v0.45.x/user-guide/configuration-reference/#ecstargetgroupobject

Besides, I would not modify the `ECSTargetGroupObject` itself soon for redesigning ECS.

khanhtc1202 (approved):
👍

ffjlabo (approved):
🚀

PR #4742, opened 2023-12-26
Replace deprecated io/ioutil package

minhquang4334:
[io/ioutil ](https://pkg.go.dev/io/ioutil) package is deprecated, use corresponding 'os' package instead.

refs: https://pkg.go.dev/io/ioutil#ReadFile

Fixes #


khanhtc1202 (approved):
Nice catch 👍

ffjlabo (approved):
Thank you!

Issue #4643, opened 2023-10-31
Introduce script run stage


ffjlabo:
**What would you like to be added**:
follow: https://github.com/pipe-cd/pipecd/issues/1089

implement stage to execute any command.

- [x] alpha: only for sync https://github.com/pipe-cd/pipecd/pull/4720
- [ ] beta: include thinking for rollback
  - [x] k8s https://github.com/pipe-cd/pipecd/pull/4743
  - [ ] ecs
  - [ ] cloud run
  - [ ] lambda
  - [ ] terraform

**Why is this needed**:

ffjlabo:
TODO
- [ ] just executing commands not considered about rollbacking https://github.com/pipe-cd/pipecd/pull/4603
- [ ] consider about rollback (fix rfc and implement) 
- [ ] consider about managing commands

ffjlabo:
📝 
- SCRIPT_RUN stage can cancel on the Web UI
- commands are executed in the dir for current application

ffjlabo:
refer the implementation
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/executor/customsync/customsync.go#L55

ffjlabo:
📝 Share my current opinion when to rollback :)

**Behavior during rollback of SCRIPT_RUN**

- Execute the command to rollback SCRIPT_RUN to the point where the deployment was canceled or failed.
- When there are multiple SCRIPT_RUN stages to be rolled back, they are executed in the same order as SCRIPT_RUN on the pipeline.

For example, consider when deployment proceeds in the following order from 1 to 7.
```
1. K8S_CANARY_ROLLOUT
2. WAIT
3. SCRIPT_RUN
4. K8S_PRIMARY_ROLLOUT
5. SCRIPT_RUN
6. K8S_CANARY_CLEAN
7. SCRIPT_RUN
```

Then
- If 4 is canceled or fails while running, only SCRIPT_RUN of 3 will be rolled back.
- If 6 is canceled or fails while running, only SCRIPT_RUNs 3 and 5 will be rolled back.

---
**How to realize it**

We can choose 2 ways to realize the rollback.

1. execute command on ROLLBACK stage
2. execute command on another stage e.g. ROLLBACK_SCRIPTRUN

khanhtc1202:
[...]

PR #4743, opened 2023-12-28
Implement rollback for script run as pre defined stage

ffjlabo:

Implement rollback for script run as pre defined stage.

Part of #4643

Users can use `onRollback` feature for `SCRIPT_RUN` stage.
They can execute any command when the deployment failed and the executed `SCRIPT_RUN` stage is included in the pipeline

### DIFF
--- a/pkg/app/piped/controller/scheduler.go
+++ b/pkg/app/piped/controller/scheduler.go
@@ -369,34 +369,37 @@ func (s *scheduler) Run(ctx context.Context) error {
 	// we start rollback stage if the auto-rollback option is true.
 	if deploymentStatus == model.DeploymentStatus_DEPLOYMENT_CANCELLED ||
 		deploymentStatus == model.DeploymentStatus_DEPLOYMENT_FAILURE {
-		if stage, ok := s.deployment.FindRollbackStage(); ok {
+
+		if rollbackStages, ok := s.deployment.FindRollbackStages(); ok {
 			// Update to change deployment status to ROLLING_BACK.
 			if err := s.reportDeploymentStatusChanged(ctx, model.DeploymentStatus_DEPLOYMENT_ROLLING_BACK, statusReason); err != nil {
 				return err
 			}
 
-			// Start running rollback stage.
-			var (
-				sig, handler = executor.NewStopSignal()
-				doneCh       = make(chan struct{})
-			)
-			go func() {
-				rbs := *stage
-				rbs.Requires = []string{lastStage.Id}
-				s.executeStage(sig, rbs, func(in executor.Input) (executor.Executor, bool) {
-					return s.executorRegistry.RollbackExecutor(s.deployment.Kind, in)
-				})
-				close(doneCh)
-			}()
+			for _, stage := range rollbackStages {
+				// Start running rollback stage.
+				var (
+					sig, handler = executor.NewStopSignal()
+					doneCh       = make(chan struct{})
+				)
+				go func() {
+					rbs := *stage
+					rbs.Requires = []string{lastStage.Id}
+					s.executeStage(sig, rbs, func(in executor.Input) (executor.Executor, bool) {
+						return s.executorRegistry.RollbackExecutor(s.deployment.Kind, in)
+					})
+					close(doneCh)
+				}()
 
-			select {
-			case <-ctx.Done():
-				handler.Terminate()
-				<-doneCh
-				return nil
+				select {
+				case <-ctx.Done():
+					handler.Terminate()
+					<-doneCh
+					return nil
 
-			case <-doneCh:
-				break
+				case <-doneCh:
+					break
+				}
 			}
 		}
 	}
--- a/pkg/app/piped/controller/scheduler.go
+++ b/pkg/app/piped/controller/scheduler.go
@@ -436,6 +436,24 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
 		lp.Complete(time.Minute)
 	}()
 
+	// Check whether to execute the script rollback stage or not.
+	// If the base stage is executed, the script rollback stage will be executed.
+	if ps.Name == model.StageScriptRunRollback.String() {
+		baseStageID := ps.Metadata["baseStageID"]
+		if baseStageID == "" {
+			return
+		}
+
+		baseStageStatus, ok := s.stageStatuses[baseStageID]
+		if !ok {
+			return
+		}
+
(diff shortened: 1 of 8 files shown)


pkg/app/piped/controller/scheduler.go
```diff
@@ -369,34 +369,37 @@ func (s *scheduler) Run(ctx context.Context) error {
...
+
+				select {
+				case <-ctx.Done():
+					handler.Terminate()
+					<-doneCh
+					return nil
+
+				case <-doneCh:
+					break
+				}
```
ffjlabo, line 402: 📝 Changed to execute multiple kinds of rollback stages.
The changes are to do logic just like the same as before with each rollback stage.

pkg/app/piped/planner/kubernetes/pipeline.go
```diff
@@ -114,6 +115,31 @@ func buildProgressivePipeline(pp *config.DeploymentPipeline, autoRollback bool,
...
+		// Add a stage for rolling back script run stages.
+		for i, s := range pp.Stages {
+			if s.Name == model.StageScriptRun {
+				// Use metadata as a way to pass parameters to the stage.
+				envStr, _ := json.Marshal(s.ScriptRunStageOptions.Env)
+				metadata := map[string]string{
+					"baseStageID": out[i].Id,
+					"onRollback":  s.ScriptRunStageOptions.OnRollback,
+					"env":         string(envStr),
+				}
```
ffjlabo, line 128: 📝 I use metadata for the stage to recognize below on executing it.
- The commands which are executed on the rollback script run stage. (onRollback, env)
- The stage IDs of script run stages, which are used to check whether to rollback it. (baseStageID)
    - It is because I want to execute the command to rollback SCRIPT_RUN to the point where the deployment was canceled or failed.

pkg/app/piped/controller/scheduler.go
```diff
@@ -433,6 +436,24 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
...
+
+		baseStageStatus, ok := s.stageStatuses[baseStageID]
+		if !ok {
+			return
+		}
+
+		if baseStageStatus == model.StageStatus_STAGE_NOT_STARTED_YET || baseStageStatus == model.StageStatus_STAGE_SKIPPED {
+			return
+		}
+	}
```
ffjlabo, line 455: 📝 Check the status of the script run stage to decide whether to rollback or not.
It is because the target stages should be already executed.

pkg/app/piped/executor/kubernetes/rollback.go
```diff
@@ -39,7 +44,8 @@ func (e *rollbackExecutor) Execute(sig executor.StopSignal) model.StageStatus {
 	switch model.Stage(e.Stage.Name) {
 	case model.StageRollback:
 		status = e.ensureRollback(ctx)
-
+	case model.StageScriptRunRollback:
+		status = e.ensureScriptRunRollback(ctx)
```
ffjlabo, line 48: 📝 Added the logic to rollback executor for each application (firstly, I added on the k8s's one) because rollback stages are executed per the kind of application not per stage.

ffjlabo:
**The PR summary**
- Add ROLLBACK_SCRIPT_RUN stage and enable to plan[be09a10](https://github.com/pipe-cd/pipecd/pull/4743/commits/be09a102860c7612bd2a76393a317b2acc00828e)
- Enable to execute multiple rollback stages [48be28e](https://github.com/pipe-cd/pipecd/pull/4743/commits/48be28e4e2f78b00861f2c6e53fc7f868745bab4)
- Add script run rollback logic to k8s app [0d0ecee](https://github.com/pipe-cd/pipecd/pull/4743/commits/0d0ecee2eadf495b17dcbb4ba2e538414b5fe2b8)

**some problems**
I don't have any idea to separate the rollback for application and the rollback for the stage...

pkg/app/piped/executor/kubernetes/rollback.go
```diff
@@ -171,3 +179,45 @@ func (e *rollbackExecutor) ensureRollback(ctx context.Context) model.StageStatus
 	}
 	return model.StageStatus_STAGE_SUCCESS
 }
+
+func (e *rollbackExecutor) ensureScriptRunRollback(ctx context.Context) model.StageStatus {
+	e.LogPersister.Infof("Runnnig commands for rollback...")
```
khanhtc1202, line 184:
```suggestion
	e.LogPersister.Info("Runnnig commands for rollback...")
```

pkg/app/piped/executor/kubernetes/rollback.go
```diff
@@ -171,3 +179,45 @@ func (e *rollbackExecutor) ensureRollback(ctx context.Context) model.StageStatus
...
+
+	onRollback, ok := e.Stage.Metadata["onRollback"]
+	if !ok {
+		e.LogPersister.Error("onRollback metadata is missing")
+		return model.StageStatus_STAGE_FAILURE
+	}
```
khanhtc1202, line 190: I just want to ensure I understand this right: This means that if there is no user-defined `onRollback` script, this script run rollback will return an error, right?

khanhtc1202 (approved):
I'm okay with this approach, let's rethink about making this script rollback kind independent later when we introduce app kind plugin pattern

pkg/app/piped/executor/kubernetes/rollback.go
```diff
@@ -171,3 +179,45 @@ func (e *rollbackExecutor) ensureRollback(ctx context.Context) model.StageStatus
...
+
+	if onRollback == "" {
+		e.LogPersister.Info("No commands to run")
+		return model.StageStatus_STAGE_SUCCESS
+	}
```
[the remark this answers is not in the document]
khanhtc1202: @ffjlabo What I mean here is that we may make this:
1. Add rollbackable of type bool to specify whether the current script run can be rollbacked or not, default set to false 🤔 
2. In case of rollbackable stage, if `onRollback` is not set, re-run script instead or return error rollback script is missing
wdyt?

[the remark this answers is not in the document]
khanhtc1202: On the second thought, I got your idea.
- If users did not define `onRollback` script, then there is nothing to run
- If users defined the `onRollback` script, then run that script
Let's keep this idea, thanks 👍

khanhtc1202 (approved):
Let's go 🚀

khanhtc1202:
@t-kikuc Please check 🙏

t-kikuc (approved):
Go 🚀

PR #4748, opened 2024-01-11
Fix copyright year

t-kikuc:


t-kikuc:
> Note: File service.pb.auth.go, which is generated by our protoc-gen-auth needs to be regenerated later using the updated tool/codegen image published after this PR is merged in order to pass the action/code check (I know it's quite meta 😂), will do it later by separated PR.

PR #4752, opened 2024-01-18
feat: exposeService detail logs

Forget-C:
before run `pipectl quickstart `
WX20240118-110004@2x
Due to network reasons, `exposeService` will block the installation process for a long time and eventually time out. During this period, the user does not know whether the program is running and what the program is waiting for.
after run `pipectl quickstart `
image
Added the "what to wait for" log of the program and the status scrolling prompt of rollout. Subsequent logs and prompts have not changed.

Fixes #


Forget-C:
> Hi @Forget-C 👋 Thank you for the nice contribution! plz signoff for commits with `git commit -s` 🙏 ref: https://github.com/pipe-cd/pipecd/blob/master/CONTRIBUTING.md#contributor-license-agreement

I have modified the submission

ffjlabo (approved):
LGTM 🚀

Forget-C:
@ffjlabo Because codecov failed, please review the new testing commit

pkg/app/pipectl/cmd/quickstart/quickstart.go
```diff
@@ -29,11 +29,10 @@ import (
 	"time"
 
 	"github.com/manifoldco/promptui"
-	"github.com/spf13/cobra"
-
 	"github.com/pipe-cd/pipecd/pkg/backoff"
 	"github.com/pipe-cd/pipecd/pkg/cli"
 	"github.com/pipe-cd/pipecd/pkg/version"
+	"github.com/spf13/cobra"
```
ffjlabo, line 35: [nit] For convention, should add this line to the above block 🙏 

The import lines are ordered like this ↓
1. standard library
2. Third-party library
3. our library (packages from this repo)

pkg/cache/memorycache/ttl_cache_test.go
```diff
@@ -19,22 +19,36 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pipe-cd/pipecd/pkg/cache"
```
ffjlabo, line 22: [nit] Reorder the line 🙏 
> The import lines are ordered like this ↓
> 1. standard library
> 2. Third-party library
> 3. our library (packages from this repo)

pkg/cache/memorycache/ttl_cache_test.go
```diff
@@ -19,22 +19,36 @@ import (
...
+		err := c.Put("key-1", "value-1")
+		require.NoError(t, err)
+		cancel()
+		err = c.Put("key-2", "value-2")
+		require.NoError(t, err)
+		<-time.After(6 * time.Second)
+		value, err := c.Get("key-2")
+		require.NoError(t, err)
+		assert.Equal(t, "value-2", value)
+	})
```
ffjlabo, line 40: Thank you for adding a test case :) 

[imo] It would be better to describe what is the behavior of the test case. 👀 
Also, it would be enough just to add one key-value pair and check it. :)

```suggestion
	t.Run("remain a value after context canceled", func(t *testing.T) {
		ctx, cancel := context.WithCancel(context.Background())
		c := NewTTLCache(ctx, 0, 5*time.Second)
		err := c.Put("key-1", "value-1")
		require.NoError(t, err)
		cancel()
		<-time.After(6 * time.Second)
		value, err := c.Get("key-1")
		require.NoError(t, err)
		assert.Equal(t, "value-1", value)
	})
```

ffjlabo (commented):
@Forget-C Thank you for adding a test case 🙏 I added some nit comments.
I will approve after these changes :)

Forget-C:
> Great contribution! Thank you 🚀 CodeCov failed, but it looks like this is not related to the changes.

Because changes in code style lead to changes in coverage, there is no actual impact

khanhtc1202 (approved):
Great, thank you 🚀

PR #4785, opened 2024-02-13
Support AWS Secrets Manager as a source of piped-config

t-kikuc:

Enabled to use a secret in AWS Secrets Manager as the piped-config.

usage:  
- `launcher --config-from-aws-secret=true --aws-secret-id=arn:aws:secretsmanager:ap-xxx:xxx:secret:my-piped-config`

- `piped --config-aws-secret=arn:aws:secretsmanager:xxx:xxx:secret:my-piped-config`

This enables uses to run a piped on App Runner, with storing the piped-config securely in AWS Secrets Manager.

Related to https://github.com/pipe-cd/pipecd/discussions/4765

### DIFF
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -578,8 +578,9 @@ func (p *piped) createAPIClient(ctx context.Context, address, projectID, pipedID
 
 // loadConfig reads the Piped configuration data from the specified source.
 func (p *piped) loadConfig(ctx context.Context) (*config.PipedSpec, error) {
-	if p.configFile != "" && p.configGCPSecret != "" && p.configAWSSecret != "" {
-		return nil, fmt.Errorf("only one of config-file, config-gcp-secret or config-aws-secret could be set")
+	// HACK: When the version of cobra is updated to >=v1.8.0, this should be replaced with https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsOneRequired.
+	if err := p.hasTooManyConfigFlags(); err != nil {
+		return nil, err
 	}
 
 	extract := func(cfg *config.Config) (*config.PipedSpec, error) {
@@ -907,4 +908,17 @@ func stopCommandHandler(ctx context.Context, cmdLister commandstore.Lister, logg
 	}
 
 	return true, nil
+}
+
+func (p *piped) hasTooManyConfigFlags() error {
+	cnt := 0
+	for _, v := range []string{p.configFile, p.configGCPSecret, p.configAWSSecret} {
+		if v != "" {
+			cnt++
+		}
+	}
+	if cnt > 1 {
+		return fmt.Errorf("only one of config-file, config-gcp-secret or config-aws-secret could be set")
+	}
+	return nil
 }
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -30,6 +30,8 @@ import (
 	"time"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	awssecretsmanager "github.com/aws/aws-sdk-go-v2/service/secretsmanager"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/spf13/cobra"
@@ -86,6 +88,7 @@ type piped struct {
 	configFile      string
 	configData      string
 	configGCPSecret string
+	configAWSSecret string
 
 	insecure                             bool
 	certFile                             string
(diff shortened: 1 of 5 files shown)


pkg/app/piped/cmd/piped/piped.go
```diff
@@ -574,8 +578,8 @@ func (p *piped) createAPIClient(ctx context.Context, address, projectID, pipedID
 
 // loadConfig reads the Piped configuration data from the specified source.
 func (p *piped) loadConfig(ctx context.Context) (*config.PipedSpec, error) {
-	if p.configFile != "" && p.configGCPSecret != "" {
-		return nil, fmt.Errorf("only config-file or config-gcp-secret could be set")
+	if p.configFile != "" && p.configGCPSecret != "" && p.configAWSSecret != "" {
+		return nil, fmt.Errorf("only one of config-file, config-gcp-secret or config-aws-secret could be set")
```
ffjlabo, line 582: [ask] This if statement is true only when all of `p.configFile`, `p.configGCPSecret`, and `p.configAWSSecret` exist. 👀 
How is the behavior when the `p.configFile` and `p.configGCPSecret` exist?

ffjlabo (commented):
💬

pkg/app/piped/cmd/piped/piped.go
t-kikuc: Thanks, my impl was wrong!
When two of the flags are set, it should return err.
I'll fix.

t-kikuc: @ffjlabo I fixed!

[IMO] cobra should be updated in the future for useful options.

pkg/app/piped/cmd/piped/piped_test.go
```diff
@@ -0,0 +1,68 @@
...
+		{
+			title: "two configs are set",
+			p: &piped{
+				configFile:      "config.yaml",
+				configGCPSecret: "xxx",
+				configAWSSecret: "",
+			},
+			expectErr: true,
+		},
+	}
```
ffjlabo, line 58: [nit] add a case for all of them are set 👀

pkg/app/piped/cmd/piped/piped.go
```diff
@@ -574,8 +578,9 @@ func (p *piped) createAPIClient(ctx context.Context, address, projectID, pipedID
 
 // loadConfig reads the Piped configuration data from the specified source.
 func (p *piped) loadConfig(ctx context.Context) (*config.PipedSpec, error) {
-	if p.configFile != "" && p.configGCPSecret != "" {
-		return nil, fmt.Errorf("only config-file or config-gcp-secret could be set")
+	// HACK: When the version of cobra is updated to >=v1.8.0, this should be replaced with https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsOneRequired.
+	if err := p.hasTooManyConfigFlags(); err != nil {
```
ffjlabo, line 582: I agree with updating cobra's version 👍 

Btw, the function might not work as you expected 👀 
I think you expected to check whether there is exactly one flag with `MarkFlagsOneRequired`, right?

https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsOneRequired

> MarkFlagsOneRequired marks the given flags with annotations so that Cobra errors if the command is invoked without at least one flag from the given set of flags.

This means that `MarkFlagsOneRequired` returns like below when the given flags are {a, b, c}.
- {}: false
- {a}: true
- {a b}: true
- {a b c}: true
Also, `!MarkFlagsOneRequired()` means all of them are not set 👀

khanhtc1202: Could you open an issue for updating cobra version later?

pkg/app/piped/cmd/piped/piped_test.go
t-kikuc: @ffjlabo Thanks!! I added a case of three.

pkg/app/piped/cmd/piped/piped.go
t-kikuc: @ffjlabo 
I'm really sorry.
The correct func is [`MarkFlagsMutuallyExclusive`](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive).

So I fixed [the comment](https://github.com/pipe-cd/pipecd/pull/4785/files#diff-a6436ea98eb0426cf2f50c52d74885ac49f3c87c22f0d2a41c3c1a8975d661ccR581) to below:

> // HACK: When the version of cobra is updated to >=v1.8.0, this should be replaced with https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive.

t-kikuc: @khanhtc1202 
Yeah, I opened an issue https://github.com/pipe-cd/pipecd/issues/4788 now.

ffjlabo: Wow, that's good!
I didn't know the method 👀

ffjlabo (approved):
LGTM 🚀

khanhtc1202 (approved):
Let's go 🚀

t-kikuc:
@ffjlabo @khanhtc1202 
Thank you so much for your early and great reviews 🙏

PR #4804, opened 2024-03-05
Fix the Lambda encryptedSecrets key used in the example

ouchi2501:
- as Title
- Specified in function.yaml is .encryptedSecrets.key
  - https://github.com/pipe-cd/examples/blob/master/lambda/secret-management/function.yaml#L10

Fixes #


khanhtc1202 (approved):
Nice catch ⚾

t-kikuc (approved):
Great, thanks!

PR #4805, opened 2024-03-05
Support deploying Lambda functions depending on Lambda layers

### LINKED ISSUE #4803: Support deploying Lambda functions depending on Lambda layers

- Enable deploying AWS Lambda functions which depend on Lambda layers.

**Why is this needed**:

- Currently, PipeCD cannot deploy functions which use Lambda layers because the input of `create-function` does not have `Layers` attribute as below:
https://github.com/pipe-cd/pipecd/blob/9a298a8a7d2dcc4071ec2f63ab7c9d2fb1156ae5/pkg/app/piped/platformprovider/lambda/client.go#L111-L121

- However, some users need to deploy functions which depend on layers in order to share packages.

t-kikuc:

Enabled to deploy Lambda functions which use layers.

The config is like below:

function.yaml
```yaml
kind: LambdaFunction
spec:
  ...
  layers: # NEW ATTRIBUTE.  not necessary.
    - arn:aws:lambda:REGION:ACCOUNT_ID:layer:layer-A:3
    - arn:aws:lambda:REGION:ACCOUNT_ID:layer:layer-B:1
```

Fixes #4803 

**Does this PR introduce a user-facing change?**:  yes (addition only)

- **How are users affected by this change**:  add attributes if needed

### DIFF
--- a/pkg/app/piped/platformprovider/lambda/client.go
+++ b/pkg/app/piped/platformprovider/lambda/client.go
@@ -529,6 +529,30 @@ func (c *client) untagFunction(ctx context.Context, functionArn string, tags map
 	return nil
 }
 
+func (c *client) publishLayer(ctx context.Context, in *lambda.PublishLayerVersionInput) error {
+	// in := &lambda.PublishLayerVersionInput{
+	// 	Content: &types.LayerVersionContentInput{
+	// 		S3Bucket:        aws.String(""),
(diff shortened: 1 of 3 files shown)


khanhtc1202 (approved):
Nice improvement 👍

ffjlabo (approved):
Sorry to be late 🙏

PR #4826, opened 2024-03-13
Relocate plugin proto

khanhtc1202:

Fixes #


pkg/plugin/platform/api/api.proto
```diff
@@ -14,8 +14,8 @@
 
 syntax = "proto3";
 
-package grpc.pluggin.applicationkind;
-option go_package = "github.com/pipe-cd/pipecd/pkg/app/pipedv1/pluggin/applicationkind/api";
+package grpc.plugin.platformapi;
+option go_package = "github.com/pipe-cd/pipecd/pkg/plugin/platform/platformapi";
```
ffjlabo, line 18:
```suggestion
option go_package = "github.com/pipe-cd/pipecd/pkg/plugin/platform/api";
```

ffjlabo (commented):
commented 👀

t-kikuc (dismissed):

pkg/plugin/platform/api/api.proto
ffjlabo: @khanhtc1202 
I got your point. So how about this? `/api/platform` and `/api/stage` like googleapis.
https://github.com/googleapis/googleapis/tree/master/google/api/serviceusage/v1

We can avoid using the import aliases. Also, we can do versioning as `/api/platform/v1` when we fix the plugin API.

khanhtc1202: Nice point, let's me adopt this 👍

ffjlabo: Ah, also we might consider about the plugin versioning like `/plugin/v1alpha1/api` 👀
WDYT?

ffjlabo: ref: k8s versioning https://kubernetes.io/docs/reference/using-api/#api-versioning

khanhtc1202:
Updated, please re-approve @ffjlabo @t-kikuc

t-kikuc (approved):

pkg/plugin/platform/api/api.proto
ffjlabo: 📝 talked outside github: We decided to use `/plugin/api/v1alpha1/platform` to do versioning only the package exposed for the users.

ffjlabo (approved):

PR #4836, opened 2024-03-17
Update the deprecated func for google.golang.org/grpc package

### LINKED ISSUE #4824: Update the deprecated func for `google.golang.org/grpc` package

ref:
- https://github.com/pipe-cd/pipecd/blob/master/pkg/rpc/rpcclient/option.go#L70-L74
- https://github.com/pipe-cd/pipecd/blob/master/pkg/rpc/rpcclient/option.go#L52-L56

**Why is this needed**:

Exzrgs:
Change deprecated functions for google.golang.org/grpc package to recommended ones

Fixes #4824

No

### DIFF
--- a/pkg/rpc/rpcclient/option.go
+++ b/pkg/rpc/rpcclient/option.go
@@ -19,6 +19,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 )
 
 type option struct {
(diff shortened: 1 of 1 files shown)


khanhtc1202:
@Exzrgs, thanks for your contribution 👍 Just want to mention that there are maybe other APIs that are marked as deprecated by `google.golang.org/grpc`. So, if possible, could you also check other APIs as well. It's okay to address that by separated PRs, in such case, please change the PR description to `Part of #4824` in order to keep the issue open. 😄

Exzrgs:
I searched GitHub of PipeCD for everything deprecated in the `google.golang.org/grpc` documentation, but it didn't seem to be used other than the ones I fixed👍
If I have misunderstood something, please let me know🙏

khanhtc1202:
Cool, I just want to make sure all other APIs from the package are checked. Thanks for investigation 🙏

khanhtc1202 (approved):
Thank you 👍

PR #4855, opened 2024-04-01
Fix unable to use SecretEncryption and Attachment features at the same time

khanhtc1202:

In this PR, I reimplemented the sourceprosser logic by introducing the `SourceProcessor` interface.

Previously, we called two functions, `DecryptSecret` and `Attach`, to perform templating data separately. Both of them use golang `text/template` functions, which causes conflict on key not found whenever using both `{{ .attachment }}` and `{{ .secretEncrytion }}` for those two features at the same time.

By introducing the SourceProcessor interface, updated the logic to call golang `text/template` function once. Both SecretDecryption and Attachment functions would only focus on building the templating data, and then `SourceProcesser.Process` would perform the templating task.

Fixes #

### DIFF
--- a/pkg/app/piped/deploysource/deploysource.go
+++ b/pkg/app/piped/deploysource/deploysource.go
@@ -188,7 +188,7 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
 			fmt.Fprintf(lw, "Unable to process the source files (%v)\n", err)
 			return nil, err
 		}
-		fmt.Fprintln(lw, "Successfully process the source files")
+		fmt.Fprintln(lw, "Successfully processed the source files")
 	}
 
 	return &DeploySource{
--- a/pkg/app/piped/deploysource/deploysource.go
+++ b/pkg/app/piped/deploysource/deploysource.go
@@ -171,21 +171,24 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
 	}
 	fmt.Fprintln(lw, "Successfully loaded the application configuration file")
 
+	var templProcessors []sourceprocesser.SourceTemplateProcessor
 	// Decrypt the sealed secrets if needed.
-	if gac.Encryption != nil && p.secretDecrypter != nil && len(gac.Encryption.DecryptionTargets) > 0 {
-		if err := sourceprocesser.DecryptSecrets(appDir, *gac.Encryption, p.secretDecrypter); err != nil {
-			fmt.Fprintf(lw, "Unable to decrypt the secrets (%v)\n", err)
-			return nil, err
-		}
-		fmt.Fprintf(lw, "Successfully decrypted secrets: %v\n", gac.Encryption.DecryptionTargets)
+	if gac.Encryption != nil && p.secretDecrypter != nil {
+		templProcessors = append(templProcessors, sourceprocesser.NewSecretDecrypterProcessor(gac.Encryption, p.secretDecrypter))
+	}
+	// Attach the data if needed.
+	if gac.Attachment != nil {
+		templProcessors = append(templProcessors, sourceprocesser.NewAttachmentProcessor(gac.Attachment))
 	}
 
-	if gac.Attachment != nil && len(gac.Attachment.Targets) > 0 {
-		if err := sourceprocesser.AttachData(appDir, *gac.Attachment); err != nil {
-			fmt.Fprintf(lw, "Unable to attach the data (%v)\n", err)
+	// Process templating source files.
+	if len(templProcessors) > 0 {
+		sp := sourceprocesser.NewSourceProcessor(appDir, templProcessors...)
+		if err := sp.Process(); err != nil {
+			fmt.Fprintf(lw, "Unable to process the source files (%v)\n", err)
 			return nil, err
 		}
-		fmt.Fprintf(lw, "Successfully attached data: %v\n", gac.Attachment.Targets)
+		fmt.Fprintln(lw, "Successfully process the source files")
 	}
 
 	return &DeploySource{
--- a/pkg/app/piped/sourceprocesser/attacher.go
+++ b/pkg/app/piped/sourceprocesser/attacher.go
@@ -18,6 +18,9 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
 
 	"github.com/pipe-cd/pipecd/pkg/config"
 )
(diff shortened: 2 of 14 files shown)


khanhtc1202:
Note: All references from `pipedv1` are pointed to `piped/sourceprosser` package in this PR only to ensure the pipedv1 (which is under development) could work as same as piped in the mean time. I will remove `pipedv1/sourceprocesser` later.

khanhtc1202:
> This feature can be used in combo with PipeCD [SecretManagement feature](https://pipecd.dev/docs-v0.46.x/user-guide/managing-application/secret-management). You can encrypt your secret data using PipeCD secret encryption function, it will be decrypted and placed in your configuration files; then the PipeCD attachment feature will attach that decrypted configuration to the manifest of resource, which requires that configuration.

@ffjlabo Your right, I missed this part for the current implementation. If we only templating once, then there is no way that the source processor can perform something like this 😢 Need to think another way to handle this case 👍

khanhtc1202:
> This might be the current expectation ? 🤔
> 
> https://pipecd.dev/docs-v0.46.x/user-guide/managing-application/manifest-attachment/
> 
> > Tip:
> > This feature can be used in combo with PipeCD [SecretManagement feature](https://pipecd.dev/docs-v0.46.x/user-guide/managing-application/secret-management). You can encrypt your secret data using PipeCD secret encryption function, it will be decrypted and placed in your configuration files; then the PipeCD attachment feature will attach that decrypted configuration to the manifest of resource, which requires that configuration.

@ffjlabo  To enable the logic to adopt both use-cases
- 2 features being used in the same file (files contain both attachment and encryption keys)
- 2 features being used in a chain (perform decryption first, then attach the decrypted data to files)
This is my approach
- Initially build the templating data for "all" processors (we make that to avoid key not found errors on perform descriptions with files containing attachments, and vice versa)
- Force rebuild templating data before performing processor data templating to ensure it will template fresh/latest data (which could be updated after previous processor execution)

ref: https://github.com/pipe-cd/pipecd/pull/4855/files#diff-db7074f59efd9faf6eb5bed299a908d916999b8657186aa506e1965936c3869eR51-R80

ffjlabo:
@khanhtc1202 
Thank you for trying another pattern :)  This is a tough problem 🤔 

[ASK]
> Initially build the templating data for "all" processors (we make that to avoid key not found errors on perform descriptions with files containing attachments, and vice versa)

This pattern might not be able to cover the pattern in which the secret management templates are in the attachment source file. 🤔 
This pattern is shown in the example below↓
https://github.com/pipe-cd/examples/tree/master/ecs/attachment

Could you add a test case and try it?

t-kikuc (commented):
After fixing codes, we need to copy the changes into `pkg/app/pipedv1/sourceprocesser/*` too.

pkg/app/piped/sourceprocesser/sourceprocesser.go
```diff
@@ -0,0 +1,83 @@
...
+package sourceprocesser
+
+import (
+	"fmt"
+)
+
+type SourceTemplateProcessor interface {
+	// BuildTemplateData returns the data that will be used to template target files
+	BuildTemplateData(appDir string) (map[string]string, error)
+	// TemplateSource performs the templating prepared data to the source files
```
t-kikuc, line 24: You mean this?
```suggestion
	// TemplateSource applies prepared template data to the source files
```

t-kikuc (commented):
tough issue, tough PR...

I commented two nits.

khanhtc1202:
> After fixing codes, we also need to copy the changes into `pkg/app/pipedv1/sourceprocesser/*`.

I got your point but let me ignore this in this PR, if you look closely to the v1 package source, all refer to the non-v1 package right now. I will update this later 🙏

khanhtc1202:
> @khanhtc1202 Thank you for trying another pattern :) This is a tough problem 🤔
> 
> [ASK]
> 
> > Initially build the templating data for "all" processors (we make that to avoid key not found errors on perform descriptions with files containing attachments, and vice versa)
> 
> This pattern might not be able to cover the pattern in which the secret management templates are in the attachment source file. 🤔 This pattern is shown in the example below↓ https://github.com/pipe-cd/examples/tree/master/ecs/attachment
> 
> Could you add a test case and try it?

@ffjlabo Did you mean this?
https://github.com/pipe-cd/pipecd/pull/4855/files#diff-8600c6fb95c75a5ce5151428a80c506c5e32a1fc65cf67d71e2818a4a70bfb41R91-R116

I think I added that already 👀

pkg/app/piped/deploysource/deploysource.go
```diff
@@ -171,21 +171,24 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
...
-			fmt.Fprintf(lw, "Unable to attach the data (%v)\n", err)
+	// Process templating source files.
+	if len(templProcessors) > 0 {
+		sp := sourceprocesser.NewSourceProcessor(appDir, templProcessors...)
+		if err := sp.Process(); err != nil {
+			fmt.Fprintf(lw, "Unable to process the source files (%v)\n", err)
 			return nil, err
 		}
-		fmt.Fprintf(lw, "Successfully attached data: %v\n", gac.Attachment.Targets)
+		fmt.Fprintln(lw, "Successfully process the source files")
```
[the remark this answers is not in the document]
khanhtc1202: Thanks, addressed by https://github.com/pipe-cd/pipecd/pull/4855/commits/25acbb46d0b44cd51d48a6d659356d91ac98a528

pkg/app/piped/sourceprocesser/sourceprocesser.go
khanhtc1202: Thanks, addressed by https://github.com/pipe-cd/pipecd/pull/4855/commits/25acbb46d0b44cd51d48a6d659356d91ac98a528

ffjlabo:
Sorry, my branch checkouted in local is old 🙏 I'll recheck soon

pkg/app/piped/sourceprocesser/attacher.go
```diff
@@ -25,53 +25,61 @@ import (
...
 		src[k] = string(buff)
 	}
-	data := map[string](map[string]string){
-		"attachment": src,
-	}
+	return src, nil
+}
+
+func (p *attachmentProcessor) TemplateKey() string {
+	return "attachment"
```
ffjlabo, line 57: [IMO] define it as const value 👀

pkg/app/piped/sourceprocesser/sourceprocesser.go
```diff
@@ -0,0 +1,83 @@
...
+		data[tp.TemplateKey()] = pdata
+
+		// Perform the templating.
+		if err := tp.TemplateSource(p.appDir, data); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ffjlabo, line 83: Got it! For my understanding, wrote my understanding of the code  (sorry if duplicate former your comment 🙏 )

At first, create values for templating. If the data includes the partial template, treat it as a value to avoid the parse error like "map has no entry for key".

Then iterate in the order below
- embed values to the templates
- then rebuild values to get up-to-date data after previous template processing

ffjlabo: [nit]
it would be better to execute `tp.TemplateSource` first, then execute `tp.BuildTemplateData` to understand it easily 👀

ffjlabo: [nit] Refer to the background with comments because this is tricky. 👀 

example↓
- We want to multiple processors step by step, but currently, text/template can't realize it by default (occurs the missing key error).
- So then create values for templating. If the data includes the partial template, treat it as a value.

📝 found the go issue about missing key error
ref: https://github.com/golang/go/issues/60073

Also, later it might be better to explain on docs because the behavior is a little different from the default go template's one 👀 (it's ok on another PR)

ffjlabo (commented):
@khanhtc1202 Sorry to be late 🙏  I didn't come up with this idea! nice!
left some nit comment 🙏

pkg/app/piped/sourceprocesser/attacher.go
khanhtc1202: this object function acts as constant for this "object" already, so I don't think we need to make new constant 🌝

pkg/app/piped/sourceprocesser/sourceprocesser.go
khanhtc1202: > [nit]
> 
> it would be better to execute `tp.TemplateSource` first, then execute `tp.BuildTemplateData` to understand it easily 👀 
> 
> 

I don't think so. Build/prepare data then use that data to templating file is more natural to me. wdyt?

khanhtc1202: Also I don't think docs is needed here, this is implementation and tests cover what we mean/expected to be run already. And I can't image how can I write some docs for this too 😅

ffjlabo: > I don't think so. Build/prepare data then use that data to template file is more natural to me. wdyt?

I see, I got your point and agreed 👍

> Also I don't think docs is needed here, this is implementation and tests cover what we mean/expected to be run already. And I can't image how can I write some docs for this too 😅

I see, also agreed! keep in my mind for test first 👍

ffjlabo (approved):
Thank you!

PR #4858, opened 2024-04-05
Fix to get the resource key from the annoation

### LINKED ISSUE #4269: prune cluster scoped resources does not work when specify namespace in app config
**What happened**:
prune cluster scoped resources (such as `clusterrole`) does not work when specify namespace in app config

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

ffjlabo:

Use the resource key from the annotation to store it as a livestate.
This fix is expected to be able to delete the cluster scoped resource.

Fixes #4269

### DIFF
--- a/pkg/app/piped/livestatestore/kubernetes/reflector.go
+++ b/pkg/app/piped/livestatestore/kubernetes/reflector.go
@@ -249,7 +249,11 @@ func (r *reflector) start(_ context.Context) error {
 
 func (r *reflector) onObjectAdd(obj interface{}) {
 	u := obj.(*unstructured.Unstructured)
-	key := provider.MakeResourceKey(u)
+	key := provider.GetResourceKeyFromActualResource(u)
+	if anoKey := u.GetAnnotations()[provider.LabelResourceKey]; anoKey != "" {
+		key, _ = provider.DecodeResourceKey(anoKey)
+
+	}
 
 	// Ignore all predefined ones.
 	if _, ok := ignoreResourceKeys[key.String()]; ok {
@@ -283,7 +287,7 @@ func (r *reflector) onObjectUpdate(oldObj, obj interface{}) {
 	oldU := oldObj.(*unstructured.Unstructured)
 
 	// Ignore all predefined ones.
-	key := provider.MakeResourceKey(u)
+	key := provider.GetResourceKeyFromActualResource(u)
 	if _, ok := ignoreResourceKeys[key.String()]; ok {
 		kubernetesmetrics.IncResourceEventsCounter(
 			kubernetesmetrics.LabelEventUpdate,
@@ -312,7 +316,7 @@ func (r *reflector) onObjectUpdate(oldObj, obj interface{}) {
 
 func (r *reflector) onObjectDelete(obj interface{}) {
 	u := obj.(*unstructured.Unstructured)
-	key := provider.MakeResourceKey(u)
+	key := provider.GetResourceKeyFromActualResource(u)
 
 	// Ignore all predefined ones.
 	if _, ok := ignoreResourceKeys[key.String()]; ok {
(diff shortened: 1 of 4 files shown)


ffjlabo:
I tested it on the quick sync with `ClusterRole.` and work well.

When `spec.input.namespace` is set in the app.pipecd.yaml
- create ClusterRole
- update ClusterRole
- delete ClusterRole

When not set in the app.pipecd.yaml
- create ClusterRole
- update ClusterRole
- delete ClusterRole

ffjlabo:
Also, I tested it on the quick sync with ClusterRole. and worked well.

When spec.input.namespace is set in the app.pipecd.yaml

- ✅ create Deployment
- ✅ update Deployment
- 😇 delete ClusterRole

ffjlabo:
I re-tested the fix on the quick sync with ClusterRole (Cluster scoped resource). and worked well.

**When `spec.input.namespace` is set in the app.pipecd.yaml**
- create ClusterRole
- update ClusterRole
- delete ClusterRole

**`spec.input.namespace` is not set**
- create ClusterRole
- update ClusterRole
- delete ClusterRole

ffjlabo:
Also, I re-tested the fix on the quick sync with Deployment (Namespace scoped resource). and worked well.

**When `spec.input.namespace` is set in the app.pipecd.yaml**
- create Deployment
- update Deployment
- delete Deployment

**`spec.input.namespace` is not set**
- create Deployment
- update Deployment
- delete Deployment

khanhtc1202:
@ffjlabo Nice approach to solve this 👍 Just want to confirm, could you check how this change effected the livestate funtion? I mean, our livestate function currently build the applications resource graph based on its manifests' annotation, so just want to make sure what can be effected or not. You can test that by check the livestate graph in the application detail page of application with cluster resource 👀

ffjlabo:
@khanhtc1202 
Thank you for the comment! I found the strange behavior 👀 
I launched the app, including a ClusterRole and a Deployment. 
Then, The livestate show the 2 Deployment. One should be a Replicaset. 
PipeCD

ffjlabo:
I'll investigate it 👍

ffjlabo:
https://github.com/pipe-cd/pipecd/pull/4858#issuecomment-2046933419
This is because the annotation `pipecd.dev/resource-key` of the Replicaset is for Deployment's one.

The convenience of the Deployment controller causes this.
Currently, the annotations set to Deployment are taken over the Replicaset.
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L235

Also, the resource key is used to build the k8s livestate.
So the application detail page show the livestate of the Replicaset as Deployment.
https://github.com/pipe-cd/pipecd/blob/4436527f569374ea223cbb34654a67eab5ab0a92/pkg/app/piped/platformprovider/kubernetes/state.go#L33C1-L53C23

We need to think how to avoid this 🤔

ffjlabo:
I made some candidates below.

1. avoid using the resource key in the codebase as much as possible
- use the raw k8s obj instead of it in a situation like https://github.com/pipe-cd/pipecd/blob/4436527f569374ea223cbb34654a67eab5ab0a92/pkg/app/piped/platformprovider/kubernetes/state.go#L33C1-L53C23

2. deprecate the annotation `pipecd.dev/resource-key` 
- fix not to use the annotation in the codebase
- Current logic related to the annotation (searched by the code jump)
  - addBuiltinAnnotations https://github.com/pipe-cd/pipecd/blob/fda7cf70d99d79d44fd6eacd7f16008b984fc820/pkg/app/piped/executor/kubernetes/kubernetes.go#L223
  - reflector onObjectAdd https://github.com/pipe-cd/pipecd/blob/fda7cf70d99d79d44fd6eacd7f16008b984fc820/pkg/app/piped/livestatestore/kubernetes/reflector.go#L253
  - applier Delete https://github.com/pipe-cd/pipecd/blob/fda7cf70d99d79d44fd6eacd7f16008b984fc820/pkg/app/piped/platformprovider/kubernetes/applier.go#L161

ffjlabo:
I tried the solution on the another PR https://github.com/pipe-cd/pipecd/pull/4891

PR #4857, opened 2024-04-03
Truncate `deploymentStatus` metrics after reporting stats

t-kikuc:
**What this PR does / why we need it**:

Delete `deploymentStatus` metrics after reporting stats
in order to avoid the error of excess grpc message size by accumulated `deploymentStatus` records.

**Which issue(s) this PR fixes**:

Fixes #4786

**Does this PR introduce a user-facing change?**: no

- **How are users affected by this change**: no
- **Is this breaking change**: no
- **How to migrate (if breaking change)**: no

PR #4869, opened 2024-04-11
Add a caution comment of scrape_interval

t-kikuc:

Add a caution comment against changing `scrape_interval` of Prometheus to be larger than 1m.

Related to https://github.com/pipe-cd/pipecd/pull/4857

### DIFF
--- a/manifests/pipecd/templates/configmap.yaml
+++ b/manifests/pipecd/templates/configmap.yaml
@@ -201,7 +201,7 @@ data:
 
       - job_name: pipecd-ops
         # This scrape_interval must not be larger than the interval of piped's stats-reporting.
-        # ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/statsreporter/reporter.go#L53
+        # ref: https://github.com/pipe-cd/pipecd/blob/a73d0913643ef1da05bc8898cf055c9893f16e0b/pkg/app/piped/statsreporter/reporter.go#L53
         scrape_interval: 1m
         kubernetes_sd_configs:
           - role: endpoints
--- a/manifests/pipecd/templates/configmap.yaml
+++ b/manifests/pipecd/templates/configmap.yaml
@@ -200,7 +200,7 @@ data:
             replacement: $1
 
       - job_name: pipecd-ops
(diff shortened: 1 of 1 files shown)


manifests/pipecd/templates/configmap.yaml
```diff
@@ -200,6 +200,8 @@ data:
             replacement: $1
 
       - job_name: pipecd-ops
+        # This scrape_interval must not be larger than 1m, which is the interval of piped's stats-reporting.
```
khanhtc1202, line 203: Should mention it as not less than piped stat-reporting interval is enough. We may change the interval of that piped stat reporting later

t-kikuc: @khanhtc1202 
Thanks, I agree with you.
I fixed as below:

>    # This scrape_interval must not be larger than the interval of piped's stats-reporting.

manifests/pipecd/templates/configmap.yaml
```diff
@@ -200,6 +200,8 @@ data:
             replacement: $1
 
       - job_name: pipecd-ops
+        # This scrape_interval must not be larger than the interval of piped's stats-reporting.
+        # ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/statsreporter/reporter.go#L53
```
ffjlabo, line 204: [IMO]
The link leads to the code for the latest master. So, there is a possibility of changing the position if someone change the file (e.g. delete or add some lines)

So how about adding the interval description to the docs and referring to it on this line?

ffjlabo: Sure, so just finish the PR and later make docs PR 👍

ffjlabo: > If you worry about the master branch reference, let fix it as snapshot (commit hash) branch reference

@t-kikuc plz change only it 🙏

t-kikuc: @khanhtc1202 @ffjlabo 
thanks, I fixed using a73d0913643ef1da05bc8898cf055c9893f16e0b (v0.47.0 tag's hash)

khanhtc1202 (approved):
Thank you 🚀 🌔

PR #4871, opened 2024-04-11
Enable support for both EncryptedSecrets and AttachmentData in a single file

HoangNguyen689:

## Problem I faced

When I used AttachmentData and EncryptedSecret in the same file (`taskdef.yaml`), I received this error: `map has no entry for key "attachment".` So, I can only use one feature on a single file.

## What is the problem

AttachmentData and EncryptedSecret have similar code handling: Get the value and load it into template target files.
It is handled sequence by sequence: EncryptedSecret first and AttachmentData after. When we use both, the EncryptedSecret handling part would not have any information about the AttachmentData; parsing using the template file raised an error because of missing data.

## What do I do

In the case of AttachmentData and EncryptedSecret, I prepare data for both parts and pass it to the template target files.

## Does this change affect other parts

It keeps the original handling of `DecryptSecrets` and `AttachData`,  it doesn't affect other parts.

## About test

Because there isn't any new logic, test cases in `decrypter` and `attacher` have not been written again. Only both AttachmentData and EncryptedSecret data exist in a single file test is added.

##  Other concerns

I also see the handling of `decrypter` and  `attacher` in `driftdetector` package, but I don't see the occurrence of ECSApp, so I guess it is unnecessary to change that part (at least now)

Fixes #

### DIFF
--- a/pkg/app/piped/deploysource/deploysource.go
+++ b/pkg/app/piped/deploysource/deploysource.go
@@ -28,6 +28,20 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/model"
 )
 
+type EmbeddedType string
+
+const (
+	EmbeddedTypeSecretOnly     EmbeddedType = "EMBEDDED_TYPE_SECRET_ONLY"
+	EmbeddedTypeAttachmentOnly EmbeddedType = "EMBEDDED_TYPE_ATTACHMENT_ONLY"
+	EmbeddedTypeBoth           EmbeddedType = "EMBEDDED_TYPE_BOTH"
+	EmbeddedTypeNone           EmbeddedType = "EMBEDDED_TYPE_NONE"
+)
+
+type embedded struct {
+	secret     bool
+	attachment bool
+}
+
 type DeploySource struct {
 	RepoDir                  string
 	AppDir                   string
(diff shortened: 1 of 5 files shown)


khanhtc1202:
@HoangNguyen689 Thanks for the contribution. About this change, possible affected parts could be
- DeploySource logic
- DriftDetector logic (for Terraform / K8s / CloudRun)
We have to update that all with your new logic for combination.

Screenshot 2024-04-12 at 11 07 36

pkg/app/piped/deploysource/deploysource.go
```diff
@@ -28,6 +28,20 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/model"
 )
 
+type EmbeddedType string
+
+const (
+	EmbeddedTypeSecretOnly     EmbeddedType = "EMBEDDED_TYPE_SECRET_ONLY"
+	EmbeddedTypeAttachmentOnly EmbeddedType = "EMBEDDED_TYPE_ATTACHMENT_ONLY"
+	EmbeddedTypeBoth           EmbeddedType = "EMBEDDED_TYPE_BOTH"
+	EmbeddedTypeNone           EmbeddedType = "EMBEDDED_TYPE_NONE"
```
khanhtc1202, line 37: This should be defined in sourceprocessor package

pkg/app/piped/deploysource/deploysource.go
```diff
@@ -171,16 +185,40 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
...
+		embedded.attachment = true
+	}
+
+	if embedded.secret && embedded.attachment {
+		embeddedType = EmbeddedTypeBoth
+	} else if embedded.secret {
+		embeddedType = EmbeddedTypeSecretOnly
+	} else if embedded.attachment {
+		embeddedType = EmbeddedTypeAttachmentOnly
+	}
```
khanhtc1202, line 207: I understand this approach, but can we have some better for future embedding combinations? I mean, currently, we have attachment and secretEncrypter so that we can have if/else with BOTH flag, but if I add one more processor, this could be too complicated to deal with wdyt?

khanhtc1202: In fact, I have implemented an approach that supports what I mentioned above, but it still needs to be finished. You can check it here 👀
My implementation has a problem with `only run go template once`, so it can't perform encrypt->attach 😓

khanhtc1202:
duplicated PR #4855

khanhtc1202:
@HoangNguyen689 I will close this PR due to the problem is resolved by PR https://github.com/pipe-cd/pipecd/pull/4855 and we will release it in patch v0.47.1. Thanks a lot for your contribution 🙏

PR #4873, opened 2024-04-15
Update configuration validation logic for Attachment and Encryption features

khanhtc1202:

Ensure the configuration of Attachment and Encryption features contain targets filepath by adding validate rule

Fixes #

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -554,6 +554,9 @@ func (e *SecretEncryption) Validate() error {
 			return fmt.Errorf("value field of %s in encryptedSecrets must not be empty", k)
 		}
 	}
+	if len(e.DecryptionTargets) == 0 {
+		return fmt.Errorf("no target files to be decrypted")
+	}
 	return nil
 }
 
@@ -564,14 +567,17 @@ type Attachment struct {
 	Targets []string `json:"targets"`
 }
 
-func (e *Attachment) Validate() error {
-	for k, v := range e.Sources {
+func (a *Attachment) Validate() error {
+	for k, v := range a.Sources {
 		if k == "" {
 			return fmt.Errorf("key field in sources must not be empty")
 		}
 		if v == "" {
 			return fmt.Errorf("value field in sources must not be empty")
 		}
+	}
+	if len(a.Targets) == 0 {
+		return fmt.Errorf("no target files to be embedded")
 	}
 	return nil
 }
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -555,7 +555,7 @@ func (e *SecretEncryption) Validate() error {
 		}
 	}
 	if len(e.DecryptionTargets) == 0 {
-		return fmt.Errorf("no target files to be decrypted")
+		return fmt.Errorf("derecryptionTargets must not be empty")
 	}
 	return nil
 }
@@ -577,7 +577,7 @@ func (a *Attachment) Validate() error {
 		}
 	}
 	if len(a.Targets) == 0 {
-		return fmt.Errorf("no target files to be embedded")
+		return fmt.Errorf("attachment targets must not be empty")
 	}
 	return nil
 }
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -546,6 +546,9 @@ type SecretEncryption struct {
 }
 
 func (e *SecretEncryption) Validate() error {
+	if len(e.DecryptionTargets) == 0 {
+		return fmt.Errorf("derecryptionTargets must not be empty")
+	}
 	for k, v := range e.EncryptedSecrets {
 		if k == "" {
 			return fmt.Errorf("key field in encryptedSecrets must not be empty")
@@ -553,9 +556,6 @@ func (e *SecretEncryption) Validate() error {
 		if v == "" {
 			return fmt.Errorf("value field of %s in encryptedSecrets must not be empty", k)
 		}
-	}
-	if len(e.DecryptionTargets) == 0 {
-		return fmt.Errorf("derecryptionTargets must not be empty")
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/config/application.go
```diff
@@ -554,6 +554,9 @@ func (e *SecretEncryption) Validate() error {
 			return fmt.Errorf("value field of %s in encryptedSecrets must not be empty", k)
 		}
 	}
+	if len(e.DecryptionTargets) == 0 {
+		return fmt.Errorf("no target files to be decrypted")
```
t-kikuc, line 558: [IMO] How about "decryptionTargets must not be empty"?

pkg/config/application.go
```diff
@@ -564,15 +567,18 @@ type Attachment struct {
...
+	for k, v := range a.Sources {
 		if k == "" {
 			return fmt.Errorf("key field in sources must not be empty")
 		}
 		if v == "" {
 			return fmt.Errorf("value field in sources must not be empty")
 		}
 	}
+	if len(a.Targets) == 0 {
+		return fmt.Errorf("no target files to be embedded")
```
t-kikuc, line 580: [IMO] How about "attachment's targets must not be empty"?

pkg/config/application.go
t-kikuc: That's because I think users can more easily find the cause&solution if "decryptionTargets" is in the message.

pkg/config/application.go
t-kikuc: That's because I think users can more easily find the cause&solution if "attachment's targets" is in the message.

khanhtc1202:
@t-kikuc Updated, thanks for comments 🙏

t-kikuc (approved):
Thank you so much!!!!

pkg/config/application.go
```diff
@@ -564,15 +567,18 @@ type Attachment struct {
...
 		if k == "" {
 			return fmt.Errorf("key field in sources must not be empty")
 		}
 		if v == "" {
 			return fmt.Errorf("value field in sources must not be empty")
 		}
 	}
+	if len(a.Targets) == 0 {
+		return fmt.Errorf("attachment targets must not be empty")
+	}
```
ffjlabo, line 581: [IMO] It would be nice to check `a.Targets` because we can avoid executing the for sentence 👀

ffjlabo, line 581: [IMO] It would be nice to check a.Targets because we can avoid executing the for sentence 👀

pkg/config/application.go
```diff
@@ -554,6 +554,9 @@ func (e *SecretEncryption) Validate() error {
 			return fmt.Errorf("value field of %s in encryptedSecrets must not be empty", k)
 		}
 	}
+	if len(e.DecryptionTargets) == 0 {
+		return fmt.Errorf("derecryptionTargets must not be empty")
+	}
```
ffjlabo, line 559: > [nit] It would be nice to check a.Targets because we can avoid executing the for sentence 👀

Also this one 👀

pkg/config/application_test.go
```diff
@@ -210,11 +210,13 @@ func TestValidateEncryption(t *testing.T) {
 	testcases := []struct {
 		name             string
 		encryptedSecrets map[string]string
+		targets          []string
 		wantErr          bool
 	}{
 		{
 			name:             "valid",
 			encryptedSecrets: map[string]string{"password": "pw"},
+			targets:          []string{"secret.yaml"},
```
ffjlabo, line 219: [IMO] add `target` for the other patterns "invalid because key is empty", "invalid because value is empty" for the correct test case 👀

pkg/config/application_test.go
```diff
@@ -227,18 +229,65 @@ func TestValidateEncryption(t *testing.T) {
...
+		{
+			name:    "invalid because key is empty",
+			sources: map[string]string{"": "config-data"},
+			wantErr: true,
+		},
+		{
+			name:    "invalid because value is empty",
+			sources: map[string]string{"config": ""},
+			wantErr: true,
+		},
```
ffjlabo, line 272: > [IMO] add target for the other patterns "invalid because key is empty", "invalid because value is empty" for the correct test case 👀

Also 👀

pkg/config/application_test.go
```diff
@@ -227,18 +229,65 @@ func TestValidateEncryption(t *testing.T) {
...
+		{
+			name:    "no target files sepcified",
+			sources: map[string]string{"config": "config.yaml"},
+			wantErr: true,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
```
ffjlabo, line 280: [nit] add tc := tc

pkg/config/application_test.go
```diff
@@ -227,18 +229,65 @@ func TestValidateEncryption(t *testing.T) {
...
 			wantErr:          true,
 		},
+		{
+			name:             "no target files sepcified",
+			encryptedSecrets: map[string]string{"password": "pw"},
+			wantErr:          true,
+		},
 	}
 	for _, tc := range testcases {
 		t.Run(tc.name, func(t *testing.T) {
```
ffjlabo, line 239: [nit] add tc := tc

ffjlabo (commented):
Thank you for the fix ! commented 💬

pkg/config/application.go
khanhtc1202: @ffjlabo  Sorry, I don't understand what do you mean. Could you explain this?

khanhtc1202: Ah, I get it now; you are missing the word "before," lol. Nice idea 👍

khanhtc1202:
@ffjlabo Thanks for comments, addressed by https://github.com/pipe-cd/pipecd/pull/4873/commits/daf0d7551ca83da2c5c563b36358b852b676192a

ffjlabo:
@t-kikuc Plz re-approve if you have time 🙏

t-kikuc (approved):

PR #4881, opened 2024-04-19
[ECS] Support Plan Preview for ECS

### LINKED ISSUE #4870: [ECS] Support Plan Preview for ECS
This issue includes implementing `Diff()` func, which compares manifests.

t-kikuc:

- Support Plan Preview for ECS
- Including `Diff()` func, which compares manifests(i.e. serviceDef&taskDef in ECS) for non-k8s resources like Lambda

Fixes #4870

**How it works**:

image

Plan Preview is a feature for visualizing model-level diff, not file-level diff(`Files Changed` on GitHub covers it).
So we can ignore following points in the output:
- Each key becomes UpperCamelCase.
   e.g. `image` in taskdef.yaml ->`Image` in planpreview result
- The result also shows attributes that a user did not specify in taskdef.yaml or servicedef.yaml.
   e.g. `EnableExecuteCommand` is shown above but I did not specify it.


pkg/diff/diffbycommand.go
```diff
@@ -0,0 +1,81 @@
...
+	newBytes, err := yaml.Marshal(new)
+	if err != nil {
+		return nil, err
+	}
+
+	return DiffBytesByCommand(command, oldBytes, newBytes)
+}
+
+// DiffBytesByCommand runs the command to compare the given bytes.
+func DiffBytesByCommand(command string, oldBytes, newBytes []byte) ([]byte, error) {
```
khanhtc1202, line 41: What is the point of separate this function from `DiffByCommand`? 🤔

pkg/diff/diff.go
```diff
@@ -97,6 +98,42 @@ func DiffUnstructureds(x, y unstructured.Unstructured, key string, opts ...Optio
 	return d.result, nil
 }
 
+// DiffStructs calulates the diff between two struct objects,
+// which are not k8s manifests.
```
khanhtc1202, line 102: I can imagine this related to this line in your PR description.

> Including Diff() func, which compares manifests(i.e. serviceDef&taskDef in ECS) for non-k8s resources like Lambda

But it is still hard to imagine when to use `DiffUnstructureds` and `DiffStructs`. Maybe we need a better name, or we need to add comments that describe these two functions better.

khanhtc1202: Maybe we also should rename this to `DiffStructureds` either ways 🤔

pkg/diff/diff.go
```diff
@@ -97,6 +98,42 @@ func DiffUnstructureds(x, y unstructured.Unstructured, key string, opts ...Optio
...
+		path = []PathStep{}
+		vx   = reflect.ValueOf(map_x)
+		vy   = reflect.ValueOf(map_y)
+		d    = &differ{result: &Result{}}
+	)
+	for _, opt := range opts {
+		opt(d)
+	}
+
+	d.initIgnoredPaths(key)
```
khanhtc1202, line 127: Not sure this would work? 🤔 Or could you introduce the possible key for this? Since this key in case of k8s manifests could look like this https://pipecd.dev/docs-v0.47.x/user-guide/managing-application/configuration-drift-detection/#ignore-drift-detection-for-specific-fields

pkg/diff/diffbycommand.go
t-kikuc: I merged `DiffBytesByCommand()` into `DiffByCommand()`.

I used `DiffBytesByCommand()` from outside in previous commits, but it is not used now.

t-kikuc:
@khanhtc1202 Thank you for your reviewing, I fixed 3 points.

t-kikuc:
@khanhtc1202 
I fixed:
- Moved LoadECSManifest() to the test code
- Modified tests of DiffByCommand()
- Renamed 'ECSManifest' to 'ECSManifests'

pkg/diff/diffbycommand.go
```diff
@@ -0,0 +1,76 @@
...
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"sigs.k8s.io/yaml"
+)
+
+// DiffByCommand converts the given objects into yaml and then runs the command to compare them.
+func DiffByCommand(command string, old, new interface{}) ([]byte, error) {
```
ffjlabo, line 27: > [IMO]
It would be better to define it as renderByCommand.
I feel the responsibility of the method is to render the diffResult.

Similar to the reason above, how about renaming it as `RenderByCommand` and moving it in the `render.go` like below↓?
https://github.com/pipe-cd/pipecd/blob/master/pkg/diff/renderer.go#L54

I think the DiffXX methods in the diff pkg return `diff.Result` and have the responsibility to extract the differences. It is ideal to unify the interface and the responsivility as much as possible.

WDYT? I want your opinion @t-kikuc @khanhtc1202

pkg/diff/diffbycommand.go
```diff
@@ -0,0 +1,76 @@
...
+	oldBytes, err := yaml.Marshal(old)
```
ffjlabo, line 28: [nit] Check whether the command exists or not.

ffjlabo (commented):
Thank you for the great work! I left some nit and imo comment!

pkg/diff/diffbycommand.go
t-kikuc: @ffjlabo 
Thanks, I agree with you.
TBH, when I referred to CloudRun's diff.go, I thought it was confusing and should be renamed and moved.
cf. https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/cloudrun/diff.go

pkg/app/piped/planpreview/ecsdiff.go
```diff
@@ -0,0 +1,135 @@
...
+
+	summary := fmt.Sprintf("%d changes were detected", len(result.Diff.Nodes()))
+	details := result.Render(provider.DiffRenderOptions{
+		UseDiffCommand: true,
+	})
+	fmt.Fprintf(buf, "--- Last Deploy\n+++ Head Commit\n\n%s\n", details)
+
+	return &diffResult{
+		summary: summary,
+	}, nil
```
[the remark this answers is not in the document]
t-kikuc: thanks, I fixed

[3fee826](https://github.com/pipe-cd/pipecd/pull/4881/commits/3fee82612cc4f296c75b36d93977a4e7234f8e92)

pkg/app/piped/platformprovider/ecs/diff.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+
+	return bytes.Join([][]byte{
+		[]byte("# 1. ServiceDefinition"),
+		serviceDiff,
+		[]byte("\n# 2. TaskDefinition"),
+		taskDiff,
+	}, []byte("\n")), nil
+}
```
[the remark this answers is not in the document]
t-kikuc: thanks, I fixed it. [0cb5d29](https://github.com/pipe-cd/pipecd/pull/4881/commits/0cb5d29436de848192c1ae45b3a6bc9ed2bd33d1)

`renderByCommand` is better especially because this func contains the following code.

```
return bytes.Join([][]byte{
		[]byte("# 1. ServiceDefinition"),
		serviceDiff,
		[]byte("\n# 2. TaskDefinition"),
		taskDiff,
	}, []byte("\n")), nil
```

pkg/diff/diffbycommand.go
t-kikuc: thanks, I fixed it. [c9764e4](https://github.com/pipe-cd/pipecd/pull/4881/commits/c9764e436d27c8dca67789f6b31bcae507a56e0a)

pkg/diff/renderer_test.go
```diff
@@ -204,3 +204,141 @@ func TestRenderPrimitiveValue(t *testing.T) {
...
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := RenderByCommand(tc.command, tc.old, tc.new)
+			if tc.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
```
ffjlabo, line 340: [nit]
```suggestion
			assert.Equal(t, tc.expectErr, err != nil)
```

ffjlabo (commented):
@t-kikuc thanks! added one nit comment for test! I'll approve after the fix :)

pkg/diff/renderer_test.go
t-kikuc: @ffjlabo 
Thanks!!! I fixed it: [64efe0a](https://github.com/pipe-cd/pipecd/pull/4881/commits/64efe0a6bfacb595c8638d9c4ffd4e9860f77ffc)

ffjlabo (approved):
Thank you 🚀

khanhtc1202 (approved):
Nice work, can't wait to see this on prod 🚀

PR #4891, opened 2024-04-22
Remove checking the resource key with the annotation because their generation logic are different

ffjlabo:

context: https://github.com/pipe-cd/pipecd/issues/4269#issuecomment-2031079364

Removed checking the resource key with the annotation because their generation logic differs.

Previously, I tried to keep using the annotation `pipecd.dev/resource-key`, but found that some k8s resources inherit the parent's annotation and can't be attached correct resource key.
ref: https://github.com/pipe-cd/pipecd/pull/4858#issuecomment-2048982687

So I decided to remove the checking.

Part of https://github.com/pipe-cd/pipecd/issues/4269
Previous PR https://github.com/pipe-cd/pipecd/pull/4858

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/applier.go
+++ b/pkg/app/piped/platformprovider/kubernetes/applier.go
@@ -138,7 +138,6 @@ func (a *applier) ReplaceManifest(ctx context.Context, manifest Manifest) error 
 }
 
 // Delete deletes the given resource from Kubernetes cluster.
-// If the resource key is different, this returns ErrNotFound.
 func (a *applier) Delete(ctx context.Context, k ResourceKey) (err error) {
 	a.initOnce.Do(func() {
 		a.kubectl, a.initErr = a.findKubectl(ctx, a.getToolVersionToRun())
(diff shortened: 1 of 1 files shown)


ffjlabo:
## cluster scoped resource

- ✅ namespace exists in app.pipecd.yaml　
- ✅ namespace doesn't exist in app.pipecd.yaml

## namespace scoped resource

- ✅ namespace exists in app.pipecd.yaml　
- ✅ namespace doesn't exist in app.pipecd.yaml

ffjlabo:
There is a problem about the resource key of cluster scoped resources. I'll consider it later to modify the effect scope.

pkg/app/piped/platformprovider/kubernetes/applier.go
```diff
@@ -147,21 +146,17 @@ func (a *applier) Delete(ctx context.Context, k ResourceKey) (err error) {
 		return a.initErr
 	}
 
-	m, err := a.kubectl.Get(
+	// Check the resource exists or not using the resource key.
```
t-kikuc, line 149:
```suggestion
	// Check whether the resource exists or not using the resource key.
```

ffjlabo:
close this PR because I decided another solution for it. 
context: https://github.com/pipe-cd/pipecd/issues/4269#issuecomment-2104055928
follow the PR → https://github.com/pipe-cd/pipecd/pull/4916

PR #4892, opened 2024-04-22
Support PipelineSync on OUT_OF_SYNC

golemiso:

Currently, sync strategy on OUT_OF_SYNC is QUICK_SYNC and this is unchangeable feature.
This can be fixed by using SyncStrategy_AUTO to entrust strategy detection to planner.

NONE

  ppl who configure pipeline sync and expect to use OnOutOfSync trigger forcibly QUICK_SYNC, experience PIPELINE_SYNC (I believe this is rare case)
  I don't think so

  NONE

### DIFF
--- a/pkg/app/piped/trigger/trigger.go
+++ b/pkg/app/piped/trigger/trigger.go
@@ -287,10 +287,6 @@ func (t *Trigger) checkRepoCandidates(ctx context.Context, repoID string, cs []c
 			deploymentChainID = c.command.GetChainSyncApplication().DeploymentChainId
 			deploymentChainBlockIndex = c.command.GetChainSyncApplication().BlockIndex
 
-		case model.TriggerKind_ON_OUT_OF_SYNC:
-			strategy = model.SyncStrategy_QUICK_SYNC
-			strategySummary = "Quick sync to attempt to resolve the detected configuration drift"
-
 		default:
 			strategy = model.SyncStrategy_AUTO
 		}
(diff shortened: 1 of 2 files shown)


pkg/app/piped/trigger/trigger.go
```diff
@@ -287,10 +287,6 @@ func (t *Trigger) checkRepoCandidates(ctx context.Context, repoID string, cs []c
 			deploymentChainID = c.command.GetChainSyncApplication().DeploymentChainId
 			deploymentChainBlockIndex = c.command.GetChainSyncApplication().BlockIndex
 
-		case model.TriggerKind_ON_OUT_OF_SYNC:
-			strategy = model.SyncStrategy_QUICK_SYNC
-			strategySummary = "Quick sync to attempt to resolve the detected configuration drift"
-
```
khanhtc1202, line 293: I think we still need this strategySummary anyway, so how about this

```go
		case model.TriggerKind_ON_OUT_OF_SYNC:
			strategy = model.SyncStrategy_AUTO
			strategySummary = "Sync to attempt to resolve the detected configuration drift"
```

golemiso: strategySummary is unused when SyncStrategy is AUTO.
This is the same logic as with OnCommit trigger.

khanhtc1202: I got it, you're right 👍

khanhtc1202 (approved):
Thank you @golemiso 👍

khanhtc1202:
> ppl who configure pipeline sync and expect to use OnOutOfSync trigger forcibly QUICK_SYNC, experience PIPELINE_SYNC (I believe this is rare case)

@golemiso I totally agree with this point. Basically, the current spec for this part is always making the deployment quick sync to ensure that one diff happens between cluster and manifest, and sync is applied as soon as possible if users enable the `onOutOfSync` trigger (since they want that, we think). Maybe, after this PR, we can introduce a new flag/config to the `onOutOfSync` trigger like `alwaysQuickSync` or `alwaysPipelineSync`, but it's beyond this PR context. Thanks for contribution 🙏

t-kikuc:
[JFYI] https://github.com/pipe-cd/pipecd/pull/2795 might be the background of the current spec.

t-kikuc (approved):
I think there are no problem.

PR #4903, opened 2024-04-30
feat: add support mention slack groups for NotificationReceiverSlack and NotificationMention

### LINKED ISSUE #4730: Support group mentions for slack notification
I think it would be better to support slack group mentions. 

**Why is this needed**:
If someone is assigned to and leaves the development team, it is difficult to change the PipeCD notification settings every time.

Currently [pipeCD is supporting individual user mention](https://pipecd.dev/docs-v0.45.x/user-guide/managing-piped/configuration-reference/#notificationreceiverslack), but the syntax for individual user mentions and group mentions is different in Slack. 
refs: https://api.slack.com/reference/surfaces/formatting#mentioning-groups

As a workaround, if I put the format <!subteam^ID> into mentionedAccounts, it gets displayed as <@@subteam_name> in the Mention To section of the Slack message, with double @@ and even angle brackets <>

hungran:
as mentioned from the issue #4730, currently PipeCD [pipeCD is supporting individual user mention](https://pipecd.dev/docs-v0.47.x/user-guide/managing-piped/configuration-reference/#notificationreceiverslack) 

According on difference syntax that slack support mention in message 
- [user](https://api.slack.com/reference/surfaces/formatting#mentioning-users) 
- [group](https://api.slack.com/reference/surfaces/formatting#mentioning-groups)

This PR will have feature to add field `mentionedGroups` for [NotificationReceiverSlack](https://pipecd.dev/docs-v0.47.x/user-guide/managing-piped/configuration-reference/#notificationreceiverslack) and `slackGroups` for [NotificationMention](https://pipecd.dev/docs-v0.47.x/user-guide/configuration-reference/#notificationmention)

Fixes #4730 


pkg/model/notificationevent.pb.go
t-kikuc, line 1: As the comment in the file says as following, 
```go
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
// 	protoc-gen-go v1.27.1
// 	protoc        v3.18.1
// source: pkg/model/notificationevent.proto
```

you need to edit `notificationevent.proto` instead and run `make gen/code` to generate protobuf files.

hungran: Thanks @t-kikuc  i reverted the change `notificationevent.proto` and successfully ran `make gen/code`

pkg/app/pipedv1/notifier/slack.go
t-kikuc, line 1: It seems you need to copy the changes of `notifier/slack.go` & `slack_test.go` to `piped/notifier/` 🙇

pkg/app/pipedv1/notifier/slack.go
```diff
@@ -228,90 +229,100 @@ func (s *slack) buildSlackMessage(event model.NotificationEvent, webURL string)
...
+		generateDeploymentEventData(md.Deployment, getAccountsAsString(md.MentionedAccounts), getGroupsAsString(md.MentionedGroups))
 
 	case model.NotificationEventType_EVENT_DEPLOYMENT_PLANNED:
 		md := event.Metadata.(*model.NotificationEventDeploymentPlanned)
 		md.MentionedAccounts = append(md.MentionedAccounts, s.config.MentionedAccounts...)
+		md.MentionedGroups = append(md.MentionedGroups, s.config.MentionedGroups...)
 		title = fmt.Sprintf("Deployment for %q was planned", md.Deployment.ApplicationName)
 		text = md.Summary
-		generateDeploymentEventData(md.Deployment, getAccountsAsString(md.MentionedAccounts))
+		generateDeploymentEventData(md.Deployment, getAccountsAsString(md.MentionedAccounts), getAccountsAsString(md.MentionedGroups))
```
t-kikuc, line 266: You mean this??
```suggestion
		generateDeploymentEventData(md.Deployment, getAccountsAsString(md.MentionedAccounts), getGroupsAsString(md.MentionedGroups))
```

pkg/app/pipedv1/notifier/slack.go
```diff
@@ -205,20 +205,21 @@ func (s *slack) buildSlackMessage(event model.NotificationEvent, webURL string)
 		fields            []slackField
 	)
 
-	generateDeploymentEventData := func(d *model.Deployment, accounts string) {
+	generateDeploymentEventData := func(d *model.Deployment, accounts string, groups string) {
```
t-kikuc, line 208: [IMO]
How about calling getAccountsAsString(), getGroupsAsString() here to simplify the switch part?
```suggestion
	generateDeploymentEventData := func(d *model.Deployment, accounts, groups []string) {
	accountsStr := getAccountsAsString(accounts)
	groupsStr := getGroupsAsString(groups)
```

pkg/app/pipedv1/notifier/slack.go
```diff
@@ -205,20 +205,21 @@ func (s *slack) buildSlackMessage(event model.NotificationEvent, webURL string)
...
 			{"Triggered By", d.TriggeredBy(), true},
-			{"Mention To", accounts, true},
+			{"Mention To Users", accounts, true},
+			{"Mention To Groups", groups, true},
 			{"Started At", makeSlackDate(d.CreatedAt), true},
 		}
 	}
 
-	generateDeploymentEventDataForTriggerFailed := func(app *model.Application, hash, msg, accounts string) {
+	generateDeploymentEventDataForTriggerFailed := func(app *model.Application, hash, msg, accounts string, groups string) {
```
t-kikuc, line 222: [IMO]
How about calling getAccountsAsString(), getGroupsAsString() here to simplify the switch part?

```suggestion
	generateDeploymentEventDataForTriggerFailed := func(app *model.Application, hash, msg string, accounts, groups []string) {
	accountsStr := getAccountsAsString(accounts)
	groupsStr := getGroupsAsString(groups)
```

t-kikuc (commented):
Thank you for your great contribution!

pkg/config/piped.go
```diff
@@ -971,6 +972,14 @@ func (n *NotificationReceiverSlack) Validate() error {
 		formatMentionedAccount := strings.TrimPrefix(mentionedAccount, "@")
 		mentionedAccounts = append(mentionedAccounts, formatMentionedAccount)
 	}
+	mentionedGroups := make([]string, 0, len(n.MentionedGroups))
+	for _, mentionedGroup := range n.MentionedGroups {
+		formatMentionedGroup := fmt.Sprintf("<!subteam^%s>", strings.TrimPrefix(mentionedGroup, "@"))
+		mentionedGroups = append(mentionedGroups, formatMentionedGroup)
+	}
```
ffjlabo, line 979: Is this logic necessary here?  It seems also in the func `getGroupsAsString` 👀
 If possible, it would be nice to put it in the `getGroupsAsString` because this func is just for validating. :)

khanhtc1202:
Ping @minhquang4334 👋

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
```diff
@@ -83,6 +83,23 @@ spec:
...
+      - name: integration-slack-api-with-oauthTokenData-and-mentioned-both-accounts-and-groups
+        slack:
+          oauthTokenData: token
+          channelID: testid
+          mentionedAccounts:
+            - 'user1'
+            - '@user2'
+          mentionedGroups:
+            - 'group1'
+            - '@group2'
```
minhquang4334, line 102: IMO, we should not use "@" to mention a slack group because the `handle` field after '@' (eg. `group2`) is mutable, it is not convenient since changing the handle involves updating configuration.
refs: https://slack.com/help/articles/212906697-Create-a-user-group

Thus, I suggest using `<!subteam^ID>` for group mentioning, since the fields are immutable and can be used according to [Slack's official docs](https://api.slack.com/reference/surfaces/formatting#mentioning-groups).

pkg/config/piped.go
hungran: correct me if im wrong, they are difference thing `mentionedGroups` is for [NotificationReceiverSlack](https://pipecd.dev/docs-v0.47.x/user-guide/managing-piped/configuration-reference/#notificationreceiverslack) under piped config

the remain for deployment mention slack group (a new field) 
https://pipecd.dev/docs-v0.47.x/user-guide/configuration-reference/#notificationmention

pkg/model/notificationevent.pb.go
hungran: @t-kikuc 
correct me if im wrong
instead change `notificationevent.pb.go`, we add field at `pkg/model/notificationevent.proto` ?

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
hungran: @minhquang4334  thank you very much for your support, yes I agreed with your point
so should be support both `<!subteam^ID>` or only `ID`

pkg/app/pipedv1/notifier/slack.go
hungran: @t-kikuc thank you, yes, calling that way should be visible!

pkg/app/pipedv1/notifier/slack.go
hungran: i dont get this 
- why we remove data type of `accounts`
- why add data type groups as list string?

hungran: i believed we get return all method here as `string`

pkg/app/pipedv1/notifier/slack.go
hungran: same question above?

pkg/config/piped.go
hungran: um not sure but i've seen we already have same thing for user
```
	mentionedAccounts := make([]string, 0, len(n.MentionedAccounts))
	for _, mentionedAccount := range n.MentionedAccounts {
		formatMentionedAccount := strings.TrimPrefix(mentionedAccount, "@")
		mentionedAccounts = append(mentionedAccounts, formatMentionedAccount)
	}
```
should we already remove it as well because `getAccountsAsString` already have it?

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
hungran: @minhquang4334 just from the time being, as slack's client, using slack to mention user and group has exactly same way (by `@`)

pkg/app/pipedv1/notifier/slack.go
hungran: yes, done

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
```diff
@@ -83,6 +83,23 @@ spec:
...
+      - name: integration-slack-api-with-oauthTokenData-and-mentioned-both-accounts-and-groups
+        slack:
+          oauthTokenData: token
+          channelID: testid
+          mentionedAccounts:
+            - 'user1'
+            - '@user2'
+          mentionedGroups:
+            - 'group1'
+            - '<!subteam^group2>'
```
minhquang4334, line 102: Using the groupID or teamID in the documents makes it easier for users to use this feature.

```suggestion
            - 'groupID1'
            - '<!subteam^groupID2>'
```

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
minhquang4334: thank you @hungran

pkg/config/application.go
```diff
@@ -588,14 +588,21 @@ type DeploymentNotification struct {
...
-func (n *DeploymentNotification) FindSlackAccounts(event model.NotificationEventType) []string {
+func (n *DeploymentNotification) FindSlackAccountsAndGroups(event model.NotificationEventType) []string {
 	as := make(map[string]struct{})
 	for _, m := range n.Mentions {
 		if m.Event != allEventsSymbol && "EVENT_"+m.Event != event.String() {
 			continue
 		}
-		for _, s := range m.Slack {
-			as[s] = struct{}{}
+		if len(m.Slack) > 0 {
```
minhquang4334, line 597: It would be better if we changed `Slack` to `SlackUsers` to separate it from `SlackGroups`.

hungran: @minhquang4334 thanks, will have a look

hungran: how if we keep `Slack` field and mark it as deprecated field and add `SlackUsers`
wdyt? @minhquang4334

hungran: also if we keep `Slack`, thinking 3 option
- combine 2 those field to final list users mention
- set priority to use new field only
- throw error from validate() that only support one of those at same time
wdyt?

minhquang4334: > how if we keep Slack field and mark it as deprecated field and add SlackUsers
wdyt? @minhquang4334

LGTM

> set priority to use new field only

IMO, this one seems easier for users to use, and removing the Slack field minimizes the impact of breaking changes 🤔

hungran: @minhquang4334  thanks for your feed back
update the PR https://github.com/pipe-cd/pipecd/pull/4903/commits/e957946f41e50b145ae5556aef03b4b27f5d1090
please have a look

minhquang4334: LGTM

minhquang4334 (approved):
🐳

hungran:
build failed?
maybe needed `pkg/model/notificationevent.pb.go` @minhquang4334

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
Warashi: @hungran @minhquang4334 
As far as I know, we can only use immutable ID for the `<!subteam^ID>` format when we post messages by API.
If we want to provide a way to use the handle for group mentioning, we can use the web API `usergroups.list` method to retrieve the ID.
We can use group mentions only with paid workspace, so I'll verify the current implementation of this PR with my Slack workspace, which I use to work.
I must gain permission to register for the Slack application, so the verification takes time.

hungran: thank you Warashi, yes we are following immutable ID `<!subteam^ID>`

Warashi: Oops, thank you.
Then, please change `groupname` to `groupID` in this document?
- `docs/content/en/docs-dev/user-guide/managing-piped/configuration-reference.md`

Warashi: I checked the behavior with this configuration.
```
spec:
  notification:
    mentions:
      - event: DEPLOYMENT_TRIGGERED
        slackGroups:
          - <!subteam^S07A5J95MHR>
          - S07A5J95MHR
          - <!subteam^warashi-test>
          - warashi-test
```

And I got this notification.
(I set the keyword notification, so please ignore `warashi` and `sawada` is highlighted)

image

There are extra `<@` and `>` around the mention, and configuration only with the ID is not recognized as a mention.

Warashi:
@hungran 
> build failed?
> maybe needed pkg/model/notificationevent.pb.go

Please run `make gen/code`?
This command generates needed files.

docs/content/en/docs-dev/user-guide/managing-piped/configuring-notifications.md
hungran: wired, they should be in `Mention To Group` let me double check

hungran: Hi @Warashi a bit hard from my site to reproduce this error since i didnt have Slack workspace ent
@Warashi @khanhtc1202 @minhquang4334 could you help a look

Warashi (commented):
I investigated the problem of all mentions listed in `Mention To Users` and the problem of a group mention specified with only ID not being recognized as a group mention in the Slack message.
I left comments on that.

pkg/app/piped/trigger/trigger.go
```diff
@@ -473,7 +473,7 @@ func (t *Trigger) GetLastTriggeredCommitGetter() LastTriggeredCommitGetter {
 func (t *Trigger) notifyDeploymentTriggered(_ context.Context, appCfg *config.GenericApplicationSpec, d *model.Deployment) {
 	var mentions []string
 	if n := appCfg.DeploymentNotification; n != nil {
-		mentions = n.FindSlackAccounts(model.NotificationEventType_EVENT_DEPLOYMENT_TRIGGERED)
+		mentions = n.FindSlackAccountsAndGroups(model.NotificationEventType_EVENT_DEPLOYMENT_TRIGGERED)
```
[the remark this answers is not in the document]
hungran: thanks, you are corrected,
i dont think atm i could find a best way instead of duplicate code for logic groups mention

pkg/config/application.go
```diff
@@ -619,16 +631,40 @@ func (n *DeploymentNotification) FindSlackAccounts(event model.NotificationEvent
...
+		if !strings.Contains(slackGroup, "!subteam^") {
+			formatSlackGroup := fmt.Sprintf("<!subteam^%s>", slackGroup)
+			slackGroups = append(slackGroups, formatSlackGroup)
+		} else {
+			slackGroups = append(slackGroups, slackGroup)
+		}
+	}
+	if len(slackGroups) > 0 {
+		n.SlackGroups = slackGroups
+	}
```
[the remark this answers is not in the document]
hungran: updated logic and remove format string as it was handled in `slack.go`

pkg/config/application.go
```diff
@@ -598,14 +598,26 @@ type DeploymentNotification struct {
...
+		}
+		if len(m.SlackUsers) > 0 {
+			for _, su := range m.SlackUsers {
+				as[su] = struct{}{}
+			}
+		}
+		if len(m.SlackGroups) > 0 {
+			for _, sg := range m.SlackGroups {
+				as[sg] = struct{}{}
+			}
```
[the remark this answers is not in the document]
hungran: yes, atm i dont think i have better option that duplicate the function logic to find slackgroups

hungran:
@Warashi  seem a lot of duplicate code here for handle slackGroups, do you have any idea? @minhquang4334 @khanhtc1202 @Warashi

Warashi:
I checked the behavior, and this PR worked as expected! Thank you!

By the way, I have an idea about duplicated codes.
How about defining the wrapper codes for `getMentionedUsers` and `getMentionedGroups`?

I think the codes are below, and I think these codes are defined in the config package.

```go
type Getter interface {
	Get(string) (string, bool)
}

func GetApplicationNotificationMentions(getter Getter, event model.NotificationEventType) (users []string, groups []string, error) {
	n, ok := getter.Get(model.MetadataKeyDeploymentNotification)
	if !ok {
		return []string{}, nil
	}

	var notification DeploymentNotification
	if err := json.Unmarshal([]byte(n), &notification); err != nil {
		return nil, fmt.Errorf("could not extract mentions config: %w", err)
	}

	return notification.FindSlackUsers(event), notification.FindSlackGroups(event), nil
}
```

WDYT? @hungran  @khanhtc1202  @ffjlabo  @t-kikuc  @minhquang4334

hungran:
> I checked the behavior, and this PR worked as expected! Thank you!
> 
> By the way, I have an idea about duplicated codes. How about defining the wrapper codes for `FindSlackUsers` and `FindSlackGroups`?
> 
> I think the codes are below, and I think these codes are defined in the config package.
> 
> ```go
> type Getter interface {
> 	Get(string) (string, bool)
> }
> 
> func GetApplicationNotificationMentions(getter Getter, event model.NotificationEventType) (users []string, groups []string, error) {
> 	n, ok := getter.Get(model.MetadataKeyDeploymentNotification)
> 	if !ok {
> 		return []string{}, nil
> 	}
> 
> 	var notification DeploymentNotification
> 	if err := json.Unmarshal([]byte(n), &notification); err != nil {
> 		return nil, fmt.Errorf("could not extract mentions config: %w", err)
> 	}
> 
> 	return notification.FindSlackUsers(event), notification.FindSlackGroups(event), nil
> }
> ```
> 
> WDYT? @hungran @khanhtc1202 @ffjlabo @t-kikuc @minhquang4334

@Warashi thank you very much for checking, 
for me avoid duplicate code in this case is greats, but better to see more view point, advice from team

ffjlabo:
Thank you for asking.

@Warashi 's idea seems to have two purposes.
- to unify the process to get users and groups in the same method
- to unify the above processes on the scheduler, planner, and executor in the same method

I totally agree with the idea, but I am concerned about the second purpose.
For me, creating the interface `Getter` is a little bit overly complex because it only abstracts the `metadataStore.Get`.

So how about solving the first one at the moment? 
It means to create `getApplicationNotificationMentions` on the scheduler, planner, executor like this.

WDYT? @hungran @Warashi @khanhtc1202 @t-kikuc @minhquang4334

```golang
func (s *scheduler) getApplicationNotificationMentions(n *DeploymentNotification , event model.NotificationEventType) ([]string, []string, error) {
  n, ok := s.metadataStore.Shared().Get(model.MetadataKeyDeploymentNotification)
  if !ok {
    return []string{}, nil
  }
  var notification config.DeploymentNotification
  if err := json.Unmarshal([]byte(n), &notification); err != nil {
    return nil, fmt.Errorf("could not extract mentions users config: %w", err)
  }

  return notification.FindSlackUsers(event), notification.FindSlackGroups(event), nil
}
```

```golang
func (p *planner) getApplicationNotificationMentions(n *DeploymentNotification , event model.NotificationEventType) ([]string, []string, error) {
  n, ok := s.metadataStore.Shared().Get(model.MetadataKeyDeploymentNotification)
  if !ok {
    return []string{}, nil
  }
  var notification config.DeploymentNotification
  if err := json.Unmarshal([]byte(n), &notification); err != nil {
    return nil, fmt.Errorf("could not extract mentions users config: %w", err)
  }

  return notification.FindSlackUsers(event), notification.FindSlackGroups(event), nil
}
```

```golang
func (e *Executor) getApplicationNotificationMentions(n *DeploymentNotification , event model.NotificationEventType) ([]string, []string, error) {
  n, ok := s.metadataStore.Shared().Get(model.MetadataKeyDeploymentNotification)
  if !ok {
    return []string{}, nil
  }
  var notification config.DeploymentNotification
  if err := json.Unmarshal([]byte(n), &notification); err != nil {
    return nil, fmt.Errorf("could not extract mentions users config: %w", err)
  }

  return notification.FindSlackUsers(event), notification.FindSlackGroups(event), nil
}
```

Warashi:
>  For me, creating the interface Getter is a little bit overly complex because it only abstracts the metadataStore.Get.

I agree with you.

I found `metadatastore.Getter` interface is already defined. How about using this instead of defining another one?
https://github.com/pipe-cd/pipecd/blob/6400c898654e41ac80b6c2ed5a345a8b0bc080a7/pkg/app/piped/metadatastore/store.go#L27-L30

ffjlabo:
@Warashi 
I agree with using the interface. Also, I got another concern 🙏 

I think there are two processes in the `GetApplicationNotificationMentions`.
- get the value (type `*config.DeploymentNotification`) from the metadataStore
- get the users and groups from `*config.DeploymentNotification`

For the first one, it would be nice to separate it as an another component which has the responsibility to get the proper values from metadataStore. I think it is simpler.

But we should consider it on the other issue. So we focus on fixing to unify only `getMentionedUsers` 
and `getMentionedGroups` to the `getApplicationNotificationMentions` on the scheduler, planner, executor.

WDYT?

Warashi:
@ffjlabo 
I agree. Let's start by unifying only getMentionedUsers and getMentionedGroups.

hungran:
@ffjlabo @Warashi Thank you so much for your feedback



@ffjlabo 
> I think there are two processes in the `GetApplicationNotificationMentions`.
> 
> * get the value (type `*config.DeploymentNotification`) from the metadataStore
> * get the users and groups from `*config.DeploymentNotification`

agreed unify the vaule  (type `*config.DeploymentNotification`) from the metadataStore should be unify by another component/function but here im clear how to continue with the way to focus on fixing `getMentionedUsers`
and `getMentionedGroups`

hungran:
@ffjlabo  @Warashi thank you for your feedback, updated follow idea to unify process to get Users and Groups for planner, scheduler, executor

Warashi (approved):
I checked the behavior, and it worked as expected.
LGTM
Thank you!

ffjlabo (approved):
Great work! Thank you 🚀

PR #4916, opened 2024-05-15
Change resource key logic for k8s

ffjlabo:

This fix will solve the problem that the cluster-scoped resource can't be deleted when we set the namespace in app.pipecd.yaml.
context: https://github.com/pipe-cd/pipecd/issues/4269#issuecomment-2104055928

**expected behavior**
- We can delete the cluster-scoped resource.

**livestate side**
- use "" for the cluster-scoped resource
- use the namespace for the namespace scoped resource

**When reading the manifests from git** 
- use "" for the cluster-scoped resource
-  for the namespace scoped resource,
   - use `spec.input.namespace` in app.pipecd.yaml if it is set.
   - use the namespace on the manifest from git if it is set.
   - use `default` if  both of `spec.input.namespace` and the namespace are not set

---

Currently, we use the resource key to identify each k8s resource.
It is created by `MakeResourceKey`, and the rule is below.

**livestate side**
- use `default` when the resource obj doesn't have the namespace.
- use the namespace when the resource obj has the namespace.

**When reading the manifests from git** 
- use `spec.input.namespace` in app.pipecd.yaml if it is set.
- use the namespace on the manifest from git if it is set.
- use `default` when the namespace on the manifest is "".

This rule doesn't consider the cluster-scoped resource.
So for example, cluster-scoped resource don't have any namespace, but if we set the `spec.input.namespace` in the app.pipecd.yaml, it sets the value as the namespace to the resource key.

This causes the problem that can't prune the resource because the resource key can't identify the resource correctly.
So I fixed the the logic like below.

**livestate side**
- use "" for the cluster-scoped resource
- use the namespace for the namespace scoped resource

**When reading the manifests from git** 
- use "" for the cluster-scoped resource
-  for the namespace scoped resource,
   - use `spec.input.namespace` in app.pipecd.yaml if it is set.
   - use the namespace on the manifest from git if it is set.
   - use `default` if  both of `spec.input.namespace` and the namespace are not set

Part of #4269

- **Does this PR introduce a user-facing change?**:

### DIFF
--- a/pkg/app/piped/driftdetector/kubernetes/detector.go
+++ b/pkg/app/piped/driftdetector/kubernetes/detector.go
@@ -24,8 +24,6 @@ import (
 
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/pipe-cd/pipecd/pkg/app/piped/livestatestore/kubernetes"
 	provider "github.com/pipe-cd/pipecd/pkg/app/piped/platformprovider/kubernetes"
@@ -126,33 +124,13 @@ func (d *detector) Run(ctx context.Context) error {
 }
 
 func (d *detector) check(ctx context.Context) {
-	// Use discovery to discover APIs supported by the Kubernetes API server.
-	// This should be run periodically with a low rate because the APIs are not added frequently.
-	// https://godoc.org/k8s.io/client-go/discovery
-	cp := d.provider
-	kubeConfig, err := clientcmd.BuildConfigFromFlags(cp.KubernetesConfig.MasterURL, cp.KubernetesConfig.KubeConfigPath)
+	isNamespacedResources, err := provider.GetIsNamespacedResources(d.provider.KubernetesConfig)
 	if err != nil {
-		d.logger.Error("failed to build kube config", zap.Error(err))
+		d.logger.Error("failed to get isNamespacedResources", zap.Error(err))
 		return
 	}
-	discoveryClient, err := discovery.NewDiscoveryClientForConfig(kubeConfig)
-	if err != nil {
-		d.logger.Error("failed to create discovery client", zap.Error(err))
-		return
-	}
-	groupResources, err := discoveryClient.ServerPreferredResources()
-	if err != nil {
-		d.logger.Error("failed to fetch preferred resources", zap.Error(err))
-		return
-	}
-	d.logger.Info(fmt.Sprintf("successfully preferred resources that contains for %d groups", len(groupResources)))
 
-	for _, gr := range groupResources {
-		for _, resource := range gr.APIResources {
-			gvk := schema.FromAPIVersionAndKind(gr.GroupVersion, resource.Kind)
-			d.isNamespacedResources[gvk] = resource.Namespaced
-		}
-	}
+	d.isNamespacedResources = isNamespacedResources
 
 	appsByRepo := d.listGroupedApplication()
 
--- a/pkg/app/piped/driftdetector/kubernetes/detector.go
+++ b/pkg/app/piped/driftdetector/kubernetes/detector.go
@@ -23,6 +23,9 @@ import (
 	"time"
 
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/pipe-cd/pipecd/pkg/app/piped/livestatestore/kubernetes"
 	provider "github.com/pipe-cd/pipecd/pkg/app/piped/platformprovider/kubernetes"
@@ -67,8 +70,9 @@ type detector struct {
 	secretDecrypter   secretDecrypter
 	logger            *zap.Logger
 
-	gitRepos   map[string]git.Repo
-	syncStates map[string]model.ApplicationSyncState
+	gitRepos              map[string]git.Repo
+	syncStates            map[string]model.ApplicationSyncState
+	isNamespacedResources map[schema.GroupVersionKind]bool
 }
 
 func NewDetector(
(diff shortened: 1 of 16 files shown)


ffjlabo:
# The Overview of the fix

**refine the namespace by `loader.refineNamespace` when reading the manifests from git repo**
- fixed to obtain information to determine whether each k8s resource is cluster scoped when executor, planner, plan preview, and detector start. (`provider.GetIsNamespacedResources`)

**remove the logic to determine the namespace from `MakeResourceKey`**
- I think `MakeResourceKey` shouldn't determine the namespace. This is  only loader's responsibility.

ffjlabo:
@khanhtc1202 @t-kikuc 
This issue is complex and my explanation may be insufficient. If you have any questions, please feel free to ask. :) 🙏

khanhtc1202:
@ffjlabo some testcases are failed :eyes:

ffjlabo:
@khanhtc1202 Sorry, I fixed some of them. 🙏 
There are some failings, but these are under the /pipecdv1. I want to fix it after the current implementation looks good.

khanhtc1202:
@ffjlabo Please rebase this PR due to pipedv1 package change 👀

ffjlabo:
The failing test is solved by this PR 🙏 https://github.com/pipe-cd/pipecd/pull/4926

pkg/app/piped/executor/kubernetes/kubernetes.go
```diff
@@ -75,6 +77,18 @@ func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
 	ctx := sig.Context()
 	e.commit = e.Deployment.Trigger.Commit.Hash
 
+	cp, ok := e.PipedConfig.FindPlatformProvider(e.Deployment.PlatformProvider, model.ApplicationKind_KUBERNETES)
+	if !ok {
+		e.LogPersister.Errorf("provider %s was not found", e.Deployment.PlatformProvider)
+		return model.StageStatus_STAGE_FAILURE
+	}
```
khanhtc1202, line 84: Is this change related to PR purposes?

pkg/app/piped/executor/kubernetes/rollback.go
```diff
@@ -41,6 +46,39 @@ func (e *rollbackExecutor) Execute(sig executor.StopSignal) model.StageStatus {
...
+	if err != nil {
+		e.LogPersister.Errorf("failed to create discovery client: %v", zap.Error(err))
+		return model.StageStatus_STAGE_FAILURE
+	}
+	groupResources, err := discoveryClient.ServerPreferredResources()
+	if err != nil {
+		e.LogPersister.Errorf("failed to fetch preferred resources: %v", zap.Error(err))
+		return model.StageStatus_STAGE_FAILURE
+	}
+	e.LogPersister.Info(fmt.Sprintf("successfully preferred resources that contains for %d groups", len(groupResources)))
```
khanhtc1202, line 72:
```suggestion
	e.LogPersister.Infof("successfully preferred resources that contains for %d groups", len(groupResources))
```
And remove "fmt" package

pkg/app/piped/platformprovider/kubernetes/loader.go
```diff
@@ -166,16 +167,50 @@ func (l *loader) LoadManifests(ctx context.Context) (manifests []Manifest, err e
...
-	if namespace == "" {
-		return
+// refineNamespace fix the namespace of the given manifest.
+// The priority is as follows:
+// If the resource is cluster-scoped, it returns an empty string.
+// Otherwise, it is the namespace-scoped resource and the namespace is determined by the following order:
+// 1. The namespace set in the application configuration.
+// 2. The namespace set in the manifest.
+// 3. The default namespace.
+func (l *loader) refineNamespace(m *Manifest) error {
```
khanhtc1202, line 190: Should it be named `determineResourceNamespace` or `determineNamespace`? 🤔

ffjlabo: Thanks! fixed [5d22c75](https://github.com/pipe-cd/pipecd/pull/4916/commits/5d22c7587fec159fb2a8e926ade63ded2eead06b)

pkg/app/piped/platformprovider/kubernetes/loader.go
```diff
@@ -166,16 +167,50 @@ func (l *loader) LoadManifests(ctx context.Context) (manifests []Manifest, err e
...
+		if l.input.Namespace != "" {
+			namespace = l.input.Namespace
+		}
 	}
+
+	m.Key.Namespace = namespace
+	m.u.SetNamespace(namespace)
+
+	return nil
 }
```
ffjlabo, line 214: 📝 Check the current logic

ffjlabo:
I tested it using QuickSync with the scenario below, and it works well.

scenario:
tried the following operations with and without setting `spec.input.namespace` in `app.pipecd.yaml` for cluster scoped resource and namespaced resource.
- create a k8s app
- update manifests
- fix the actual resource with `kubectl edit`
- delete manifests

I chose and tried them with ClusterRole (cluster scoped) and Deployment (namespaced) resources.

checkpoint:
- create, update, and delete the resources successfully
- reflect the operations result on the application livestate (check on the Web UI)
- detect OutOfSync after `kubectl edit`

pkg/app/piped/driftdetector/kubernetes/detector.go
```diff
@@ -87,18 +89,19 @@ func NewDetector(
...
+		gitClient:             gitClient,
+		stateGetter:           stateGetter,
+		reporter:              reporter,
+		appManifestsCache:     appManifestsCache,
+		interval:              time.Minute,
+		config:                cfg,
+		secretDecrypter:       sd,
+		gitRepos:              make(map[string]git.Repo),
+		syncStates:            make(map[string]model.ApplicationSyncState),
+		isNamespacedResources: make(map[schema.GroupVersionKind]bool),
```
Warashi, line 103: nits

isNamespacedResources seems to be an immutable map. After construction, there is no operation to add an element to this map.
In this case, a nil map is enough to express an empty map.

example code
https://go.dev/play/p/A8qeO9dCyuw

Warashi (approved):
Seems good to me.
I commented on one nitpick for your information.

pkg/app/piped/driftdetector/kubernetes/detector.go
ffjlabo: After that, I removed `isNamespacedResources` from the struct detector because it currently gets it every time it executes the detector's check logic. 🙏 
[a885bab](https://github.com/pipe-cd/pipecd/pull/4916/commits/a885bab8f238a6cf0dd224155eda5a556eeb3f25)

Warashi (approved):
LGTM

khanhtc1202 (approved):
Great work, thank you 🪨

t-kikuc (approved):
greatest work

ffjlabo:
Thank you for reviewing 🙏

PR #4922, opened 2024-05-22
Support skipping stages by path pattern or commit message prefix

### LINKED ISSUE #4899: [feat] Skip stages when the commit is not important (Analysis,WaitApproval,Wait,ScriptRun)
## What would you like to be added

- A new feature of skipping the `ANALYSIS` or `WAIT_APPROVAL` stages when the commit diff is not important.

- The configurable conditions of skipping the stage:
    - changed file path patterns (cf. [GitHub Actions > paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore))
    - commit message (prefix)

## Why is it needed

- Some users would like to skip the `ANALYSIS` or `WAIT_APPROVAL` stages for faster and automatic deployment when the commit diff does not deserve analysis or approval.

- use cases:
  (a) When you modify only the `HorizontalPodAutoscaler` of K8s resources in your config repo, skip the `WAIT_APPROVAL` stage.
  (b) When you modify only `go.mod` dependencies in your source repo(_not pipecd's config repo_), skip the `ANALYSIS` stage.
  (c) Skip the `WAIT_APPROVAL` stage when the change in your source repo is small, but skip both the `WAIT_APPROVAL` and `ANALYSIS` stages when the change is tiny.

## How to realize it

- Add the `skip` configuration to app.pipecd.yaml.

```yaml
spec:
  pipeline:
    stages:
      # When changes are only `autoscaler.yaml` in any directory, skip this WAIT_APPROVAL.
      - name: WAIT_APPROVAL
        skip:
           paths:
              - "**/autoscaler.yaml"
      # When the commit message starts with "[skip analysis]", skip this ANALYSIS.
      - name: ANALYSIS
        skip:
            commitMessagePrefixes:
             - "[skip analysis]"
```

### Example1. How to realize the use case (b)

1. Configure `eventWatcher` and `pipeline` in your app.pipecd.yaml.
```yaml
spec:
  eventWatcher:
    - matcher:
        name: skip-analysis
      handler:
        type: GIT_UPDATE
        config:
          commitMessage: "[skip analysis] update xxx"
  pipeline:
      stages:
      - name: ANALYSIS
        skip:
            commitMessagePrefixes:
             - "[skip analysis]"
```
2. Let your CI publish an event named `skip-analysis` if only `go.mod` is changed.

### Example2. How to realize the use case (c)

1. Configure `eventWatcher` and `pipeline` in your app.pipecd.yaml.

```yaml
spec:
  eventWatcher:
    - matcher:
        name: small-change
      handler:
        type: GIT_UPDATE
        config:
          commitMessage: "[small change] update xxx"
    - matcher:
        name: tiny-change
      handler:
        type: GIT_UPDATE
        config:
          commitMessage: "[tiny change] update yyy"
  pipeline:
      stages:
      - name: WAIT_APPROVAL
        skip:
           commitMessagePrefixes:
              - "[small change]"
              - "[tiny change]"
      - name: ANALYSIS
        skip:
            commitMessagePrefixes:
             - "[tiny change]"
```

2. Let your CI publish an event named `small-change` to skip only `WAIT_APPROVAL`, and `tiny-change` to skip both `WAIT_APPROVAL` and `ANALYSIS`.

## Possible Extensions

- In addition to `commitMessagePrefixes`, it would be better to have other options like:
    -  `skip.commitMessageOrBodyContains: []string`
       - It skips the stage when the title or body of the commit contains words you specify.
    - `skip.commitMessageOrBodyContainsCommand: bool`
       - It skips the stage when the title or body of the commit contains specific commands like `[skip analysis]`, `[skip stage]`, etc.
       - cf. [GitHub Actions > Skipping workflow runs](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs)

t-kikuc:

This PR enables skipping stages (Analysis,WaitApproval,Wait,ScriptRun) when the commit matches
configured path patterns or commit message prefixes.

Fixes #4899

**Does this PR introduce a user-facing change?**: N/A (only append)

**How to use**

Configuration: add `with.skipOn` section in stages to skip.

- `paths`: When ALL changes of the commit match them, the stage will be skipped.
- `commitMessagePrefixes`: When the prefix of the commit's message matches ANY of them, the stage will be skipped.

```yaml
  pipeline:
    stages:
      - name: ECS_TRAFFIC_ROUTING
        with:
          canary: 30
      - name: WAIT_APPROVAL
        with:
          skipOn:
            commitMessagePrefixes:
              - [skip-approval] # When the commit message starts with '[skip-approval]', this stage will be skipped
      - name: WAIT
        with:
          duration: 60s
          skipOn:
            paths:
              - "**/*.yaml" # Any yaml file under any directory
      - name: ANALYSIS
        with:
          duration: 60s
          skipOn:
            paths:
              - "**/app.pipecd.yaml" # Any app.pipecd.yaml under any directory
              - "**/*.md" # Any markdown file under any directory
      - name: ECS_TRAFFIC_ROUTING
        with:
          primary: 100
```

When all `skipOn` conditions of each stage passed, the deployment will be:
image


t-kikuc:
I’ll remove the `wip` label after fixing the CI error.

docs/content/en/docs-dev/user-guide/configuration-reference.md
```diff
@@ -578,6 +578,13 @@ Note: The available values are identical to those found in the aws-sdk-go-v2 Typ
 | Field | Type | Description | Required |
 |-|-|-|-|
 
+## SkipOptions
+
+| Field | Type | Description | Required |
+|-|-|-|-|
+| commitMessagePrefixes | []string | List of prefixes. When the prefix of the commit's message matches any of them, the stage will be skipped. Empty means the stage will not be skipped by this condition. | No |
```
khanhtc1202, line 585:
```suggestion
| commitMessagePrefixes | []string | List of commit message's prefixes. The stage will be skipped when the commit message matches any of them. Empty means the stage will not be skipped by this condition. | No |
```
nits

docs/content/en/docs-dev/user-guide/configuration-reference.md
```diff
@@ -578,6 +578,13 @@ Note: The available values are identical to those found in the aws-sdk-go-v2 Typ
...
+| paths | []string | List of directories or files. When all changes of the commit match them, the stage will be skipped. Empty means the stage will not be skipped by this condition. Regular expression can be used. | No |
```
khanhtc1202, line 586:
```suggestion
| paths | []string | List paths to directories or files. When all commit changes match them, the stage will be skipped. Empty means the stage will not be skipped by this condition. Regular expression can be used. | No |
```
nits

pkg/app/piped/controller/scheduler.go
```diff
@@ -537,6 +538,23 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
 		return model.StageStatus_STAGE_FAILURE
 	}
 
+	// Skip the stage if needed based on the skip config.
+	skip, err := s.checkSkipStage(sig.Context(), input)
```
khanhtc1202, line 542: If we don't need the executor to check whether this stage is skippable, we should run it before L532 to reduce the waste.

pkg/app/piped/controller/scheduler.go
```diff
@@ -537,6 +538,23 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
...
+	if err != nil {
+		lp.Errorf("failed to check whether skipping the stage", zap.Error(err))
+		if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_FAILURE, ps.Requires); err != nil {
+			s.logger.Error("failed to report stage status", zap.Error(err))
+		}
+		return model.StageStatus_STAGE_FAILURE
```
khanhtc1202, line 548:
```suggestion
		if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_FAILURE, ps.Requires); err != nil {
			s.logger.Error("failed to report stage status", zap.Error(err))
			return model.StageStatus_STAGE_FAILURE
		}
		lp.Errorf("failed to check whether skipping the stage", zap.Error(err))
		return model.StageStatus_STAGE_SKIPPED
```

pkg/app/piped/controller/scheduler.go
```diff
@@ -537,6 +538,23 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
...
+	}
+	if skip {
+		lp.Infof("The stage was successfully skipped due to the skip configuration of the stage.")
+		if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_SKIPPED, ps.Requires); err != nil {
+			s.logger.Error("failed to report stage status", zap.Error(err))
+		}
+		return model.StageStatus_STAGE_SKIPPED
```
khanhtc1202, line 555:
```suggestion
		if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_SKIPPED, ps.Requires); err != nil {
			s.logger.Error("failed to report stage status", zap.Error(err))
			return model.StageStatus_STAGE_FAILURE
		}
		lp.Info("The stage was successfully skipped due to the skip configuration of the stage.")
		return model.StageStatus_STAGE_SKIPPED
```

pkg/app/piped/controller/scheduler.go
```diff
@@ -537,6 +538,23 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
 		return model.StageStatus_STAGE_FAILURE
 	}
 
+	// Skip the stage if needed based on the skip config.
+	skip, err := s.checkSkipStage(sig.Context(), input)
+	if err != nil {
+		lp.Errorf("failed to check whether skipping the stage", zap.Error(err))
```
khanhtc1202, line 544:
```suggestion
		lp.Errorf("failed to check whether skipping the stage: %v", err.Error())
```

khanhtc1202: @t-kikuc This `lp` is log persister, a printer that prints log as stage log and is viewable from web UI. While the `logger` is a zap logger, which prints the execution log in stdout of the piped running container. So basically, we can't use the `zap` format with log persister and vice versa. Please refer other place to understand how to use those loggers

pkg/app/piped/controller/scheduler.go
```diff
@@ -687,6 +705,25 @@ func (s *scheduler) reportDeploymentCompleted(ctx context.Context, status model.
 	return err
 }
 
+func (s *scheduler) checkSkipStage(ctx context.Context, in executor.Input) (skip bool, err error) {
```
khanhtc1202, line 708:
```suggestion
func (s *scheduler) shouldSkipStage(ctx context.Context, in executor.Input) (skip bool, err error) {
```
nits

pkg/app/piped/executor/analysis/analysis.go
```diff
@@ -339,7 +339,7 @@ func (e *Executor) buildAppArgs(customArgs map[string]string) argsTemplate {
 	return args
 }
 
-func (e *Executor) checkSkipped(ctx context.Context) bool {
+func (e *Executor) checkSkippedFromCmd(ctx context.Context) bool {
```
khanhtc1202, line 342:
```suggestion
func (e *Executor) checkSkippedByCmd(ctx context.Context) bool {
```
nits

pkg/git/repo.go
```diff
@@ -39,6 +39,7 @@ type Repo interface {
 	ListCommits(ctx context.Context, visionRange string) ([]Commit, error)
 	GetLatestCommit(ctx context.Context) (Commit, error)
 	GetCommitHashForRev(ctx context.Context, rev string) (string, error)
+	GetCommitForRev(ctx context.Context, rev string) (Commit, error)
```
khanhtc1202, line 42: By adding this, I think we should mark the above `GetCommitHashForRev` as deprecated and remove it later, since we can get commit hash if we have commit object provided by this new interface 👀

t-kikuc: thanks, I'll delete it now.
I noticed `GetCommitForRev` is called only from `pkg/git/repo_test.go`.
Deleting it does not affect production code.

t-kikuc:
@khanhtc1202 
Thank you so much for your great reviewing! I fixed all!

khanhtc1202 (commented):
Overall LGTM, only the part "should we separate the skipstage as a new package/module or just place it in scheduler code" is left. IMO, we should place it in the scheduler.go code. Happy to hear other thoughts here 👀 
cc @ffjlabo @Warashi

pkg/app/piped/controller/scheduler.go
```diff
@@ -528,6 +529,24 @@ func (s *scheduler) executeStage(sig executor.StopSignal, ps model.PipelineStage
...
+			s.logger.Error("failed to report stage status", zap.Error(err))
+		}
+		return model.StageStatus_STAGE_FAILURE
+	}
+	if skip {
+		if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_SKIPPED, ps.Requires); err != nil {
+			s.logger.Error("failed to report stage status", zap.Error(err))
+			return model.StageStatus_STAGE_FAILURE
+		}
+		lp.Infof("The stage was successfully skipped due to the skip configuration of the stage.")
```
khanhtc1202, line 546:
```suggestion
		lp.Info("The stage was successfully skipped due to the skip configuration of the stage.")
```
nits. There is no need to use formater here 👀

t-kikuc: Thanks, I fixed it now. [e2e35f4](https://github.com/pipe-cd/pipecd/pull/4922/commits/e2e35f46f4f0fe56cee9c4126b4e34e8f4cac64d)

t-kikuc:
> "should we separate the skipstage as a new package/module or just place it in scheduler code"

I quit separating `skipstage` package because I also think we don't need to separate package here.


- [cab3539](https://github.com/pipe-cd/pipecd/pull/4922/commits/cab353926c7ff57a738d1a38b526200084f2aa49)

- [7eda335](https://github.com/pipe-cd/pipecd/pull/4922/commits/7eda335753b450a3021b05fdc223ca61ff74fee4)

Warashi:
IMO, I agree with khanhtc1202. Because we have no need to hide functions from other controller codes.

pkg/app/piped/controller/skipstage.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func commitMessageHasAnyPrefix(commitMessage string, prefixes []string) bool {
+	for _, prefix := range prefixes {
+		if strings.HasPrefix(commitMessage, prefix) {
+			return true
+		}
+	}
+	return false
+}
```
ffjlabo, line 74: [imo] It might not be a func, just put it in the `skipByCommitMessagePrefixes` ?

pkg/app/piped/controller/skipstage.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	for _, changedFile := range changedFiles {
+		if !matcher.Matches(changedFile) {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
```
ffjlabo, line 103: [imo] It might not be a func, just put it in the `skipByPathPattern`?

pkg/app/piped/controller/skipstage.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/pipe-cd/pipecd/pkg/filematcher"
+	"github.com/pipe-cd/pipecd/pkg/git"
+)
+
+// checkSkipStage checks whether the stage should be skipped or not.
+func checkSkipStage(ctx context.Context, in executor.Input, opt config.SkipOptions) (skip bool, err error) {
+	if len(opt.Paths) == 0 && len(opt.CommitMessagePrefixes) == 0 {
+		// When no condition is specified.
+		return false, nil
+	}
```
ffjlabo, line 32: [imo] It might not be a func, just put it in the `shouldSkipStage` ?

ffjlabo (commented):
Sorry for the late review 🙏 Added some comments.

ffjlabo:
Sorry to be late. I agree with it.
https://github.com/pipe-cd/pipecd/pull/4922#pullrequestreview-2087774132
> should we separate the skipstage as a new package/module or just place it in scheduler code" is left. IMO, we should place it in the scheduler.go code. Happy to hear other thoughts here 👀

pkg/app/piped/controller/skipstage.go
t-kikuc: Thanks, I did it here [8a96a8c](https://github.com/pipe-cd/pipecd/pull/4922/commits/8a96a8c09fb01c1230275795fa3537e38574b7d7)

pkg/app/piped/controller/scheduler.go
```diff
@@ -687,6 +705,25 @@ func (s *scheduler) reportDeploymentCompleted(ctx context.Context, status model.
...
+	case model.StageAnalysis:
+		skipOptions = stageConfig.AnalysisStageOptions.SkipOn
+	case model.StageWait:
+		skipOptions = stageConfig.WaitStageOptions.SkipOn
+	case model.StageWaitApproval:
+		skipOptions = stageConfig.WaitApprovalStageOptions.SkipOn
+	case model.StageScriptRun:
+		skipOptions = stageConfig.ScriptRunStageOptions.SkipOn
+	default:
+		return false, nil
```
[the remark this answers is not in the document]
t-kikuc: I think the current one is enough because
- `StageConfig.GetSkipConfig()` will be used only in `skipstage` package 
- adding the func to each stage model seems redundant

WDYT?

pkg/app/piped/controller/skipstage.go
t-kikuc: You mean we need to mock `repo.ChangedFiles` in `skipByCommitMessagePrefixes`?

pkg/app/piped/controller/scheduler.go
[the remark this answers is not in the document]
ffjlabo: OK, I got it 👍

pkg/app/piped/controller/skipstage.go
ffjlabo: I feel `commitMessageHasAnyPrefix` doesn't have much logic, so IMO, it would be nice to gather the logic to the `skipByCommitMessagePrefixes`.

> You mean we need to mock repo.ChangedFiles in skipByCommitMessagePrefixes?

When testing, we might need to mock as you said.

t-kikuc: @ffjlabo thanks, I merged the func and used mock in tests.
TBH, I missed git.mock.go https://github.com/pipe-cd/pipecd/blob/master/pkg/git/gittest/git.mock.go

pkg/app/piped/controller/skipstage.go
t-kikuc: @ffjlabo thanks, I merged the func and used mock in tests.

Warashi (commented):
If I push several commits at once, this PR's code seems to process only the last commit.
I think all pushed commits should be handled the same way as the last commit. What do you think?

t-kikuc:
@Warashi 

Thank you, that's important.

**a. Path Pattern:** 

This PR already checks ALL changes from the running commit to the latest commit as below.

```go
changedFiles, err := repo.ChangedFiles(ctx, runningRev, targetRev)
```


**b. Commit Message:**

As you said, only the latest commit is checked.

Cases when the latest commit matches the condition but not all commits match it would be:

- Commit&merge frequently 
- Not using a squash merge

cf. `commitMatcher` only checks the triggered commit.
https://github.com/pipe-cd/pipecd/blob/812907842c79b6c26f48fc56cf533d9a47cc24a2/pkg/app/pipedv1/planner/kubernetes/kubernetes.go#L158


@peaceiris What do you think?  Should ALL commits match the condition?

peaceiris:
Only the latest commit is good enough for my use case 👍

t-kikuc:
@peaceiris I got it, thank you so much!

khanhtc1202:
I think the same way, we should only treat the last commit message 👍

khanhtc1202 (approved):
LGTM, can't wait for this in action 👍

Warashi (approved):
Thank you. There is a consensus to see only the last commit message, so LGTM.

PR #4927, opened 2024-05-29
Refactor and add tests for git package

Warashi:

This PR does four things to improve git package.
1. adds test for repository copy method.
2. remove unused private method `getLatestRemoteHashForBranch` from git client
3. fix the variable used in test because it seems typo
4. test not private method but its exported wrapper method

Fixes #

### DIFF
--- a/pkg/git/repo_test.go
+++ b/pkg/git/repo_test.go
@@ -242,3 +242,38 @@ func Test_setGCAutoDetach(t *testing.T) {
 
 	assert.Equal(t, false, got)
 }
+
+func TestCopy(t *testing.T) {
+	faker, err := newFaker()
+	require.NoError(t, err)
+	defer faker.clean()
+
+	var (
+		org      = "test-repo-org"
+		repoName = "repo-copy"
+		ctx      = context.Background()
+	)
+
+	err = faker.makeRepo(org, repoName)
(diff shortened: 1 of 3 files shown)


t-kikuc (approved):
Thank you!!  What a clear commit log!

ffjlabo (approved):
Thank you!

pkg/git/repo_test.go
```diff
@@ -206,7 +206,7 @@ func Test_setGCAutoDetach(t *testing.T) {
 	}
 
 	getGCAutoDetach := func(ctx context.Context, repo *repo) (bool, error) {
-		cmd := exec.CommandContext(ctx, r.gitPath, "config", "--get", "gc.autoDetach")
+		cmd := exec.CommandContext(ctx, repo.gitPath, "config", "--get", "gc.autoDetach")
```
khanhtc1202, line 209: @Warashi Sorry for the late comment, look like the `r` variable in the below line (L210) has the same issue with this fixed one. It works because both refer to the outer `r` variable 👀 Would be great if you could open another PR to fix this.

Warashi: I opened another PR to fix this with other improvements.
https://github.com/pipe-cd/pipecd/pull/4932

PR #4932, opened 2024-05-31
Define in-test function at the head of test

Warashi:

This PR moves the definition of the in-test function to the top of the test.
This makes the in-test function unable to capture variables other than `*testing.T`.

Fixes #

### DIFF
--- a/pkg/git/repo_test.go
+++ b/pkg/git/repo_test.go
@@ -188,6 +188,21 @@ func TestCommitChanges(t *testing.T) {
 }
 
 func Test_setGCAutoDetach(t *testing.T) {
+	getGCAutoDetach := func(ctx context.Context, repo *repo) (bool, error) {
+		cmd := exec.CommandContext(ctx, repo.gitPath, "config", "--get", "gc.autoDetach")
+		cmd.Dir = repo.dir
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			return false, err
+		}
+		v, err := strconv.ParseBool(strings.TrimSuffix(string(out), "\n"))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/git/repo_test.go
```diff
@@ -188,6 +188,21 @@ func TestCommitChanges(t *testing.T) {
...
+		if err != nil {
+			return false, err
+		}
+		v, err := strconv.ParseBool(strings.TrimSuffix(string(out), "\n"))
+		if err != nil {
+			return false, err
+		}
+
+		return v, nil
+	}
```
khanhtc1202, line 204: TBH, I prefer to push the variable/function declaration right before placing it to use, which increases the readability. The closure function outer variable shadowing or reference is a normal thing we have to face if we use it, so I think it's better to capture it in the review process rather than placing it at the head of the function. wdyt?
Happy to hear other opinions on this cc @ffjlabo @t-kikuc

Warashi: IMO, there is a trade-off between the readability of the closure and the codes that use the closure.
When I read the closure itself, closures placed at the top has more readability because there is no captured variable.
On the other hand, when I read codes that use the closure, places at just before it is used has more readability.

In this case, it's not so far between the top of the test function and the line it uses because test is 55 lines of code, not so long.
I think it is short enough to switch eye between definition and call.

khanhtc1202: Okey, this is just a small change, let's choose the win-win option 🙌

khanhtc1202 (approved):
Here you go 🪨

t-kikuc (approved):
🚀

PR #4944, opened 2024-06-05
Setup DCO workflow

Warashi:

This PR adds a workflow to check DCO.
I added this workflow because the app to check DCO is now down.

Fixes #

### DIFF
--- a/.github/workflows/dco.yaml
+++ b/.github/workflows/dco.yaml
@@ -1,0 +1,23 @@
+name: DCO
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Python 3.x
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
+      - name: Check DCO
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          pip3 install -U dco-check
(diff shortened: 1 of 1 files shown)


t-kikuc:
Thanks!
https://github.com/dcoapp/app/issues/211#issuecomment-2148058964 is related.

After the DCO recovers, this action will be removed, right? @Warashi @khanhtc1202

Warashi:
> After the DCO recovers, this action will be removed, right? @Warashi @khanhtc1202

I think so.
I noticed some actions in this workflow is not allowed in this repository, so I'll push some fix.

.github/workflows/dco.yaml
```diff
@@ -0,0 +1,23 @@
+name: DCO
```
t-kikuc, line 1: [IMO] How about specifying this DCO is a workaround?

```suggestion
name: DCO-workaround
```

t-kikuc (commented):
Thank you so much, I commented a nit.

Warashi:
This workflow runs as this.
https://github.com/pipe-cd/pipecd/actions/runs/9377057788/job/25817994492?pr=4944

@khanhtc1202 Since I cannot modify branch protection rules, please change the required checks.

khanhtc1202:
@Warashi Updated the branch protection rules to remove DCO workflow as required step 👍

khanhtc1202 (approved):
Thank you 👍

ffjlabo (approved):
Thanks!

Warashi:
@khanhtc1202 
I merged this PR, so please update the branch protection rules to add DCO-workaround as required.

PR #4946, opened 2024-06-05
Fix gen-contributions.sh: Not to remove FOSSA Status

t-kikuc:

To keep 'FOSSA Status' even after `make gen/contributions` because it needs update due to 
https://github.com/pipe-cd/pipecd/commit/8326ad0668608f041af5a3a1f1757c68aba17056


khanhtc1202 (approved):
Nice catch 👍

Warashi (approved):
Thank you!
I ran `make gen/contributions` locally, and the result seems correct.

PR #4952, opened 2024-06-06
Remove DCO workaround

Warashi:

Since the DCO GitHub App has recovered, we can remove the workaround.

Fixes #


t-kikuc (approved):
Amazing!!!!!!

We need to force the original DCO again.

ffjlabo (approved):
Thank you!

PR #4969, opened 2024-06-13
Remove the annotation for pipecd.dev/resource-key

ffjlabo:

Remove the annotation `pipecd.dev/resource-key` because it is no longer used.

Part of #4269

- **How are users affected by this change**: -


Warashi (commented):
The changes themselves seem good.
By the way, will users face a lot of Out of Sync after merging this PR?
So I think this is a change that needs to be announced carefully.

ffjlabo:
@Warashi 
Thank you for the opinion!

I think the annotation is added when creating or updating the resource.
The old resource holds the annotation at the time even though this PR is merged.
So it does not affect the users.

But I will check the behavior just in case👍

khanhtc1202 (approved):
LGTM 👍

khanhtc1202:
Let's close this due to the revert PR #4997 4997

PR #4971, opened 2024-06-13
Add an RFC for exporting deployment traces

Warashi:

I'd be happy to consider another way to collect deployment statuses.

Part of #4977 


khanhtc1202:
Hi, @Warashi, Thanks for the idea 👍
Could you add more information about which metrics you plan to extract/collect, and how we can use it in order to monitor the deployment? (aka. How will we monitor the deployment with otlp?)

Warashi:
@khanhtc1202 Thank you for the advice!
I added some examples.
I have not planned a detailed way to send traces from the collector to hosting, so please give me time to investigate the collector's configurations.

Warashi:
After implementing PoC on my local machine, I changed the way piped is authenticated.
In the previous proposal, OpenTelemetry Collector had to implement a custom authenticator.
This prevented users from using their custom collector.
The envoy's external authentication filter is used to authenticate piped in the new proposal.
So, users can use any collector with a built-in OTLP gRPC receiver.

docs/rfcs/0013-export-deployment-traces-with-otlp.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Detailed design
+
+## Architecture
+
+Collect spans at piped
+→ send to pipecd-gateway envoy
+→ proxies to OpenTelemetry Collector
+→ send anywhere with exporters
+
```
ffjlabo, line 26: [ask]
This is just curious.
In this RFC, do you consider the case when users want to use their own otel collector?

ffjlabo (commented):
Thank you for the interesting RFC! I added the asking comment.

docs/rfcs/0013-export-deployment-traces-with-otlp.md
Warashi: If you mean 'users' is control plane managers, the answer is 'Yes.'
If you mean 'users' is piped managers, the answer is 'No.'

When control plane managers want to send traces to SaaS, the collector must include exporters to send them. They may need to select the collector image and configure it.

ffjlabo: > If you mean 'users' is control plane managers, the answer is 'Yes.'
If you mean 'users' is piped managers, the answer is 'No.'

OK, I got your point!
So it might be nice to add an explanation that this is for the control plane managers.

ffjlabo: I thought some 'piped managers' wanted to check the traces and send their own maintained otel-collector to check the deployment.

Warashi: I added mention of users this RFC targets.
https://github.com/pipe-cd/pipecd/pull/4971/commits/c59f4f857ffff3b16e92bec4cb17663251f5d70a

Warashi: > I thought some 'piped managers' wanted to check the traces and send their own maintained otel-collector to check the deployment.

I agree, but I want to implement the method for control plane managers first.

ffjlabo (approved):
LGTM 👍

khanhtc1202 (approved):
LGTM 👍

PR #5002, opened 2024-06-28
Add default env for script run

ffjlabo:
**What this PR does / why we need it**:

Added default env value for the deployment information (such as deployment id, application id...)

Current values are below.

| Name | Description | Example |
|-|-|-|
|SR_DEPLOYMENT_ID| The deployment id | 877625fc-196a-40f9-b6a9-99decd5494a0 |
|SR_APPLICATION_ID| The application id | 8d7609e0-9ff6-4dc7-a5ac-39660768606a |
|SR_APPLICATION_NAME| The application name | example |
|SR_TRIGGERED_AT| The timestamp when the deployment is triggered  | 1719571113 |
|SR_TRIGGERED_COMMIT_HASH| The commit hash that triggered the deployment | 2bf969a3dad043aaf8ae6419943255e49377da0d |
|SR_REPOSITORY_URL| The repository url configured in the piped config  | git@github.com:org/repo.git, https://github.com/org/repo |
|SR_SUMMARY| The summary of the deployment | Sync with the specified pipeline because piped received a command from user via web console or pipectl|
|SR_CONTEXT_RAW| The json encoded string of above values | {"deploymentID":"877625fc-196a-40f9-b6a9-99decd5494a0","applicationID":"8d7609e0-9ff6-4dc7-a5ac-39660768606a","applicationName":"example","triggeredAt":1719571113,"triggeredCommitHash":"2bf969a3dad043aaf8ae6419943255e49377da0d","repositoryURL":"git@github.com:org/repo.git","labels":{"env":"example","team":"product"}} |
|SR_LABEL_XXX| The label attached to the deployment. The env name depends on the label name. For example, if a deployment has the labels `env:prd` and `team:server`, `SR_LABEL_ENV` and `SR_LABEL_TEAM` are registered.  | prd, server |

PipeCD

example of app.pipecd.yaml
```
apiVersion: pipecd.dev/v1beta1
kind: KubernetesApp
spec:
  name: script-run-like-jenkins
  labels:
    env: example
    team: product
  pipeline:
    stages:
      - name: K8S_CANARY_ROLLOUT
        with:
          replicas: 10%
      - name: SCRIPT_RUN
        with:
          run: |
            echo $SR_DEPLOYMENT_ID
            echo $SR_APPLICATION_ID
            echo $SR_APPLICATION_NAME
            echo $SR_TRIGGERED_AT
            echo $SR_TRIGGERED_COMMIT_HASH
            echo $SR_REPOSITORY_URL
            echo $SR_RAW
[...]

Issue #4814, opened 2024-03-10
Use current deployment context in SCRIPT_RUN stage


minhquang4334:
**What would you like to be added**:
Use current deployment context (like DeploymentTrigger, DeploymentStatus, etc.) in SCRIPT_RUN stage.
refs: https://pipecd.dev/docs-v0.46.x/user-guide/managing-application/customizing-deployment/script-run/

**Why is this needed**:
Since we don't know about the current deployment, it is difficult to write a script for customization of the deployment process. 

This is similar to GitHubAction, where we can access job and step contexts through a defined property.
refs: https://docs.github.com/en/actions/learn-github-actions/contexts#job-context

peaceiris:
I look forward to this feature!

In our use case, we want to get the `Summary` including a docker image tag.

ffjlabo:
@minhquang4334 @peaceiris 
Sorry for the late response 🙏 

We can realize it by adding the default environment variable for `SR_XX` (like `SR_DEPLOYMENT_ID`).
How about this?

```
apiVersion: pipecd.dev/v1beta1
kind: KubernetesApp
spec:
  name: script-run
  labels:
    env: example
    team: product
  pipeline:
    stages:
      - name: K8S_CANARY_ROLLOUT
        with:
          replicas: 10%
      - name: WAIT
        with:
          duration: 10s
      - name: SCRIPT_RUN
        with:
          run: |
            echo $SR_DEPLOYMENT_ID
      - name: K8S_PRIMARY_ROLLOUT
      - name: K8S_CANARY_CLEAN
```

peaceiris:
> environment variable for SR_XX (like SR_DEPLOYMENT_ID).

LGTM ʕ◔ϖ◔ʔ

minhquang4334:
> We can realize it by adding the default environment variable for SR_XX (like SR_DEPLOYMENT_ID).

LGTM @ffjlabo

ffjlabo:
Thanks, both 👍 I will try it.

Also, I received a comment from @khanhtc1202 about the implementation.
We should check whether the env values are shared by multiple deployments.

So at first, I will investigate it :)

ffjlabo:
After the investigation, I found out the way is possible. 👍 I will implement it later.

**Detail**
As the official document, `os/exec` package executes the command as a new process.
> Package exec runs external commands. It wraps os.StartProcess to make it easier to remap stdin and stdout, connect I/O with pipes, and do other adjustments.
[...]

PR #5009, opened 2024-07-01
Install jq in piped-base images

Warashi:

This PR installs jq to the piped-base images.

PR #5002 introduces environment variables in the SCRIPT_RUN stage.
One of these environment variables has JSON format, so we need jq to deal with it.
After merging this PR, we need to update other Dockefiles to use new piped-base images.

Part of #4814 


ffjlabo (approved):
Thank you 🙏

khanhtc1202 (approved):
🚀

PR #5018, opened 2024-07-02
Fix the docs for the script run feature status

ffjlabo:

Fixed the description of the script run feature.
Currently, we can use the script run stage without the rollback feature on all kinds of applications.

**background**
I implemented the script run stage without rollback feature In this PR https://github.com/pipe-cd/pipecd/pull/4720.
But later, I implemented the rollback feature for the k8s app: https://github.com/pipe-cd/pipecd/pull/4743.

Part of #4643


t-kikuc (approved):
🚀

Warashi (approved):
👍🏻

Issue #4980, opened 2024-06-19
Plugin architectured piped (pipedv1) supports deployment for applications


khanhtc1202:
**What would you like to be added**:

This issue is for managing the process of adopting plugin architecture to piped. The ideal result is to make all platform providers' logic independent from the piped core logic (the controller package).

Tasks pool break down 💪 

- [x] Revise the plugin planner interface <- @khanhtc1202 @Warashi 
- [x] Implement piped plugin grpc service <- @khanhtc1202 
- [x] Implement piped plugin grpc secret decryption endpoint <- @khanhtc1202 
- [x] Implement piped planner which calls to the plugin deployment service <- @khanhtc1202 
- [x] Implement k8s plugin: PlannerService interface implementation <- @Warashi 
- [x] Update model.PipelineStage model:
  - [x] mark `Predefined` and `Visible` as deprecated
  - [x] add `rollback` field
  - [x] add `timeout` field

**Why is this needed**:

khanhtc1202:
As the current stage, the interface for the plugin of the piped as below
We have 2 services:
The PlannerService
```proto
// PlannerService defines the public APIs for remote planners.
service PlannerService {
    // BuildPlan builds plan for the given deployment.
    rpc BuildPlan(BuildPlanRequest) returns (BuildPlanResponse) {}
}

message BuildPlanRequest {
    string working_dir = 1 [(validate.rules).string.min_len = 1];
    // Last successful commit hash and config file name.
    // Use to build deployment source object for last successful deployment.
    string last_successful_commit_hash = 2;
    string last_successful_config_file_name = 3;
    // The configuration of the piped that handles the deployment.
    bytes piped_config = 4 [(validate.rules).bytes.min_len = 1];
    // The deployment to build a plan for.
    model.Deployment deployment = 5 [(validate.rules).message.required = true];
}

message BuildPlanResponse {
    // The built deployment plan.
    DeploymentPlan plan = 1;
}

message DeploymentPlan {
    model.SyncStrategy sync_strategy = 1;
    // Text summary of planned deployment.
    string summary = 2;
    repeated model.ArtifactVersion versions = 3;
    repeated model.PipelineStage stages = 4;
}
```
And, the ExecutorService
```proto
service ExecutorService {
    // Execute executes the given stage of the deployment plan.
    rpc ExecuteStage(ExecuteStageRequest) returns (stream ExecuteStageResponse) {}
}

message ExecuteStageRequest {
[...]

PR #5020, opened 2024-07-03
Implement QuickSyncPlan for k8s plugin

Warashi:

This PR implements QuickSyncPlan RPC for the Kubernetes plugin.

Part of https://github.com/pipe-cd/pipecd/issues/4980


pkg/app/pipedv1/plugin/sourcecloner/cloner.go
```diff
@@ -0,0 +1,82 @@
...
+	if _, err := git.RunGitCommand(ctx, c.gitPath, "", os.Environ(), "clone", c.remotePath, dest); err != nil {
+		return err
+	}
+	if c.revision == "" {
+		if _, err := git.RunGitCommand(ctx, c.gitPath, "", os.Environ(), "-C", dest, "checkout", c.revision); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
Warashi, line 72: I want to write tests for this method with [git.faker](https://github.com/pipe-cd/pipecd/blob/281e85470a5c762cd9f755271d6f16e1d6755812/pkg/git/client_test.go#L95-L98).
However, the PR will become larger if I make git.faker public.
So, I decide not to write the test in this PR.
I'll make git.faker public later, and then write the test.

pkg/app/pipedv1/plugin/sourcecloner/cloner.go
```diff
@@ -0,0 +1,82 @@
...
+	return path, nil
+})
+
+// Cloner is a source cloner for git repositories.
+type Cloner struct {
+	gitPath      string
+	remotePath   string
+	revision     string
+	revisionName string
+}
```
khanhtc1202, line 43: @Warashi I have an idea with this sourcecloner package 🤔 
Basically, our plan is to use the `piped/deploysource` as the portal for all git resource operations (indeed, the name is deploysource and it emits the internal using git stub).
I think this newly created package `sourcecloner` should follow that approach and more. What package dose is to fetch the already cloned source (prepared by deploysource), then I think we should emits the internal using git as same as we do in deploysource.

khanhtc1202: Instead of a `Clone` interface, which could lead to confusion (if the reviewer doesn't know we don't clone or any, just fetch the cloned local source prepared by deploysource), how about to reuse the SourceCloner interface. aka this: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/deploysource/sourcecloner.go#L24-L28
I can see we have some implementation of this interface looks pretty fits our use case here: 
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/deploysource/sourcecloner.go#L69-L94

Warashi: Oops, I missed the localSourceCloner. Thank you!
I think we can use this, but I wonder why this `Clone` implementation only invokes the Copy method of the internal git repo.
If the internal git repo's current HEAD doesn't match localSourceCloner's revision, this Clone method will give the wrong results.
Or have I misunderstood the `Revision()` of SourceCloner?

khanhtc1202: @Warashi That's a good catch 👍 I think up to now, the localSourceCloner not yet raised any issues because its caller doesn't require any other revision than the HEAD 🤔 So, basically I think it's a bug potential logic and should be fixed, but as far as I investigated, the only place which we use this localSourceCloner is `planpreview/builder` (ref: this code: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/planpreview/builder.go#L228), which basically only request for the HEAD of the current branch - due to the feature plan preview logic.
So, my point is to fix this localSourceCloner and use it for whatever "NOT git action pull directly but pull/fetch from a local source". WDYT? 🤔

khanhtc1202: My point is: we should not create too many interfaces/packages that do similar tasks (clone source from git), or it will lead to confusion for the caller side. In this case, it would be better to have only one SourceCloner interface, one pull/fetch to the outer git, one pull/fetch to the local repo that was pulled by the previous implementation.

Warashi: I agree with you.
I'll fix the localSourceCloner and use it.

pkg/app/pipedv1/plugin/inputs.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"os/exec"
+
+	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/deploysource"
+	"github.com/pipe-cd/pipecd/pkg/git"
+	"github.com/pipe-cd/pipecd/pkg/plugin/api/v1alpha1/platform"
+)
+
+func GetPlanSourceCloner(input *platform.PlanPluginInput) (deploysource.SourceCloner, error) {
```
Warashi, line 11: 🤔 I wonder where we should place utilities like this function.
I place it here because all plugins may need this utility, but should I create another package under the plugin package?

pkg/git/client.go
```diff
@@ -252,6 +252,11 @@ func (c *client) envsForRepo(remote string) []string {
 	return append(envs, c.gitEnvs...)
 }
 
+// RunGitCommand runs a git command with the given arguments.
```
khanhtc1202, line 255: we don't use this anymore 👀

pkg/app/pipedv1/plugin/inputs.go
```diff
@@ -0,0 +1,24 @@
+package plugin
```
khanhtc1202, line 1: Please add the file header copyright

pkg/app/pipedv1/plugin/inputs.go
khanhtc1202: I got your point, let's find out later. In this PR, place it here is LGTM 👍

pkg/git/client.go
Warashi: Thank you. I removed it on this commit
https://github.com/pipe-cd/pipecd/pull/5020/commits/355ce05ef71a28865fb680c31fe98ac0a6a5556d

pkg/app/pipedv1/plugin/inputs.go
Warashi: Thank you.
I added it with this commit
https://github.com/pipe-cd/pipecd/pull/5020/commits/0b562e4cb0aed3d16f412c0e2222f42a7df566ff

khanhtc1202 (dismissed):
LGTM with some nits, thank you 🚀 
(we can fix nits with other PRs)

pkg/app/pipedv1/plugin/platform/kubernetes/planner/server.go
```diff
@@ -63,7 +77,39 @@ func (ps *PlannerService) DetermineStrategy(ctx context.Context, in *platform.De
...
+	cloner, err := plugin.GetPlanSourceCloner(in.GetInput())
+	if err != nil {
+		return nil, err
+	}
+
+	d, err := os.MkdirTemp("", "") // TODO
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare temporary directory (%w)", err)
+	}
+	defer os.RemoveAll(d)
```
khanhtc1202, line 91: @Warashi Question:
Should we add workingDir as PlanPluginInput? 🤔 Currently, the planner in the controller uses workingDir passed at the time controller.planner object is created, which mean we have 1:1 for relationship between workingDir & planner which is used to build plan for a deployment
ref: 
https://github.com/pipe-cd/pipecd/blob/6400c898654e41ac80b6c2ed5a345a8b0bc080a7/pkg/app/piped/controller/controller.go#L438-L443
https://github.com/pipe-cd/pipecd/blob/6400c898654e41ac80b6c2ed5a345a8b0bc080a7/pkg/plugin/api/v1alpha1/platform/api.proto#L63

khanhtc1202: If we make it that way, then the piped.planner object will create the workingDir then clone/pull the deploy source, then we pass the workingDir (which contains cloned source) to the planner plugin via PlanPluginInput.

pkg/app/pipedv1/plugin/platform/kubernetes/planner/server.go
```diff
@@ -63,7 +77,39 @@ func (ps *PlannerService) DetermineStrategy(ctx context.Context, in *platform.De
...
+		cloner,
+		*in.GetInput().GetDeployment().GetGitPath(),
+		ps.Decrypter,
+	)
+
+	ds, err := p.GetReadOnly(ctx, io.Discard /* TODO */)
+	if err != nil {
+		return nil, err
+	}
+
```
khanhtc1202, line 104: nits
```go
	cfg := ds.ApplicationConfig.KubernetesApplicationSpec
	if cfg == nil {
		err = fmt.Errorf("missing KubernetesApplicationSpec in application configuration")
		return
	}
```
Should check spec available to avoid nil pointer 👀

pkg/app/pipedv1/plugin/platform/kubernetes/planner/server.go
Warashi: That sounds nice.
In this PR, we clone the deploy source manually.
However, it's needed for each plugin implementation, so it's simpler if we handle it at piped and only use the prepared source in the plugin.

khanhtc1202: Ok, let address it by another PR 👍

pkg/app/pipedv1/plugin/platform/kubernetes/planner/server.go
```diff
@@ -63,7 +77,39 @@ func (ps *PlannerService) DetermineStrategy(ctx context.Context, in *platform.De
...
+	autoRollback := false
+	if a := ds.ApplicationConfig.KubernetesApplicationSpec.Input.AutoRollback; a != nil {
+		autoRollback = *a
+	}
```
khanhtc1202, line 108: nits, we have a default value set as true, so no need to initialize autoRollback var as false here, since the `ds.ApplicationConfig.KubernetesApplicationSpec.Input.AutoRollback` is true as init
ref: ds.ApplicationConfig.KubernetesApplicationSpec.Input.AutoRollback

Warashi: Thank you.
I noticed that I can view the document here.
https://pkg.go.dev/github.com/pipe-cd/pipecd/pkg/config#KubernetesDeploymentInput

pkg/app/pipedv1/plugin/platform/kubernetes/planner/server.go
Warashi: I fixed it on this commit.
https://github.com/pipe-cd/pipecd/pull/5020/commits/d845f7a1e293dea974f1544db818dff83fbe5531

pkg/app/pipedv1/plugin/platform/kubernetes/planner/server.go
Warashi: I fixed it on this commit.
https://github.com/pipe-cd/pipecd/pull/5020/commits/d845f7a1e293dea974f1544db818dff83fbe5531

ffjlabo (approved):
🚀

khanhtc1202 (approved):
👍👍

PR #5022, opened 2024-07-03
Update release docs

khanhtc1202:

Fixes #

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -48,13 +48,15 @@ This refers to the release of new features or breaking changes.
 - Create a tagged release from the master branch. The release should start with "v" and be followed by the version number.
 
 ## Minor release
-This refers to the release of new features.
+This refers to the release of new features (breaking change may be included).
 
 Please refer to [Major release](https://github.com/pipe-cd/pipecd/blob/master/RELEASES.md#major-release) for the processes.
 
 ## Patch release
 This refers to the release of critical bug fixes. \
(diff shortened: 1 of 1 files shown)


t-kikuc (approved):
thanks

RELEASES.md
```diff
@@ -48,14 +48,16 @@ This refers to the release of new features or breaking changes.
 - Create a tagged release from the master branch. The release should start with "v" and be followed by the version number.
 
 ## Minor release
-This refers to the release of new features.
+This refers to the release of new features (breaking change may be included).
```
Warashi, line 51: What do you think about breaking changes in a minor version after v1.0.0?
It's recommended not to release that after v1, but we are allowed before v1.
So I suggest mentioning breaking changes in minor versions are limited to before v1.

```suggestion
This refers to the release of new features (breaking change may be included before v1.0.0).
```

khanhtc1202: Hmm, I think it's a bit hard to ensure that all minor releases (even after v1.0.0) do not contain any breaking changes. Since this forces us to only release breaking changes as v2.0.0 and so on, I think it's really difficult to archive. WDYT? @Warashi

khanhtc1202: @Warashi, FYI, we call any changes that require users' action on the upgrades a breaking change (for example, https://github.com/pipe-cd/pipecd/releases/tag/v0.33.0). But, with this change like this as a new major version - I don't think so 🤔

Warashi: I agree. I don't like to release a new major version because of changes like this.

Warashi (approved):
Thank you

PR #5025, opened 2024-07-04
Fix: Update latest version of docs to v0.48.x

t-kikuc:

The latest version of docs is still v0.47.x although we released v0.48.0.
The cause is not clear yet.

I executed the following command to fix it.
https://github.com/pipe-cd/pipecd/blob/6439f977a8c3a327bdb84d86eceb7686119d9ad6/hack/gen-release-docs.sh#L63-L64

N/A

### DIFF
--- a/docs/main.go
+++ b/docs/main.go
@@ -28,7 +28,7 @@ import (
 const dir = "/public"
 
 // Don't update here manually. /hack/gen-release-docs.sh does.
-const latestPath = "/docs-v0.47.x/"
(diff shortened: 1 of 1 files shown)


Warashi (approved):
Thank you

Warashi:
The cause is ./hack/gen-release-docs.sh expects sed is the BSD version, it's the default of macOS.
I'm using the GNU version, and its `-i` option behavior is different from the BSD version.

ffjlabo (approved):
🚀

PR #5031, opened 2024-07-10
Add notes of linking target groups and ELB listener rules in advance

t-kikuc:

Add notes that a user needs to link target groups and ELB listener rules in advance.

If not, a piped will not control traffic weights and a new taskset cannot receive traffic.

-

- **Is this breaking change**:no


khanhtc1202 (approved):
LGTM

Warashi (approved):
LGTM

PR #5035, opened 2024-07-12
Make it clear to attach IAM policies to ECS task-role

t-kikuc:

Make it clear which role to attach IAM policies to because ECS's `task-role` and `task-execution-role` are confusing.

### DIFF
--- a/docs/content/en/docs-dev/installation/install-piped/required-permissions.md
+++ b/docs/content/en/docs-dev/installation/install-piped/required-permissions.md
@@ -8,6 +8,8 @@ description: >
 
 A Piped requires some permissions to deploy applications, depending on the platform.
 
+Note: If you run a piped as an ECS task, you need to attach the permissions on the piped task's `task role`, not `task execution role`.
+
 ## For ECSApp
 
 You need IAM actions like the following example. You can restrict `Resource`.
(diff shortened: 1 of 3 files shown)


khanhtc1202 (approved):
👍

Warashi (commented):
How about removing this line?
https://github.com/pipe-cd/pipecd/blob/7ca5c05b57b7f8e25006b708be799b2fb44223ac/docs/content/en/docs-dev/installation/install-piped/required-permissions.md?plain=1#L49

t-kikuc:
@Warashi 

> How about removing this line?
> 
> https://github.com/pipe-cd/pipecd/blob/7ca5c05b57b7f8e25006b708be799b2fb44223ac/docs/content/en/docs-dev/installation/install-piped/required-permissions.md?plain=1#L49

It’s mandatory.
When a piped creates a new task, the piped needs to pass task-execution-role to the task.

Warashi:
> It’s mandatory.
> When a piped creates a new task, the piped needs to pass task-execution-role to the task.

Sorry, I misread.

As I understood, the `Resource` field at this line does not mean the role to which we add permission but the role we use to run the tasks.
So we need this line.

Warashi (approved):
LGTM

PR #5043, opened 2024-07-16
Add working_dir to PlanPluginInput

Warashi:

[As discussed in this comment](https://github.com/pipe-cd/pipecd/pull/5020#discussion_r1673704611), I add the `working_dir` field to the PlanPluginInput message.

Fixes #


pkg/plugin/api/v1alpha1/platform/api.proto
```diff
@@ -71,6 +71,9 @@ message PlanPluginInput {
     string last_successful_config_file_name = 4;
     // The configuration of plugin that handles the deployment.
     bytes plugin_config = 5;
+    // The workking directory to use in plugin.
+    // This directory contains the cloned deploy source with the directory name "deploy_source".
+    string working_dir = 6;
```
Warashi, line 76: @khanhtc1202 
I think the working dir passed to the plugin should contain the deploy source.
On the other hand, this working_dir is used with `os.MkdirTemp`, so we should not place the deploy source at the root of this directory.
So, I documented that the deploy source is placed within the directory named "deploy_source."
WDYT?

khanhtc1202: I think that's fine 👍

Warashi:
I think it's simpler than this PR to pass the DeploySource information directory to the plugin.
So we start from that point and then pass workspace_dir, working_dir, temp_dir, or something similar if we need it.

PR #5047, opened 2024-07-16
Add DetermineVersions rpc to PlannerService

khanhtc1202:

Follow PR #5015 


Warashi (approved):
LGTM

MEMO:
The implementation will be something like this.
https://github.com/pipe-cd/pipecd/blob/5acc6e125206c923bcb6b6ef6ce81396e35aa2b2/pkg/app/piped/planner/kubernetes/kubernetes.go#L501-L503

ffjlabo (approved):
🚀

PR #5049, opened 2024-07-17
Add a labeler rule of area/pipedv1

t-kikuc:

- This rule will tag `area/pipedv1` when a PR contains any change under pkg/app/pipedv1/ recursively.
- I want to clarify whether a PR is made for pipedv1.

N/A


khanhtc1202 (approved):
thanks 👍

Warashi (approved):
thank you

PR #5050, opened 2024-07-17
Pass the DeploySource directly to the plugin

Warashi:

This PR modifies the plugin input.
The piped passes the deploy source information directly to the plugin, and the plugin expects the piped to prepare the deploy source before invoking the RPC methods.
The application spec is defined in bytes because we don't want to define the spec's structure for each plugin.

This PR does not touch the deploysource package because it's widely used, and the type `DeploySource` is not only used in the pipedv1 but also in other packages.

Fixes #


pkg/plugin/api/v1alpha1/platform/api.proto
```diff
@@ -74,14 +74,34 @@ message PipelineSyncPlanResponse {
...
+
+message DeploySource {
+    // The repository directory to deploy for.
+    string repo_dir = 1;
+    // The applicaiton directory to deploy for.
+    string app_dir = 2;
+    // The commit hash of the deployment source.
+    string revision = 3;
+    // The application configuration of the deployment.
+    ApplicationConfig config = 4;
```
khanhtc1202, line 93: I'm not sure why we should not make this `bytes config = 4` instead? In the revoking side, we have to import `pkg/config` package anyway 🤔 (ref: https://github.com/pipe-cd/pipecd/pull/5050/files#diff-74d9cfa389294bac8e44026eba99fca2d6e82e7827efcddec084ab732491656aR23)

pkg/plugin/api/v1alpha1/platform/api.proto
```diff
@@ -74,14 +74,34 @@ message PipelineSyncPlanResponse {
...
-    // Use to build deployment source object for last successful deployment.
-    string last_successful_commit_hash = 3;
-    string last_successful_config_file_name = 4;
     // The configuration of plugin that handles the deployment.
-    bytes plugin_config = 5;
+    bytes plugin_config = 2;
+    // The source of the deployment.
+    DeploySource deploy_source = 3;
+    // The source of the last successful deployment.
+    DeploySource last_successful_deploy_source = 4;
```
khanhtc1202, line 82:
```suggestion
    // The source of the target state.
    DeploySource target_deploy_source = 3;
    // The source of the running state.
    DeploySource running_deploy_source = 4;
```
How about renaming like this 👀

pkg/plugin/api/v1alpha1/platform/api.proto
khanhtc1202: I get your point now 👍

Warashi:
I have not been able to work on this PR for a while.
It is tough to resolve conflicts, so I will close it and work on it again.

PR #5051, opened 2024-07-17
Delete platform specific codes from pipedv1

Warashi:

This PR deletes the platform-specific codes from pipedv1.
Because pipedv1 uses a plugin architecture to implement platform-specific operations, platform-specific codes are not needed except for plugin implementations.

Fixes #


khanhtc1202 (approved):
🚀

ffjlabo (approved):

PR #5052, opened 2024-07-17
Pipedv1 using auto rollback config from spec.Planner instead

khanhtc1202:

The auto rollback configuration is generic for all applications kinds, so it should be placed in `spec.Planner` instead of `spec(Kind).Input`

Fixes #

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -64,6 +64,9 @@ type DeploymentPlanner struct {
 	// Disable auto-detecting to use QUICK_SYNC or PROGRESSIVE_SYNC.
 	// Always use the speficied pipeline for all deployments.
 	AlwaysUsePipeline bool `json:"alwaysUsePipeline"`
+	// Automatically reverts all deployment changes on failure.
+	// Default is true.
+	AutoRollback *bool `json:"autoRollback,omitempty" default:"true"`
 }
 
 type Trigger struct {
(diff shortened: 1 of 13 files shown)


t-kikuc (dismissed):

pkg/config/application.go
```diff
@@ -64,6 +64,9 @@ type DeploymentPlanner struct {
 	// Disable auto-detecting to use QUICK_SYNC or PROGRESSIVE_SYNC.
 	// Always use the speficied pipeline for all deployments.
 	AlwaysUsePipeline bool `json:"alwaysUsePipeline"`
+	// Automatically reverts all deployment changes on failure.
+	// Default is true.
+	AutoRollback *bool `json:"autoRollback,omitempty" default:"true"`
```
ffjlabo, line 69: [imo] How about adding the comment to explain this is only for the v1 for now? It's because we avoid using it on the current version.

khanhtc1202: I think we can add a deprecated note to the `Input.AutoRollback` 👍

khanhtc1202: @ffjlabo Done by https://github.com/pipe-cd/pipecd/pull/5052/commits/ebe5f5f786717c793a64cab0ac7a008e09cbfce8

ffjlabo (approved):
🚀

t-kikuc (approved):

PR #5062, opened 2024-07-21
Implement POC for planner of pipedv1

khanhtc1202:

This PR provides POC implementation for the Planner of pipedv1 (piped side). The logic of building plan as follow
```go
// The strategy determination logic is based on the following order:
//   - Direct trigger via web console
//   - Force quick sync if there is no pipeline specified
//   - Force pipeline if the `spec.planner.alwaysUsePipeline` was configured
//   - CommitMatcher ensure pipeline/quick sync based on the commit message
//   - Force quick sync if there is no previous deployment (aka. this is the first deploy)
//   - Based on PlannerService.DetermineStrategy returned by plugins
```

Part of #4980

### DIFF
--- a/pkg/app/pipedv1/controller/planner.go
+++ b/pkg/app/pipedv1/controller/planner.go
@@ -210,6 +210,14 @@ func (p *planner) Run(ctx context.Context) error {
 	return p.reportDeploymentPlanned(ctx, out)
 }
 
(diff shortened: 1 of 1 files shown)


khanhtc1202:
Close due to interface changed :ok:

PR #5063, opened 2024-07-22
Support `EphemeralStorage` in TaskDefinition [ECS]

t-kikuc:

Enabled to configure `ephemeralStorage` in TaskDefinition.

Because it is not supported yet.

N/A

**How to use it**

You can configure like this in your taskdef.yaml:
image


khanhtc1202 (approved):
👍

Warashi (approved):
🚀

PR #5065, opened 2024-07-23
Remove platform provider migration ops task

khanhtc1202:

This migration task has existed over 2 years, and users are expected to finished the migration/update already 😉 

Fixes #


t-kikuc (approved):
Thanks!

I understood the `runner.go` was added in https://github.com/pipe-cd/pipecd/pull/3767 on Jun 30, 2022.

We should release this in the next minor with some announce.

ffjlabo (approved):

PR #5096, opened 2024-07-30
Set fetch-depth as 0 to create correct patches during git cherry-pick as much as possible

ffjlabo:

Set fetch-depth for actions/checkout as 0 to create correct patches during git cherry-pick as much as possible.
https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches

By default, actions/checkout set fetch-depth as 1. It means that only the latest commit is fetched.
In the cherry-pick process, some patches are created based on the commits in the fetched repo and merged into the target branch.
It would be nice to fetch commits as much as possible to avoid conflict with cherry-pick.

Fixes #

### DIFF
--- a/.github/workflows/cherry_pick.yaml
+++ b/.github/workflows/cherry_pick.yaml
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


khanhtc1202 (approved):
👍

t-kikuc (approved):
👍

PR #5108, opened 2024-08-02
Use LRUCache for Application Manifests Cache

Warashi:

Use LRUCache for the application manifests cache instead of TTLCache to control memory consumption.

N/A

### DIFF
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -103,6 +103,7 @@ type piped struct {
 	addLoginUserToPasswd                 bool
 	launcherVersion                      string
 	maxRecvMsgSize                       int
+	appManifestCacheCount                int
 }
 
 func NewCommand() *cobra.Command {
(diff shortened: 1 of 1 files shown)


khanhtc1202 (dismissed):

pkg/app/piped/cmd/piped/piped.go
```diff
@@ -355,7 +358,11 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 	analysisResultStore := analysisresultstore.NewStore(apiClient, input.Logger)
 
 	// Create memory caches.
-	appManifestsCache := memorycache.NewTTLCache(ctx, time.Hour, time.Minute)
+	appManifestsCache, err := memorycache.NewLRUCache(150)
```
ffjlabo, line 361:
```suggestion
	appManifestsCache, err := memorycache.NewLRUCache(p.appManifestCacheCount)
```

ffjlabo (changes requested):

pkg/app/piped/cmd/piped/piped.go
ffjlabo: This is to use the cache count from flag. 🙏

Warashi: Thanks. I commited the suggestion.

ffjlabo (approved):
LGTM 🙆‍♂️

khanhtc1202 (approved):

PR #5111, opened 2024-08-05
POC Piped v1 planner with plugin deployment interface

khanhtc1202:

Part of #4980


pkg/config/application.go
```diff
@@ -223,6 +223,7 @@ type PipelineStage struct {
 	Name    model.Stage
 	Desc    string
 	Timeout Duration
+	With    json.RawMessage
```
khanhtc1202, line 226: Only this part is the only change that not under the pipedv1 directory. This is for packing stage configuration (not unmarshall) and pass it as is to the plugin.

Warashi (approved):
Thank you

t-kikuc (approved):

PR #5112, opened 2024-08-06
Add Deploysource as proto model

Warashi:

Retry of #5050 
I noticed that the piped also uses the Deploysource, so I defined it not in the plugin proto but in the model proto.

Part of #4980


pkg/model/deployment_source.proto
```diff
@@ -0,0 +1,45 @@
...
+
+message DeploymentSource {
+  // The application directory where the source code is located.
+  string application_directory = 1;
+  // The git commit revision of the source code.
+  string revision = 2;
+  // The configuration of the application which is independent for plugins.
+  GenericApplicationSpec generic_application_config = 3;
+  // The configuration of the application which is specific for plugins.
+  PluginApplicationSpec application_config = 4;
```
khanhtc1202, line 28: For these two, I think we need the data structure under the `pkg/config` package. So, how about passing this as bytes to the plugin, and the plugin will import the pipecd `pkg/config` package to unmarshal it? In that case, we can reuse the logic of the marshal/unmarshal config we currently have.

khanhtc1202: I mean, reuse this whole: https://github.com/pipe-cd/pipecd/blob/master/pkg/config/application.go#L29-L61

pkg/model/deployment_source.proto
```diff
@@ -0,0 +1,45 @@
...
+  // The git commit revision of the source code.
+  string revision = 2;
+  // The configuration of the application which is independent for plugins.
+  GenericApplicationSpec generic_application_config = 3;
+  // The configuration of the application which is specific for plugins.
+  PluginApplicationSpec application_config = 4; 
+}
+
+message GenericApplicationSpec {
+  // TODO: Add more fields to support generic application spec.
```
khanhtc1202, line 32: Screenshot 2024-08-06 at 11 15 31
I just realized the plugin does not need the GenericApplicationSpec at all 🤔

Warashi: I removed GenericApplicationSpec from the model because the plugin does not need it, as you pointed out.

pkg/model/deployment_source.proto
Warashi: I removed GenericApplicationSpec.
PluginApplicationSpec's fields are api_version, kind, and spec.
The api_version and kind are common for all plugins, so I defined as fields.
The field spec is defined as bytes, as you pointed out.

pkg/model/deployment_source.proto
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package model;
+
+option go_package = "github.com/pipe-cd/pipecd/pkg/model";
+
+message DeploymentSource {
```
khanhtc1202, line 21: Note: repoDir is removed because it is only being used in the deploysource package. We will change our way to clone/prepare source later.

pkg/model/deployment_source.proto
```diff
@@ -0,0 +1,40 @@
...
+  // The application directory where the source code is located.
+  string application_directory = 1;
+  // The git commit revision of the source code.
+  string revision = 2;
+  // The configuration of the application which is specific for plugins.
+  PluginApplicationSpec application_config = 3;
```
khanhtc1202, line 27: I don't get the point of having this type of definition instead of just using bytes. Could you help me understand more about this 👀

khanhtc1202: cc @Warashi

Warashi: I think kind and api_version exist for all configurations, so I defined them separately.
Additionally, these definitions reflect the limitation of the config.

khanhtc1202: @Warashi Thanks for the explanation 🙏 I think this place we should not define a new type for lower maintenance costs. It could be better to just use bytes type here and both piped and plugin source code will use the `config.ApplicationConfig` type and marshall function defined in config package. wdyt? 🤔

Warashi: @khanhtc1202 
Do you mean [`config.Config`](https://github.com/pipe-cd/pipecd/blob/0e7a17704ab6c21895482826466cc02fc21ae457/pkg/config/config.go#L72-L89)?
I think the plugin source code will not use this because the config struct will be defined in the plugin code, not in the piped code.

Warashi: > I think this place we should not define a new type for lower maintenance costs.

I got your point 👍🏻 
I'll try to define a function or struct/method used by both piped and plugins.
I think this definition will use the generic type parameter.

Warashi: @khanhtc1202 
I fixed the type as `bytes` as you pointed.
In this PR, I want to define DeploymentSource only.
I'll implement config.ApplicationConfig or a similar one in another PR when I need it.

khanhtc1202: @Warashi Thank you 👍

khanhtc1202 (approved):
Thank you 💯

ffjlabo (approved):

PR #5121, opened 2024-08-08
Remove piped config bytes from planner pipedv1

khanhtc1202:

As the title, I previously planned to pass the piped config to the plugin so that the plugin could build gitClient or decrypt by itself. However, my plan changed, and now we should not share such information with the plugins.

Part of #4980 


t-kikuc (approved):
only pipedv1 is affected

Warashi (approved):

PR #5122, opened 2024-08-08
Support Drift Detection for ECS

### LINKED ISSUE #5005: [ECS] Support Drift Detection for ECS

t-kikuc:

Supported DriftDetection for ECS with some restrictions.

When synced:

image

When out-of-sync:
image

Restrictions: 
- If `desiredCount` of headManifest is 0 or not set, which means Autoscale may be enabled), diff of `desiredCount` will be ignored. 
    - cf. https://github.com/pipe-cd/pipecd/pull/5030
- Some parameters are ignored. (e.g. Tags, LoadBalancers)
    - See `detector.go` > `ignoreParameters()`
- Even if unintended drift tasksets exist and the diff is detected, automatic deployment will NOT remove the tasksets. 
    - If needed, we should fix the deployment logic of ECS.
    - If diff of other fields exists, `desiredCount`'s diff will also be shown (as the above image).

I will add docs about them in another PR.

Fixes #5005

**Does this PR introduce a user-facing change?**: Yes,  users can detect drift for ECS

### DIFF
--- a/pkg/app/piped/driftdetector/detector.go
+++ b/pkg/app/piped/driftdetector/detector.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/pipe-cd/pipecd/pkg/app/piped/driftdetector/cloudrun"
+	"github.com/pipe-cd/pipecd/pkg/app/piped/driftdetector/ecs"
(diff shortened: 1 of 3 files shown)


ffjlabo (approved):
🚀

khanhtc1202 (commented):
LGTM 👍

khanhtc1202 (approved):
LGTM 👍

PR #5146, opened 2024-08-19
Discard the otel logs

Warashi:

If the control plane disables the monitoring feature, piped's OTEL TracerProvider logs some errors.
It's not a problem, but we don't want to see error logs.
So we discard the otel logs.

Fixes #

### DIFF
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -33,6 +33,7 @@ import (
 	secretmanagerpb "cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	awssecretsmanager "github.com/aws/aws-sdk-go-v2/service/secretsmanager"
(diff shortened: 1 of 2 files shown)


pkg/app/piped/cmd/piped/piped.go
```diff
@@ -241,6 +242,11 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 	}
 	otel.SetTracerProvider(tracerProvider)
 
+	// if control plane's flag monitoring.enabled is false, otel provider logs errors.
+	// it's no problem but we don't want to see it.
+	// so we discard the logger.
+	otel.SetLogger(logr.Discard())
+
```
ffjlabo, line 249: 📝 It's ok for now. However, if users want to investigate the otel behavior, it might be necessary. 
If so, it would be nice to add the flag on whether to show the log.

ffjlabo (approved):

khanhtc1202 (approved):
Got it, thanks 👍

Issue #5148, opened 2024-08-19
Easier quickstart for PipeCD


khanhtc1202:
**What would you like to be added**:

We have `pipectl quickstart` command, but it still requires many things to be prepared before touching pipecd. Say, we need
1. install pipectl
2. having a k8s cluster and kubectl
3. install helm
4. run...

Most of our users may already have (2), so it could be better to remove (1) and (3). I want to serve pre-prepared quickstart manifests, and users can simply run `kubectl apply`, which could be way easier.

**Why is this needed**:

khanhtc1202:
Tasks:
- [ ] Make `quickstart` dir under the main repo containing raw Kubernetes manifests for quickstart (build by helm template command)
- [ ] Add CI step update the quickstart manifests on release

PR #5152, opened 2024-08-20
Add docs for quickstart with raw manifests on README

ffjlabo:

as title

Part of https://github.com/pipe-cd/pipecd/issues/5148

### DIFF
--- a/quickstart/README.md
+++ b/quickstart/README.md
@@ -3,6 +3,39 @@
 This directory contains configuration files for `control-plane` and `piped` for [Quickstart](https://pipecd.dev/docs/quickstart) Guide.
 
 
+
+### Quickstart with raw manifests
+
+1. create namespace `pipecd`
+
+```
+kubectl create namespace pipecd
+```
+
+2. deploy Control Plane to the namespace: pipecd
+
+```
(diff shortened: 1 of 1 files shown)


quickstart/README.md
```diff
@@ -3,6 +3,39 @@
...
+kubectl create namespace pipecd
+```
+
+2. deploy Control Plane to the namespace: pipecd
+
+```
+kubectl apply -n pipecd -f ./manifests/control-plane.yaml
+```
+
+3. create piped on the Web UI
```
Warashi, line 21: I think it's nice for users to show kubectl port-forward command and URL of the Web UI.
WDYT?

ffjlabo: @Warashi 
It's a nice idea. But I wonder if I should do so 🤔 
For me, this doc is like a memo to describe whats the manifest is. 
Also there is the actual process of the quickstart on the docs. https://pipecd.dev/docs-v0.48.x/quickstart/

Warashi: OK, I got it. I agree that we have other documents about these manifests and don't have to show detailed commands here.

khanhtc1202 (approved):
Thank you 👍

PR #5154, opened 2024-08-22
Fix the kubernetes manifest diff

Warashi:

Normalize both old and new manifests when calculating diff.

Without this PR, the unexpected difference is calculated when resource/limit numbers have floating-point values such as `cpu: "1.5"`
The diff signs 1.5 and 1500m are different.
This occurs because the new manifests are normalized, but the old ones are not.

N/A

Users can get the correct plan-preview diffs.
Some deployments triggered by incorrect diffs become not triggered.

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/diff.go
+++ b/pkg/app/piped/platformprovider/kubernetes/diff.go
@@ -63,13 +63,13 @@ func Diff(old, new Manifest, logger *zap.Logger, opts ...diff.Option) (*diff.Res
 
 	normalizedOld, err := remarshal(old.u)
 	if err != nil {
-		logger.Info("compare manifests directly since it was unable to remarshal Kubernetes manifest to normalize special fields", zap.Error(err))
+		logger.Info("compare manifests directly since it was unable to remarshal old Kubernetes manifest to normalize special fields", zap.Error(err))
 		return diff.DiffUnstructureds(*old.u, *new.u, key, opts...)
 	}
 
 	normalizedNew, err := remarshal(new.u)
 	if err != nil {
-		logger.Info("compare manifests directly since it was unable to remarshal Kubernetes manifest to normalize special fields", zap.Error(err))
+		logger.Info("compare manifests directly since it was unable to remarshal new Kubernetes manifest to normalize special fields", zap.Error(err))
 		return diff.DiffUnstructureds(*old.u, *new.u, key, opts...)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/app/piped/platformprovider/kubernetes/diff_test.go
```diff
@@ -405,3 +405,168 @@ spec:
...
+					t.Log(node.ValueY)
+					t.Log("---")
+				}
+			}
+			for _, add := range result.Adds {
+				t.Log(add.Key)
+			}
+			for _, delete := range result.Deletes {
+				t.Log(delete.Key)
+			}
```
ffjlabo, line 569: [ask] This codes are for checking the diff when `result.NoChange()` is false?

pkg/app/piped/platformprovider/kubernetes/diff.go
```diff
@@ -60,13 +60,20 @@ func Diff(old, new Manifest, logger *zap.Logger, opts ...diff.Option) (*diff.Res
 	}
 
 	key := old.Key.String()
-	normalized, err := remarshal(new.u)
+
+	normalizedOld, err := remarshal(old.u)
+	if err != nil {
+		logger.Info("compare manifests directly since it was unable to remarshal Kubernetes manifest to normalize special fields", zap.Error(err))
```
ffjlabo, line 66: [nit] It would be nice to show old or new when investigating with this log.
It is the same for the remarshal new one.

```suggestion
		logger.Info("compare manifests directly since it was unable to remarshal old Kubernetes manifest to normalize special fields", zap.Error(err))
```

ffjlabo (commented):
Thank you! This is the nice improvement!
I left nit comments.

khanhtc1202 (approved):
Nice catch, thank you 👍

pkg/app/piped/platformprovider/kubernetes/diff_test.go
Warashi: Yes, and these are logs for debugging, not assertions.

pkg/app/piped/platformprovider/kubernetes/diff.go
Warashi: Fixed by this commit
https://github.com/pipe-cd/pipecd/pull/5154/commits/ead1e14988b9982a4d541234d8d270e653bca194

ffjlabo (approved):
🚀

khanhtc1202 (approved):
ready to go 🚀

PR #5155, opened 2024-08-23
remove a validation check in notification-mention

nnnkkk7:

This PR removes an  validation check in the Validate method of the NotificationMention struct. 
This contradicts our documentation( https://pipecd.dev/docs-v0.48.x/user-guide/configuration-reference/#notificationmention ), which states these fields are optional, and the omitempty JSON tags on the struct fields.

Fixes #

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -672,9 +672,6 @@ type NotificationMention struct {
 }
 
 func (n *NotificationMention) Validate() error {
-	if len(n.Slack) == 0 && len(n.SlackGroups) == 0 && len(n.SlackUsers) == 0 {
-		return fmt.Errorf("slack, slackusers or slackGroups must not be empty")
-	}
 	if n.Event == allEventsSymbol {
 		return nil
 	}
@@ -779,7 +776,7 @@ func (dd *DriftDetection) Validate() error {
 }
 
 func LoadApplication(repoPath, configRelPath string, appKind model.ApplicationKind) (*GenericApplicationSpec, error) {
-	var absPath = filepath.Join(repoPath, configRelPath)
+	absPath := filepath.Join(repoPath, configRelPath)
 
 	cfg, err := LoadFromYAML(absPath)
 	if err != nil {


pkg/config/application.go
```diff
@@ -672,9 +672,6 @@ type NotificationMention struct {
 }
 
 func (n *NotificationMention) Validate() error {
-	if len(n.Slack) == 0 && len(n.SlackGroups) == 0 && len(n.SlackUsers) == 0 {
```
khanhtc1202, line 675: I think the point of this check is to ensure at least one of the configurations must not be empty. So it's not a no meaning 👀

nnnkkk7: ok 
but when we don't need Slack,  SlackGroups or SlackUsers settings, this check return err and we can't deploy by pipecd...

khanhtc1202: can you share the configuration which failed the validation test in your usecase?

nnnkkk7: @khanhtc1202 
ok!
This configuration causes it..

```
kind: LambdaApp
spec:
  name: dev-hoge
  labels:
    env: dev
  notification:
    mentions:
      - event: DEPLOYMENT_SUCCEEDED
      - event: DEPLOYMENT_FAILED
  eventWatcher:
    - matcher:
        name: hoge-update
        labels:
          env: dev
      handler:
        type: GIT_UPDATE
        config:
          replacements:
            - file: function.yaml
              regex: 00000.dkr.ecr.ap-northeast-1.amazonaws.com/hoge:fuga
```
↓

スクリーンショット 2024-08-26 午前10 35 47

khanhtc1202:
```
  notification:
    mentions:
      - event: DEPLOYMENT_SUCCEEDED
      - event: DEPLOYMENT_FAILED
```

@nnnkkk7 Thanks for the detailed configuration 👍 In such a case, where would you want the piped to send a notification to? Without slack mentions, there is no meaning for this `notification.mentions` config. Or could you share your expected behavior of piped you want in this case?

nnnkkk7: @khanhtc1202 

I see. 
Notification settings in piped-config.yaml are as follows. Is this sufficient for  notifications?🤔
I want to send a notification to this channel..
```
analysisProviders:
〜
notifications:
  receivers:
  - name: receiver
    slack:
      channelID: ""
      hookURL: '******'
      oauthToken: ""
      oauthTokenData: ""
      oauthTokenFile: ""
  routes:
  - events:
    - DEPLOYMENT_FAILED
    labels:
      env: dev
    name: dev
    receiver: receiver
  - events:
    - DEPLOYMENT_APPROVED
    - DEPLOYMENT_WAIT_APPROVAL
    - DEPLOYMENT_SUCCEEDED
    - DEPLOYMENT_FAILED
    labels:
      env: stg
    name: stg
    receiver: receiver
  - events:
    - DEPLOYMENT_APPROVED
    - DEPLOYMENT_WAIT_APPROVAL
    - DEPLOYMENT_ROLLING_BACK
    - DEPLOYMENT_SUCCEEDED
    - DEPLOYMENT_FAILED
    - DEPLOYMENT_CANCELLED
    labels:
      env: prd
    name: prd
    receiver: receiver
〜

```

khanhtc1202: Thanks, I get it now. In this case, your expected behavior of piped is just to send the notification in those cases to target channel without mention anyone, is it right?

khanhtc1202 (approved):
👍

pkg/config/application.go
khanhtc1202: @nnnkkk7 Thank you for clarifying. I approved this change after discuss with the team 👍

t-kikuc (approved):

PR #5162, opened 2024-08-27
refactor: flushAll method to wait for all log flushes to complete

### LINKED ISSUE #5156: logpersister graceful shutdown
The `flushAll` method in the `logpersister` package should wait for all `flushFromLastCheckpoint` calls to complete before proceeding. Currently, it initiates these calls asynchronously using `go sp.flushFromLastCheckpoint(ctx)`, which may lead to the `cancel` being executed immediately, not allowing the expected persister.gracePeriod(30-second timeout) to function correctly. 

https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/logpersister/persister.go#L159
**Why is this needed**:

naonao2323:

Fixes #5156

### DIFF
--- a/pkg/app/piped/logpersister/persister.go
+++ b/pkg/app/piped/logpersister/persister.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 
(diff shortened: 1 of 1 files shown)


khanhtc1202 (approved):
LGTM 👍

Warashi (approved):
👍🏻

PR #5169, opened 2024-08-30
Update piped config for quickstart

khanhtc1202:

Update piped config to make the setup for install piped in quickstart less prepare steps

Part of #5148

### DIFF
--- a/manifests/piped/templates/configmap.yaml
+++ b/manifests/piped/templates/configmap.yaml
@@ -13,12 +13,12 @@ data:
     spec:
       projectID: quickstart
       pipedID: {{ .Values.quickstart.pipedId }}
-      pipedKeyFile: /etc/piped-secret/piped-key
+      pipedKeyData: {{ .Values.quickstart.pipedKeyData }}
       apiAddress: pipecd:8080
       syncInterval: 1m
       repositories:
         - repoId: examples
-          remote: {{ .Values.quickstart.gitRepoRemote }}
+          remote: https://github.com/pipe-cd/examples.git
           branch: master
 {{- else }}
   {{ .Values.config.fileName }}: |-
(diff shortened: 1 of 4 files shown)


manifests/piped/templates/configmap.yaml
```diff
@@ -13,12 +13,12 @@ data:
...
       projectID: quickstart
       pipedID: {{ .Values.quickstart.pipedId }}
-      pipedKeyFile: /etc/piped-secret/piped-key
+      pipedKeyData: {{ .Values.quickstart.pipedKeyData }}
       apiAddress: pipecd:8080
       syncInterval: 1m
       repositories:
         - repoId: examples
-          remote: {{ .Values.quickstart.gitRepoRemote }}
+          remote: https://github.com/pipe-cd/examples.git
```
Warashi, line 21: If we use this repository as the quickstart config, the users cannot try the git commit hooked deployments.
I think it's nice to place this value as the default value, not as the constant value.

khanhtc1202: To make the setup more accessible, I'm thinking about only making the quickstart scenario like
- prepare k8s cluster
- install pipecd controllplane
- install piped (with minimal setting)
- add app and trigger quicksync as default

For other things, like trigger canary, it could be better to make it elsewhere. Also, we can refer to the docs, which show how to change the piped setting for adding other platform providers, adding other remote repos, etc.

khanhtc1202: For example, we have this docs: https://pipecd.dev/docs-v0.48.x/user-guide/managing-piped/adding-a-git-repository/

Warashi: I got your point, and I noticed that we have [the tutorial](https://github.com/pipe-cd/tutorial), so we can try other remote repos with this.

Warashi (approved):
Thank you

t-kikuc (approved):

ffjlabo (approved):
👌

PR #5008, opened 2024-06-29
Support OIDC for the SSO

kumo-rn5s:
**What this PR does / why we need it**:

This PR introduces a new OIDC login method to enhance user authentication options by supporting the basic OIDC Authorization Code Flow. It also includes support for common OIDC providers like Keycloak, Okta, and Cognito.

**Which issue(s) this PR fixes**:

Fixes #4906

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**: Users will have the option to log in using OIDC, providing a more secure and versatile authentication method.
- **Is this breaking change**: No
- **How to migrate (if breaking change)**: No need

TODO:
- [x] Adding key commits for basic OIDC Authorization Code Flow.
- [x] Testing with major OIDC providers
   - [x] Keycloak
   - [x] Auth0
   - [x] Cognito
- [x] Adding OIDC documentation.

PR #5172, opened 2024-09-02
Show username on the header

t-kikuc:

As title.  
Since avatar images are not shown on the header for OIDC users, users want to know which account they logged in to.

- GitHub:  GitHub user ID

    image

- OIDC: username from IdP

     image

I think `role` is not necessary to show.

Related to #5008

See https://github.com/pipe-cd/pipecd/pull/5008#issuecomment-2277603291

**Does this PR introduce a user-facing change?**: Users can see their username.

**For Reviewers**

See here to understand why `subject` is used.
https://github.com/pipe-cd/pipecd/blob/034054133543b7ae9d57efd360e1b0a9aae7c7db/pkg/app/server/httpapi/callback.go#L92-L103

### DIFF
--- a/web/src/components/header/index.tsx
+++ b/web/src/components/header/index.tsx
@@ -197,9 +197,10 @@ export const Header: FC = memo(function Header() {
                 aria-controls="user-menu"
                 aria-haspopup="true"
                 onClick={(e) => setUserAnchorEl(e.currentTarget)}
-              >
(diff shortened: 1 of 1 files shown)


Warashi (approved):
👍🏻

ffjlabo (approved):
🚀

PR #5177, opened 2024-09-03
Improve Terraform plan result handling

golemiso:

- Make it possible to handle terraform plan results when:
  - adding new outputs
  - removing resources from state using removed block
  - refactoring *.tf files using moved block

without these modifications, pipecd returns `no changes` or `unable to parse plan output`.

Fixes #


Warashi (approved):
Thanks! Great work!
I checked the piped-side behavior, and it worked successfully.
It's difficult to check the actions-plan-preview side behavior without merging this PR, so we should check it after merging it.

khanhtc1202 (approved):
LGTM 👍 
Note: We also have to release tool/actions-plan-preview in order to test this. Let's release a rc for testing the GitHub actions behaviors.

PR #5179, opened 2024-09-03
Add document for force-sync-by-replace annotation

### LINKED ISSUE #5174: Support kubectl replace with force option
The `kubectl replace` with the `--force` option.

**Why is this needed**:
We want to run a Kubernetes job during deployment.
After running a job, we have to replace it with the `--force` option to run the same job.
There is the `sync-by-replace` annotation option, but this option does not run kubectl replace with the `--force` option.
So we may add a `force-sync-by-replace` annotation option to run kubectl replace with `--force`.

Warashi:

Add documentation of `force-sync-by-replace` annotation.

Fixes #5174 
Follows #5175 


t-kikuc (approved):
Let's merge this after releasing https://github.com/pipe-cd/pipecd/pull/5175.

because users cannot use the annotation in v0.48.x yet.

ffjlabo (approved):

khanhtc1202 (approved):
👍

PR #5186, opened 2024-09-04
[Lambda] Support Drift Detection for Lambda

### LINKED ISSUE #5182: [Lambda] Support Drift Detection for Lambda

t-kikuc:

Supported DriftDetection for Lambda

When OutOfSync:
    image

Restrictions: 
- The following fields in [function.yaml](https://pipecd.dev/docs-v0.48.x/user-guide/configuration-reference/#specific-functionyaml) are ignored because livestates don't have them.
  - `source`
  - `s3Bucket`
  - `s3Key`
  - `s3ObjectVersion`

Fixes #5182

**Does this PR introduce a user-facing change?**: Users can see drifts.

**Note:**

To use this feature, users need to add the following IAM Actions to their Piped's IAM Role.

- [GetFunction](https://docs.aws.amazon.com/lambda/latest/dg/API_GetFunction.html)
- [ListFunctions](https://docs.aws.amazon.com/lambda/latest/dg/API_ListFunctions.html)

Later I'll describe the details in another PR.

### DIFF
--- a/pkg/app/piped/driftdetector/lambda/detector.go
+++ b/pkg/app/piped/driftdetector/lambda/detector.go
@@ -207,8 +207,13 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 	}
 	d.logger.Info(fmt.Sprintf("application %s has a live function manifest", app.Id))
 
-	ignoreAndSortParameters(&liveManifest.Spec, &headManifest.Spec)
+	ignoreAndSortParameters(&headManifest.Spec)
 
+	// WithIgnoreAddingMapKeys option ignores all of followings:
+	//  - default value of Architecture
+	//  - default value of EphemeralStorage
+	//  - environments added in live states
+	//  - tags added in live states, including pipecd managed tags
 	result, err := provider.Diff(
 		liveManifest,
 		headManifest,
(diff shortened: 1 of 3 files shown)


ffjlabo (dismissed):
LGTM

pkg/app/piped/driftdetector/lambda/detector.go
```diff
@@ -13,3 +13,385 @@
...
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+
+	sorted := make(map[string]string)
+	for _, k := range keys {
+		sorted[k] = m[k]
+	}
+	return sorted
+}
```
Warashi, line 276: The for-range for map[string]string is always looped in random order.
So, this code may not work as expected.

Warashi (commented):
I commented about sorting maps.
I think the code I commented on is not needed because it does not work as expected, but the test you wrote is passed.

pkg/app/piped/driftdetector/lambda/detector.go
t-kikuc: I understood, what a meaningless code... 😅 
let me re-think...

t-kikuc: @Warashi 
Thank you, I removed the unnecessary sort.  [be00911](https://github.com/pipe-cd/pipecd/pull/5186/commits/be009116934f76d32d349bc5258eb5f48d701440)

Some slices need to be sorted, but maps do NOT need to be sorted.

Warashi (approved):
LGTM

t-kikuc:
@ffjlabo Would you please review again?

This is the only one change: https://github.com/pipe-cd/pipecd/pull/5186#discussion_r1745009233

ffjlabo:
Sorry for the rough review🙏
I rechecked the new fix. It is fine.

ffjlabo (approved):
🚀

PR #5193, opened 2024-09-06
Discard otel errors

Warashi:

This PR discards all errors and logs from the otel library.
Without this PR, we will see error logs when an error occurs, like the one below.

```
2024/09/06 13:16:53 traces export: rpc error: code = PermissionDenied desc =
```

Fixes #

### DIFF
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -237,17 +237,18 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 		return err
 	}
 
+	// if control plane's flag monitoring.enabled is false, otel provider logs errors.
+	// it's no problem but we don't want to see it.
+	// so we discard the errors and the logs.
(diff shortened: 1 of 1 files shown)


pkg/app/piped/cmd/piped/piped.go
```diff
@@ -237,18 +237,19 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 		return err
 	}
 
+	// if control plane's flag monitoring.enabled is false, otel provider logs errors.
+	// it's no problem but we don't want to see it.
+	// so we discard the errors and the logs.
+	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {}))
+	otel.SetLogger(logr.Discard())
```
Warashi, line 244: we have to set global ErrorHandler and Logger before we create the tracer providers.

ffjlabo (approved):
🚀

t-kikuc (approved):

PR #5194, opened 2024-09-06
Set the envoy retry policy for external-authz filter

Warashi:

I saw the envoy access log sometimes records UAEX (UnauthorizedExternalService) errors even if the ext-authz service returns them as OK.
I checked the admin port of the envoy and realized the tx_reset had occurred.
I don't know how to reduce the tx_reset, so I want to try the retry_policy.
When we use the retry_policy, we have to bump the envoy version.
So, I bump the envoy version to the latest, then I configured the retry_policy.

Fixes #

### DIFF
--- a/manifests/pipecd/templates/deployment.yaml
+++ b/manifests/pipecd/templates/deployment.yaml
@@ -25,7 +25,7 @@ spec:
       {{- end }}
       containers:
         - name: envoy
-          image: envoyproxy/envoy-alpine:{{ .Values.gateway.imageTag }}
+          image: envoyproxy/envoy:{{ .Values.gateway.imageTag }}
           imagePullPolicy: IfNotPresent
           command:
           - envoy
--- a/manifests/pipecd/templates/envoy-configmap.yaml
+++ b/manifests/pipecd/templates/envoy-configmap.yaml
@@ -42,8 +42,12 @@ data:
                   transport_api_version: V3
                   include_peer_certificate: false
               - name: envoy.filters.http.grpc_web
+                typed_config:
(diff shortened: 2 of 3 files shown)


manifests/pipecd/templates/envoy-configmap.yaml
```diff
@@ -38,18 +38,31 @@ data:
...
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors
 {{- end }}
               - name: envoy.filters.http.grpc_stats
                 typed_config:
                   "@type": type.googleapis.com/envoy.extensions.filters.http.grpc_stats.v3.FilterConfig
                   stats_for_all_methods: true
                   enable_upstream_stats: true
               - name: envoy.filters.http.router
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
```
Warashi, line 65: We have to set these to bump the envoy version

manifests/pipecd/templates/deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
       {{- end }}
       containers:
         - name: envoy
-          image: envoyproxy/envoy-alpine:{{ .Values.gateway.imageTag }}
+          image: envoyproxy/envoy:{{ .Values.gateway.imageTag }}
```
Warashi, line 28: envoy-alpine has stopped updating 2 years ago

khanhtc1202 (approved):
Thanks for detailed explanation 🫡

ffjlabo (approved):
LGTM 🙆‍♂️

PR #5203, opened 2024-09-09
Remove retry_policy and make timeout longer

Warashi:

Make ext-authz-filter timeout longer, and remove retry_policy.

I realized that the Authz RPC is sometimes slow in the dev environment. So, I made the timeout longer.
After I checked the envoy log, the retry did not occur when the request was timeouted. So, I removed the retry_policy.

Fixes #


ffjlabo (approved):
LGTM

khanhtc1202 (approved):
👍

PR #5208, opened 2024-09-12
Enable the persistence of the Grafana Tempo

Warashi:

Without this PR, the traces are lost when the tempo pod is evicted.
The StatefulSet manages the tempo's pod, so usually it does not occur, but it occurs with the times such as when the configmap is changed.

Fixes #

**Does this PR introduce a user-facing change?**: Yes only if users use the control plane version v0.49.0-rc0

When users using control plane helm chart with monitoring.enabled as true, they cannot upgrade the control plane without manual operation.

- **Is this breaking change**: Yes
Delete the Grafana Tempo's StatefulSet before applying the Helm Chart.


t-kikuc (approved):
Thank you

PR #5215, opened 2024-09-17
Fix to use target dsp in script run rollback

### LINKED ISSUE #5214: Can't use the script file on the SCRIPT_RUN_ROLBACK stage
**What happened**:

When using the script file on the `onRollback` field in app.pipecd.yaml, can't execute it on the SCRIPT_RUN_ROLLBACK stage.
PipeCD
(We can use the script file on the SCRIPT_RUN stage when setting the `run` option of this stage)

**file structure**
```
- app.pipecd.yaml
- script.sh
```

**app.pipecd.yaml**
```
kind: KubernetesApp
spec:
  name: script-run-like-jenkins
  labels:
    env: example
    team: product
  pipeline:
    stages:
      - name: SCRIPT_RUN
        with:
          run: |
            sleep 10
            sh script.sh
          onRollback: |
            echo $SR_DEPLOYMENT_ID
            echo $SR_APPLICATION_ID
            echo $SR_APPLICATION_NAME
            echo $SR_TRIGGERED_AT
            echo $SR_TRIGGERED_COMMIT_HASH
            echo $SR_REPOSITORY_URL
            echo $SR_SUMMARY
            echo $SR_CONTEXT_RAW
            sh script.sh
```

It would be nice to execute the script file when SCRIPT_RUN_ROLLBACK stage is executed.

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

ffjlabo:

Fixed to use TargetDSP on the SCRIPT_RUN_ROLLBACK stage.
With these fixes, we can refer to the files located in the app dir when executing the SCRIPT_RUN_ROLLBACK stage.

I chose TargetDSP, not RunningDSP, because the commands on the `onRollback` are set in app.pipecd.yaml used when Deployment is executed. 
The purpose of the SCRIPT_RUN_ROLLBACK stage is to run whatever command is currently set in onRollback during rollback.

Part of #5214

### DIFF
--- a/pkg/app/piped/executor/kubernetes/rollback.go
+++ b/pkg/app/piped/executor/kubernetes/rollback.go
@@ -195,6 +195,13 @@ func (e *rollbackExecutor) ensureScriptRunRollback(ctx context.Context) model.St
 		return model.StageStatus_STAGE_SUCCESS
 	}
 
+	ds, err := e.TargetDSP.Get(ctx, e.LogPersister)
+	if err != nil {
+		e.LogPersister.Errorf("Failed to prepare target deploy source data (%v)", err)
+		return model.StageStatus_STAGE_FAILURE
+	}
+	e.appDir = ds.AppDir
+
 	envStr, ok := e.Stage.Metadata["env"]
 	env := make(map[string]string, 0)
 	if ok {


pkg/app/piped/deploysource/deploysource.go
```diff
@@ -90,7 +90,7 @@ func (p *provider) Get(ctx context.Context, lw io.Writer) (*DeploySource, error)
 
 	if !p.done {
 		p.source, p.err = p.prepare(ctx, lw)
-		p.done = true
+		p.done = p.err == nil // If there is an error, we should re-prepare it next time.
```
ffjlabo, line 93: 📝 This modification is necessary to avoid an error when executing a cancellation while executing TargetDSP.Get in another stage.

I encountered such a situation when canceling the deployment via WebUI during the SCRIPT_RUN stage execution.

Cursor_と_PipeCD__1__png

ffjlabo: I split the fixes to the another PR https://github.com/pipe-cd/pipecd/pull/5216

Warashi (approved):
👍🏻

t-kikuc (approved):
thank you

PR #5216, opened 2024-09-18
Regard DSP as done only when preparation is successful considering of reuse the provider

ffjlabo:

This modification is necessary to avoid an error when executing a cancellation while executing deploySourceProvider.Get in another stage.

I encountered such a situation when canceling the deployment via WebUI during the SCRIPT_RUN stage execution. 
I reproduced the behavior when checking the behavior for the PR https://github.com/pipe-cd/pipecd/pull/5215 like this↓ 
Cursor_と_PipeCD__1__png

Part of https://github.com/pipe-cd/pipecd/issues/5214

### DIFF
--- a/pkg/app/piped/deploysource/deploysource.go
+++ b/pkg/app/piped/deploysource/deploysource.go
@@ -90,7 +90,7 @@ func (p *provider) Get(ctx context.Context, lw io.Writer) (*DeploySource, error)
 
 	if !p.done {
 		p.source, p.err = p.prepare(ctx, lw)
-		p.done = true
+		p.done = p.err == nil // If there is an error, we should re-prepare it next time.
 	}
 
 	if p.err != nil {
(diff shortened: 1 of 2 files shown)


khanhtc1202 (dismissed):
Nice catch, thank you 👍

t-kikuc (commented):
[ASK] How about the following point?

https://github.com/pipe-cd/pipecd/blob/6b004fa1090cf4d16749eede2843e64631c9da54/pkg/app/piped/deploysource/deploysource.go#L117

ffjlabo:
@t-kikuc Thank you! This also might lead to the same error. So I fixed it 🙏 [f87754a](https://github.com/pipe-cd/pipecd/pull/5216/commits/f87754ac2634b4c6713e0ec194f2cb9a6c378721)

t-kikuc (approved):
thank you!

Warashi (approved):
thank you

PR #5217, opened 2024-09-18
Make controller's os.RemoveAll single-threaded

Warashi:

This PR modifies the planner/scheduler cleanup to be single-threaded in order to reduce the file IO loads.
I did a load test with the following scenario and discovered that os.RemoveAll creates a high load for piped:

- A single piped instance manages 400 apps
- All 400 apps are triggered by one commit

In this scenario, the piped requires about 20 minutes to complete all deployments.
After this PR, the piped requires about 15 minutes to complete.

Fixes #

    - The piped requires larger disk space.
    - The piped performance becomes better

### DIFF
--- a/pkg/app/piped/controller/controller.go
+++ b/pkg/app/piped/controller/controller.go
@@ -138,6 +138,9 @@ type controller struct {
 	// WaitGroup for waiting the completions of all planners, schedulers.
 	wg sync.WaitGroup
 
+	// workingDirRemovalCh is used to single-threaded removal of working directory.
+	workingDirRemovalCh chan string
+
 	workspaceDir string
 	syncInternal time.Duration
 	gracePeriod  time.Duration
@@ -186,6 +189,8 @@ func NewController(
 		mostRecentlySuccessfulCommits:         make(map[string]string),
 		mostRecentlySuccessfulConfigFilenames: make(map[string]string),
 
+		workingDirRemovalCh: make(chan string),
+
 		syncInternal: 10 * time.Second,
 		gracePeriod:  gracePeriod,
 		logger:       lg,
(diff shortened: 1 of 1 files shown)


pkg/app/piped/controller/controller.go
```diff
@@ -219,6 +224,16 @@ func (c *controller) Run(ctx context.Context) error {
 		close(lpStoppedCh)
 	}()
 
+	// Start workspace cleaner.
+	// This will remove the workspace directory of the completed planner/scheduler.
+	go func() {
+		for ws := range c.workingDirRemovalCh {
```
ffjlabo, line 230: [Ask] It looks like there is nothing to close `c.workingDirRemovalCh`. Does this lead to the goroutine leak?

ffjlabo (commented):
Thank you for the fix! I left a comment.

pkg/app/piped/controller/controller.go
Warashi: I'm not sure if we would call this pattern a goroutine leak.
This goroutine's lifetime is the same as that of the piped.
As there is only one controller for each piped, there is also only one instance of this goroutine.
I believe this code is acceptable because the number of goroutines doesn't increase.

ffjlabo: @Warashi 

> This goroutine's lifetime is the same as that of the piped.
As there is only one controller for each piped, there is also only one instance of this goroutine.

I see. I got your point 👍 
But I think it would be better to make the lifetime of the channel explicit.

It would be better for the person calling controller.Run to not be aware of that.

WDYT?

Warashi: > It would be better for the person calling controller.Run to not be aware of that.

I got your point. I'll try to make the lifetime explicit.

Warashi: Fixed in this commit
https://github.com/pipe-cd/pipecd/pull/5217/commits/efa4e6d60cca68d7fa8b35896f5a9ad0290ce9e9

ffjlabo (approved):
LGTM 🚀

khanhtc1202 (approved):
LGTM 👍

PR #5218, opened 2024-09-19
Add InstallTool RPC method for PluginService

Warashi:

With pipedv1's plugin architecture, we have to provide a way to install the tool used in the plugin.
I considered the following ways and propose the first way.

1. provide InstallTool RPC at the PluginService.
2. provide the directory path to the plugins in which the tool should be installed
3. provide nothing to the plugins, and the plugin does all of the installation, such as making the temp dir to install the tool, ...etc

I propose the first way because we should consider multiple plugins that want to install the same tool/version at the same time.
In such a situation, we should prevent race conditions, and we can handle this with a `sync.Mutex` or something similar in the first way.

Part of #4980 


t-kikuc (approved):
Only v1 is affected

ffjlabo (approved):

PR #5225, opened 2024-09-24
Use `make release` for major/minor releases

t-kikuc:

1. Uses `make release` instead of `make release/init` and `make release/docs` for major/minor releases.
   cf. https://github.com/pipe-cd/pipecd/blob/909fd76a4976cb0c81f2ad00ed52f398ec49a156/Makefile#L218-L231

2. Directly specifies to use `make release/init` for patch releases.

Fixes #


Warashi (approved):
👍🏻

ffjlabo (approved):
LGTM

PR #5237, opened 2024-09-27
Copy pkg/config to pkg/configv1

Warashi:

This PR prepares the pkg/configv1 package to develop pipedv1.
This makes us to be able to reduce mistakenly breaking piped (v0).

The files under pkg/configv1 are copies of pkg/config with no modifications.

Part of #4980


t-kikuc (approved):

khanhtc1202 (approved):
Here you go 👍

PR #5239, opened 2024-09-27
Add requires for stages in case of multi plugins quicksync pattern

khanhtc1202:

In case of multi plugins for a single application, and the deployment of the application is triggered in quick sync strategy, we need to add requires for stages to ensure they run in order instead of all at one.

Part of #4980 
Follow PR #5231 


khanhtc1202:
cc @Warashi We need this or all forward stages in quicksync triggered deployment will run at one 👀

Warashi (approved):
thank you

t-kikuc (approved):

PR #5240, opened 2024-09-27
Fix panic in ECS driftdetection when a taskdef in livestates does not exist

t-kikuc:

I added a nil check.

TaskDefs in livestates sometimes do not exist. 
e.g. When a service is created but the taskset does not exist yet.

cf. LiveTask is set here:
https://github.com/pipe-cd/pipecd/blob/1477e76f14bcfae5f4ba88b516b15f0e303c833c/pkg/app/piped/livestatestore/ecs/store.go#L66-L99

N/A


khanhtc1202 (approved):
Good catch 👍

Warashi (approved):
👍🏻

PR #5243, opened 2024-09-30
Add k8s plugin's toolregistry implementation

Warashi:

The plugin has to pass the tool install script to the piped.
I don't want to pass a script everywhere the plugin tries to install a tool.
So I wrote a wrapper package to hide install script from other implementations.

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/toolregistry/scripts.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/toolregistry/scripts.go
@@ -1,0 +1,33 @@
+// Copyright 2024 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/kubernetes/toolregistry/scripts.go
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package toolregistry
+
+const kubectlInstallScript = `
+cd {{ .TmpDir }}
+curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{ .Version }}/bin/{{ .Os }}/{{ .Arch }}/kubectl
+mv kubectl {{ .OutPath }}
```
Warashi, line 20: These template variables come from the below comments.
https://github.com/pipe-cd/pipecd/blob/c1a06aa922a994340220a4bc5de62f6630c29b8a/pkg/app/pipedv1/cmd/piped/service/service.proto#L46-L51

t-kikuc (approved):

khanhtc1202 (approved):
Thank you 👍

PR #5248, opened 2024-10-02
Remove an unused field 'configFilePathInGitRepo'

t-kikuc:

as title.

It's introduced in https://github.com/pipe-cd/pipecd/pull/2557 but seems to be never used.

N/A


Warashi (approved):

khanhtc1202 (approved):
Nice catch ⚾

PR #5249, opened 2024-10-02
Support fetching a piped config from AWS SSM Parameter Store

t-kikuc:

I added the following flags:
- launcher: `--aws-ssm-parameter`, `--config-from-aws-ssm-parameter-store`
- piped: `--config-aws-ssm-parameter `

The motivation is that some users want to store configs in AWS SSM Parameter Store rather than AWS Secrets Manager in order to reduce the cost of storing secrets.

N/A


Warashi (approved):
👍🏻

PR #5254, opened 2024-10-04
Sort subnet in live manifest of ECS Drift Detection

HoangNguyen689:

**Issue**: When viewing the details of the drift-detector on the PipeCD console, the order of subnets in the live state (red) matches the defined resource (servicedef.yaml), but the Git state (green) shows subnets in reverse order.

**Investigation**: After reviewing the [AWS DescribeServices documentation](https://awscli.amazonaws.com/v2/documentation/api/2.0.33/reference/ecs/describe-services.html), I found no evidence that AWS guarantees any sorting of subnets in the returned live state. It appears that PipeCD fetches the live state using DescribeServices.

**Proposed Solution**: I suggest adding explicit sorting of subnets in PipeCD before performing comparisons in the drift detector to avoid confusion.

Fixes #

- **How are users affected by this change**: The diff will be changed on the ECS app.

### DIFF
--- a/pkg/app/piped/driftdetector/ecs/detector.go
+++ b/pkg/app/piped/driftdetector/ecs/detector.go
@@ -280,12 +280,18 @@ func ignoreParameters(liveManifests provider.ECSManifests, headManifests provide
 	}
 	if headService.NetworkConfiguration != nil && headService.NetworkConfiguration.AwsvpcConfiguration != nil {
 		awsvpcCfg := headService.NetworkConfiguration.AwsvpcConfiguration
-		slices.Sort(awsvpcCfg.Subnets) // Livestate's Subnets are sorted by ECS. (SecurityGroups and ContainerDefinitions are not sorted)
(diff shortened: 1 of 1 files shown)


t-kikuc:
@HoangNguyen689 
thank you so much.  let me check again. 


the solution might be one of:
a. remove sorting as you mentioned
b. sort both livestate and head

HoangNguyen689:
@t-kikuc 
I just changed to sort both the live state and the head manifest!

t-kikuc:
@HoangNguyen689 
thank you!
I also think no sorting is ideal,
but let me ask AWS support.

t-kikuc:
@HoangNguyen689 
I asked AWS Support and it turned out that
sorting the `subnets` occurs but **that might be changed** in the future. (so that is not described in any docs)

That's why, sorting both live and head manifests is the best.

t-kikuc (approved):
Thank you.

Sorting both live and head is safer for the future.

Warashi (approved):
Thank you!

PR #5260, opened 2024-10-07
Add workflow steps to build and publish quickstart manifests

khanhtc1202:

This PR updates the workflow to automate the quickstart manifest release process. A PR to update quickstart manifests will be opened after the release process finishes updating the binary to the PipeCD release page.

Part of #5148

### DIFF
--- a/.github/workflows/publish_binary.yaml
+++ b/.github/workflows/publish_binary.yaml
@@ -7,7 +7,6 @@ on:
 
 env:
   GO_VERSION: 1.22.4
-  HELM_VERSION: 3.8.2
 
 jobs:
   gh_release:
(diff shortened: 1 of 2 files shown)


t-kikuc (approved):

.github/workflows/publish_binary.yaml
```diff
@@ -54,3 +55,28 @@ jobs:
...
             ./.artifacts/pipectl_${{ env.PIPECD_VERSION }}_darwin_arm64
+
+      # Building and publishing quickstart manifests.
+      - name: Install Helm
+        uses: Azure/setup-helm@v1
+        with:
+          version: ${{ env.HELM_VERSION }}
+      - name: Build quickstart manifests
+        run: |
+          helm template pipecd oci://ghcr.io/pipe-cd/chart/pipecd --version ${{ env.PIPECD_VERSION }} -n pipecd -f quickstart/control-plane-values.yaml > quickstart/manifests/control-plane.yaml
```
Warashi, line 66: `oci://ghcr.io/pipe-cd/chart/pipecd` is pushed in publish_image_chart.yaml. So I think we should place these codes in it.

https://github.com/pipe-cd/pipecd/blob/f017312df3e14251f9e65dfceb14e83c608b5805/.github/workflows/publish_image_chart.yaml#L134-L139

khanhtc1202: @Warashi Got your point. I think I only want to trigger this workflow on release (the publish_image_chart trigger on every pull request which is merged to master), so I guess we should make a new workflow instead of adding to an existing one. WDYT? 🤔

khanhtc1202: Changed my mind. I moved these steps to the publish_image_chart workflow as your suggestion, with some of the conditions checked in the steps' definition. Please recheck when you have time.

.github/workflows/publish_image_chart.yaml
```diff
@@ -6,6 +6,8 @@ on:
       - master
     tags:
       - 'v*'
+  release:
+    types: [created]
```
Warashi, line 10: I don't know how this workflow works when triggered with a release event.
I think it will work as expected. However, it will run twice simultaneously because we usually push a tag by creating a new release.

I suggest to use `startsWith(github.ref, 'refs/tags/')` to run steps only if the tags are pushed.

ref; https://docs.github.com/ja/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#startswith
ref; https://docs.github.com/ja/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context (`github.ref`)

khanhtc1202: Thanks for the comment, addressed by https://github.com/pipe-cd/pipecd/pull/5260/commits/6c7786ed582d252ff0766df48a9b4efa9b538676 👀

Warashi (approved):
LGTM 👍🏻

t-kikuc (approved):

PR #5262, opened 2024-10-08
Fix quickstart workflow not to run by matrix

Warashi:

Move steps, which creates PR to update quickstarts, as another job.
Without this PR, the step runs multiple times using matrix runs.

Fixes #

### DIFF
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -147,14 +147,31 @@ jobs:
           labels: helmRepo=pipecd
           data: ${{ env.PIPECD_VERSION }}
 
+  update-quickstart:
+    runs-on: ubuntu-latest
+    needs: artifacts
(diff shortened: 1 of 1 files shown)


t-kikuc (dismissed):

.github/workflows/publish_image_chart.yaml
```diff
@@ -147,14 +147,31 @@ jobs:
           labels: helmRepo=pipecd
           data: ${{ env.PIPECD_VERSION }}
 
+  update-quickstart:
```
khanhtc1202, line 150:
```suggestion
  release-quickstart-manifests:
```
How about rename like this?

Warashi: I think its better. Thank you.

khanhtc1202 (approved):
Thanks 👍

t-kikuc (approved):
better name

PR #5266, opened 2024-10-08
Fix the workflow publishes quickstart manifests

Warashi:

1. Add pull-requests write permission to the job
    ref; https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#token
1. Set the base branch of PR
    ref; https://github.com/pipe-cd/pipecd/actions/runs/11229319582/job/31214779847#step:6:52

Fixes #

### DIFF
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -187,3 +187,4 @@ jobs:
           delete-branch: true
           signoff: true
           token: ${{ secrets.GITHUB_TOKEN }}
+          base: master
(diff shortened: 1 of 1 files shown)


khanhtc1202 (dismissed):
Nice catch 👍

Warashi:
I found a better solution. I'll change the actions/checkout arguments, not the create-pull-request arguments.

khanhtc1202 (approved):
Looks better 👍

t-kikuc (approved):

PR #5268, opened 2024-10-08
Remove pipectl quickstart command

### LINKED ISSUE #5148: Easier quickstart for PipeCD

khanhtc1202:

We will use kubectl with quickstart manifests for quickstart, instead of pipectl quickstart command

Fixes #5148


Warashi (approved):
LGTM

PR #5279, opened 2024-10-11
Remove Cloud Providers related docs

khanhtc1202:
**What this PR does**:

**Why we need it**:

We deprecated this config 2 years ago

Fixes #


docs/content/en/docs-v0.49.x/user-guide/managing-piped/adding-a-cloud-provider.md
Warashi, line 1: IMHO, it's better to leave v0.49.x docs because we promise the patch releases don't include breaking changes.
https://github.com/pipe-cd/pipecd/blob/45e18f7f553d02cfef33569e5a2b68518841c544/RELEASES.md?plain=1#L50

Warashi (commented):
IMHO, the deletion of the deprecated config is not a breaking change, but it's a similar one. So we should not include them as a patch release.

docs/content/en/docs-v0.49.x/user-guide/managing-piped/adding-a-cloud-provider.md
khanhtc1202: In this PR, I only want to not mention the CloudProviders in docs to less confuse the newcomers. The old users have already known about the CloudProviders -> PlatformProviders for a while now (we could assume that since this config was marked as deprecated 2 years ago). So I think remove the docs is okay, it's also not a breaking change or any

Warashi: Okay, I got your point.

Warashi (approved):

PR #5290, opened 2024-10-23
Move model.DeploymentSource to deployment.DeploymentSource

Warashi:
**What this PR does**:

as title.

**Why we need it**:

I don't want plugins to depend directly on `pkg/model`.

Both pipedv1 and plugins depend on `DeploymentSource,` so I moved it.

Dependencies remain in Go codes because proto files under `pkg/plugin` depend on `pkg/model`, and the plugin has to build its structs to return them as the gRPC responses.

Part of #4980 


ffjlabo (approved):
LGTM

khanhtc1202 (approved):
LGTM 👍

PR #5291, opened 2024-10-24
Implement LoadManifests with Kustomize template

Warashi:
**What this PR does**:

as title

**Why we need it**:

We must support the kustomize templating method in the k8s plugin because the pipedv0 supports it.

Part of #4980 


khanhtc1202 (approved):
Here you go 👍

PR #5292, opened 2024-10-24
Remove order for stages in case of quick sync

khanhtc1202:
**What this PR does**:

Remove stages' order (indexes) if a quick sync deployment is triggered.

**Why we need it**:

Previously, we planned to make plugins' order as stages order in case of multi-plugin applications. However, it is difficult to decide the order of the application's plugins, and the application's plugins may change, so it is better not to depend on that.

This reverts what we did in #5231 and #5229


Warashi (approved):
Thank you

ffjlabo (approved):
🚀

PR #5297, opened 2024-10-28
Copy pkg/diff to pkg/plugin/diff except DiffStructureds and RenderByCommand

Warashi:
**What this PR does**:

as title.

**Why we need it**:

We want to use the diff package to implement plugins, so I copied them under the plugin package.

**Discussion point**:

1. Do we have to copy the package? Or can we use the `pkg/diff` from plugin implementations directly?
    I thought we should not use `pkg/diff` directly from plugin implementations, so I copied them under the plugin package to represent the package as a prepared utility for plugins.
1. Do we have to copy the whole package at once? Or can we copy the functionality each time we need it?
    I thought we should copy the smallest functionalities set at this time, so I copied not all of the package, but it became almost all. I copied the whole except DiffStructureds and RenderByCommand.

Part of #4980


khanhtc1202:
@Warashi, I got your point 👍 
(2) Tbh, the reason why we should copy only needed functions at a time is that the codebase currently contains some "not using anywhere functions," which we want to remove while implementing the new pipedv1 to avoid potential technical debt at the early stage. So, if you can actually understand that the copied code will be used, just not in the moment but in the next step while implementing, I think copying the whole is fine (if you feel moving/copying a part is annoying). But please remember my point of avoiding dead/useless code while coping.

(1) Same as we discussed before, should this be placed under `pkg/plugin/diff` instead of `pkg/app/pipedv1/plugin/diff`? 🤔

Warashi:
@khanhtc1202 
Thanks for commenting.
I noticed that we have packages such as pkg/app/pipedv1/plugin/toolregistry, so I placed diff under pkg/app/pipedv1/plugin, but now I think it's better we place this at pkg/plugin/diff.
I'll replace it.

khanhtc1202 (approved):
LGTM, thank you 👍

t-kikuc:
@Warashi 
The new title is "Copy pkg/diff to **pkg/plugin/diff** except RenderByCommand", 
not "Copy pkg/diff to pkg/app/pipedv1/plugin/diff except RenderByCommand", right?

Warashi:
@t-kikuc 
Thanks. I changed the PR title.

t-kikuc (approved):

PR #5300, opened 2024-10-31
Delete platform-specific specs from configv1

Warashi:
**What this PR does**:

as title

**Why we need it**:

We move platform-specific config to plugins, so we don't want these at the configv1 package.
I copied them on #5237, but I realized it was not a good way to copy all of them, so I removed some parts on this PR.

Part of #4980 


pkg/configv1/config.go
```diff
@@ -104,30 +105,10 @@ func (c *Config) init(kind Kind, apiVersion string) error {
 	c.APIVersion = apiVersion
 
 	switch kind {
-	case KindApplication:
+	case KindApplication, KindKubernetesApp, KindTerraformApp, KindCloudRunApp, KindLambdaApp, KindECSApp:
```
Warashi, line 108: To have backward compatibility, we treat them as the same as the KindApplication.

khanhtc1202 (approved):
LGTM 👋

ffjlabo (approved):
LGTM

PR #5301, opened 2024-10-31
[docs] Remove Compatibilities from ECS examples

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

`Compatibilities` cannot be used because `RegisterTaskDefinitionInput` does not have it.
It's confusing for users.

cf. 
- https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecs@v1.46.2#RegisterTaskDefinitionInput
- Although `types.TaskDefinition` has it, we cannot pass it when registering.
https://github.com/pipe-cd/pipecd/blob/ad3c8779717d91d0415baec5128f91508a8f9fc0/pkg/app/piped/platformprovider/ecs/client.go#L180-L195

### DIFF
--- a/examples/ecs/attachment/taskdef.yaml
+++ b/examples/ecs/attachment/taskdef.yaml
@@ -9,8 +9,6 @@ containerDefinitions:
     name: web
     portMappings:
       - containerPort: 80
-compatibilities:
-  - FARGATE
 requiresCompatibilities:
   - FARGATE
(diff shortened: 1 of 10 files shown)


Warashi (approved):
👍🏻

ffjlabo (approved):
LGTM 👍 I didn't know that

PR #5302, opened 2024-10-31
Make configv1.Config generic type

Warashi:
**What this PR does**:

This PR makes configv1.Config generic type, and makes LoadFromYAML and DecodeYAML as generic functions.

**Why we need it**:

I want to decode the plugin-specific config in the same way as the piped configs.
The piped-side implementations have no information about plugin-specific config, so I made the decoding function generic.
The plugins can use LoadFromYAML and DecodeYAML with their original type as type parameters.

Part of #4980 


khanhtc1202 (approved):
Looks great, thanks 👍

t-kikuc (approved):
simpler code, thank you

PR #5306, opened 2024-11-01
[ECS] Fix unexpected diff of Plan Preview

t-kikuc:
**What this PR does**:

- Clone manifests in DriftDetection not to modify the original ones in AppManifestsCache. 

_NOTE: I'll fix Lambda too, it has the same problem._

**Why we need it**:

- Without this PR, unexpected diff appears in ECS PlanPreview. 

    e.g. `Subnets`, `Essential`, and `Protocol` are unexpected.
    image

- The cause is that `ignoreAndSortParameters()` in ECS DriftDetection unintentionally overwrites ECS manifests in AppManifestsCache.
    - PlanPreview loads manifests from AppManifestsCache. The cache is shared with DriftDetection.
        https://github.com/pipe-cd/pipecd/blob/3a264ccdf57521bd433f6599203cc6eb51252d09/pkg/app/piped/planpreview/ecsdiff.go#L59
    - It is **values of pointer targets** in the cache that are overwritten.  Cache values themselves are not overwritten.

### DIFF
--- a/pkg/app/piped/driftdetector/ecs/detector.go
+++ b/pkg/app/piped/driftdetector/ecs/detector.go
@@ -209,11 +209,11 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 	d.logger.Info(fmt.Sprintf("application %s has live ecs definition files", app.Id))
 
 	// Ignore some fields whech are not necessary or unable to detect diff.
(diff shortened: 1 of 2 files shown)


Warashi (approved):
LGTM

khanhtc1202 (approved):
Nice catch 👍

PR #5309, opened 2024-11-05
Add execute stage interface to deployment service api

khanhtc1202:
**What this PR does**:

Add `ExecuteStage` rpc to deployment service API. I investigated which current ExecutorInput should contain in case of the `piped-plugin` pattern. Read this note for the detailed field to be added (ref: https://github.com/pipe-cd/pipecd/issues/4980#issuecomment-2452071105)

**Why we need it**:

Part of #4980 


Warashi (approved):
LGTM

ffjlabo (approved):
👍

PR #5310, opened 2024-11-05
Skip commit and push when no replacement happens in EventWatcher

t-kikuc:
**What this PR does**:

- Fix the following bug in EventWatcher by handling no-diff.

**Why we need it**:

- Without this PR, when no diff occurs by an event, the event will fail.

Before fix: `Failed to push changed files: err: exit status 128, out: fatal: invalid refspec ''`
image

After fix: `Nothing to commit`
image

- The cause is that when no diff occurs, `branchName` will be nil as below.

https://github.com/pipe-cd/pipecd/blob/d7f43699085c9da2ddcf913d8866d881cf8038c3/pkg/app/piped/eventwatcher/eventwatcher.go#L643-L645

### DIFF
--- a/pkg/app/piped/eventwatcher/eventwatcher.go
+++ b/pkg/app/piped/eventwatcher/eventwatcher.go
@@ -391,7 +391,8 @@ func (w *watcher) execute(ctx context.Context, repo git.Repo, repoID string, eve
 			switch handler.Type {
 			case config.EventWatcherHandlerTypeGitUpdate:
 				branchName, err := w.commitFiles(ctx, latestEvent, matcher.Name, handler.Config.CommitMessage, e.GitPath, handler.Config.Replacements, tmpRepo, handler.Config.MakePullRequest)
-				if err != nil && !errors.Is(err, errNoChanges) {
+				noChange := errors.Is(err, errNoChanges)
+				if err != nil && !noChange {
 					w.logger.Error("failed to commit outdated files", zap.Error(err))
 					handledEvent := &pipedservice.ReportEventStatusesRequest_Event{
 						Id:                latestEvent.Id,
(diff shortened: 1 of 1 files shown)


khanhtc1202 (dismissed):
Got your point, thanks 👍

pkg/app/piped/eventwatcher/eventwatcher.go
```diff
@@ -388,7 +391,7 @@ func (w *watcher) execute(ctx context.Context, repo git.Repo, repoID string, eve
 			switch handler.Type {
 			case config.EventWatcherHandlerTypeGitUpdate:
 				branchName, err := w.commitFiles(ctx, latestEvent, matcher.Name, handler.Config.CommitMessage, e.GitPath, handler.Config.Replacements, tmpRepo, handler.Config.MakePullRequest)
-				if err != nil {
+				if err != nil && !errors.Is(err, errNoChanges) {
```
ffjlabo, line 394: [IMO] How about using the bool variable like `change` to minimize the scope of the error variable and also for the readability?

```suggestion
				noChange := false
				if err != nil {
				    // 
				    if errors.Is(err, errNoChanges) {
				        noChange = true
				    } else {
				     ...   
				     continue
				    }
				}
				
				if noChange {
				        handledEvent.StatusDescription = "Nothing to commit"
					gitNoChangeEvents = append(gitNoChangeEvents, handledEvent)
				} else {
```

ffjlabo (commented):
Thank you for the imporvement! I left an imo comment

pkg/app/piped/eventwatcher/eventwatcher.go
t-kikuc: thank you , got it, i fixed in [7b155e6](https://github.com/pipe-cd/pipecd/pull/5310/commits/7b155e62a7e7a994b97e6b4ac31a6d082cd415f5)

ffjlabo (approved):
LGTM

PR #5312, opened 2024-11-06
Execute git clean partially when drift detection for every app is done

### LINKED ISSUE #5124: When deploying the helm chart with kustomize, sometimes the helm chart version cannot be updated
**What happened**:

In the PipeCD Application that uses Helm charts, an issue occurred where the version of the Helm chart was not updated even if the version was updated and deployed via PipeCD.

**How to reproduce it**:

**Environment**:
- `piped` version: v0.48.3
- `control-plane` version: v0.48.4
- k8s version: 1.28.11
- Others:

ffjlabo:
**What this PR does**:

Fixed to execute `git clean -f` for every app dir after doing k8s drift detection.

**Why we need it**:

During drift detection, a Git repository is cloned and reused, and a common manifest cache is utilized for various operations like drift detection, plan preview, plan, and deploy.
If an incorrect manifest is cached, it can affect other processes. When running kustomize build --enable-helm, a charts directory is created, and the Helm chart is downloaded locally.
However, before kustomize v5.3.0, this directory is not updated if the Helm chart version in kustomization.yaml changes, causing drift detection to load outdated manifests.
These manifests are cached based on commit hash, leading to incorrect manifests being applied during Plan or Deploy if the same commit hash is used.

For the context, see https://github.com/pipe-cd/pipecd/issues/5124#issuecomment-2277628373

Fixes #5124

### DIFF
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -43,6 +43,7 @@ type Repo interface {
 	Checkout(ctx context.Context, commitish string) error
 	CheckoutPullRequest(ctx context.Context, number int, branch string) error
 	Clean() error
+	CleanPartially(ctx context.Context, relativePath string) error
 
 	Pull(ctx context.Context, branch string) error
 	MergeRemoteBranch(ctx context.Context, branch, commit, mergeCommitMessage string) error
@@ -257,6 +258,15 @@ func (r *repo) CommitChanges(ctx context.Context, branch, message string, newBra
 // Clean deletes all local git data.
 func (r repo) Clean() error {
 	return os.RemoveAll(r.dir)
+}
+
+// CleanPartially deletes data in the given relative path in the repo with git clean.
+func (r repo) CleanPartially(ctx context.Context, relativePath string) error {
+	out, err := r.runGitCommand(ctx, "clean", "-f", relativePath)
+	if err != nil {
+		return formatCommandError(err, out)
+	}
+	return nil
 }
 
 func (r *repo) checkoutNewBranch(ctx context.Context, branch string) error {
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -43,7 +43,7 @@ type Repo interface {
 	Checkout(ctx context.Context, commitish string) error
 	CheckoutPullRequest(ctx context.Context, number int, branch string) error
 	Clean() error
-	CleanPartially(ctx context.Context, relativePath string) error
+	CleanPath(ctx context.Context, relativePath string) error
 
 	Pull(ctx context.Context, branch string) error
 	MergeRemoteBranch(ctx context.Context, branch, commit, mergeCommitMessage string) error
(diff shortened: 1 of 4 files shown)


Warashi (approved):
LGTM

pkg/git/repo.go
```diff
@@ -259,6 +260,15 @@ func (r repo) Clean() error {
 	return os.RemoveAll(r.dir)
 }
 
+// CleanPartially deletes data in the given relative path in the repo with git clean.
+func (r repo) CleanPartially(ctx context.Context, relativePath string) error {
```
khanhtc1202, line 264: Tbh I think `partially` is not suitable for the function name; how about renaming it `CleanPath`?

ffjlabo: @khanhtc1202 
Thank you, got it. Fixed in [ab2b33c](https://github.com/pipe-cd/pipecd/pull/5312/commits/ab2b33c326d2581481523bf43da80afd7242bc55)

pkg/git/repo.go
```diff
@@ -259,6 +260,15 @@ func (r repo) Clean() error {
 	return os.RemoveAll(r.dir)
 }
 
+// CleanPath deletes data in the given relative path in the repo with git clean.
+func (r repo) CleanPath(ctx context.Context, relativePath string) error {
+	out, err := r.runGitCommand(ctx, "clean", "-f", relativePath)
```
khanhtc1202, line 265: Q: Since this is a relative path, do we need to check whether the given path points to an under dir or outer dir? (If it contains `../` or `/,` can the git clean command handle it carefully?)

ffjlabo: @khanhtc1202 
git clean command checks whether the path is under the repo or not.
So IMO, we don't need the check. WDYT?
```
~/oss/pipe-cd/pipecd [fujiwo]
% git clean -f ../tutorial                    
fatal: ../tutorial: '../tutorial' is outside repository at '$HOME/oss/pipe-cd/pipecd
```

ffjlabo: 📝 I tried `git clean` at repo2 with both relative and absolute path.
```
at ~/test/repo2 
% git clean -f ../repo1/                                                         
fatal: ../repo1/: '../repo1/' is outside repository at '/Users/s14218/test/repo2'

at ~/test/repo2 
% git clean -f /Users/s14218/test/repo1                                          
fatal: /Users/s14218/test/repo1: '/Users/s14218/test/repo1' is outside repository at '/Users/s14218/test/repo2'
```

```
```
% tree .
.
├── repo1
│   └── README.md
├── repo2
       └── README.md
```
```

ffjlabo: Also added the test case for them [f1ef1a9](https://github.com/pipe-cd/pipecd/pull/5312/commits/f1ef1a996164b5ec5f054763fa629d7f33185d46)

khanhtc1202 (approved):
LGTM 👍

pkg/git/repo.go
khanhtc1202: That's nice 👍

ffjlabo:
@Warashi Sorry, plz re-approve 🙏

Warashi (approved):
LGTM

PR #5319, opened 2024-11-08
Add function to find ConfigMaps and Secrets in manifests

Warashi:
**What this PR does**:

This PR adds the function `findConfigs` to find ConfigMaps and Secrets in manifests.

**Why we need it**:

To determine sync strategies, we need to find updated ConfigMaps or Secrets. This PR adds functionalities for finding these candidates.

Part of #4980

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/deployment/determine.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/deployment/determine.go
@@ -156,3 +156,16 @@ func findUpdatedWorkloads(olds, news []provider.Manifest) []workloadPair {
 	}
 	return pairs
 }
+
+func findConfigs(manifests []provider.Manifest) map[provider.ResourceKey]provider.Manifest {
+	configs := make(map[provider.ResourceKey]provider.Manifest)
+	for _, m := range manifests {
+		if m.Key.IsConfigMap() {
+			configs[m.Key] = m
+		}
+		if m.Key.IsSecret() {
+			configs[m.Key] = m
+		}
+	}
+	return configs
+}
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/kubernetes/deployment/determine.go
```diff
@@ -156,3 +156,16 @@ func findUpdatedWorkloads(olds, news []provider.Manifest) []workloadPair {
 	}
 	return pairs
 }
+
+func findConfigs(manifests []provider.Manifest) map[provider.ResourceKey]provider.Manifest {
```
ffjlabo, line 160: 📝 Original method
https://github.com/pipe-cd/pipecd/blob/c2bf91d0e02165e1dbeff671d2f9d10d8a2bfda5/pkg/app/piped/planner/kubernetes/kubernetes.go#L388C1-L400C2

ffjlabo (dismissed):
LGTM

pkg/app/pipedv1/plugin/kubernetes/deployment/determine.go
khanhtc1202: Should we add a function comment like "findConfigs returns ConfigMap and Secret resources" to the head of this function? TBH, I don't actually like this implementation (even the original method) since we have to mention that this function returns both ConfigMap and Secret resources while its name makes the reader think it only returns config.

Warashi: @khanhtc1202 
I got your point. How about renaming this as `findConfigsAndSecrets`?

khanhtc1202: That sounds nice 👍 @Warashi

Warashi: I renamed it and added a comment.
https://github.com/pipe-cd/pipecd/pull/5319/commits/dedd8fc5702e3ce0a8d3c71e27c63633d91ff8f9

khanhtc1202 (approved):
LGTM, thanks 👍

ffjlabo (approved):

PR #5331, opened 2024-11-13
Remove global tracer provider to avoid non-expected spans are sent

Warashi:
**What this PR does**:

as title

**Why we need it**:

There are the spans sent, which we didn't implement and don't expect to send.
It seems the otelhttp library sends the trace with the global tracer provider, so I remove the global tracer provider.

Fixes #


khanhtc1202 (approved):
Here you go 👍

t-kikuc (approved):

Issue #5028, opened 2024-07-08
Link application repository PR and the PipeCD deployment related to it.


ffjlabo:
**What would you like to be added**:

When separating the app code repo and manifests repo by using event watcher, 
it would be nice to link the app repo PR and PipeCD deployment.

**Why is this needed**:

There are some use cases.
For example,
- rollback the app code repo quickly when the incident occurs.
- record the time from when a PR is closed until the PipeCD deployment is triggered

Warashi:
I investigated the plan to solve this issue.

## Pre-condition:
- The event operation flow is as follows, so we can only use the commit information of the manifest repository for the deployment operation.
  1. Commit to the Application repository
  1. `pipe-cd/actions-event-register` is invoked
    - In this action, `pipectl event register` is invoked
  1. piped receives the event from the control plane server and commits to the manifest repository
  1. piped pull the manifest repository and detect changes
  1. piped deploy changes to the infrastructure, we can use only the information of the pulled commits.
- We can use the `text/template` feature in the commit message through the event GIT_UPDATE; this is undocumented.
  - https://github.com/pipe-cd/pipecd/blob/713fb032551ce01dd8d15c3ffa7063d91cf4cca1/pkg/app/piped/eventwatcher/eventwatcher.go#L772-L789

## Plan:
### Overview: Pass the context information through actions-event-register -> pipectl -> control-plane-server -> piped -> manifest repo -> piped
### Detail
1. collect GitHub repo/commit context in actions-event-register
1. create a new option `--context` to the pipectl and pass contexts to it
1. add `map<string,string> context` here and pass contexts
  - https://github.com/pipe-cd/pipecd/blob/713fb032551ce01dd8d15c3ffa7063d91cf4cca1/pkg/app/server/service/apiservice/service.proto#L228-L232
1. add `map<string,string> context` here and store contexts
  - https://github.com/pipe-cd/pipecd/blob/713fb032551ce01dd8d15c3ffa7063d91cf4cca1/pkg/model/event.proto#L29-L55
1. add `Context map[string]string` here and pass contexts
  - https://github.com/pipe-cd/pipecd/blob/713fb032551ce01dd8d15c3ffa7063d91cf4cca1/pkg/app/piped/eventwatcher/eventwatcher.go#L766-L770
1. add a document about the `text/template` feature of the GIT_UPDATE event
[...]

PR #5339, opened 2024-11-14
Validate event context key

ffjlabo:
**What this PR does**:

Added validation for event context key.

The event context is used as `--trailer`. The trailer key is currently allowed to use `-` alphabet, number, ' ', '\t' but it is unclear the actual format. 
ref: https://github.com/git/git/blob/25b0f41288718625b18495de23cc066394c09a92/trailer.c#L630-L646

Also tried some formats below.
```
% git commit -m "Test" --trailer "github.repository: test" --trailer "github_repository: test" --trailer "github-repository: test"
% git show                                                                                         
commit bedaed31b10c454e730a7aa06abb118a4cd3c264 (HEAD -> test-trailer)
Author: Yoshiki Fujikane <ffjlabo@gmail.com>
Date:   Tue Nov 12 18:19:55 2024 +0900

    Test

    github.repository: test:
    github_repository: test:
    github-repository: test
```

So I added the restriction for the event context key format like kebab case such as`Test-hoge` `test-hoge` for now.

**Why we need it**:

Part of https://github.com/pipe-cd/pipecd/issues/5028


ffjlabo:
# The pattern

OK (tried `test-hoge`)

```
% ./pipectl_v0.49.3_darwin_arm64 event register --address=localhost:8080 --api-key=xxx --name=simple --data="gcr.io/pipecd/helloworld:v0.40.0" --contexts="test-hoge=fuga" --insecure=true
Successfully registered event	{"id": "cc64b59d-82d1-4d3d-a669-b6051e9fe33b"}
```

Invalid key format (tried `test.hoge`)
```
% ./pipectl_v0.49.3_darwin_arm64 event register --address=localhost:8080 --api-key=xxx --name=simple --data="gcr.io/pipecd/helloworld:v0.40.0" --contexts="test.hoge=fuga" --insecure=true
2024/11/14 16:56:22 failed to register event: rpc error: code = InvalidArgument desc = invalid request: invalid RegisterEventRequest.Contexts[test.hoge]: value does not match regex pattern "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$"
```

Key is nothing
```
% ./pipectl_v0.49.3_darwin_arm64 event register --address=localhost:8080 --api-key=xxx --name=simple --data="gcr.io/pipecd/helloworld:v0.40.0" --contexts="=fuga" --insecure=true
2024/11/14 16:46:29 failed to register event: rpc error: code = InvalidArgument desc = invalid request: invalid RegisterEventRequest.Contexts[]: value length must be at least 1 rune
```

pkg/model/event.go
```diff
@@ -18,11 +18,16 @@ import (
...
+	"regexp"
 	"sort"
 	"strings"
 )
 
-var hashFunc = sha256.New
+var (
+	hashFunc = sha256.New
+
+	eventKeyFormatRegex = regexp.MustCompile(`^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$`)
```
Warashi, line 29: [nits] This variable seems unused

pkg/app/server/service/apiservice/service.proto
```diff
@@ -229,7 +229,7 @@ message RegisterEventRequest {
     string name = 1 [(validate.rules).string.min_len = 1];
     string data = 2 [(validate.rules).string.min_len = 1];
     map<string,string> labels = 3 [(validate.rules).map.keys.string.min_len = 1, (validate.rules).map.values.string.min_len = 1];
-    map<string,string> contexts = 4;
+    map<string,string> contexts = 4 [(validate.rules).map.keys.string.min_len = 1, (validate.rules).map.keys.string.pattern = "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$", (validate.rules).map.values.string.min_len = 1];
```
Warashi, line 232: The regexp might be wrong.
The git implementations seem to allow double or more hyphens continuously like `test--hoge` or start with a hyphen like `-test`.
The pattern might be like `^[\-a-zA-Z0-9]+$`.

Warashi (commented):
commented on regexp and nits.

IMHO, the error message like below is a bit confusing.
The validation error occurs at the key of the map, but the message says "value."
```
2024/11/14 16:46:29 failed to register event: rpc error: code = InvalidArgument desc = invalid request: invalid RegisterEventRequest.Contexts[]: value length must be at least 1 rune
```

pkg/app/server/service/apiservice/service.proto
ffjlabo: @Warashi 
Yes, I know that.
I don't plan to do the same as validation for git.
It is because the trailer format is not so concrete.

So I wanted to express the format like a kebab case `test-hoge` in PipeCD.

Warashi: Okay, I got your point.

pkg/model/event.go
ffjlabo: Fixed 🙏 [341ce20](https://github.com/pipe-cd/pipecd/pull/5339/commits/341ce208e39165a07744faf376c687250861be95)

ffjlabo:
@Warashi 
> commented on regexp and nits.

> IMHO, the error message like below is a bit confusing.
The validation error occurs at the key of the map, but the message says "value."

I see. Thank you for finding it. This is based on the code generated by `protoc-gen-validate`.
In my understanding, many validations are based on it.
Is there any idea?

Warashi:
@ffjlabo 
> I see. Thank you for finding it. This is based on the code generated by protoc-gen-validate.
> In my understanding, many validations are based on it.
> Is there any idea?

How about implementing validations in pipectl code with go's regexp?
In my understanding, many parts of error messages from protoc-gen-validate will not be seen by users, but this is seen by users usually. So it's better that we can control the messages.

ffjlabo:
@Warashi 
Thanks

> In my understanding, many parts of error messages from protoc-gen-validate will not be seen by users, but this is seen by users usually. So it's better that we can control the messages.

I got it. Sounds nice!
So we would accept the message on the server side and implement the validation in pipectl.

📝 It seems that these errors occurred when using pipectl.
e.g. 
```
./pipectl_v0.49.3_darwin_arm64 event register --address=localhost:8080 --api-key=xxx --name=simple --data="gcr.io/pipecd/helloworld:v0.40.0" --labels="=fuga" --insecure=true
2024/11/14 18:06:51 failed to register event: rpc error: code = InvalidArgument desc = invalid request: invalid RegisterEventRequest.Labels[]: value length must be at least 1 runes
```

Warashi (approved):
LGTM

pkg/app/pipectl/cmd/event/register.go
```diff
@@ -79,3 +86,13 @@ func (r *register) run(ctx context.Context, input cli.Input) error {
 	)
 	return nil
 }
+
+func (r *register) validateEventContexts() error {
```
khanhtc1202, line 90: This function looks testable, please add tests for this later (another PR is ok, since this one is for rc release)

khanhtc1202 (approved):
👌

pkg/app/pipectl/cmd/event/register.go
khanhtc1202: cc @ffjlabo

PR #5340, opened 2024-11-15
Add notification event DEPLOYMENT_STARTED

Warashi:
**What this PR does**:

as title

The users can receive notification when the deployments started.

I tested with Slack notification.

スクリーンショット 2024-11-15 9 33 10

**Why we need it**:

I want to know when the Deployment status changed to RUNNING.

Part of #2002 

Users can configure the piped to send notifications when deployments start.


pkg/model/notificationevent.proto
```diff
@@ -63,6 +64,12 @@ message NotificationEventDeploymentPlanned {
     repeated string mentioned_groups = 5;
 }
 
+message NotificationEventDeploymentStarted {
+    Deployment deployment = 1 [(validate.rules).message.required = true];
+    repeated string mentioned_accounts = 3;
+    repeated string mentioned_groups = 4;
+}
```
Warashi, line 71: I skipped field No. 2 because other codes skipped it.
I don't know why, but I think it's better to do the same as the other codes.

ffjlabo (approved):
Thank you! LGTM

khanhtc1202 (approved):
LGTM 👍
Note: We should have some kind of state machine chart on docs to show the event state order

PR #5341, opened 2024-11-15
 Add checkReplicasChange function to detect changes in replicas field 

Warashi:
**What this PR does**:

- [Add checkReplicasChange function to detect changes in replicas field](https://github.com/pipe-cd/pipecd/commit/609e3f344b4dd976cc4e8669f6edaaf0cbebef26)
  - there is a bug in original checkReplicasChange, so [I backported the fix](https://github.com/pipe-cd/pipecd/pull/5341/commits/26616b6ac5d5e95a214e1bf15769ade221ee3ce8)
- [Add support for rendering interface and pointer types in RenderPrimit…](https://github.com/pipe-cd/pipecd/commit/3ea6403feed6d364c3a8e55bbe9672a1b0d362b5)
  - We need this because reflect.Interface are returned as their kind when the replicas field is absent either of before/after manifests

**Why we need it**:

We chose a quick sync strategy when the diffs are only replicas.
We must build a summary message with the workload changed, so we must check it.

Part of #4980 

When users delete or add the spec.replicas field on the Deployment manifest, they can view the plan summary as expected. Without this PR, they will see an invalid value.

### DIFF
--- a/pkg/app/piped/planner/kubernetes/kubernetes.go
+++ b/pkg/app/piped/planner/kubernetes/kubernetes.go
@@ -426,24 +426,25 @@ func checkReplicasChange(ns diff.Nodes) (before, after string, changed bool) {
 	const replicasQuery = `^spec\.replicas$`
 	node, err := ns.FindOne(replicasQuery)
 	if err != nil {
-		// not found or unknown error
+		// no difference between the before and after manifests, or unknown error occurred.
 		return "", "", false
 	}
 
-	before = node.StringX()
 	if node.TypeX == nil {
-		// The replicas field is not found in the old manifest.
-		before = ""
+		// The replicas field is not found in the before manifest.
+		// There is difference between the before and after manifests, So it means the replicas field is added in the after manifest.
+		// So the replicas field in the before manifest is nil, we should return "<nil>" as the before value.
+		return "<nil>", node.StringY(), true
 	}
 
-	after = node.StringY()
 	if node.TypeY == nil {
-		// The replicas field is not found in the new manifest.
-		after = ""
+		// The replicas field is not found in the after manifest.
+		// There is difference between the before and after manifests, So it means the replicas field is removed in the after manifest.
+		// So the replicas field in the after manifest is nil, we should return "<nil>" as the after value.
+		return node.StringX(), "<nil>", true
 	}
 
-	changed = true
-	return before, after, changed
+	return node.StringX(), node.StringY(), true
 }
 
 type containerImage struct {
--- a/pkg/app/piped/planner/kubernetes/kubernetes.go
+++ b/pkg/app/piped/planner/kubernetes/kubernetes.go
@@ -421,17 +421,29 @@ func checkImageChange(ns diff.Nodes) (string, bool) {
 	return desc, true
 }
 
+// checkReplicasChange checks if the replicas field is changed.
 func checkReplicasChange(ns diff.Nodes) (before, after string, changed bool) {
 	const replicasQuery = `^spec\.replicas$`
 	node, err := ns.FindOne(replicasQuery)
 	if err != nil {
-		return
+		// not found or unknown error
+		return "", "", false
 	}
 
 	before = node.StringX()
+	if node.TypeX == nil {
+		// The replicas field is not found in the old manifest.
+		before = ""
+	}
+
 	after = node.StringY()
+	if node.TypeY == nil {
+		// The replicas field is not found in the new manifest.
+		after = ""
+	}
+
 	changed = true
-	return
+	return before, after, changed
 }
 
 type containerImage struct {
--- a/pkg/app/pipedv1/plugin/kubernetes/deployment/determine.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/deployment/determine.go
@@ -193,3 +193,28 @@ func checkImageChange(ns diff.Nodes) (string, bool) {
 	desc := fmt.Sprintf("Sync progressively because of updating %s", strings.Join(images, ", "))
 	return desc, true
 }
+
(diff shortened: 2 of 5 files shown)


pkg/app/piped/planner/kubernetes/kubernetes.go
```diff
@@ -421,17 +421,29 @@ func checkImageChange(ns diff.Nodes) (string, bool) {
...
-		return
+		// not found or unknown error
+		return "", "", false
 	}
 
 	before = node.StringX()
+	if node.TypeX == nil {
+		// The replicas field is not found in the old manifest.
+		before = ""
+	}
```
Warashi, line 437: 🤔 oops, we should put this before as `"nil"`?

Warashi: @khanhtc1202 @t-kikuc @ffjlabo 
WDYT about this?

ffjlabo: @Warashi 
When does `node.TypeX == nil` occur?
I think we should clarify the situation to determine it.

Warashi: @ffjlabo Sorry for forgetting to write it.
When the replicas field is not found in the old manifests, `node.TypeX` becomes `nil` and `node.ValueX` becomes invalid value.

ffjlabo: @Warashi 

> @ffjlabo Sorry for forgetting to write it.
When the replicas field is not found in the old manifests, node.TypeX becomes nil and node.ValueX becomes invalid value.

Thanks.

> When the replicas field is not found in the old manifests,

It seems that the `node` is the one that has `spec.replicas` extracting by the logic before.
So we might not need to check it. 

```
	const replicasQuery = `^spec\.replicas$`
	node, err := ns.FindOne(replicasQuery)
```

I'm sorry if I misunderstood. 🙏

Warashi: Hmm, I think one side of diff defines spec.replicas, and the other side does not; this occurs.

Warashi: @ffjlabo 
Without this check, this test case fails.
https://github.com/pipe-cd/pipecd/pull/5341/files#diff-4c74652886416486eb7b42cc641e5fd09672015dbaeeb19e3dddefaf81e913dbR1351-R1381

ffjlabo: @Warashi 
> Hmm, I think one side of diff defines spec.replicas, and the other side does not; this occurs.

I got your point. You are right. Sorry, I misunderstood. 🙏 
We need it in such a case.

IMO, It might be better to init `before` as empty first. I felt easy to guess the value.

```
before = ""
if node.TypeX != nil {
    // found spec.replicas in the old manifests
    before = node.StringX()
}
```

Warashi: I revised the whole, how about this code?

```go
// checkReplicasChange checks if the replicas field is changed.
func checkReplicasChange(ns diff.Nodes) (before, after string, changed bool) {
	const replicasQuery = `^spec\.replicas$`
	node, err := ns.FindOne(replicasQuery)
	if err != nil {
		// no difference between the before and after manifests, or unknown error occurred.
		return "", "", false
	}

	if node.TypeX == nil {
		// The replicas field is not found in the before manifest.
		// There is difference between the before and after manifests, So it means the replicas field is added in the after manifest.
		return "", node.StringY(), true
	}

	if node.TypeY == nil {
		// The replicas field is not found in the after manifest.
		// There is difference between the before and after manifests, So it means the replicas field is removed in the after manifest.
		return node.StringX(), "", true
	}

	return node.StringX(), node.StringY(), true
}
```

ffjlabo: @Warashi It's also ok 👍 I can understand the code because of the comments.

Warashi: Thanks 👍🏻 

BTW, I realized it's better to return `"<nil>"` instead of an empty string because the before and after are used in this context.
https://github.com/pipe-cd/pipecd/blob/23eddac42ab78523e522ac534437b8718f200ba1/pkg/app/piped/planner/kubernetes/kubernetes.go#L311-L313

Warashi: fixed on this commit
https://github.com/pipe-cd/pipecd/pull/5341/commits/13625cacf2d14b5ff0d3ab0c9701503e8f6dfd50

ffjlabo: @Warashi
Nice catch! I agree with it!  The case is when there is nothing `spec.replicas`.

I think the `<nil>` value should be defined as const value. WDYT?
We might use it to express the same situation. e.g. checkImageTags.

Warashi: @ffjlabo 
I think it's too early to make `"<nil>"` as constant because we use this only two places in one function, and we should use `"<nil>"` carefully because whether the `"<nil>"` or an empty string should be used is case-by-case.

For example, I think the image `nginx` should be just `nginx`, not `nginx:<nil>`.

ffjlabo: @Warashi 
I agree with it. Thanks. So let's rethink it when the same situation will occur. 👌

ffjlabo (approved):
LGTM 🆗

khanhtc1202 (approved):
Here you go 👍

PR #5342, opened 2024-11-15
Introduce plugin logpersister package

khanhtc1202:
**What this PR does**:

Introduce plugin logpsersister package

**Why we need it**:

In pipedv0, we use logPersister as the writer for stage logs (which will later be used to show stage logs on UI). In this PR, I created `pkg/plugin/logpersister` package, with the same responsibility, but instead of sending the stage log directly to the controlplane, it sends to piped plugin-api then piped will send it to the controlplane.

Fixes #


Warashi (approved):
LGTM

ffjlabo (approved):
LGTM

PR #5343, opened 2024-11-18
Implement the tool registry in the piped side

Warashi:
**What this PR does**:

as title

**Why we need it**:

We decided the installation tools should be done on the piped side, not the plugin. ref; https://github.com/pipe-cd/pipecd/pull/5218

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/cmd/piped/grpcapi/tool_registry.go
+++ b/pkg/app/pipedv1/cmd/piped/grpcapi/tool_registry.go
@@ -1,0 +1,133 @@
+// Copyright 2024 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package grpcapi
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"text/template"
+
+	"golang.org/x/sync/singleflight"
+)
+
+type toolRegistry struct {
+	toolsDir string
+	tmpDir   string
+	group    singleflight.Group
+}
+
+type templateValues struct {
+	Name    string
+	Version string
+	OutPath string
+	TmpDir  string
+	Arch    string
+	Os      string
+}
+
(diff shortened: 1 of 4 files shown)


pkg/app/pipedv1/cmd/piped/grpcapi/tool_registry.go
```diff
@@ -0,0 +1,133 @@
...
+	"text/template"
+
+	"golang.org/x/sync/singleflight"
+)
+
+type toolRegistry struct {
+	toolsDir string
+	tmpDir   string
+	group    singleflight.Group
+}
```
khanhtc1202, line 34: Since the tool is not plugin-specific, I think we should have some kind of cache to avoid redundant installation. If the name and version are the same, then we should return the installed path immediately instead of running the script, wdyt?

Warashi: I agree. I'll add the cache like below.
https://github.com/pipe-cd/pipecd/blob/23eddac42ab78523e522ac534437b8718f200ba1/pkg/app/piped/toolregistry/registry.go#L116-L121

Warashi: @khanhtc1202 
fixed on this commit, thanks
https://github.com/pipe-cd/pipecd/pull/5343/commits/75d7daaf84f259431f11e67ec79472b3b7d9025d

pkg/app/pipedv1/cmd/piped/grpcapi/tool_registry.go
```diff
@@ -0,0 +1,150 @@
...
+	Name    string
+	Version string
+	OutPath string
+	TmpDir  string
+	Arch    string
+	Os      string
+}
+
+func newToolRegistry(toolsDir, tmpDir string) (*toolRegistry, error) {
+	tmpDir, err := os.MkdirTemp(tmpDir, "tool-registry")
```
ffjlabo, line 49: [Ask] Is it enough just to create the tmp dir for the tool installation in the newToolRegistry.

Warashi: Do you mean to call like `os.MkdirTemp("", "tool-registry")` and do not pass the `tmpDir` as the argument?

ffjlabo: Yes, I mean fixing as newToolRegistry(toolsDir string) and call os.MkdirTemp inside it as you said.

Warashi: I got your point! That's enough.
I'll change the codes.

Warashi: fixed on this commit
https://github.com/pipe-cd/pipecd/pull/5343/commits/f9f7bf041c6ba8227261cdf65ac4cc7118dbf218

ffjlabo (approved):
LGTM

khanhtc1202 (approved):
Good to go 👍

PR #5344, opened 2024-11-18
Handle ECS Apps DeploymentConfiguration and containers Environments drift detection

### LINKED ISSUE #5336: Drift Detection Issue with ECS Apps
**What happened**:

After upgrading PipeCD to v0.49.3, all ECS applications in the PipeCD console are displayed as "Out Of Sync". 
The detected drift includes:
- Configuration settings that are not specified in the `servicedef.yaml` file.
- Environment variables that have not been changed but are still flagged as differences.

- Unchanged environment variables should not be flagged as drift.
- Drift detection should only report differences for properties explicitly managed by `servicedef.yaml`
- Configuration settings not specified in `servicedef.yaml` should be ignored by the drift detection process.

**How to reproduce it**:

- Navigate to the PipeCD console and view any ECS application.
- Notice that all ECS applications are marked as "Out Of Sync".
- Inspect the detected diff to see unexpected changes.

**Environment**:
- `piped` version: v0.49.3
- `control-plane` version:
- Others:

**Potential Cause**

The issue is likely due to the incorrect handling of the order of environment variables.

**Caution Notes** (advised from @t-kikuc)
- service.DeploymentConfiguration: 
  - If `head` is nil, then `live` should also be set to nil. This is to prevent issues caused by unexpected default values. 
  - Add an item to the func comment. (Under `TODO: Maybe we should check diff of following fields`)
- environment
  - Make sure to sort both live and head environment variables. There’s a possibility that this behaviour might change in the future, so it’s important to handle this now.
  - It’s recommended that the sorting logic be extracted into a separate function, for example: `func sortKeyPairs(kps []types.KeyValuePair)`
  - When sorting, use slices.Clone() to create a copy of the Environment array before performing operations. This prevents modifications to the original array, which could otherwise cause issues in the PlanPreview process.

HoangNguyen689:
**What this PR does**:

**Why we need it**:

Fixes #5336

### DIFF
--- a/pkg/app/piped/driftdetector/ecs/detector.go
+++ b/pkg/app/piped/driftdetector/ecs/detector.go
@@ -236,6 +236,7 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 //   - taskDefinition.ContainerDefinitions[].PortMappings[].HostPort
 //
 // TODO: Maybe we should check diff of following fields when not set in the head manifests in some way. Currently they are ignored:
+//   - service.DeploymentConfiguration
 //   - service.PlatformVersion
 //   - service.RoleArn
 func ignoreParameters(liveManifests provider.ECSManifests, headManifests provider.ECSManifests) (live, head provider.ECSManifests) {
@@ -262,10 +263,7 @@ func ignoreParameters(liveManifests provider.ECSManifests, headManifests provide
 		liveTask.Revision = 0 // TODO: Find a way to compare the revision if possible.
 		liveTask.TaskDefinitionArn = nil
 		for i := range liveTask.ContainerDefinitions {
-			// Sort the live environment variables by name.
-			envs := slices.Clone(liveTask.ContainerDefinitions[i].Environment)
-			envs = sortKeyPairs(envs)
-			liveTask.ContainerDefinitions[i].Environment = envs
+			liveTask.ContainerDefinitions[i].Environment = sortKeyPairs(liveTask.ContainerDefinitions[i].Environment)
 
 			for j := range liveTask.ContainerDefinitions[i].PortMappings {
 				// We ignore diff of HostPort because it has several default values. See https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html#ECS-Type-ContainerDefinition-portMappings.
@@ -305,7 +303,6 @@ func ignoreParameters(liveManifests provider.ECSManifests, headManifests provide
 		liveService.NetworkConfiguration = &types.NetworkConfiguration{AwsvpcConfiguration: &awsvpcCfg}
 	}
 
-	// Ignore the diff of DeploymentConfiguration if it's not specified.
 	if headService.DeploymentConfiguration == nil {
 		liveService.DeploymentConfiguration = nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/app/piped/driftdetector/ecs/detector.go
```diff
@@ -480,3 +497,13 @@ func ignoreAutoScalingDiff(r *provider.DiffResult) bool {
 		r.New.ServiceDefinition.DesiredCount == 0 && // When desiredCount is 0 or not defined in the head manifest, autoscaling may be enabled.
 		r.Old.ServiceDefinition.DesiredCount != r.New.ServiceDefinition.DesiredCount
 }
+
+func sortKeyPairs(kps []types.KeyValuePair) []types.KeyValuePair {
+	sorted := make([]types.KeyValuePair, len(kps))
+	copy(sorted, kps)
```
t-kikuc, line 503: How about cloning here and removing `slices.Clone` when calling? That will reduce copying slices.

```suggestion
	sorted := slices.Clone(kps)
```

We can notice this function won't have side effects from a caller side because it returns slices.

pkg/app/piped/driftdetector/ecs/detector.go
```diff
@@ -316,6 +327,12 @@ func ignoreParameters(liveManifests provider.ECSManifests, headManifests provide
 			// Essential is true by default. See https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html#ECS-Type-ContainerDefinition-es.
 			cd.Essential = aws.Bool(true)
 		}
+
+		// Sort the head environment variables by name.
+		envs := slices.Clone(cd.Environment)
+		envs = sortKeyPairs(envs)
+		cd.Environment = envs
```
t-kikuc, line 334: After refactoring `sortKeyPairs()`

```suggestion
		cd.Environment = sortKeyPairs(cd.Environment)
```

pkg/app/piped/driftdetector/ecs/detector.go
```diff
@@ -299,6 +305,11 @@ func ignoreParameters(liveManifests provider.ECSManifests, headManifests provide
 		liveService.NetworkConfiguration = &types.NetworkConfiguration{AwsvpcConfiguration: &awsvpcCfg}
 	}
 
+	// Ignore the diff of DeploymentConfiguration if it's not specified.
```
t-kikuc, line 308: Would you add `- service.DeploymentConfiguration` to the function's comment?

```go
// TODO: Maybe we should check diff of following fields when not set in the head manifests in some way. Currently they are ignored:
//   - service.DeploymentConfiguration                   <--------- HERE
//   - service.PlatformVersion
//   - service.RoleArn
func ignoreParameters(liveManifests provider.ECSManifests, headManifests provider.ECSManifests) (live, head provider.ECSManifests) {
```

HoangNguyen689: added in https://github.com/pipe-cd/pipecd/pull/5344/commits/930b46fc7f1fdf5523793cb7d503330cbdd6e7bb

pkg/app/piped/driftdetector/ecs/detector.go
HoangNguyen689: fixed in https://github.com/pipe-cd/pipecd/pull/5344/commits/930b46fc7f1fdf5523793cb7d503330cbdd6e7bb

t-kikuc (approved):

ffjlabo (approved):
Thank you for the contribution!

Issue #5252, opened 2024-10-04
Make PipeCD control plane support plugin-arched piped


ffjlabo:
**What would you like to be added**:

This issue tracks the progress for deprecating Platform Provider and Kind toward piped plugin architecture in pipedv1. 

The purpose is to manage application kind and platform providers on the PipeCD plugin side in the future.

Considering the impact on users, pay attention to the following points:
- In Pipedv1, platform providers, kind remain on the data and do not create logic with them. Use the value you define instead.
- minify the migration for users.

**current update process**

1. Update data for the pipedv1
  -  Add the label `kind` based on the current Application Kind to the Application.
  - Add `deployTarget`  based on the current Application Platform Provider to the Application.
2. update to pipedv1 (users can use existing application at the time)

### For Platform Provider

Instead of Platform Provider, we plan to introduce the config for the plugin and define deployTargets.

**piped config**
```yaml
apiVersion: pipecd.dev/v1beta1
kind: Piped
spec:
...
  plugin:
    - name: k8s_plugin
      port: 8081
      deployTargets:
        - name: dev
          labels: 
            env: dev
          config: # depends on plugins
            masterURL: http://cluster-dev
            kubeConfigPath: ./kubeconfig-dev
```

```golang
type PipedDeployTarget struct {
	Name   string                     `json:"name"`
	Labels map[string]string          `json:"labels,omitempty"`
	Config json.RawMessage            `json:"config"`
}
```

We also plan to deploy the app to multiple targets at once in a multicluster feature for k8s.
So, we define `DeployTargets` as an array in Application and Deployment.

**Application**

```proto
message Application {
    reserved 3;
    ...
    // TODO: Add validation for this field.
    string platform_provider = 15;
    // 
    repeated string deploy_targets = 16;
    ...
}
```

**Deployment**

```proto
message Deployment {
    reserved 4;
    ...
    // The name of platform provider where to deploy this application.
    // This must be one of the provider names registered in the piped.
    string platform_provider = 11;
    
    repeated string deploy_targets = 12;
}
```

#### For the backward compatibility

Before updating the piped, users should migrate the current data on DB.

<details>

<summary>Former idea</summary>

During the migration, there might be both platform providers and deploy targets in the piped config.
So we need to convert the platform providers to deploy targets internally.

**Refer the Platform Provider or Deploy Target**

- If the ApplicationKind is `Application`, just use `DeployTarget`
[...]

PR #5358, opened 2024-11-21
Add kind APPLICATION

ffjlabo:
**What this PR does**:

Added new kind `APPLICATION` as ApplicationKind and the method to get the kind as string. This is for the app in pipedv1.

In the pipedv1, we use labels["kind"] instead of the current Kind.

See more: https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0015-pipecd-plugin-arch-meta.md#for-kind

**Why we need it**:

Part of https://github.com/pipe-cd/pipecd/issues/5252


pkg/model/application.go
```diff
@@ -78,6 +78,17 @@ func (a *Application) ContainLabels(labels map[string]string) bool {
 	return true
 }
 
+func (a *Application) GetKindString() string {
```
ffjlabo, line 81: I wanted to use `GetKind`. But there is the same name method generated by protobuf.

Warashi (approved):
👍🏻

ffjlabo:
After a discussion with @khanhtc1202, I found that this is no use for pipedv1.

We don't use the `Kind` and `Platform Provider` because we migrate the DB data before updating pipedv1.
After updated, the pipedv1 refers `labels["kind"]` and `daployTarget` instead of `Kind` and `Platform Provider`.

So, close this PR.

FYI, draft migration process.
1. Update data for the pipedv1
  -  Add the label `kind` based on the current Application Kind to the Application.
  - Add `deployTarget`  based on the current Application Platform Provider to the Application.
2. update to pipedv1 (users can use existing application at the time)

PR #5360, opened 2024-11-22
Refactor launcher package to lifecycle

Warashi:
**What this PR does**:

This PR adds the package "lifecycle".
The "lifecycle" package is from launcher codes, and the launcher codes become using this package.

**Why we need it**:

We want to maintain the plugin lifecycle with these codes.

Part of #4980 


khanhtc1202 (approved):
Nice 👍

t-kikuc (approved):

Issue #5363, opened 2024-11-25
Livestate feature in plugin architecture


ffjlabo:
**What would you like to be added**:

This issue is the meta (epic) issue for the livestate feature in plugin architecture.

TODO
- [x] make RFC
- [x] define an interface for livestate
- [ ] ~implement livestate store on the plugin side~
- [x] implement livestate reporter on the piped side
- [x] Implement Web UI

**Why is this needed**:

ffjlabo:
completed

PR #5364, opened 2024-11-25
Add draft rfc for livestate feature in plugin architecture

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

Part of #5363


t-kikuc (approved):
Great flow chart!

It seems complex to handle some points which are specific to some plugins.
e.g. `pass it to the executor.Input and use it on the k8s stage`

Warashi (approved):
LGTM

ffjlabo:
> It seems complex to handle some points which are specific to some plugins.
e.g. pass it to the executor.Input and use it on the k8s stage

That's true. In my understanding, livestatestore is used only in the executor.
So I plan to use livestatestore directly from the executor in the plugin side.

PR #5365, opened 2024-11-25
Modify rfc for deprecating Kind and Platform Provider

ffjlabo:
**What this PR does**:

Modified the way to deprecate Kind and Platform Provider in the RFC for plugin architecture.

**Why we need it**:

We don't use the `Kind` and `Platform Provider` because we migrate the DB data before updating pipedv1.
After updated, the pipedv1 refers `labels["kind"]` and `daployTarget` instead of `Kind` and `Platform Provider`.

Part of #5252
Related to #5358


Warashi (approved):
I got your point. LGTM.

khanhtc1202 (approved):
👍 LGTM

PR #5371, opened 2024-11-29
Partial implementation of the K8S_SYNC stage

Warashi:
**What this PR does**:

This PR adds a partial implementation of the K8S_SYNC stage.

The unimplemented operations are below.
- manifests caching
    - duplicate manifests to avoid modifying cached manifests
- handle QuickSync.AddVariantLabelToSelector
- add config hash annotations to workloads
    - to ensure the workloads are restarted when the config changes
- construct applier from plugin configurations
- resource pruning

**Why we need it**:

We have to implement deploying k8s manifests.

Part of #4980 


pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -175,3 +181,89 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
...
+	// TODO: handle cfg.Spec.QuickSync.AddVariantLabelToSelector
+
+	// Add builtin annotations for tracking application live state.
+	addBuiltinAnnotations(
+		manifests,
+		variantLabel,
+		primaryVariant,
+		input.GetDeployment().GetTrigger().GetCommit().GetHash(),
+		input.GetDeployment().GetPipedId(),
+		input.GetDeployment().GetApplicationId(),
```
khanhtc1202, line 232: For these annotations, how about adding it under loadManifests logic? This is quite common and should be force-added to any manifests under pipecd management.

Warashi: It's nice. I'll change the codes. Thanks.

Warashi: @khanhtc1202 
fixed on this commit. Please check it.
https://github.com/pipe-cd/pipecd/pull/5371/commits/390bcabf79f3001ed6e9bf3c9425d510255f427b

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -175,3 +184,72 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
...
+		return a.executeK8sSyncStage(ctx, request.GetInput())
+	case StageK8sRollback.String():
+		return a.executeK8sRollbackStage(ctx, request.GetInput())
+	default:
+		return nil, status.Error(codes.InvalidArgument, "unimplemented or unsupported stage")
+	}
+}
+
+func (a *DeploymentService) executeK8sSyncStage(ctx context.Context, input *deployment.ExecutePluginInput) (*deployment.ExecuteStageResponse, error) {
+	lp := a.LogPersister.StageLogPersister(input.GetDeployment().GetId(), input.GetStage().GetId())
```
khanhtc1202, line 200: Be sure to call `lp.Complete(timeout)` once we finish this stage execution so that the log persisted bulk can be clean up properly
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/controller/scheduler.go#L509 (for how to close in case of pipedv0)

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -51,13 +52,18 @@ type applier interface {
...
+}
+
 type DeploymentService struct {
 	deployment.UnimplementedDeploymentServiceServer
 
 	RegexPool    *regexpool.Pool
 	Logger       *zap.Logger
 	ToolRegistry toolRegistry
 	Loader       loader
+	LogPersister logPersister
```
khanhtc1202, line 66: I do not actually remember the reason why these properties should be set as public 🤔

Warashi: I think it's a mistake, so I'll make these private on another PR.

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
Warashi: @khanhtc1202
Thank you for letting me know that point.
It's okay to call lp.Complete on the plugin side. but I think it's a control-flow operation, so it's nice to be done on the piped side. WDYT?

khanhtc1202: The current implementation of log persister on the piped side is just a proxy to push logs sent by plugins to the control plane (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/pipedv1/cmd/piped/service/service.proto#L30-L33), it has no idea with the log persister object, so I can not image how the piped can send a signal to complete a log persister object which is created in the plugin. WDYT? 👀 
Also, since a logpersister (lp) object is a stage-specific object, it could be easy to "close" it when the stage execution is completed. @Warashi

Warashi: I read the codes and realized that I misunderstood the log persister. As you said, it's challenging to do on the piped side. Thank you.

Warashi: @khanhtc1202 
fixed on this commit.
https://github.com/pipe-cd/pipecd/pull/5371/commits/3233eaffb4352672e4c25c642c2c4025a1adb0c0

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -175,3 +185,81 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
...
+		return a.executeK8sSyncStage(ctx, request.GetInput())
+	case StageK8sRollback.String():
+		return a.executeK8sRollbackStage(ctx, request.GetInput())
+	default:
+		return nil, status.Error(codes.InvalidArgument, "unimplemented or unsupported stage")
+	}
+}
+
+func (a *DeploymentService) executeK8sSyncStage(ctx context.Context, input *deployment.ExecutePluginInput) (response *deployment.ExecuteStageResponse, err error) {
+	// TODO: move this to the ExecuteStage function and pass the log persister as an argument?
```
khanhtc1202, line 201: Good insight 👍 Let's try this later!

khanhtc1202 (approved):
LGTM 👍

ffjlabo (approved):
👍

PR #5373, opened 2024-11-29
Fix unable to load manifest for PlanPreview in case of ECS standalone tasks

### LINKED ISSUE #5370: [ECS] PlanPreview of standalone tasks failed
**What happened**:

PlanPreview for ECS standalone tasks failed with the following error:

```
failed to load ecs manifests at the head commit (read /tmp/plan-preview-builder-xxx/deploysource3xxxx/repo-xxx/xxx/xxx: is a directory)
```

Not failed and only TaskDefinitions are compared.

**How to reproduce it**:

1. ADD an application of an ECS standalone task
2. Create a PullRequest with diff of the standalone task
3. Call PlanPreview

**Environment**:
- `piped` version: v0.49.4
- `control-plane` version: v.049.4
- Others:

**The cause**

Currently, PlanPreview for ECS tries to load both serviceDef and taskDef.
However, standalone tasks don't have serviceDef.
That's why, all PlanPreview for ECS standalone tasks will fail.

https://github.com/pipe-cd/pipecd/blob/57f0672d34a907e1d13ff6b2f45e66ae973f9fe9/pkg/app/piped/planpreview/ecsdiff.go#L41-L45

cuongvd23:
**What this PR does**:

Skip load `serviceDef` on the ECS standalone task since there is no `serviceDef` file while running planpreview logic

**Why we need it**:

Fixes #5370

### DIFF
--- a/pkg/app/piped/planpreview/ecsdiff.go
+++ b/pkg/app/piped/planpreview/ecsdiff.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 	"github.com/pipe-cd/pipecd/pkg/app/piped/deploysource"
 	provider "github.com/pipe-cd/pipecd/pkg/app/piped/platformprovider/ecs"
 	"github.com/pipe-cd/pipecd/pkg/diff"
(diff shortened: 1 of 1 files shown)


khanhtc1202 (approved):
Good catch 👍

pkg/app/piped/planpreview/ecsdiff.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
```
t-kikuc, line 23:
```suggestion
	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
	
```

t-kikuc (approved):
It worked well, thank you!
image

PR #5375, opened 2024-11-30
refactor: Applied docker build practices

ponkio-o:
**What this PR does**: To follow docker best practices.

This PR applied the following practices:

- [JSONArgsRecommended](https://docs.docker.com/reference/build-checks/json-args-recommended/)
- [LegacyKeyValueFormat](https://docs.docker.com/reference/build-checks/legacy-key-value-format/)
- [StageNameCasing](https://docs.docker.com/reference/build-checks/stage-name-casing/)
- [NoEmptyContinuation](https://docs.docker.com/reference/build-checks/no-empty-continuation/)
- [SecretsUsedInArgOrEnv](https://docs.docker.com/reference/build-checks/secrets-used-in-arg-or-env/)

**Why we need it**: To avoid future errors

**Which issue(s) this PR fixes**: None

- **How are users affected by this change**: No impact on users

### DIFF
--- a/tool/piped-base-okd/Dockerfile
+++ b/tool/piped-base-okd/Dockerfile
@@ -1,3 +1,4 @@
+# check=skip=SecretsUsedInArgOrEnv
 # https://github.com/pipe-cd/pipecd/pkgs/container/piped-base/103341355?tag=v0.44.0-alpha-52-ge881cac
 FROM ghcr.io/pipe-cd/piped-base@sha256:18a5a6af6507c8c70c5442d19651a2000672c78fc5dd91cfd3c0af26c899c1d3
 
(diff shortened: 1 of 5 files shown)


tool/piped-base-okd/Dockerfile
```diff
@@ -1,3 +1,4 @@
+# check=skip=SecretsUsedInArgOrEnv
```
ponkio-o, line 1: The following code was pointed out to me, but I ignored it because it was a false positve.
https://github.com/pipe-cd/pipecd/blob/a5327183a47189ec530cead2d8b7a4093d6895cc/tool/piped-base-okd/Dockerfile#L39

ref. https://docs.docker.com/build/checks/#skip-checks

t-kikuc (approved):
Thank you so much!!!  Really helpful.

Warashi (approved):
Great! Thank you!

PR #5377, opened 2024-12-02
Remove plugin config from plugin deployment api

khanhtc1202:
**What this PR does**:

Remove plugin config from plugin deployment APIs.

**Why we need it**:

We can pass the plugin config to the plugin once the piped starts the plugins; we don't need to pass the plugin config each time the piped calls the plugins' interfaces.

Fixes #


ffjlabo (approved):
LGTM 👍 
Let's merge after CI has passed

Warashi (approved):
Thanks.
BTW, do we have to add an RPC like `InitializePlugin`?

khanhtc1202:
> Thanks. BTW, do we have to add an RPC like `InitializePlugin`?

I will think about this while implementing the piped start plugins logic 👍

PR #5378, opened 2024-12-02
Add deploy target

ffjlabo:
**What this PR does**:

Added deployTarget to the model and piped config to use them on the pipedv1.

**Why we need it**:

Part of #5252


pkg/model/application.proto
```diff
@@ -43,6 +43,8 @@ message Application {
     // This must be one of the provider names registered in the piped.
     // TODO: Add validation for this field.
     string platform_provider = 15;
+
+    repeated string deploy_targets = 16;
```
Warashi, line 47: How about adding comments and a validation?
The validation may be like `min_len = 1`

Warashi (commented):
almost LGTM
I commented on one nitpick

pkg/model/application.proto
khanhtc1202: This should not have validation, or application which created by pipedv0 will be invalided :eyes:

Warashi: Oops, that's right 🙏🏻

khanhtc1202: Also, could you add comment to explain what does this field do or be used for @ffjlabo

khanhtc1202:
@ffjlabo You may need to update dummyApplication to pass the web test 👀

pkg/model/application.proto
ffjlabo: Forgot to  add comment 🙏  will do it.

ffjlabo: @Warashi  @khanhtc1202 fied f31a246

ffjlabo:
@khanhtc1202 fixed for lint/web [3bc49da](https://github.com/pipe-cd/pipecd/pull/5378/commits/3bc49dae6970d88d3635200c5c2970757e80ba65)

ffjlabo:
@khanhtc1202 @Warashi Sorry for my sloppy fixes 🙏 Fixed and passed the CI

Warashi (approved):
Thanks. LGTM 👍🏻

khanhtc1202 (approved):
Here you go 👍

PR #5379, opened 2024-12-02
Change kubectl DL host to dl.k8s.io

### LINKED ISSUE #5369: Can't use kubectl v1.30.5 because of the depracation of the gs://kubernetes-release
**What happened**:

The deployment failed when using kubectl v1.30.5.
It seems that the origin bucket is deprecated. ref: https://github.com/kubernetes/k8s.io/issues/2396

```
Failed to apply manifest: name="xxx", kind="ConfigMap", namespace="default", apiVersion="v1" (%!w(*fmt.wrapError=&{failed to apply:  (fork/exec /home/piped/.piped/tools/kubectl-1.30.5: exec format error) 0xc011f0c270}))
```

https://github.com/kubernetes/k8s.io/issues/2396

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

ffjlabo:
**What this PR does**:

Changed the URL for installing the kubectl. It seems that some versions are not hosted on the older URL. ( e.g. v1.30.5

I tested whether to use kubectl 1.30.5 and it works well.
```
...
  platformProviders:
    - name: kubernetes-dev
      type: KUBERNETES
      labels:
        env: local
      config:
        masterURL: https://127.0.0.1:56247
        kubeConfigPath: kubeconfig
        kubectlVersion: 1.30.5
```

**Why we need it**:

We should use https://dl.k8s.io/ instead of https://storage.googleapis.com/kubernetes-release because the old one is deprecated.

ref: https://github.com/kubernetes/kubernetes/issues/127796#issuecomment-2392019607

Fixes #5369

### DIFF
--- a/pkg/app/piped/toolregistry/tool_darwin.go
+++ b/pkg/app/piped/toolregistry/tool_darwin.go
@@ -16,7 +16,7 @@ package toolregistry
 
 var kubectlInstallScript = `
 cd {{ .WorkingDir }}
-curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{ .Version }}/bin/darwin/amd64/kubectl
+curl -LO https://dl.k8s.io/release/v{{ .Version }}/bin/darwin/amd64/kubectl
 mv kubectl {{ .BinDir }}/kubectl-{{ .Version }}
 chmod +x {{ .BinDir }}/kubectl-{{ .Version }}
 {{ if .AsDefault }}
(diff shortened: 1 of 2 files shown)


t-kikuc (approved):
Nice!!!

Warashi (approved):
I checked that we can download kubectl v1.18.2 from this host.
This version is our default kubectl version.

PR #5381, opened 2024-12-03
Refactor the k8s plugin

Warashi:
**What this PR does**:

1. make the DeploymentService fields private and remove unused fields.
1. add arguments of NewDeploymentService to make tool registry and log persister as arguments

**Why we need it**:

1. Making the variable scope narrow is better
1. we have to create gRPC conn for piped-side service, which may be done out of the initialization of DeploymentService. So we have to pass them as arguments.

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
@@ -21,6 +21,7 @@ import (
 
 	kubeconfig "github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/kubernetes/config"
 	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/kubernetes/provider"
+	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/kubernetes/toolregistry"
 	config "github.com/pipe-cd/pipecd/pkg/configv1"
 	"github.com/pipe-cd/pipecd/pkg/model"
 	"github.com/pipe-cd/pipecd/pkg/plugin/api/v1alpha1/deployment"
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -60,21 +60,21 @@ type logPersister interface {
...
 	logger *zap.Logger,
+	toolRegistry toolRegistry,
+	logPersister logPersister,
 ) *DeploymentService {
 	return &DeploymentService{
-		RegexPool:    regexpool.DefaultPool(),
-		Logger:       logger.Named("planner"),
-		ToolRegistry: nil, // TODO: set the tool registry
+		logger:       logger.Named("planner"),
+		loader:       provider.NewLoader(toolregistry.NewRegistry(toolRegistry)),
```
ffjlabo, line 76: IMO: How about changing the interface name of `toolRegistry` to `toolClient`?
At first, I was confused that it created a new tool Registry with tool registry.

ffjlabo (commented):
Thank you for the refactor. I left imo comment 🙏

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
Warashi: I agree that it's confusing. I'll rename it as you suggested. Thanks.

Warashi: @ffjlabo 
fixed on https://github.com/pipe-cd/pipecd/pull/5381/commits/ee332f54b0720e0612c4b98cc1f8c31fbbb548ae
Thanks.

ffjlabo (approved):
🚀

t-kikuc (approved):

PR #5384, opened 2024-12-03
Add context to RunBinary interface

khanhtc1202:
**What this PR does**:

Add context to the RunBinary interface to have more control.
Also, add test for the DownloadBinary function.

**Why we need it**:

Instead of `context.Background`, use context passed by the run binary caller side would gain the caller more control on the binary lifecycle.

Part of #4980
Follows #5361


Warashi (approved):
LGTM 👍🏻

t-kikuc (approved):

PR #5385, opened 2024-12-03
Implement applier for the kubernetes plugin

Warashi:
**What this PR does**:

This PR implements the Kubernetes applier functionalities.
Almost of this PR is copy of [pkg/app/piped/platformprovider/kubernetes/applier.go](https://github.com/pipe-cd/pipecd/blob/1907301ca6de307381db05bd594768d610591a6b/pkg/app/piped/platformprovider/kubernetes/applier.go).

I made the modifications below with the above file.
- make kubectl wrapper as constructor arguments
    - before, the kubectl wrapper is initialized in the first call of the applier method
- make the struct public
    - before, the Applier interface was public and the struct is private
    - ref; [Go Code Review Comments #Interfaces](https://go.dev/wiki/CodeReviewComments#interfaces).
- Add unit test for methods

**Why we need it**:

- to implement k8s stages

Part of #4980


t-kikuc (approved):
lgtm

khanhtc1202 (approved):
Here you go 👍

PR #5386, opened 2024-12-03
Fix missing call to cancel context in backoff test

khanhtc1202:
**What this PR does**:

Fix missing call to cancel context in backoff test

**Why we need it**:

My go lint yields this error potential context leak due to not calling context cancel in all paths

Fixes #


Warashi (approved):
Nice catch! LGTM 👍🏻

ffjlabo (approved):

PR #5387, opened 2024-12-03
Implement initialization of piped service client in k8s plugin

Warashi:
**What this PR does**:

as title

**Why we need it**:

We need the piped's plugin service client to call the APIs.

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/server.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/server.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"net"
 	"strconv"
 	"time"
@@ -46,7 +47,7 @@ type server struct {
 func NewServerCommand() *cobra.Command {
 	s := &server{
 		apiPort:                10000,
-		pipedPluginServicePort: 9087,
+		pipedPluginServicePort: -1, // default as error value
 		gracePeriod:            30 * time.Second,
 	}
 	cmd := &cobra.Command{
@@ -56,7 +57,7 @@ func NewServerCommand() *cobra.Command {
 	}
 
 	cmd.Flags().IntVar(&s.apiPort, "api-port", s.apiPort, "The port number used to run a grpc server for external apis.")
-	cmd.Flags().IntVar(&s.pipedPluginServicePort, "piped-plugin-service-port", s.pipedPluginServicePort, "The port number used to connect to the piped plugin service.")
+	cmd.Flags().IntVar(&s.pipedPluginServicePort, "piped-plugin-service-port", s.pipedPluginServicePort, "The port number used to connect to the piped plugin service.") // TODO: we should discuss about the name of this flag, or we should use environment variable instead.
 	cmd.Flags().DurationVar(&s.gracePeriod, "grace-period", s.gracePeriod, "How long to wait for graceful shutdown.")
 
 	cmd.Flags().BoolVar(&s.tls, "tls", s.tls, "Whether running the gRPC server with TLS or not.")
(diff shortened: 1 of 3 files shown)


ffjlabo (dismissed):
LGTM

pkg/app/pipedv1/plugin/kubernetes/server.go
```diff
@@ -16,31 +16,38 @@ package main
...
 	enableGRPCReflection bool
 }
 
 // NewServerCommand creates a new cobra command for executing api server.
 func NewServerCommand() *cobra.Command {
 	s := &server{
-		apiPort:     10000,
-		gracePeriod: 30 * time.Second,
+		apiPort:                10000,
+		pipedPluginServicePort: 9087,
```
t-kikuc, line 49: [ASK] Where does `9087` come from?

Do other plugins need to avoid using `9087`?

t-kikuc (commented):
I wanna ask one point

pkg/app/pipedv1/plugin/kubernetes/server.go
Warashi: Sorry for not commenting about this.
This comes from the default of pipedv1's plugin service port.
https://github.com/pipe-cd/pipecd/blob/4986dd5158d424bc841c19713094e611b1235479/pkg/app/pipedv1/cmd/piped/piped.go#L103

t-kikuc: I got it, thank you.
I think it'd be better to avoid hard coding (e.g., by using const) because all plugins need to specify 9087 directly.

If it's not easy now, let's do it in another PR later.

Warashi: I came to one idea.
The piped will always specify the port.
Then, making the default `-1` and the plugin raises the error if the piped doesn't specify the port.
WDYT?

t-kikuc: It seems good!

You mean a piped passes the piped's port to each plugin, right?  If not passed, an error will occur by `-1`

Warashi: Yes 👍🏻 
I realized the way to pass this port must be the same across all the plugins.
We have to discuss how to pass it (e.g., as a command option or an environment variable, and these names, too).

Warashi: @t-kikuc
I changed the default to -1 and added validation.
In this PR, I want to merge it with this implementation.

https://github.com/pipe-cd/pipecd/pull/5387/commits/81bef1814ee9a775d9a92d34c4c7c8b89e9012f6

t-kikuc: @Warashi Thank you!

t-kikuc (dismissed):

pkg/app/pipedv1/plugin/kubernetes/server.go
```diff
@@ -68,15 +77,24 @@ func (s *server) run(ctx context.Context, input cli.Input) (runErr error) {
 
 	group, ctx := errgroup.WithContext(ctx)
 
-	// TODO: create the piped plugin gRPC client here.
+	if s.pipedPluginServicePort == -1 {
+		input.Logger.Error("piped-plugin-service-port is required")
+		return errors.New("piped-plugin-service-port is required")
+	}
```
ffjlabo, line 83: We can use `MarkFlagRequired` to require the flag 👀 
https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired

ffjlabo (commented):
Thanks! Left a commet.

pkg/app/pipedv1/plugin/kubernetes/server.go
Warashi: @ffjlabo 
Thank you for notifying me 👍🏻 
Fixed on this commit
https://github.com/pipe-cd/pipecd/pull/5387/commits/b66474b27b40adb0cb0bac05768a76d10c78f836

t-kikuc: @Warashi 

Then the following line is unnecessary, right?

```go
		pipedPluginServicePort: -1, // default as error value
```

Warashi: oops, that's right. I'll change

Warashi: thanks. fixed on this commit
https://github.com/pipe-cd/pipecd/pull/5387/commits/2d1da06cf08b164298c627d9b58a902adc64ab9b

t-kikuc (approved):

ffjlabo (approved):
🚀

PR #5388, opened 2024-12-03
Using backoff Do instead of WaitNext function to avoid potential panic

khanhtc1202:
**What this PR does**:

Using backoff.Do in pipedv1 codebase instead of backoff.WaitNext

**Why we need it**:

In the backoff package, we have two functions that handle retry logic: WaitNext and Do. The main difference between them (aside from the interface) is that WaitNext does not raise an error in case the context is canceled/stopped, while Do does. So if using WaitNext and the returned value of the caller are used as a check for further logic, it may cause panic due to no error returned while the value is nil either.

Fixes #

### DIFF
--- a/pkg/app/pipedv1/controller/scheduler_test.go
+++ b/pkg/app/pipedv1/controller/scheduler_test.go
@@ -295,7 +295,7 @@ func TestExecuteStage_SignalCancelled(t *testing.T) {
 					Id:     "stage-id",
 					Name:   "stage-name",
 					Index:  0,
-					Status: model.StageStatus_STAGE_RUNNING,
+					Status: model.StageStatus_STAGE_NOT_STARTED_YET,
 				},
 			},
 		},
--- a/pkg/app/pipedv1/controller/scheduler_test.go
+++ b/pkg/app/pipedv1/controller/scheduler_test.go
@@ -270,7 +270,7 @@ func TestExecuteStage_SignalTerminated(t *testing.T) {
 
 	handler.Terminate()
 	finalStatus := s.executeStage(sig, s.deployment.Stages[0])
-	assert.Equal(t, model.StageStatus_STAGE_RUNNING, finalStatus)
+	assert.Equal(t, model.StageStatus_STAGE_FAILURE, finalStatus)
 }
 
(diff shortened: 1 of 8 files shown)


pkg/app/pipedv1/controller/scheduler_test.go
```diff
@@ -270,7 +270,7 @@ func TestExecuteStage_SignalTerminated(t *testing.T) {
 
 	handler.Terminate()
 	finalStatus := s.executeStage(sig, s.deployment.Stages[0])
-	assert.Equal(t, model.StageStatus_STAGE_RUNNING, finalStatus)
+	assert.Equal(t, model.StageStatus_STAGE_FAILURE, finalStatus)
```
khanhtc1202, line 273: This test assertion is changed due to the previous version of `s.reportStageStatus` use WaitNext, which return no error on context cancel, this should be STAGE_FAILURE as mentioned in this change.
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/pipedv1/controller/scheduler.go#L463

pkg/app/pipedv1/controller/scheduler_test.go
```diff
@@ -306,5 +306,5 @@ func TestExecuteStage_SignalCancelled(t *testing.T) {
 
 	handler.Cancel()
 	finalStatus := s.executeStage(sig, s.deployment.Stages[0])
-	assert.Equal(t, model.StageStatus_STAGE_CANCELLED, finalStatus)
+	assert.Equal(t, model.StageStatus_STAGE_FAILURE, finalStatus)
```
khanhtc1202, line 309: Same as https://github.com/pipe-cd/pipecd/pull/5388/files#r1867429259

ffjlabo:
📝 
backoff.WaitNext handles context canceling inside and returns bool.
https://github.com/pipe-cd/pipecd/blob/master/pkg/backoff/backoff.go#L76-L91

It happened the bug related to the `WaitNext` before.
https://github.com/pipe-cd/pipecd/pull/5388

ffjlabo (approved):
Thank you!

PR #5389, opened 2024-12-04
Implement preparing manifest applier in kubernetes plugin

Warashi:
**What this PR does**:

This PR implements the preparation of a manifest applier in the Kubernetes plugin.

**Why we need it**:

Without this PR, we cannot execute the K8S_SYNC stage.

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
@@ -15,7 +15,9 @@
 package deployment
 
 import (
+	"cmp"
 	"context"
+	"encoding/json"
 	"errors"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -33,10 +35,20 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	defaultKubectlVersion = "1.18.2"
```
Warashi, line 39: This is copied from current implementation.
https://github.com/pipe-cd/pipecd/blob/40dae444572e9073bdc68e20d13c24c1911b64a5/pkg/app/piped/toolregistry/install.go#L29

ffjlabo (approved):
LGTM

t-kikuc (approved):
lgtm

PR #5390, opened 2024-12-04
Implement rollback stage for kubernetes plugin

Warashi:
**What this PR does**:

- Implement executeK8sRollbackStage
- refactor duplicates of executeK8sSyncStage and executeK8sRollbackStage
    - There are some duplicates remaining, but these parts has unimplemented TODOs. so I didn't refactor all of them.

**Why we need it**:

We have to support rollback of k8s stages

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
@@ -206,28 +206,30 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
 	return manifests, nil
 }
 
-func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (*deployment.ExecuteStageResponse, error) {
-	switch request.GetInput().GetStage().GetName() {
-	case StageK8sSync.String():
-		return a.executeK8sSyncStage(ctx, request.GetInput())
-	case StageK8sRollback.String():
-		return a.executeK8sRollbackStage(ctx, request.GetInput())
-	default:
-		return nil, status.Error(codes.InvalidArgument, "unimplemented or unsupported stage")
-	}
-}
-
-func (a *DeploymentService) executeK8sSyncStage(ctx context.Context, input *deployment.ExecutePluginInput) (response *deployment.ExecuteStageResponse, err error) {
+func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (response *deployment.ExecuteStageResponse, _ error) {
 	// TODO: move this to the ExecuteStage function and pass the log persister as an argument?
-	lp := a.logPersister.StageLogPersister(input.GetDeployment().GetId(), input.GetStage().GetId())
+	lp := a.logPersister.StageLogPersister(request.GetInput().GetDeployment().GetId(), request.GetInput().GetStage().GetId())
 	defer func() {
 		// When the piped cancelled the RPC while the stage is still runnning, we should not mark the log persister as completed.
-		if !response.GetStatus().IsCompleted() && errors.Is(err, context.Canceled) {
+		// we use the context error to check if the RPC is cancelled by the piped.
+		// we don't use the response error to check context cancellation because the response error can be set without wrapping the context error.
+		if !response.GetStatus().IsCompleted() && errors.Is(ctx.Err(), context.Canceled) {
 			return
 		}
 		lp.Complete(time.Minute)
 	}()
 
+	switch request.GetInput().GetStage().GetName() {
+	case StageK8sSync.String():
+		return a.executeK8sSyncStage(ctx, lp, request.GetInput())
+	case StageK8sRollback.String():
+		return a.executeK8sRollbackStage(ctx, lp, request.GetInput())
+	default:
+		return nil, status.Error(codes.InvalidArgument, "unimplemented or unsupported stage")
+	}
+}
+
+func (a *DeploymentService) executeK8sSyncStage(ctx context.Context, lp logpersister.StageLogPersister, input *deployment.ExecutePluginInput) (response *deployment.ExecuteStageResponse, err error) {
 	lp.Infof("Start syncing the deployment")
 
 	cfg, err := config.DecodeYAML[*kubeconfig.KubernetesApplicationSpec](input.GetTargetDeploymentSource().GetApplicationConfig())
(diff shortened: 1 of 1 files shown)


pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -206,28 +206,29 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
...
-func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (*deployment.ExecuteStageResponse, error) {
+func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (response *deployment.ExecuteStageResponse, _ error) {
+	lp := a.logPersister.StageLogPersister(request.GetInput().GetDeployment().GetId(), request.GetInput().GetStage().GetId())
+	defer func() {
+		// When the piped cancelled the RPC while the stage is still runnning, we should not mark the log persister as completed.
+		// we use the context error to check if the RPC is cancelled by the piped.
+		// we don't use the response error to check context cancellation because the response error can be set without wrapping the context error.
+		if !response.GetStatus().IsCompleted() && errors.Is(ctx.Err(), context.Canceled) {
+			return
+		}
```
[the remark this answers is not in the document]
Warashi: I realized that we have to handle two cancellations, but we cannot determine the cause because the cancel is propagated beyond the gRPC layer.

1. cancellation occurred by graceful shutdown.
    - in this case, we should not mark the log persister as completed.
2. cancellation occurred by the user via the web UI.
    - in this case, we should return the status as canceled and mark the log persister as completed.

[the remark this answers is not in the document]
Warashi: I realized the plugin receives SIGTERM or SIGINT when the graceful shutdown is going, so I implemented the handling of them.
When the cancellation occurs without receiving such signals, the log persister should be completed.

khanhtc1202:
> I realized that the plugin should return the StageStatus and should not return an error even if the error has occurred. Because the plugin can return StageStatus_FAILURE if the error occurs. I'll change my codes.
> 
> 👉🏻 DONE

@Warashi I think we should also return the error, and in the piped side, we will show that error on piped log as well.

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -206,40 +206,45 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
 	return manifests, nil
 }
 
-func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (*deployment.ExecuteStageResponse, error) {
+func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (response *deployment.ExecuteStageResponse, _ error) {
```
khanhtc1202, line 209:
```suggestion
// ExecuteStage performs stage-defined tasks.
// It returns stage status after execution without error.
// Error only be raised if the given stage is not supported.
func (a *DeploymentService) ExecuteStage(ctx context.Context, request *deployment.ExecuteStageRequest) (response *deployment.ExecuteStageResponse, _ error) {
```

khanhtc1202 (commented):
LGTM with small nits

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
Warashi: Thanks. applied this suggestion.

khanhtc1202 (dismissed):
LGTM 👍

pkg/rpc/signal_interceptor.go
```diff
@@ -0,0 +1,31 @@
...
+	"google.golang.org/grpc"
+)
+
+// SignalHandlingInterceptor is a gRPC interceptor that cancels the context when a termination signal is received.
+func SignalHandlingInterceptor(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp any, err error) {
+	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
+	return handler(ctx, req)
+}
```
ffjlabo, line 31: @Warashi Sorry, I noticed that we can't distinguish Graceful Shutdown and context cancel for now. piped still treats both as context canceled.

We might use `signal.Notify` like this.
```suggestion
// SignalHandlingInterceptor is a gRPC interceptor that cancels the context when a termination signal is received.
var terminated atomic.Bool

// SignalHandlingInterceptor is a gRPC interceptor that cancels the context when a termination signal is received.
func SignalHandlingInterceptor(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp any, err error) {
	sigCh := make(chan os.Signal, 1)
	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)

	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	go func() {
		select {
		case <-sigCh:
			// received a termination signal
			terminated.Store(true)
			cancel()
		}
	}()

	return handler(ctx, req)
}
```

Warashi: @ffjlabo 
I implemented the determination of the Graceful Shutdown in this file. Is it not enough?

[`pkg/plugin/signalhandler/handler.go`](https://github.com/pipe-cd/pipecd/pull/5390/files#diff-97d64f7b9d2af9a5ee8ac1e6ded6c7313aba3c14ecae6cdd0596f044c3d962b6)

ffjlabo: @Warashi  I got it. Sorry, I missed it. 🙏

pkg/plugin/signalhandler/handler.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func init() {
+	ctx, cancel := signal.NotifyContext(context.Background(), signals...)
+	go func() {
+		defer cancel()
+		<-ctx.Done()
+		terminated.Store(true)
+	}()
+}
```
ffjlabo, line 44: [nit] We should add a comment to describe that this distinguishes context.Cancel and Graceful shutdown.

ffjlabo (commented):
Thank you for the fix! left a nit comment.

pkg/plugin/signalhandler/handler.go
Warashi: I added the comments. Please check whether the comments are enough.
https://github.com/pipe-cd/pipecd/pull/5390/commits/02a4df23ceaabded50ac619307e4e2f38948de75

ffjlabo (approved):
LGTM 👍

khanhtc1202 (approved):

PR #5398, opened 2024-12-09
Update lifecycle.DownloadBinary to work with local file

khanhtc1202:
**What this PR does**:

Update lifecycle.DownloadBinary logic to work with local files. After this pr, pipedv1 can fetch plugins from a remote source (via HTTP protocol) or local files from another file path.

**Why we need it**:

Part of #4980
Follow #5394

### DIFF
--- a/pkg/lifecycle/binary.go
+++ b/pkg/lifecycle/binary.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -104,7 +105,7 @@ func RunBinary(ctx context.Context, execPath string, args []string) (*Command, e
 
 // DownloadBinary downloads a file from the given URL into the specified path
 // this also marks it executable and returns its full path.
-func DownloadBinary(url, destDir, destFile string, logger *zap.Logger) (string, error) {
+func DownloadBinary(sourceURL, destDir, destFile string, logger *zap.Logger) (string, error) {
 	if err := os.MkdirAll(destDir, 0755); err != nil {
 		return "", fmt.Errorf("could not create directory %s (%w)", destDir, err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/lifecycle/binary.go
```diff
@@ -132,32 +133,52 @@ func DownloadBinary(url, destDir, destFile string, logger *zap.Logger) (string,
...
-	if err != nil {
-		return "", fmt.Errorf("HTTP GET %s failed (%w)", url, err)
-	}
-	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK {
-		return "", fmt.Errorf("HTTP GET %s failed with error %d", url, resp.StatusCode)
-	}
+	switch u.Scheme {
+	case "http", "https":
```
khanhtc1202, line 144: This makes the test easier since I don't want to mock the HTTP server. The validation logic in the piped configuration remains to only allow the "https" and "file" schemes.

ffjlabo: 📝 It is difficult to test with https server because `DownloadBinary` uses http.Client{} internally and can't inject for now. So we can't use httptest.NewTLSServer.

ffjlabo (approved):
👍

pkg/lifecycle/binary.go
t-kikuc, line 144: [ASK] Are we gonna support `http` too? or only `https`?

khanhtc1202: @t-kikuc For the question you asked 🙏

khanhtc1202: No we don't, just place http there to make testing process simpler 🙏

t-kikuc: @khanhtc1202 

I got it.
binary.go allows http and https, `PipedPlugin` disallows http.

https://github.com/pipe-cd/pipecd/blob/dcecc2d9b88201fa5228350411ae5895cbb44547/pkg/configv1/piped.go#L1328-L1330

t-kikuc (approved):

PR #5409, opened 2024-12-11
Implement functions to find referencing ConfigMaps and Secrets in Kubernetes manifests

Warashi:
**What this PR does**:

as title

**Why we need it**:

I want to implement `annotateConfigHash` in the k8s plugin.
To implement this, we need to find the configmap and secrets referenced in the deployments.
https://github.com/pipe-cd/pipecd/blob/e276b8982579f46b2e199906ddf42576266be03d/pkg/app/piped/executor/kubernetes/kubernetes.go#L582-L587

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/provider/deployment.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/provider/deployment.go
@@ -1,0 +1,129 @@
+// Copyright 2024 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package provider
+
+import (
+	"slices"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// FindReferencingConfigMaps finds all configmaps that are referenced by the given manifest.
+//
+// It looks for configmaps in the following fields:
+// - spec.template.spec.volumes.configMap.name
+// - spec.template.spec.initContainers.env.valueFrom.configMapKeyRef.name
+// - spec.template.spec.initContainers.envFrom.configMapRef.name
+// - spec.template.spec.containers.env.valueFrom.configMapKeyRef.name
+// - spec.template.spec.containers.envFrom.configMapRef.name
+func FindReferencingConfigMaps(m *unstructured.Unstructured) []string {
+	var configMaps []string
+
+	if n := nestedStringSlice(m.Object, "spec", "template", "spec", "volumes", "configMap", "name"); len(n) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/app/pipedv1/plugin/kubernetes/provider/deployment.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	slices.Sort(secrets)
+	return slices.Compact(secrets)
+}
+
+// nestedStringSlice extracts a string slice from the given object by following the fields.
+// It returns the extracted string slice.
+// If there is []map[string]any in the middle of the fields, it will be flattened.
+func nestedStringSlice(obj any, fields ...string) []string {
```
Warashi, line 96: I don't want to convert the unstructured.Unstructured to appsv1.Deployment or something, so I implemented this.
We cannot use `unstructured.NestedStringSlice` because it handles the `[]map[string]any` differently.

khanhtc1202 (approved):
LGTM 👍

pkg/app/pipedv1/plugin/kubernetes/provider/deployment.go
ffjlabo: @Warashi 
[Ask] 
> We cannot use unstructured.NestedStringSlice because it handles the []map[string]any differently.

What's the difference between `unstructured.NestedStringSlice` and `nestedStringSlice` ?

Warashi: @ffjlabo 
`unstructured.NestedStringSlice` returns an error when there is `[]map[string]any` in the middle of the fields. On the other hand, `nestedStringSlice` flattens it.

ffjlabo: @Warashi 
I got your point. I tried to use `unstructured.NestedStringSlice` with the data below.
We can't find and flatten slices in the `[]interface{}{map[string]interface{}{}}`.

```
	obj := map[string]interface{}{
		"spec": map[string]interface{}{
			"template": map[string]interface{}{
				"spec": map[string]interface{}{
					"volumes": []interface{}{
						map[string]interface{}{
							"configMap": map[string]interface{}{
								"name": "canary-by-config-change",
							},
						},
					},
				},
			},
		},
	}

	res, exists, err := unstructured.NestedStringSlice(obj, "spec", "template", "spec", "volumes[]", "configMap", "name")
	assert.True(t, exists)
	assert.NoError(t, err)
	assert.Equal(t, res, []string{"canary-by-config-change"})
```

```
=== RUN   Test
        	Error:      	Should be true
        	Error:      	Received unexpected error:
        	            	.spec.template.spec.volumes.configMap accessor error: [map[configMap:map[name:canary-by-config-change]]] is of the type []interface {}, expected map[string]interface{}
        	Error:      	Not equal:
        	            	expected: []string(nil)
        	            	actual  : []string{"canary-by-config-change"}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,4 @@
        	            	-([]string) <nil>
        	            	+([]string) (len=1) {
        	            	+ (string) (len=23) "canary-by-config-change"
        	            	+}

        	Test:       	Test
```

The test for `unstructured.NestedFieldNoCopy` (used in the `unstructured.NestedStringSlice`) shows it.
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/helpers_test.go#L122

ffjlabo (approved):
LGTM 🙆‍♂️

PR #5411, opened 2024-12-11
Fix kubectl download URL in install script

Warashi:
**What this PR does**:

This PR does the same as #5379 at the k8s plugin implementation.

**Why we need it**:

Not to use deprecated download URLs.


ffjlabo (approved):
Thanks 🙏

khanhtc1202 (approved):

PR #5412, opened 2024-12-11
Use git partial clone and worktree to reduce network/file io

Warashi:
**What this PR does**:

- Use not `cp -rf` but `git worktree add` to Copy the git.Repo
- Use git partial clone to reduce network io

**Why we need it**:

I want to reduce network/file io

Fixes #

### DIFF
--- a/pkg/app/piped/driftdetector/cloudrun/detector.go
+++ b/pkg/app/piped/driftdetector/cloudrun/detector.go
@@ -263,6 +263,8 @@ func (d *detector) loadHeadServiceManifest(app *model.Application, repo git.Repo
 			if err != nil {
 				return provider.ServiceManifest{}, fmt.Errorf("failed to copy the cloned git repository (%w)", err)
 			}
+			defer repo.Clean()
+
 			repoDir := repo.GetPath()
 			appDir = filepath.Join(repoDir, app.GitPath.Path)
 		}
--- a/pkg/app/piped/driftdetector/cloudrun/detector.go
+++ b/pkg/app/piped/driftdetector/cloudrun/detector.go
@@ -221,7 +221,7 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 	return d.reporter.ReportApplicationSyncState(ctx, app.Id, state)
 }
 
-func (d *detector) loadHeadServiceManifest(app *model.Application, repo git.Repo, headCommit git.Commit) (provider.ServiceManifest, error) {
+func (d *detector) loadHeadServiceManifest(app *model.Application, repo git.Worktree, headCommit git.Commit) (provider.ServiceManifest, error) {
 	var (
 		manifestCache = provider.ServiceManifestCache{
 			AppID:  app.Id,
--- a/pkg/app/pipedv1/deploysource/deploysource.go
+++ b/pkg/app/pipedv1/deploysource/deploysource.go
@@ -191,6 +191,13 @@ func (p *provider) copy(lw io.Writer) (*DeploySource, error) {
 		return nil, err
 	}
 
+	// To avoid the git operation on copied source.
+	// TODO: do not copy the .git directory to improve the performance.
+	if err := os.RemoveAll(filepath.Join(dest, ".git")); err != nil {
+		fmt.Fprintf(lw, "Unable to remove the .git directory from the copied deploy source (%v)\n", err)
+		return nil, err
+	}
+
 	return &DeploySource{
 		RepoDir:                  dest,
 		AppDir:                   filepath.Join(dest, p.appGitPath.Path),
--- a/pkg/app/pipedv1/deploysource/deploysource.go
+++ b/pkg/app/pipedv1/deploysource/deploysource.go
@@ -183,18 +183,20 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
 func (p *provider) copy(lw io.Writer) (*DeploySource, error) {
 	p.copyNum++
 
+	src := p.source.RepoDir
 	dest := fmt.Sprintf("%s-%d", p.source.RepoDir, p.copyNum)
-	cmd := exec.Command("cp", "-rf", p.source.RepoDir, dest)
+
+	// use tar to exclude the .git directory
+	// the tar command does not create the destination directory if it does not exist.
+	// so we need to create it before running the command.
+	if err := os.MkdirAll(dest, 0700); err != nil {
+		fmt.Fprintf(lw, "Unable to create the directory to store the copied deploy source (%v)\n", err)
+		return nil, err
+	}
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("tar c -f - -C '%s' --exclude='.git' . | tar x -f - -C '%s'", src, dest))
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		fmt.Fprintf(lw, "Unable to copy deploy source data (%v, %s)\n", err, string(out))
-		return nil, err
-	}
-
-	// To avoid the git operation on copied source.
-	// TODO: do not copy the .git directory to improve the performance.
-	if err := os.RemoveAll(filepath.Join(dest, ".git")); err != nil {
-		fmt.Fprintf(lw, "Unable to remove the .git directory from the copied deploy source (%v)\n", err)
 		return nil, err
 	}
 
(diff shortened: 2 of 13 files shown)


pkg/git/repo.go
```diff
@@ -52,6 +53,17 @@ type Repo interface {
 	CommitChanges(ctx context.Context, branch, message string, newBranch bool, changes map[string][]byte, trailers map[string]string) error
 }
 
+// Worktree provides functions to get and handle git worktree.
+// It is a separate checkout of the repository.
+// It is used to make changes to the repository without affecting the main repository.
+// Worktree always does checkout with the detached HEAD, so it doesn't affect the main repository.
+type Worktree interface {
```
Warashi, line 60: I added this to restrict operation at worktree.

pkg/app/pipedv1/deploysource/deploysource.go
```diff
@@ -191,6 +191,13 @@ func (p *provider) copy(lw io.Writer) (*DeploySource, error) {
 		return nil, err
 	}
 
+	// To avoid the git operation on copied source.
+	// TODO: do not copy the .git directory to improve the performance.
+	if err := os.RemoveAll(filepath.Join(dest, ".git")); err != nil {
+		fmt.Fprintf(lw, "Unable to remove the .git directory from the copied deploy source (%v)\n", err)
+		return nil, err
+	}
```
khanhtc1202, line 199: How about using a `tar` trick to avoid `.git` copy and remove. I think tar is available everywhere and fast enough compared to cp
```
tar -cpP --exclude=.git -C "$source_dir" . | tar -xpP -C "$target_dir"
```

khanhtc1202:
Overall LGTM. Just left a nits, want to give it a shot 👍

pkg/app/pipedv1/deploysource/deploysource.go
Warashi: Thanks. I'll try.

Warashi: @khanhtc1202 

We use the alpine as the base image for piped, so I checked the tar command in the piped image.
First, the tar command exists. But, the options differ from the GNU/BSD tar.
For example, short flags `-p` and `-P` do not exist.
IMO, it's not good to depend on this tar behavior because users may run the piped on a non-alpine-based OS.
WDYT?

Note: I checked `tar --help` in the piped container image. I paste the result below.
```
$ docker run --rm -it --entrypoint sh  ghcr.io/pipe-cd/piped:v0.50.0-23-g8f23da4@sha256:88ec60ebf2b7e3131ff38b1af4a043117ef688bc636ddd95afd6bd368293651b
/ $ tar --help
BusyBox v1.36.1 (2024-06-10 07:11:47 UTC) multi-call binary.

Usage: tar c|x|t [-ZzJjahmvokO] [-f TARFILE] [-C DIR] [-T FILE] [-X FILE] [LONGOPT]... [FILE]...

Create, extract, or list files from a tar file

        c       Create
        x       Extract
        t       List
        -f FILE Name of TARFILE ('-' for stdin/out)
        -C DIR  Change to DIR before operation
        -v      Verbose
        -O      Extract to stdout
        -m      Don't restore mtime
        -o      Don't restore user:group
        -k      Don't replace existing files
        -Z      (De)compress using compress
        -z      (De)compress using gzip
        -J      (De)compress using xz
        -j      (De)compress using bzip2
        --lzma  (De)compress using lzma
        -a      (De)compress based on extension
        -h      Follow symlinks
        -T FILE File with names to include
        -X FILE File with glob patterns to exclude
        --exclude PATTERN       Glob pattern to exclude
        --overwrite             Replace existing files
        --strip-components NUM  NUM of leading components to strip
        --no-recursion          Don't descend in directories
        --numeric-owner         Use numeric user:group
        --no-same-permissions   Don't restore access permissions
```

khanhtc1202: This is the version without -P, the more universal solution
```
tar --exclude='.git' -cf - source_dir | tar -xf - -C destination_dir
```
@Warashi

Warashi: OKay, thank you for providing this. I'll use this script.

Warashi: I tried some tar options and finally reached this commit.
I tested it with busybox's tar (the same as alpine's) and GNU tar.
https://github.com/pipe-cd/pipecd/pull/5412/commits/c5216c91bffc906f57b30133f3d0e557f1b1888f

khanhtc1202 (approved):
LGreatTM 💯 
This could be the patch we're looking for 🥇

pkg/app/piped/driftdetector/cloudrun/detector.go
```diff
@@ -263,6 +263,8 @@ func (d *detector) loadHeadServiceManifest(app *model.Application, repo git.Repo
 			if err != nil {
 				return provider.ServiceManifest{}, fmt.Errorf("failed to copy the cloned git repository (%w)", err)
 			}
+			defer repo.Clean()
```
ffjlabo, line 266: @Warashi 
it seems that the `dir` is removed after executing the `loadHeadServiceManifest`
So we might not call `repo.Clean()`. WDYT?

```
		// We have to copy repository into another directory because
		// decrypting the sealed secrets or attaching files might change the git repository.
		if attachmentUsed || encryptionUsed {
			dir, err := os.MkdirTemp("", "detector-git-processing")
			if err != nil {
				return nil, fmt.Errorf("failed to prepare a temporary directory for git repository (%w)", err)
			}
			defer os.RemoveAll(dir)

			repo, err = repo.Copy(filepath.Join(dir, "repo"))
			if err != nil {
				return nil, fmt.Errorf("failed to copy the cloned git repository (%w)", err)
			}
			defer repo.Clean()

			repoDir = repo.GetPath()
			appDir = filepath.Join(repoDir, app.GitPath.Path)
		}
``` 
https://github.com/pipe-cd/pipecd/blob/c5216c91bffc906f57b30133f3d0e557f1b1888f/pkg/app/piped/driftdetector/kubernetes/detector.go#L270C1-L281C22

pkg/git/repo.go
```diff
@@ -35,7 +35,8 @@ var (
 type Repo interface {
 	GetPath() string
 	GetClonedBranch() string
-	Copy(dest string) (Repo, error)
+	Copy(dest string) (Worktree, error)
+	CopyToModify(dest string) (Repo, error)
```
ffjlabo, line 39: [IMO] How about using the `Worktree` and `Copy`?

```
	Worktree(dest string) (Worktree, error)
	Copy(dest string) (Repo, error)
```

pkg/app/piped/driftdetector/cloudrun/detector.go
Warashi: @ffjlabo 

I think we should call `repo.Clean()` here because the 'repo.Clean()' does not only the directory deletion.
IMO, we should call some methods like `Clean()` or `Close()` everywhere because the caller doesn't know what happens with such methods. The caller only knows the cleanup methods exist.

pkg/git/repo.go
Warashi: It sounds clear, but I want to leave here as-is.
I want to clear when to use the `CopyToModify` method with its naming.
Almost all of implementation, we don't need modification of git commits/histories, so we should not use CopyToModify, and we can know from its naming.
If we rename them as Worktree and Copy, some implementation may call Copy method mistakenly even if it's enough to call the Worktree method.

WDYT?

ffjlabo: > Almost all of implementation, we don't need modification of git commits/histories, so we should not use CopyToModify, and we can know from its naming.

I got your point! I agree with it.

pkg/app/piped/driftdetector/cloudrun/detector.go
ffjlabo: @Warashi 
> I think we should call repo.Clean() here because the 'repo.Clean()' not only does the directory deletion.

I got it. The `git worktree` needs additional operation other than deleting the directory to keep the consistency of the worktree as you said. 

> IMO, we should call some methods like Clean() or Close() everywhere because the caller doesn't know what happens with such methods. The caller only knows the cleanup methods exist.

Also I agree with it.

ffjlabo: A little off topic, but the detector reuses the one repo. `loadHeadServiceManifest` also uses it as an arg. So I feel a little bit confused about cleaning at this point.

But I noticed that this section is used temporarily and different from the actual repo passed as arg. 🙏

ffjlabo (approved):
LGTM

PR #5414, opened 2024-12-12
Remove decryption grpc from pipedv1 plugin api

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

We agreed that Piped should be in charge of preparing the deploy source, and the plugin can just pull and use it. The plugins should not need to care whether the manifest is encrypted or not.

Fixes #


Warashi (approved):
LGTM

PR #5415, opened 2024-12-12
Move toolregistry package to pkg/plugin

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

All the plugin shared packages should be placed under `pkg/plugin` and we can public it as the based for creating new plugin later.

Part of #4980


Warashi (approved):
👍🏻

ffjlabo (approved):

Issue #4896, opened 2024-04-24
Quickstart: PipeCD control plane does not work on Arm architecture cluster


ponkio-o:
**What happened**:
Quickstart doesn't work on Arm architecture kubernetes cluster.

related: https://github.com/pipe-cd/pipecd/issues/4116

**What you expected to happen**:
Working pipecd correctly on kubernetes cluster using arm arch.

**How to reproduce it**:
Run `pipectl quickstart` against a Kubernetes cluster configured with only Arm architecture worker nodes.

My k8s cluster configured with only arm64 nodes.
```console
$ k version
Client Version: v1.30.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.2

$ k get node -o go-template='{{range .items}}{{.metadata.name}}  {{index .metadata.labels "kubernetes.io/arch"}}{{"\n"}}{{end}}'
instance-20211220-2249  arm64
instance-20211220-2250  arm64
instance-20211222-2325  arm64
instance-20211222-2332  arm64
```

Run `pipectl quickstart`
```console
$ pipectl quickstart --version v0.47.0
Installing the controlplane in quickstart mode...
Release "pipecd" does not exist. Installing it now.
NAME: pipecd
LAST DEPLOYED: Wed Apr 24 11:29:25 2024
NAMESPACE: pipecd
STATUS: deployed
REVISION: 1
TEST SUITE: None

Intalled the controlplane successfully!

Waiting for PipeCD control plane to be ready...
PipeCD control plane status: ContainerCreating 3 Running 1 Init:0/1 2
PipeCD control plane status: ContainerCreating 1 CrashLoopBackOff 2 Running 1 Init:0/1 2
PipeCD control plane status: ContainerCreating 1 CrashLoopBackOff 2 Running 1 Init:0/1 2
PipeCD control plane status: CrashLoopBackOff 2 Running 2 Init:0/1 2
PipeCD control plane status: Running 3 Init:0/1 2 Error 1
PipeCD control plane status: Running 3 Init:0/1 2 Error 1
PipeCD control plane status: Running 2 Init:0/1 2 Error 2
PipeCD control plane status: CrashLoopBackOff 1 Running 2 Init:0/1 2 Error 1
PipeCD control plane status: CrashLoopBackOff 2 Running 2 Init:0/1 2
PipeCD control plane status: CrashLoopBackOff 2 Running 2 Init:0/1 2
[...]

PR #5420, opened 2024-12-14
feat: added arm64 build

ponkio-o:
**What this PR does**: Added linux/arm64 build to all docker images.

**Why we need it**: To support linux/arm64

Fixes https://github.com/pipe-cd/pipecd/issues/4896

### DIFF
--- a/.github/workflows/publish_image_chart.yaml
+++ b/.github/workflows/publish_image_chart.yaml
@@ -62,7 +62,7 @@ jobs:
           fetch-depth: 0
       - name: Determine version
 
-      # Setup QEMU and Buildx.      
+      # Setup QEMU and Buildx.
       - name: Set up QEMU
         uses: docker/setup-qemu-action@8b122486cedac8393e77aa9734c3528886e4a1a8 #v2.0.0
       - name: Set up Docker Buildx
@@ -85,14 +85,14 @@ jobs:
 
       # Building and pushing container images.
       - name: Build and push ${{ matrix.image }} image
-        uses: docker/build-push-action@0565240e2d4ab88bba5387d719585280857ece09 #v5.0.0
+        uses: docker/build-push-action@48aba3b46d1b1fec4febb7c5d0c644b249a11355 # v6.10.0
         with:
           push: true
           context: .
           file: ${{ matrix.dockerfile }}
           platforms: linux/amd64,linux/arm64
           tags: ${{ matrix.container_registry }}/${{ matrix.image }}:${{ env.PIPECD_VERSION }}
-      
+
       # Building and pushing Helm charts.
       - name: Install helm
         uses: Azure/setup-helm@v1
@@ -145,9 +145,9 @@ jobs:
       - name: Publish quickstart manifests
         uses: peter-evans/create-pull-request@v6
         with:
-          title: "[bot] Publish quickstart manifests"
-          commit-message: "[bot] Publish quickstart manifests"
-          branch: "create-pull-request/publish-quickstart-manifests"
+          title: '[bot] Publish quickstart manifests'
+          commit-message: '[bot] Publish quickstart manifests'
+          branch: 'create-pull-request/publish-quickstart-manifests'
           body: |
             Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action.
             The workflow is defined [here](https://github.com/pipe-cd/pipecd/blob/master/.github/workflows/publish_image_chart.yaml).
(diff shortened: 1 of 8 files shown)


cmd/helloworld/Dockerfile
ponkio-o, line 1: Currently, binaries built for linux/amd64 on CI are added to the image, but I have changed to build them inside the Dockerfile. This will automatically configure GOARCH and other settings.

As noted in the [Issue](https://github.com/pipe-cd/pipecd/issues/4896), the linux/arm64 docker image exists, but the binary that exits in the image is linux/amd64, which is the cause of the current error.

Warashi (dismissed):
LGTM 👍🏻

cmd/pipecd/Dockerfile
```diff
@@ -1,3 +1,26 @@
...
+
+WORKDIR /app
+
+COPY go.* ./
+RUN go mod download
+
+COPY . ./
+
+RUN make build/go MOD=pipecd
+
```
Warashi, line 23: Sorry, we use `git` in our Makefile in many places, so please install git in web and go builder?
https://github.com/pipe-cd/pipecd/blob/d5ba3d2949bd1dc794020f36fafafb61ca749e44/Makefile#L21

Warashi:
@ponkio-o 
Sorry, I checked the behavior after I approved this, and I realized that the git command is absent in the builder container.

cmd/pipecd/Dockerfile
ponkio-o: @Warashi Thank you for your review. I installed `git` command to builder container on below commit.
https://github.com/pipe-cd/pipecd/pull/5420/commits/beb90e71020a14194b606a6f4d66d913c61a09b8

 And `golang:1.23.3` has been installed `git` already.
```console
$ docker run --rm -it golang:1.23.3 git --help
usage: git [-v | --version] [-h | --help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]

These are common Git commands used in various situations:

start a working area (see also: git help tutorial)
   clone     Clone a repository into a new directory
   init      Create an empty Git repository or reinitialize an existing one

work on the current change (see also: git help everyday)
   add       Add file contents to the index
   mv        Move or rename a file, a directory, or a symlink
   restore   Restore working tree files
   rm        Remove files from the working tree and from the index

examine the history and state (see also: git help revisions)
   bisect    Use binary search to find the commit that introduced a bug
   diff      Show changes between commits, commit and working tree, etc
   grep      Print lines matching a pattern
   log       Show commit logs
   show      Show various types of objects
   status    Show the working tree status

grow, mark and tweak your common history
   branch    List, create, or delete branches
   commit    Record changes to the repository
   merge     Join two or more development histories together
   rebase    Reapply commits on top of another base tip
   reset     Reset current HEAD to the specified state
   switch    Switch branches
   tag       Create, list, delete or verify a tag object signed with GPG

collaborate (see also: git help workflows)
   fetch     Download objects and refs from another repository
   pull      Fetch from and integrate with another repository or a local branch
   push      Update remote refs along with associated objects

'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
See 'git help git' for an overview of the system.
```

Warashi (approved):
Thank you! LGTM 👍🏻

.github/workflows/publish_image_chart.yaml
```diff
@@ -59,40 +57,12 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
-      - uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
-          cache: true
       - name: Determine version
```
t-kikuc, line 63: [ASK] Is this line still needed?

t-kikuc: @ponkio-o 
Sorry for the wrong selection.
I'm asking about only this line.
```yaml
  - name: Determine version
```

ponkio-o: Sorry, the below line is needed. I was revert.
```yaml
run: echo "PIPECD_VERSION=$(git describe --tags --always --abbrev=7)" >> $GITHUB_ENV
```

https://github.com/pipe-cd/pipecd/pull/5420/commits/1fb552384dda94d7c459dbdddfaa74ab6a4e88ef

Warashi (approved):
Thank you!

PR #5421, opened 2024-12-16
Add calling annotateConfigHash and refactor unmarshaling deploy target

Warashi:
**What this PR does**:

1. add calling annotateConfigHash
2. refactor the unmarshaling of the deploy target config

**Why we need it**:

1. we want to refresh workloads when the config changes

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/config/application_test.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/config/application_test.go
@@ -1,0 +1,99 @@
+// Copyright 2024 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/kubernetes/config/application_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := FindDeployTarget(tt.cfg, tt.targetName)
+			if tt.expectedErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
```
ffjlabo, line 95: [nit] 
```suggestion
			assert.Equal(t, tt.expectedErr, err != nil)
```

ffjlabo (commented):
Almost all LGTM! I left just one nit comment.

pkg/app/pipedv1/plugin/kubernetes/config/application_test.go
Warashi: I think my code provides more information when debugging the codes.
Your suggested code works fine when the test passes, but there is no information about the error when the test fails.
WDYT?

ffjlabo: OK, there was a reason. got it. I also agree with that reason.

ffjlabo (approved):
🚀

khanhtc1202 (approved):
Here you go 👍

PR #5425, opened 2024-12-16
[ECS] Enabled to configure/update `PropagateTags` and `EnableECSManagedTags`

t-kikuc:
**What this PR does**:

1. Enabled to configure `PropagateTags` of Service on creating, not the default `SERVICE`.
2. Enabled to update `PropagateTags` and `EnableECSManagedTags` of Service

**Why we need it**:

- for 1: We don't need to force `SERVICE` for `PropagateTags`.
    - We don't have to propagate tags from a Service to Tasks.
    - That's because **PipeCD managed tags of Tasks are NOT used**.
        - Those of Services and TaskSets are used.  They are explicitly attached in the code.
    - cf. It's introduced in https://github.com/pipe-cd/pipecd/pull/4140.

- for 2: There's no reason to prevent updating `PropagateTags` and `EnableECSManagedTags`.
    - `EnableECSManagedTags` has NO effect on PipeCD managed tags.


ffjlabo (approved):
👍

khanhtc1202 (approved):
Thank you 👍

PR #5431, opened 2024-12-17
Add test for executing K8S_SYNC stage at the kubernetes plugin

Warashi:
**What this PR does**:

Implement the test in the same way as #5399 at the Kubernetes plugin, then fix the failure about the resource key.

**Why we need it**:

To ensure compatibility between the current Kubernetes platform provider and the Kubernetes plugin.

Part of #4980


pkg/app/pipedv1/plugin/kubernetes/deployment/server_test.go
```diff
@@ -0,0 +1,216 @@
...
+	require.NoError(t, err)
+
+	assert.Equal(t, "simple", deployment.GetName())
+	assert.Equal(t, "simple", deployment.GetLabels()["app"])
+	assert.Equal(t, "piped", deployment.GetAnnotations()["pipecd.dev/managed-by"])
+	assert.Equal(t, "piped-id", deployment.GetAnnotations()["pipecd.dev/piped"])
+	assert.Equal(t, "app-id", deployment.GetAnnotations()["pipecd.dev/application"])
+	assert.Equal(t, "apps/v1", deployment.GetAnnotations()["pipecd.dev/original-api-version"])
+	assert.Equal(t, "apps/v1:Deployment:default:simple", deployment.GetAnnotations()["pipecd.dev/resource-key"])
+	assert.Equal(t, "0123456789", deployment.GetAnnotations()["pipecd.dev/commit-hash"])
```
Warashi, line 214: these assertions are copied from #5399

pkg/app/pipedv1/plugin/kubernetes/provider/resource.go
```diff
@@ -87,6 +87,9 @@ func MakeResourceKey(obj *unstructured.Unstructured) ResourceKey {
 		Namespace:  obj.GetNamespace(),
 		Name:       obj.GetName(),
 	}
+	if k.Namespace == "" {
+		k.Namespace = DefaultNamespace
+	}
```
Warashi, line 92: There was an incompatibility between the current Kubernetes platform provider and the Kubernetes plugin, so I fixed this.

pkg/app/pipedv1/plugin/kubernetes/deployment/server_test.go
```diff
@@ -0,0 +1,216 @@
...
+	config "github.com/pipe-cd/pipecd/pkg/configv1"
+	"github.com/pipe-cd/pipecd/pkg/model"
+	"github.com/pipe-cd/pipecd/pkg/plugin/api/v1alpha1/deployment"
+	"github.com/pipe-cd/pipecd/pkg/plugin/logpersister"
+	"github.com/pipe-cd/pipecd/pkg/plugin/toolregistry/toolregistrytest"
+)
+
+// newTestLogPersister creates a new testLogPersister for testing.
+//
+// TODO: move to a common package
```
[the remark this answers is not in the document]
Warashi: Thank you. I moved these to `pkg/plugin/logpersister/logpersistertest`.
https://github.com/pipe-cd/pipecd/pull/5431/commits/3c0a57d2d11373bf2539db0028b435e35b88aa77

pkg/app/pipedv1/plugin/kubernetes/deployment/server_test.go
```diff
@@ -0,0 +1,216 @@
...
+	d, _ := os.Getwd()
+	for {
+		if _, err := os.Stat(filepath.Join(d, "examples")); err == nil {
+			return filepath.Join(d, "examples")
+		}
+		d = filepath.Dir(d)
+	}
+}
+
+// TODO: move to a common package
```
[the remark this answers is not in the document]
Warashi: Thank you. I removed the TODO comment.

pkg/app/pipedv1/plugin/kubernetes/provider/resource.go
Warashi: Thank you for commenting. I reverted this change and fixed the assertion.
We can write the e2e test now, so I want to write tests to prevent bugs.

https://github.com/pipe-cd/pipecd/pull/5431/commits/057b9f55222d546bf363e7f252ccfb27f64f59f1
https://github.com/pipe-cd/pipecd/pull/5431/commits/e030be1f13b4a06ede8407fb2fe45bd146d710d9

pkg/app/pipedv1/plugin/kubernetes/deployment/server.go
```diff
@@ -189,7 +189,7 @@ func (a *DeploymentService) loadManifests(ctx context.Context, deploy *model.Dep
 	manifests, err := a.loader.LoadManifests(ctx, provider.LoaderInput{
 		PipedID:          deploy.GetPipedId(),
 		AppID:            deploy.GetApplicationId(),
-		CommitHash:       deploy.GetTrigger().GetCommit().GetHash(),
+		CommitHash:       deploymentSource.GetRevision(),
```
khanhtc1202, line 192: This could be the same but I like the previous version more. How about to rename `GetRevision` to `GetCommitHash` 🤔

Warashi: First, I also like `commit hash` more than `revision`.

However, this is not the same because the `deploymentSource` can be the running deploy source, not the target deploy source.

The naming `GetRevision` comes from proto definition, so how about renaming this field as `commit_hash`?
https://github.com/pipe-cd/pipecd/blob/17a5007b6f1ab40dd230376a511b2f1e1d109b40/pkg/plugin/api/v1alpha1/deployment/api.proto#L142-L143

khanhtc1202: I got your point. Also, it's a commit hash, so it should be named directly as a commit hash to reduce confusion. Let's address it by another PR (change proto and update this)

khanhtc1202 (approved):
LGTM 👍

ffjlabo (approved):
LGTM 🚀

PR #5437, opened 2024-12-20
Regenerate mock file using uber-go mock

### LINKED ISSUE #5429: Migrate to use uber-go mock instead

We're using golang/go-mock now, but the repo is archived, we should mirage to use uber-go/mock instead

**Why is this needed**:

khanhtc1202:
**What this PR does**:

Update mock files by regenerating using the updated codegen image.
Also, fix the tests to use uber-go mock instead of golang mock.

**Why we need it**:

Fixes #5429
Follow PR #5426


Warashi (approved):
👍🏻

ffjlabo (approved):
LGTM

PR #5440, opened 2024-12-23
Send notification about stage executions

### LINKED ISSUE #5393: Adding WAIT_APPROVAL_TIMEOUT webhook event
To notify WAIT_APPROVAL is failed, I want to get WAIT_APPROVAL_TIMEOUT webhook event, like #5340.

**Why is this needed**:
I use PipeCD to deploy plenty of microservices, some changes, like changing proto file, may invoke tens of deployments. In that case, I often forget to click approve with WAIT_APPROVAL phase, and also cannot know which deployment reaches WAIT_APPROVAL timeout.

Warashi:
**What this PR does**:

- Add notification event group about stage executions.
- Send notification before/after stage executions.

image
I attached the screen capture of the Slack notification above, but this PR adds webhook notifications, too.

**Why we need it**:

To know the deployment progress systematically.

Maybe fixes #5393 

- **How are users affected by this change**: Users receives new notifications.


pkg/app/piped/controller/scheduler.go
```diff
@@ -345,10 +345,53 @@ func (s *scheduler) Run(ctx context.Context) error {
...
+			case model.StageStatus_STAGE_SKIPPED:
+				s.notifier.Notify(model.NotificationEvent{
+					Type: model.NotificationEventType_EVENT_STAGE_SKIPPED,
+					Metadata: &model.NotificationEventStageSkipped{
+						Deployment: s.deployment,
+						Stage:      ps,
+					},
+				})
+			}
+
```
t-kikuc, line 394: [IMO] How about extracting into a method? 
I just want to simplify `scheduler.Run()` for readability and maintenanceability.

For example: 

```suggestion
			s.notifyStageEvent(model.StageStatus_STAGE_STARTED, ps)

			result = s.executeStage(sig, *ps, func(in executor.Input) (executor.Executor, bool) {
				return s.executorRegistry.Executor(model.Stage(ps.Name), in)
			})

                         s.notifyStageEvent(result, ps)
```

t-kikuc (commented):
It seems great!

I left a refactoring comment.

ffjlabo (approved):
LGTM

pkg/app/piped/controller/scheduler.go
t-kikuc: Thank you!
StageStart and StageEnd are really good.

t-kikuc (approved):

PR #5442, opened 2024-12-23
Copy .git in deploysource provider

Warashi:
**What this PR does**:

This PR partially reverts https://github.com/pipe-cd/pipecd/pull/5412.

**Why we need it**:

In #5412, I implemented that the piped does not copy `.git` in the deploysource provider's `copy` method.
This made a breaking change on the SCRIPT_RUN stage.
Before #5412, we can use the `git` command to retrieve manifest repository information.
After #5412, we cannot do that because there is not `.git`.
In #5412, I implemented the git client to use `git worktree` to clone from the locally cached git repository, so it's a small cost to copy the `.git` because it's a small text file.

Fixes #

### DIFF
--- a/pkg/app/piped/deploysource/deploysource.go
+++ b/pkg/app/piped/deploysource/deploysource.go
@@ -203,17 +203,8 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
 func (p *provider) copy(lw io.Writer) (*DeploySource, error) {
 	p.copyNum++
 
-	src := p.source.RepoDir
 	dest := fmt.Sprintf("%s-%d", p.source.RepoDir, p.copyNum)
-
-	// use tar to exclude the .git directory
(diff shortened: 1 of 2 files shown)


Warashi:
With this PR, we can use git command in the SCRIPT_RUN stage.
image

ffjlabo:
@Warashi 
[Ask] Is it possible to get the right output for `git log` inside the dir in which `.git` file is placed ?
It means that can we do `git log` inside `project_root_dir/hoge`?
```
.git
project_root_dir/hoge
```

Warashi:
@ffjlabo 
Yes. Do you consider we can use `git` without changing the directory like `cd ..` in the SCRIPT_RUN stage? We can do this. Here is the screenshot.
image

ffjlabo:
@Warashi 
Thanks, that's what I want to confirm.

ffjlabo (approved):
👍

t-kikuc (approved):

PR #5447, opened 2024-12-24
Add rpc interface for livestate in plugin architecture

ffjlabo:
**What this PR does**:

Added rpc definition for livestate features in the plugin architecture.

The proto files are below.
- `pkg/plugin/api/v1alpha1/livestate/api.proto`: define it as another file with deployment/api.proto to separate by features. 
- `pkg/model/application_live_state.proto`: the actual application livestate.

The rpc `GetLivestate` returns the actual state and the sync state at one time.

I'll update the RFC later.

**Why we need it**:

We will support livestate feature in the plugin architecture.

Part of #5363


pkg/model/application_live_state.proto
```diff
@@ -71,6 +72,49 @@ message LambdaApplicationLiveState {
...
+    // The health status of this resource.
+    HealthStatus health_status = 6 [(validate.rules).enum.defined_only = true];
+    // The description of the health status.
+    string health_description = 7;
+
+    // The timestamp when this resource was created.
+    int64 created_at = 10 [(validate.rules).int64.gt = 0];
+    // The timestamp of the last time when this resource was updated.
+    int64 updated_at = 11 [(validate.rules).int64.gt = 0];
+}
```
ffjlabo, line 116: The diff compared to the existing livestate resource state for k8s, ecs, lambda, cloud run.

Deleted
- `owner_ids`: no longer used by any kinds of application.
- `api_version`, `kind`, `namespace`: these naming are k8s specific and we need to adopt the platform-neutral field for plugin architecture.

**Added**
- `resource_type`: We decide to use it as `kind` to show resource type.
- `resource_metadata`: We decided to add metadata as string map to add platform specific values freely (such as `api_version`, `namespace`)

t-kikuc (approved):
GREAT

Warashi (approved):
👍🏻

PR #5450, opened 2024-12-25
Rename deploysource's revision to commit_hash

Warashi:
**What this PR does**:

as title

**Why we need it**:

Because it's a git commit hash, the name revision is not wrong, but it's clearer to name it commit_hash.

Part of #4980 


khanhtc1202 (approved):
👍

ffjlabo (approved):
👍

PR #5452, opened 2024-12-25
Add livestatereporter for the plugin architecture

ffjlabo:
**What this PR does**:

Implement livestatereporter for the plugin architecture.

It reports the application's live state and the sync state for every registered apps by the interval of 1 minute.

**Why we need it**:

We need to support livestate features on the pipedv1

Part of https://github.com/pipe-cd/pipecd/issues/5363

### DIFF
--- a/pkg/app/pipedv1/livestatereporter/livestatereporter.go
+++ b/pkg/app/pipedv1/livestatereporter/livestatereporter.go
@@ -77,9 +77,9 @@ func (r *reporter) Run(ctx context.Context) error {
 	for _, reporter := range r.reporters {
 		// Avoid starting all reporters at the same time to reduce the API call burst.
 		time.Sleep(10 * time.Second)
-		r.logger.Info("starting app live state reporter for plugin")
 
 		group.Go(func() error {
+			reporter.logger.Info("starting app live state reporter for plugin")
 			return reporter.Run(ctx)
 		})
 	}
--- a/pkg/app/pipedv1/livestatereporter/livestatereporter.go
+++ b/pkg/app/pipedv1/livestatereporter/livestatereporter.go
@@ -74,6 +74,8 @@ func NewReporter(appLister applicationLister, apiClient apiClient, plugins map[s
 func (r *reporter) Run(ctx context.Context) error {
 	group, ctx := errgroup.WithContext(ctx)
 
+	r.logger.Info(fmt.Sprintf("starting %d live state reporters of plugins", len(r.reporters)))
+
 	for _, reporter := range r.reporters {
 		// Avoid starting all reporters at the same time to reduce the API call burst.
 		time.Sleep(10 * time.Second)
--- a/pkg/app/pipedv1/livestatereporter/livestatereporter.go
+++ b/pkg/app/pipedv1/livestatereporter/livestatereporter.go
@@ -145,7 +145,6 @@ func (pr *pluginReporter) flushSnapshots(ctx context.Context) {
 		// TODO: Implement DetermineAppHealthStatus for the case of plugin architecture.
 		snapshot.DetermineAppHealthStatus()
 
-		// TODO: Fix ReportApplicationLiveState to store ApplicationLiveState.
 		if _, err := pr.apiClient.ReportApplicationLiveState(ctx, &pipedservice.ReportApplicationLiveStateRequest{
 			Snapshot: snapshot,
 		}); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/app/pipedv1/livestatereporter/livestatereporter.go
```diff
@@ -0,0 +1,171 @@
...
+			appLister:             appLister,
+			apiClient:             apiClient,
+			pluginClient:          p,
+			logger:                rlogger.With(zap.String("plugin-name", name)),
+		})
+	}
+	return r
+}
+
+func (r *reporter) Run(ctx context.Context) error {
```
ffjlabo, line 74: Implemented it based on this code.
https://github.com/pipe-cd/pipecd/blob/336d96d501311103dd9921bcea55722e2c8e774a/pkg/app/piped/livestatereporter/reporter.go#L104C1-L127C2

pkg/app/pipedv1/livestatereporter/livestatereporter.go
```diff
@@ -0,0 +1,171 @@
...
+type pluginReporter struct {
+	pluginName            string
+	snapshotFlushInterval time.Duration
+	appLister             applicationLister
+	apiClient             apiClient
+	pluginClient          pluginapi.PluginClient
+	logger                *zap.Logger
+}
+
+func (pr *pluginReporter) Run(ctx context.Context) error {
```
ffjlabo, line 107: Implemented it based on this code.
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/livestatereporter/kubernetes/reporter.go#L79C1-L110C2

pkg/app/pipedv1/livestatereporter/livestatereporter.go
```diff
@@ -0,0 +1,171 @@
...
+	group, ctx := errgroup.WithContext(ctx)
+
+	for i, reporter := range r.reporters {
+		// Avoid starting all reporters at the same time to reduce the API call burst.
+		time.Sleep(time.Duration(i) * 10 * time.Second)
+		r.logger.Info("starting app live state reporter for plugin")
+
+		group.Go(func() error {
+			return reporter.Run(ctx)
+		})
```
Warashi, line 84: If there are N plugins, this code sleeps for `N(N+1) * 10s`. It's a bit long and the expected duration seems `N * 10s`. One more, the info log should contain the plugin name. So, I suggest 2 ways below.



1
```suggestion
		// Avoid starting all reporters at the same time to reduce the API call burst.
		group.Go(func() error {
			time.Sleep(time.Duration(i) * 10 * time.Second)
			reporter.logger.Info("starting app live state reporter for plugin")))

			return reporter.Run(ctx)
		})
```

2
```suggestion
		// Avoid starting all reporters at the same time to reduce the API call burst.
		time.Sleep(10 * time.Second)
		reporter.logger.Info("starting app live state reporter for plugin")

		group.Go(func() error {
			return reporter.Run(ctx)
		})
```

Warashi (dismissed):
Almost LGTM. I commented on one nitpick

pkg/app/pipedv1/livestatereporter/livestatereporter.go
ffjlabo: @Warashi Thanks! I fixed

- wait 10s for starting every plugin's reporter [9b697dd](https://github.com/pipe-cd/pipecd/pull/5452/commits/9b697dd175e39c8bfac67336f44c0234bcdf49e1)
- Use logger of plugin reporter [24fb566](https://github.com/pipe-cd/pipecd/pull/5452/commits/24fb566fd8e4b82eb1c4e9a254adaf1153d02e08)

Also, I added the info log to notice starting main reporter. [a7217fc](https://github.com/pipe-cd/pipecd/pull/5452/commits/a7217fc818d6312215f12cd3fcce1bb2ac3015cd)

Warashi (approved):
LGTM

ffjlabo:
I noticed that we don't fix the control plane side's rpc. This is just put it as json format.
So I removed the TODO comment for `ReportApplicationLiveState` 🙏 . [e9c9370](https://github.com/pipe-cd/pipecd/pull/5452/commits/e9c937013a9a98a6654e5dc0128031a76f49f096)

- https://github.com/pipe-cd/pipecd/blob/a7217fc818d6312215f12cd3fcce1bb2ac3015cd/pkg/app/server/grpcapi/piped_api.go#L633-L649
- https://github.com/pipe-cd/pipecd/blob/a7217fc818d6312215f12cd3fcce1bb2ac3015cd/pkg/app/server/applicationlivestatestore/store.go#L81-L91

Warashi (approved):

t-kikuc (approved):
LGTM

Issue #5006, opened 2024-06-28
k8s multicluster deployment


ffjlabo:
**What would you like to be added**:

**Why is this needed**:

PR #5453, opened 2024-12-26
Add deploy target name to the resource state

ffjlabo:
**What this PR does**:

Added `deploy_target` to show the livestate grouped by it on the Web UI.

**Why we need it**:

In the multicluster support, we introduce multi-targets in one app.

Part of https://github.com/pipe-cd/pipecd/issues/5363, https://github.com/pipe-cd/pipecd/issues/5006


t-kikuc (approved):
LTGM 

It's simpler to use `deploy_target` than the root node for GENERIC UI.

Warashi (approved):
LGTM

PR #5464, opened 2024-12-27
Add default envs in ScriptRunStage: SR_TRIGGERED_COMMANDER, SR_IS_ROLLBACK

t-kikuc:
**What this PR does**:

as title 

**Why we need it**:

- SR_TRIGGERED_COMMANDER:  e.g. To judge if the deployment was triggered via UI or not
- SR_IS_ROLLBACK:  e.g. To call the different script when it's rollback

Fixes #

### DIFF
--- a/pkg/app/piped/executor/scriptrun/scriptrun.go
+++ b/pkg/app/piped/executor/scriptrun/scriptrun.go
@@ -140,6 +140,7 @@ type ContextInfo struct {
 	RepositoryURL       string            `json:"repositoryURL,omitempty"`
 	Summary             string            `json:"summary,omitempty"`
 	Labels              map[string]string `json:"labels,omitempty"`
+	IsRollback          bool              `json:"isRollback,omitempty"`
 }
 
 // NewContextInfo creates a new ContextInfo from the given deployment.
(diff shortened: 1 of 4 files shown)


ffjlabo (approved):
Thank you!

pkg/app/piped/executor/scriptrun/scriptrun.go
```diff
@@ -136,22 +136,26 @@ type ContextInfo struct {
...
 		DeploymentID:        d.Id,
 		ApplicationID:       d.ApplicationId,
 		ApplicationName:     d.ApplicationName,
 		TriggeredAt:         d.Trigger.Timestamp,
 		TriggeredCommitHash: d.Trigger.Commit.Hash,
+		TriggeredCommander:  d.Trigger.Commander,
 		RepositoryURL:       d.GitPath.Repo.Remote,
 		Summary:             d.Summary,
 		Labels:              d.Labels,
+		IsRollback:          isRollback,
```
t-kikuc, line 158: I fixed to use `isRollback` flag instead of `d.Status`.

That's because `d.Status` is ALWAYS `PLANNED` since scheduler does not update it...

Warashi (approved):
LGTM

ffjlabo (approved):

PR #5467, opened 2025-01-07
Add plugin registry

ffjlabo:
**What this PR does**:

**Why we need it**:

Fixes #

### DIFF
--- a/pkg/plugin/registry/registry.go
+++ b/pkg/plugin/registry/registry.go
@@ -1,0 +1,141 @@
+// Copyright 2024 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package planner provides a piped component
+// that decides the deployment pipeline of a deployment.
+// The planner bases on the changes from git commits
+// then builds the deployment manifests to know the behavior of the deployment.
+// From that behavior the planner can decides which pipeline should be applied.
+package registry
+
+import (
+	"context"
+	"fmt"
+
+	config "github.com/pipe-cd/pipecd/pkg/configv1"
+	pluginapi "github.com/pipe-cd/pipecd/pkg/plugin/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


Warashi (dismissed):
LGTM

[IMO]
We have 2 kinds of packages under `pkgs/plugin`, but it's confusing. So we may have to separate them.
1. packages mainly used by the plugins
2. packages mainly used by the piped, and it operates about the plugins

Let's discuss and do it on another PR.

pkg/plugin/registry/registry.go
```diff
@@ -0,0 +1,141 @@
...
+	if len(pipeline.Stages) == 0 {
+		return nil, fmt.Errorf("no stages are set in the pipeline")
+	}
+
+	plugins := make([]pluginapi.PluginClient, 0, len(pipeline.Stages))
+	for _, stage := range pipeline.Stages {
+		plugin, ok := pr.stageBasedPlugins[stage.Name.String()]
+		if ok {
+			plugins = append(plugins, plugin)
+		}
```
t-kikuc, line 113: [ASK]
What if not `ok`?  I think it should return an error.


```suggestion
		}else{
		   return nil, fmt.Errorf("no plugin found for the stage %s", stage.Name.String())
		}
```

instead of the following:
```go
	if len(plugins) == 0 {
		return nil, fmt.Errorf("no plugin found for each stages")
	}
```

The same in `getPluginClientsByNames()`.

pkg/plugin/registry/registry_test.go
```diff
@@ -0,0 +1,315 @@
...
+	pluginapi "github.com/pipe-cd/pipecd/pkg/plugin/api/v1alpha1"
+)
+
+type fakePluginClient struct {
+	pluginapi.PluginClient
+	name string
+}
+
+func TestPluginRegistry_GetPluginClientsByAppConfig(t *testing.T) {
+	tests := []struct {
```
t-kikuc, line 37: [ASK] Is it possible to add `t.Parallel()`? 

Also in `t.Run(tt.name, func(t *testing.T) {`.

ffjlabo: Thanks, added on [1e1bd50](https://github.com/pipe-cd/pipecd/pull/5467/commits/1e1bd50517d828506b5871bc23b683a1453e9f59)

pkg/plugin/registry/registry.go
ffjlabo: I see, thanks.
I fixed to return when the plugin couldn't find by the given identity (such as stage name or plugin name) 
[58f6e2a](https://github.com/pipe-cd/pipecd/pull/5467/commits/58f6e2a57758b1a48a39fb45a73bd430905c2cbe)

Warashi (approved):

ffjlabo:
@Warashi Sorry, plz re-approve 🙏

t-kikuc (approved):
Thank you!

PR #5468, opened 2025-01-07
Move PipedPluginService definition to the under the pkg/plugin

Warashi:
**What this PR does**:

as title

**Why we need it**:

It's weird that the plugins have to depend on the package under the `pkg/app/pipedv1/cmd/piped`, so I moved the package.

Part of #4980 


ffjlabo (approved):
👍

t-kikuc (approved):

PR #5474, opened 2025-01-14
Add 'unsupported' flag in DetermineStrategyResponse

t-kikuc:
**What this PR does**:

as title.

**Why we need it**:

To distinguish whether `DetermineStrategy()` is not supported or is not successful for plugins that do not have the responsibility of DetermineStrategy().

e.g. Wait Stage plugin, Analysis Stage plugin, WaitApproval Stage plugin

Part of #4980


Warashi (approved):
LGTM

PR #5476, opened 2025-01-14
Add MigrateDatabase RPC to APIService for database migration

Warashi:
**What this PR does**:

as title

**Why we need it**:

When developing pipedv1, we have to use data with new fields, such as deploy target, or we have to patch pipedv1 temporarily to use the old field.
This PR defines the API to convert the existing data to prepare the data with new fields.

Part of #4980 


ffjlabo (approved):
LGTM

t-kikuc (approved):

PR #5481, opened 2025-01-17
Do not remove resources which is not managed by piped

Warashi:
**What this PR does**:

as title

**Why we need it**:

Without this PR, the k8s plugin tries to remove the resources that are not directly managed by piped.
For example, when the piped manages a deployment, the deployment creates replica sets. The replica sets should not be removed by piped, but they will be removed without this PR.

Part of #4980 


t-kikuc (approved):
LGTM

khanhtc1202 (approved):

PR #5502, opened 2025-01-21
fix: list should have unique "key" in web application page

hongky-1994:
**What this PR does**:
Fix error log in console web
- Child in a list should have unique "key" (Application list page)
CleanShot 2025-01-21 at 16 43 59

**Why we need it**:
- To make web console clean and prevent unwanted behavior

- None

No
- **How to migrate (if breaking change)**: None

### DIFF
--- a/web/src/components/applications-page/application-list/application-list-item/index.tsx
+++ b/web/src/components/applications-page/application-list/application-list-item/index.tsx
@@ -12,7 +12,7 @@ import MenuIcon from "@material-ui/icons/MoreVert";
 import OpenInNewIcon from "@material-ui/icons/OpenInNew";
 import clsx from "clsx";
 import dayjs from "dayjs";
-import { FC, memo, useState } from "react";
+import { FC, memo, useState, Fragment } from "react";
 import { Link as RouterLink } from "react-router-dom";
 import { APPLICATION_KIND_TEXT } from "~/constants/application-kind";
 import { PAGE_PATH_APPLICATIONS } from "~/constants/path";
@@ -160,10 +160,10 @@ export const ApplicationListItem: FC<ApplicationListItemProps> = memo(
             <div className={classes.labels}>
               {app.labelsMap.length !== 0
                 ? app.labelsMap.map(([key, value]) => (
-                    <>
+                    <Fragment key={key}>
                       <span>{key + ": " + value}</span>
                       <br />
-                    </>
+                    </Fragment>
                   ))
                 : "-"}
             </div>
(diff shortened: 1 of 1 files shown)


web/src/components/applications-page/application-list/application-list-item/index.tsx
```diff
@@ -175,12 +175,12 @@ export const ApplicationListItem: FC<ApplicationListItemProps> = memo(
...
                       v.name === "" ? (
-                        <>
+                        <Fragment key={v.version}>
                           <span>{v.version}</span>
                           <br />
-                        </>
+                        </Fragment>
                       ) : (
-                        <>
+                        <Fragment key={v.version}>
```
t-kikuc, line 183: [ASK]  
`v.version` is already used in L178.  Is that OK?

(I'm sorry, but I'm not familiar with React)

hongky-1994: Thanks for asking,

Yes, @t-kikuc it is the same key for 1 record when map from `recentlyDeployment.versionsList`.

The error show that we need uniq key for each child when map from `recentlyDeployment.versionsList` so I use v.version as uniq key.

t-kikuc: @hongky-1994 Thank you! got it.

t-kikuc (dismissed):

web/src/components/applications-page/application-list/application-list-item/index.tsx
Warashi: Thank you!
Is the `v.version` the `v0.6.0` of the screenshot below?

スクリーンショット 2025-01-22 14 08 47

This column sometimes contains multiple artifacts and may contain different artifacts with the same version.
So, I think it's a bit risky to use `v.version` as the unique key.

スクリーンショット 2025-01-22 14 10 48

hongky-1994: Thank @Warashi !
I update the pull with key `${v.name}:${v.version}`

image

The key is just needs to be unique in a single cell, so the name and version of artifacts in a single row must be different, right?

Warashi: Thanks @hongky-1994 !

> the name and version of artifacts in a single row must be different, right?

Yes, I think so. The combination of name and version is safe as a unique key in a single row.

Warashi (approved):
LGTM

khanhtc1202 (approved):
👍

Issue #5367, opened 2024-11-26
[epic][plugin] Implement Stage Plugins


t-kikuc:
**What would you like to be added**:

Support the following stages in the plugin architecture.

- [x] `WAIT` Stage
    - [x] Init server #5445
    - [x] Use metadata store #5520
    - [x] Core logic of waiting (without handling the stopSignal)
    - [x] Optimize stage config location
    - [x] Planner methods #5504
- [x] `WAIT_APPROVAL` Stage #6067
- [x] `ANALYSIS` Stage #6068
- [x] `SCRIPT_RUN` Stage #5901

Crossing-tasks: 
- [x] Piped) Define and handle `skipOn` options
- [x] Piped&Plugin) Add Notifier methods (WaitApproval)
- [x] Handling Skip and Approve actions on UI
    - [x] API) Define `ListCommands()` and `ManualOperation`
    - [x] SDK) Wrap the `ListCommands()`
    - [x] Piped(commandStore) 
         - [x] backend of `ListCommands()`
         - [x] report stages after finishing
    - [x] UI) Show the skip button if the stage is skippable (Make compatible with pipedv0)
    - [x] UI) Show the approval button if the stage is approvable (Make compatible with pipedv0)
        - [x] How to restrict approvals from only specified users
    - [x] Add to the SDK

Need Design in advance
- [x] Analysis > Analysis Template #6036 
- [x] Analysis > analysisResultStore

~Note: We've not determined how to support the `SCRIPT_RUN` Stage in the plugin architecture.~

**Why is this needed**:

Stages will also become plugins, not built-in code.
That will extend the flexibility of pipelines.

t-kikuc:
> [ ] Piped&Plugin) Handle commands of skipStage and approveStage (Analysis, WaitApproval)
> [ ] UI&model) Add skippable flag to PipelineStage and switch by it on UI (Analysis)
> [ ] UI&model) Add approvable flag to PipelineStage and switch by it on UI (WaitApproval)

The overview of handling SkipStage and ApproveStage will be like this:

![Image](https://github.com/user-attachments/assets/2d7df502-4609-4e0c-b750-05ae8f59b446)




- Add `skippable` and `approvable` flags in the model `PipelineStage`.
- Set them true in `BuildPipelineSyncStages()` of Analysis and WaitApproval stage plugins.
- Define and use `ListStageCommands()` and `ReportStageCommandHandled()` APIs from a plugin to Piped to handle commands on the plugin side.
[...]

PR #5504, opened 2025-01-21
[WAIT] Implement server plan methods of Wait plugin

t-kikuc:
**What this PR does**:

Implemented the following methods:

- DetermineVersions()
- DetermineStrategy()
- BuildQuickSyncStages()
- BuildPipelineSyncStages()

**Why we need it**:

to plan wait stages

**Note**

[IMO] It'd be better to set values of the following variables of BuildQuick(or Pipeline)SyncStagesResponse on piped planner side:

- `Status`
- `CreatedAt`
- `UpdatedAt`

That's because plugins should not set wrong values and they're not related to plugin behavior.

Part of #5367


t-kikuc:
I'll create another PR using the SDK

PR #5516, opened 2025-01-27
Feat/form application piped v1

hongky-1994:
**What this PR does**:
- Create Application form V1
- Refactor Application Form V0 and Manual Form
- Update logic AddApplicaitonm Drawer and Edit Application-drawer
- Fix minor bugs: 
  - Refetch application list when form is submit successfully.
  - Dupplicate Quit warning dialog.

**Why we need it**:
- Application Create and Edit form need to be suitable for both piped v0 and piped v1 #5252

**Which issue(s) this PR fixes**:
Fixes #5509 

**Does this PR introduce a user-facing change?**:
Yes

- **How are users affected by this change**:
  - Now user will have 3 option to create an application
    - Piped v0 add from suggestion
    - Piped v1 add from suggestion
    - Add Manually
  
CleanShot 2025-01-28 at 00 20 45

- **Is this breaking change**: No
- **How to migrate (if breaking change)**:

Issue #5509, opened 2025-01-23
Update Form add and edit Application for piped v1


hongky-1994:
> @hongky-1994 will keep working on the UI update for piped v1
> 
> ### PlatformProvider; UI
> #### Purpose
> **Mostly related to 2 model applications and deployment**
> 
> Used in
> 
> * API call AddApplication, UpdateApplication
> * For showing in Application, Deployment detail page
> 
> The source fetched to used in the drawer is come from the piped configuration
> 
> https://github.com/pipe-cd/pipecd/blob/master/web/src/components/application-form/index.tsx#L58-L80
> 
> #### Modification idea
> * Keep it as in the application and deployment model
> * Also, keep the configuration and parser for pipedv0 config the same as it is (no UI affected around that fetch)
> * Make a new Add Application form/suggestion form that doesn't show any platform provider-related information
> 
>  

 _Originally posted by @khanhtc1202 in [#5252](https://github.com/pipe-cd/pipecd/issues/5252#issuecomment-2609331542)_

PR #5517, opened 2025-01-29
Update application endpoints web api

khanhtc1202:
**What this PR does**:

Make platform providers and kinds not required for application-related endpoints (AddApplication, UpdateApplication) web API.

**Why we need it**:

Follow this comment (https://github.com/pipe-cd/pipecd/pull/5516#discussion_r1930905224), we're updating the web UI to accept both pipedv0 and pipedv1, the applications which be registered by pipedv1 will have no platform nor kind, so we should update the web API to accept that.

Part of #5252 #5509


Warashi (approved):
LGTM

ffjlabo (approved):

PR #5456, opened 2024-12-26
[Wait] Implemented the core of executing Wait with WaitStageOptions

t-kikuc:
**What this PR does**:

- Created `WaitStageOptions`
- Implemented `execute()` of Wait

TODO:
- use metadataStore
- handle`SkipOn` options
- enable building pipelineSyncStages

**Why we need it**:

- To execute the Wait stage as a plugin.

**Which issue(s) this PR fixes**:

Part of #5367

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #5520, opened 2025-01-30
[WAIT] Use metadataStore in the wait stage plugin

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

- To preserve the time when the wait stage began in order to avoid waiting for too long after rebooting.
- Previously, I commented out codes where the metadataStore is used to focus on core logic in #5456

Part of #5367

### DIFF
--- a/pkg/app/pipedv1/plugin/wait/deployment/wait.go
+++ b/pkg/app/pipedv1/plugin/wait/deployment/wait.go
@@ -89,7 +89,7 @@ func wait(ctx context.Context, duration time.Duration, initialStart time.Time, s
 	}
 }
 
-func (s *deploymentServiceServer) retrieveStartTime(ctx context.Context, deploymentID, stageID string) (t time.Time) {
+func (s *deploymentServiceServer) retrieveStartTime(ctx context.Context, deploymentID, stageID string) time.Time {
 	sec, err := s.metadataStore.GetStageMetadata(ctx, &pipedservice.GetStageMetadataRequest{
 		DeploymentId: deploymentID,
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/wait/deployment/wait.go
```diff
@@ -84,27 +89,32 @@ func wait(ctx context.Context, duration time.Duration, initialStart time.Time, s
...
+		Key:          startTimeKey,
+	})
+	if err != nil {
+		s.logger.Error(fmt.Sprintf("failed to get stage metadata %s", startTimeKey), zap.Error(err))
+		return
+	}
+	ut, err := strconv.ParseInt(sec.Value, 10, 64)
+	if err != nil {
+		s.logger.Error(fmt.Sprintf("failed to parse stage metadata %s", startTimeKey), zap.Error(err))
+		return
```
Warashi, line 105: [IMO]
I don't like naked returns, and named result parameters that don't clarify the usage of their return values.
ref; https://go.dev/wiki/CodeReviewComments#named-result-parameters

t-kikuc: @Warashi 
Thank you. I agree with you and have fixed it. [7113bd3](https://github.com/pipe-cd/pipecd/pull/5520/commits/7113bd360fdaf94a37fd49a61032c02a1596592e)

Warashi (approved):
LGTM

khanhtc1202 (approved):
👍

PR #5522, opened 2025-01-30
[Stage Plugins] Add `availableOperation` in the PipelineStage model

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

- To judge if the stage can be skipped/approved on UI

cf. See here for the overview of skipStage/ApproveStage: https://github.com/pipe-cd/pipecd/issues/5367#issuecomment-2624728722 

Part of #5367


pkg/model/deployment.proto
```diff
@@ -149,6 +149,11 @@ message PipelineStage {
     int64 completed_at = 13 [(validate.rules).int64.gte = 0];
     int64 created_at = 14 [(validate.rules).int64.gt = 0];
     int64 updated_at = 15 [(validate.rules).int64.gt = 0];
+
+    // Whether this stage can be skipped by a user via UI.
+    bool skippable = 16;
+    // Whether this stage can be approved by a user via UI.
+    bool approvable = 17;
```
t-kikuc, line 156: I named it `approvable`, not like `waits_approval` because it depends on whether a plugin waits for approvals.

If you have any idea of the name, please share 🙏

Warashi: How about making these flags as enum field like `candidate_ui_action`?
We don't have a use case for both skippable and approvable are set true for now, so we can make them a single field.

t-kikuc: Thank you, it seems greater.
I fixed as below:

[b98dcad](https://github.com/pipe-cd/pipecd/pull/5522/commits/b98dcad459c9b2a20f160abca91d81c9bbca0833)


Let's make it `repeated` when needed in the future.

Warashi (approved):
LGTM

ffjlabo (approved):
👍

PR #5532, opened 2025-02-03
Ensure ApplicationKind only accept enum defined value

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

Instead of not validating the ApplicationKind, which web form submitted to the API as preparation for pipedv1 managing application, we should mark this as accepting only defined enum value. With application managing by pipedv1, this field is not required, but the web form will send value 0 (kind: Kubernetes) as default; this field would not be used in any logic on the server side of piped side in case of v1 piped.

Fllow #5517


Warashi (approved):
👍🏻

t-kikuc (approved):

PR #5535, opened 2025-02-04
[WIP] Implement multi cluster k8s sync for non template

ffjlabo:
**What this PR does**:

I fixed the K8S_SYNC to deploy an app to the multiple clusters.

**Details**
I introduced `spec.input.multiTargets` field.
We can control the settings (manifests, kubectlVersion and so on) for each deploy target.

Here is the example of the app.pipecd.yaml with it.
```yaml
kind: KubernetesApp
spec:
  name: simple
  labels:
    env: example
    team: product
  quickSync:
    prune: true
  input:
    kubectlVersion: 1.31.0
    multiTargets:
      - target:
          name: cluster1
        manifests:
          - ./cluster1/deployment.yaml
        kubectlVersion: 1.31.0
      - target:
          name: cluster2
        manifests:
          - ./cluster2/deployment.yaml
        kubectlVersion: 1.31.0
```

If multiTargets are set, the configuration for each target is applied and deployed accordingly. If multiTargets is not set, the deployment is executed for all deploy targets.

Also, each multi-target process is executed in parallel. 
K8S_SYNC is successful when all of the processes are executed successfully.

**Why we need it**:

Part of #5006

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/config/application.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/config/application.go
@@ -75,6 +75,19 @@ type KubernetesDeploymentInput struct {
 	AutoCreateNamespace bool `json:"autoCreateNamespace,omitempty"`
 
 	// TODO: Define fields for KubernetesDeploymentInput.
+	MultiTargets []KubernetesMultiTarget `json:"multiTargets,omitempty"`
+}
+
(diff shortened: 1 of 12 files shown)


ffjlabo:
We decided to implement multi-cluster feature as another k8s plugin.
So I close this PR for now.

PR #5538, opened 2025-02-04
[Stage Plugins] Define Notifier APIs of Approved and WaitApproval

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

The WaitApproval plugin needs to: 

- notify approvers when the stage starts (`NotificationEventDeploymentWaitApproval`)
- notify when approval is completed  (`NotificationEventDeploymentApproved`)

cf. https://github.com/pipe-cd/pipecd/blob/2ce020e567f97c767675a5769aefe5d9142cdf13/pkg/app/piped/executor/waitapproval/waitapproval.go#L112-L143

They depend on the current [matcher/sender configs](https://pipecd.dev/docs-v0.50.x/user-guide/managing-piped/configuring-notifications/).

**Note**

- Other notifications must NOT be triggered by plugins (e.g. DeploymentStarted, StageFinished)
- If a plugin wants to send other customized notifications, the plugin should do so within the plugin without the pipedservice.

- These pipedservice methods can be used by other plugins
   - e.g. customized notification plugin
   - The advantage of using them over full-scratch notifications is that the current matcher/sender can be used.

Part of #5367


t-kikuc:
I'll refine in another PR.
accounts/groups should not be touched in plugin side

PR #5541, opened 2025-02-05
Call `runningDSP.Get` when runningDSP  is not nil

ffjlabo:
**What this PR does**:

I fixed to call `runningDSP.Get` when runningDSP is not nil.

**Why we need it**:

When first deploy, runningDSP is nil because previous running commit is not found.
So it led panic.

Part of #4980


Warashi (approved):
LGTM

t-kikuc (approved):
like here? i got it
https://github.com/pipe-cd/pipecd/blob/edcf430c1548878c56e06d16f311edbbd8423564/pkg/app/piped/controller/scheduler.go#L257-L267

PR #5545, opened 2025-02-06
Add plugins to piped model

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

We need to use plugins' information in the application form on the UI. This PR adds the plugin information to the piped model, and when piped is fetched and shown on the UI, we can have plugin info to use.

Part of #5252


ffjlabo (approved):
LGTM

PR #5547, opened 2025-02-06
Add deploy target by plugin to application web API

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

These API will be used by forms in Web UI to register and edit application

Part of #5252


pkg/app/server/grpcapi/api.go
```diff
@@ -361,7 +361,7 @@ func (a *API) UpdateApplication(ctx context.Context, req *apiservice.UpdateAppli
 		return nil, status.Error(codes.InvalidArgument, "Requested piped does not belong to your project")
 	}
 
-	if err := a.applicationStore.UpdateConfiguration(ctx, req.ApplicationId, req.PipedId, req.PlatformProvider, req.GitPath.ConfigFilename); err != nil {
+	if err := a.applicationStore.UpdateConfiguration(ctx, req.ApplicationId, req.PipedId, req.PlatformProvider, req.GitPath.ConfigFilename, nil); err != nil {
```
khanhtc1202, line 364: Just pass nil here and not update the grpc for this API since this one is used by pipectl and outer tool (like https://github.com/pipe-cd/terraform-provider-pipecd), I will consider to update this later

Warashi (approved):
GLTM

PR #5549, opened 2025-02-07
Improve docker build performance

### LINKED ISSUE #5548: Docker image build on CI failure
**What happened**:

We see a build failure in the Docker build on CI.

for example, https://github.com/pipe-cd/pipecd/actions/runs/13126728528/job/36624384578#step:8:1

no failure occurs

**How to reproduce it**:

run CI many times.

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

Warashi:
**What this PR does**:

Build a Go binary with the host platform using the GOOS/GOARCH option and copy it to the target container image.
ref; https://docs.docker.com/build/building/multi-platform/#cross-compiling-a-go-application

**Why we need it**:

We sometimes face build failure at go build. I think this comes from QEMU and Go combination. So I changed the Dockerfile to build go binary without QEMU.
https://github.com/pipe-cd/pipecd/actions/runs/13126728528/job/36624384578#step:8:1

Maybe fixes #5548 

### DIFF
--- a/cmd/helloworld/Dockerfile
+++ b/cmd/helloworld/Dockerfile
@@ -1,4 +1,8 @@
-FROM golang:1.23.3 AS builder
+# syntax=docker/dockerfile:1
+FROM --platform=$BUILDPLATFORM golang:1.23.3 AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
 
 WORKDIR /app
 
@@ -7,7 +11,7 @@ RUN go mod download
 
 COPY . ./
 
-RUN make build/go MOD=helloworld
+RUN make build/go MOD=helloworld BUILD_OS=${TARGETOS} BUILD_ARCH=${TARGETARCH}
 
 FROM alpine:3.21.0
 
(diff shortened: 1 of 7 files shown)


khanhtc1202 (approved):
👍

ffjlabo:
@Warashi 
So we don't need the QEMU setup in the workflow?
https://github.com/pipe-cd/pipecd/blob/75db0fad0eb9d6a08914a73a003f455442a42c38/.github/workflows/publish_image_chart.yaml#L66C1-L70C91

Warashi:
> @Warashi So we don't need the QEMU setup in the workflow? https://github.com/pipe-cd/pipecd/blob/75db0fad0eb9d6a08914a73a003f455442a42c38/.github/workflows/publish_image_chart.yaml#L66C1-L70C91

@ffjlabo Thank you for your comment.
We still need QEMU because some multi-stage builds that have no platform options are run on QEMU.
We have to run them on QEMU because the final image must be built on the same platform as the image's platform.

ffjlabo (approved):
Sorry, I forgot to approve 🙏 LGTM

PR #5551, opened 2025-02-07
Fix error unique key when map - react

hongky-1994:
**What this PR does**:
- Fix error on missing key when map array - react

**Why we need it**:
- Remove error - warning
image

Fixes #


khanhtc1202 (approved):
👍

ffjlabo (approved):
👍

PR #5552, opened 2025-02-07
Pipedv1 reports plugins while report metadata

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

Piped need to report plugins as it metadata, so we can use this info on Web UI application form

Part of #5252 
Follow #5545


ffjlabo (approved):
LGTM

t-kikuc (approved):

PR #5556, opened 2025-02-07
Add DeployTargetsByPlugin to Web API

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

These API will be used by forms in Web UI to register and edit application

Part of #5252


pkg/app/server/grpcapi/api.go
```diff
@@ -360,7 +360,7 @@ func (a *API) UpdateApplication(ctx context.Context, req *apiservice.UpdateAppli
 		return nil, status.Error(codes.InvalidArgument, "Requested piped does not belong to your project")
 	}
 
-	if err := a.applicationStore.UpdateConfiguration(ctx, req.ApplicationId, req.PipedId, req.PlatformProvider, req.GitPath.ConfigFilename); err != nil {
+	if err := a.applicationStore.UpdateConfiguration(ctx, req.ApplicationId, req.PipedId, req.PlatformProvider, req.GitPath.ConfigFilename, nil); err != nil {
```
ffjlabo, line 363: Maybe add `req.DeployTargetsByPlugin` as args?

```suggestion
	if err := a.applicationStore.UpdateConfiguration(ctx, req.ApplicationId, req.PipedId, req.PlatformProvider, req.GitPath.ConfigFilename, req.DeployTargetsByPlugin); err != nil {
```

khanhtc1202: We can't do that, since this api be used by other tools like terraform-pipecd. I will consider to change this later 👀

ffjlabo: @khanhtc1202 I got it. So it would be nice to add a comment about the context.

IMO, we need to update the API for pipectl and external tool because it will update as nil when adding the app and later update via pipectl and some other tool.

khanhtc1202: @ffjlabo Ofcource, we will update pipectl commands related to applications later. I just do not want to update the API this time to confuse the outer tools dev because the pipedv1 plugin is not yet ready.
Also, I don't add comments here since it's not just this API but others like AddApplication that need to be revised later, too. The change here isn't caused by an API change but by an internal store interface change, so fewer comments are okay. We have to revise all the pipette commands later, so I think a comment is not needed.

ffjlabo (approved):
LGTM

t-kikuc (approved):

PR #5558, opened 2025-02-07
Enhanced EventWatcher logs

t-kikuc:
**What this PR does**:

Mainly: 
- add messages about "will retry in the next loop" and retry counts 
- convert log lever from Error to Warn in `retry.Do()`
   - After failing for the max count, ErrorLog will be shown
- add event-ids field 

**Why we need it**:

- To clarify that some errors can be solved in the next loop or retry
- To suppress too many logs while retry
- To enable to track which events failed


ffjlabo (dismissed):
👍

ffjlabo (approved):

khanhtc1202 (approved):
👍

PR #5563, opened 2025-02-14
Update kubectl version in deployment examples to 1.32.2

Warashi:
**What this PR does**:

as title

**Why we need it**:

We use examples in the k8s plugin test, but there is no binary release for kubectl 1.18.0 for aarch64-darwin. So, the test fails on macOS with Apple Silicon.

There is no need to specify that old version, so I upgraded it to the latest.
ref; https://kubernetes.io/releases/

Fixes #

**Does this PR introduce a user-facing change?**:　No


khanhtc1202 (approved):

t-kikuc (approved):
thank you!

PR #5564, opened 2025-02-14
Add MANUAL_OPERATION_NONE to ManualOperation enum

Warashi:
**What this PR does**:

as title

**Why we need it**:

We need a way to represent that no manual operation is available on the Web UI.
It's not unknown, so it's not a good way to use UNKNOWN to represent it.

Part of #5367


khanhtc1202 (approved):
🆗

PR #5578, opened 2025-02-17
Improve docker build performance

Warashi:
**What this PR does**:

- Use `--platform=$BUILDPLATFORM` flag to build web static assets in a docker multi-stage build
- Use docker build cache from the latest image in GitHub Actions Workflow

**Why we need it**:

- I want to reduce CI build time

Fixes #

### DIFF
--- a/.github/workflows/build_tool.yaml
+++ b/.github/workflows/build_tool.yaml
@@ -40,3 +40,6 @@ jobs:
           context: tool/${{ matrix.image }}
           tags: ${{ env.REGISTRY }}/pipe-cd/${{ matrix.image }}:${{ env.PIPECD_VERSION }}
           platforms: linux/amd64,linux/arm64
+          # parameter to use inline cache. ref; https://docs.docker.com/build/ci/github-actions/cache/#inline-cache
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/pipe-cd/${{ matrix.image }}:latest
+          cache-to: type=inline
--- a/cmd/pipecd/Dockerfile
+++ b/cmd/pipecd/Dockerfile
@@ -1,19 +1,13 @@
 # syntax=docker/dockerfile:1
 # web builder
-# because of this issue, we choose node 18
-# https://github.com/pipe-cd/pipecd/issues/5422
-# https://github.com/nodejs/docker-node/issues/1335#issuecomment-2024344411
-FROM node:18.20.5-alpine3.21 AS web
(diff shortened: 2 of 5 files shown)


cmd/pipecd/Dockerfile
```diff
@@ -1,20 +1,14 @@
 # syntax=docker/dockerfile:1
 # web builder
-# because of this issue, we choose node 18
-# https://github.com/pipe-cd/pipecd/issues/5422
-# https://github.com/nodejs/docker-node/issues/1335#issuecomment-2024344411
-FROM node:18.20.5-alpine3.21 AS web
+# because this stage builds only web assets, we can use any platform
+FROM --platform=$BUILDPLATFORM node:18.20.5-alpine3.21 AS web
```
Warashi, line 4: With this modification, we can now use node 20 or later. I want to discuss whether we should upgrade this to current LTS versions.

khanhtc1202 (approved):
Nice improvement 👍

.github/workflows/build_tool.yaml
```diff
@@ -40,3 +40,6 @@ jobs:
           context: tool/${{ matrix.image }}
           tags: ${{ env.REGISTRY }}/pipe-cd/${{ matrix.image }}:${{ env.PIPECD_VERSION }}
           platforms: linux/amd64,linux/arm64
+          # parameter to use inline cache. ref; https://docs.docker.com/build/ci/github-actions/cache/#inline-cache
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/pipe-cd/${{ matrix.image }}:latest
+          cache-to: type=inline
```
ffjlabo, line 45: [IMO] How about adding `mode=max` ? 
It seems that we can cache all layer with it. 
https://docs.docker.com/build/cache/backends/#cache-mode

It might be more efficient because our images are often built with mutl-stage build to download dependencies.

Warashi: Unfortunately, the inline cache doesn't support the max cache mode.
> However, note that the inline cache exporter only supports min cache mode.

ref; https://docs.docker.com/build/ci/github-actions/cache/#inline-cache

We can consider using the GitHub cache, but it's experimental yet. So, I want to start by using inline cache and min cache mode.

ffjlabo: I got it. Thanks!

ffjlabo (approved):

PR #5586, opened 2025-02-19
[WAIT,SDK] Implemented Wait stage with SDK

t-kikuc:
**What this PR does**:

- impl wait stage plugin based on the SDK
- refine the SDK
- Pass only `PipelineStage.With` in scheduler

**Why we need it**:

it's simpler to use SDK for stage plugins

Part of #5367


khanhtc1202 (approved):
👍

ffjlabo (approved):

pkg/configv1/application.go
```diff
@@ -182,11 +182,13 @@ func (s GenericApplicationSpec) GetStageByte(index int32) ([]byte, bool) {
...
-		return nil, false
-	}
-	return b, true
+
+	return []byte(stage.With), true
+	// b, err := json.Marshal(stage)
+	// if err != nil {
+	// return nil, false
+	// }
+	// return b, true
```
Warashi, line 191: 📝 
We should remove these lines in another PR

Warashi (approved):

Issue #5530, opened 2025-02-03
Implementing an SDK for plugin implementations


Warashi:
**What would you like to be added**:

The SDK to help implement plugins.

**Why is this needed**:

Plugin implementations have many boilerplates, so it is nice to have an SDK.

t-kikuc:
I want a way to tell the SDK users about changelog, especially breaking changes.

It's difficult because the commit hash to refer is different from this repo.
I can't find the best way. 🤨

For example:

1. Create `CHANGELOG.origin` file in the SDK dir, and `CHANGELOG` file in the SDK repo
2. Update `CHANGELOG.origin` when developing the SDK in this repo
3. When syncing to the SDK repo, sync from `CHANGELOG.origin` to `CHANGELOG` with overwriting the commit hash.

khanhtc1202:
Due to the sdk v0.1.0 release (ref: https://github.com/pipe-cd/piped-plugin-sdk-go/releases/tag/v0.1.0), we can close this issue for now 🎉

PR #5588, opened 2025-02-19
[SDK] Refine StageLogPersister handling in SDK

Warashi:
**What this PR does**:

- define type `sdk.StageLogPersister` and replace `logpersister.StageLogPersister` with it.
- Add graceful shutdown handling in ExecuteStage

**Why we need it**:

- The `Complete` method of StageLogPersister is called only in SDK, so hide it from SDK users.
- We have to handle graceful shutdown to implement plugins correctly, but it's difficult for users who don't know the piped behavior. So, I want to do it in SDK.

Part of #5530 

### DIFF
--- a/pkg/app/pipedv1/plugin/wait/wait.go
+++ b/pkg/app/pipedv1/plugin/wait/wait.go
@@ -22,7 +22,6 @@ import (
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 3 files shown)


khanhtc1202 (approved):
LGTM 👍

t-kikuc (approved):
Thank you!

PR #5591, opened 2025-02-20
Remove minimum pair validation for deploy targets in RPC

Warashi:
**What this PR does**:

as title

**Why we need it**:

We have to continue supporting application registration with a non-plugin-architectured piped.
Such a piped doesn't support deploy targets configuration, so we cannot validate that it must be present.

Part of #5252 


t-kikuc (approved):
Thank you!

In the future, it might be better to create `AddApplicationRequestV1` if the validation is complex.

ffjlabo (approved):
LGTM

PR #5595, opened 2025-02-21
Add increase operation to calc Total triggered deploys

Warashi:
**What this PR does**:

This PR fixes the Grafana dashboard to use the `increase` operation when calculating the total triggered deployments in the last 24 hours.

**Why we need it**:

The metric `grpcapi_create_deployment_total` is a counter metric, so we have to use the `increase` operation to view it correctly.

Fixes #

The control plane managers who use the monitoring feature.

### DIFF
--- a/manifests/pipecd/grafana-dashboards/control-plane/overview.json
+++ b/manifests/pipecd/grafana-dashboards/control-plane/overview.json
@@ -466,7 +466,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum(grpcapi_create_deployment_total{project=~\"$project\"})",
+          "expr": "sum(increase(grpcapi_create_deployment_total{project=~\"$project\"}[24h]))",
           "interval": "",
           "legendFormat": "",
           "refId": "A"
--- a/manifests/pipecd/grafana-dashboards/control-plane/overview.json
+++ b/manifests/pipecd/grafana-dashboards/control-plane/overview.json
@@ -531,7 +531,7 @@
       "targets": [
         {
           "exemplar": true,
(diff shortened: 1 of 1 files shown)


khanhtc1202 (dismissed):
Nice catch 👍

manifests/pipecd/grafana-dashboards/control-plane/overview.json
```diff
@@ -531,7 +531,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum by (project) (grpcapi_create_deployment_total)",
+          "expr": "sum by (project) (increase(grpcapi_create_deployment_total[24h]))",
```
khanhtc1202, line 534: Should we use `ceil(increase)` instead of just increase, since the value could be not an integer? 🤔

Warashi: I think the `increase` of a counter metric must be an integer, but I'll add the `ceil` operation for safety.

manifests/pipecd/grafana-dashboards/control-plane/overview.json
```diff
@@ -531,7 +531,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum by (project) (grpcapi_create_deployment_total)",
+          "expr": "sum by (project) (round(increase(grpcapi_create_deployment_total[24h])))",
```
Warashi, line 534: I found some floating-point numbers in the results, and adding `round` behaves as expected.

khanhtc1202 (approved):
👍

t-kikuc (approved):
Thank you!

PR #5597, opened 2025-02-21
Fix cancel handling in pipedv1 scheduler

Warashi:
**What this PR does**:

- fix the handling of errors when the context is canceled. This occurs when the user cancels the deployment from WebUI.
- remove `StageStatusCancelled` from the SDK because cancel is handled in the piped. We don't have to handle cancel in the plugin implementations.

**Why we need it**:

- to fix cancel behavior
- to clarify what we need to implement in plugins

Part of #4980 #5530

### DIFF
--- a/pkg/app/pipedv1/plugin/wait/wait.go
+++ b/pkg/app/pipedv1/plugin/wait/wait.go
@@ -78,7 +78,10 @@ func wait(ctx context.Context, duration time.Duration, initialStart time.Time, s
 
 		case <-ctx.Done(): // on cancelled
 			slp.Info("Wait cancelled")
-			return sdk.StageStatusCancelled
+			// The piped handles this case as cancelled by the user without using the plugin's result.
+			// So we don't need to consider which status should be returned.
+			// We return the failure here.
+			return sdk.StageStatusFailure
 		}
 	}
 }
--- a/pkg/app/pipedv1/plugin/wait/wait.go
+++ b/pkg/app/pipedv1/plugin/wait/wait.go
@@ -78,9 +78,8 @@ func wait(ctx context.Context, duration time.Duration, initialStart time.Time, s
 
 		case <-ctx.Done(): // on cancelled
 			slp.Info("Wait cancelled")
-			// The piped handles this case as cancelled by the user without using the plugin's result.
-			// So we don't need to consider which status should be returned.
-			// We return the failure here.
+			// We can return any status here because the piped handles this case as cancelled by a user, 
+			// ignoring the result from a plugin.
 			return sdk.StageStatusFailure
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/plugin/sdk/deployment.go
```diff
@@ -455,9 +455,8 @@ type ExecuteStageResponse struct {
 type StageStatus int
 
 const (
-	StageStatusSuccess   StageStatus = 2
-	StageStatusFailure   StageStatus = 3
-	StageStatusCancelled StageStatus = 4
+	StageStatusSuccess StageStatus = 2
+	StageStatusFailure StageStatus = 3
```
khanhtc1202, line 459: [Q] I don't remember why we made this enum start from 2; could you teach me? 👀

Warashi: Because these lines are copied from below. It's a bit confusing, so I want to make them start from 1.
https://github.com/pipe-cd/pipecd/blob/ad00a56f0974334cd6f4a0dbca5b505ab19d4f5c/pkg/model/deployment.pb.go#L109-L120

Warashi: I refactored it on this commit.
https://github.com/pipe-cd/pipecd/pull/5597/commits/a755ab8a3859edf1cac522337c67648ca2707ced

pkg/app/pipedv1/plugin/wait/wait.go
```diff
@@ -78,7 +78,10 @@ func wait(ctx context.Context, duration time.Duration, initialStart time.Time, s
 
 		case <-ctx.Done(): // on cancelled
 			slp.Info("Wait cancelled")
-			return sdk.StageStatusCancelled
```
t-kikuc, line 81: [IMO] I think `StageStatusCancelled` should remain, although it's not used in piped.
That's because plugin developers will be confused about which status to return.

If we want to remove `StageStatusCancelled`, we should remove `case <-ctx.Done():` section too. (If possible, that's ideal)

pkg/app/pipedv1/plugin/wait/wait.go
```diff
@@ -78,7 +78,10 @@ func wait(ctx context.Context, duration time.Duration, initialStart time.Time, s
...
+			// The piped handles this case as cancelled by the user without using the plugin's result.
+			// So we don't need to consider which status should be returned.
+			// We return the failure here.
```
t-kikuc, line 83: [IMO] this might be clearer

```suggestion
			// We can return any status here because the piped handles this case as cancelled by a user, 
			// ignoring the result from a plugin.
```

t-kikuc (approved):
Thank you!

khanhtc1202 (approved):
LGTM 👍

PR #5602, opened 2025-02-26
Add empty methods which need for implementing the k8s plugin with the SDK

ffjlabo:
**What this PR does**:

as title.
I plan to rewrite it with SDK as follow.
- implement the current logic with SDK into struct `sdkPlugin`.
- Replace current struct `plugin` after implementing all methods which need for the interface `DeploymentPlugin`.

**Why we need it**:

We want to rewrite plugin by using SDK while keeping the current k8s plugin work. 
Also, I want to rewrite it before implementing multi cluster plugin.

Part of https://github.com/pipe-cd/pipecd/issues/4980  https://github.com/pipe-cd/pipecd/issues/5006

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/plugin.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/plugin.go
@@ -25,11 +25,13 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/pipe-cd/pipecd/pkg/admin"
+	kubeconfig "github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/kubernetes/config"
 	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/kubernetes/deployment"
 	"github.com/pipe-cd/pipecd/pkg/cli"
 	config "github.com/pipe-cd/pipecd/pkg/configv1"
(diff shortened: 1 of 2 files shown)


Warashi (approved):

pkg/app/pipedv1/plugin/kubernetes/plugin.go
```diff
@@ -162,3 +164,47 @@ func (s *plugin) run(ctx context.Context, input cli.Input) (runErr error) {
...
 }
+
+// sdkPlugin implements the sdk.DeploymentPlugin interface.
+// TODO: Rename to 'plugin' after rewriting the current plugin logic with sdk.
+type sdkPlugin struct{}
+
+var _ sdk.DeploymentPlugin[sdk.ConfigNone, kubeconfig.KubernetesDeployTargetConfig] = (*sdkPlugin)(nil)
+
+func (p *sdkPlugin) Name() string {
+	return "kubernetes"
```
t-kikuc, line 175: let's update it to like 'kubernetes-multicluster' later

t-kikuc (approved):

pkg/app/pipedv1/plugin/kubernetes/plugin.go
ffjlabo: This is for rewriting kubernetes plugin. So I think the plugin name is correct.

PR #5610, opened 2025-02-27
Update deployment trace ui

hongky-1994:
**What this PR does**:
- Change title from 'Deployment traces' to 'Traces'
- Add hover effect for trace-item in list
- Fix bugs overflow header on mobile view.
- Not hide btn commit message when open deployment list

**Why we need it**:
- For better UI and UX
CleanShot 2025-02-27 at 17 23 17@2x
CleanShot 2025-02-27 at 17 24 29
CleanShot 2025-02-27 at 17 25 28@2x

Fixes #

### DIFF
--- a/web/src/components/deployment-trace-page/deployment-trace-item/deployment-item.tsx
+++ b/web/src/components/deployment-trace-page/deployment-trace-item/deployment-item.tsx
@@ -17,15 +17,25 @@ const useStyles = makeStyles((theme) => ({
     padding: theme.spacing(2),
     display: "flex",
     alignItems: "center",
-    // backgroundColor: theme.palette.background.paper,
+    overflow: "hidden",
+    columnGap: theme.spacing(2),
+    [theme.breakpoints.down("sm")]: {
+      flexDirection: "column",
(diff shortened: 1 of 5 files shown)


khanhtc1202:
Hi @hongky-1994, Thanks for the speed :D 
I just want to confirm one point. With the hover effect added to the Trace item, will the user be able to see the deployment list by clicking on the item or having to click on the right-size arrow?

hongky-1994:
> Hi @hongky-1994, Thanks for the speed :D I just want to confirm one point. With the hover effect added to the Trace item, will the user be able to see the deployment list by clicking on the item or having to click on the right-size arrow?

Currently user have to click btn arrow on the right side. We can implement click on item feature but, in the deployment trace item , we have commit link which is hover show cursor pointer too, this may make user doesn't know where to click.

khanhtc1202:
@hongky-1994 Got your point. Let's keep the current spec 👍

khanhtc1202 (approved):
🙌

t-kikuc (approved):

PR #5616, opened 2025-03-03
Pass DeployTarget in SDK

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

We need DeployTarget for the Deployment plugin to use platform-specific config.

Part of https://github.com/pipe-cd/pipecd/issues/5530

### DIFF
--- a/pkg/plugin/sdk/deployment.go
+++ b/pkg/plugin/sdk/deployment.go
@@ -207,11 +207,8 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
 		logPersister:  lp,
 	}
 
-	dtNames, err := request.GetInput().GetDeployment().GetDeployTargets(s.commonFields.config.Name)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get deploy targets from deployment for plugin %s: %v", s.commonFields.config.Name, err)
-	}
-
+	// Get the deploy targets set on the deployment from the piped plugin config.
+	dtNames := request.GetInput().GetDeployment().GetDeployTargets(s.commonFields.config.Name)
 	deployTargets := make([]*DeployTarget[DeployTargetConfig], 0, len(dtNames))
 	for _, name := range dtNames {
 		dt := s.commonFields.config.FindDeployTarget(name)
--- a/pkg/plugin/sdk/deployment.go
+++ b/pkg/plugin/sdk/deployment.go
@@ -206,7 +206,34 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
 		stageID:       request.GetInput().GetStage().GetId(),
 		logPersister:  lp,
 	}
-	return executeStage(ctx, s.base, &s.config, nil, client, request, s.logger) // TODO: pass the deployTargets
+
+	dtNames, err := request.GetInput().GetDeployment().GetDeployTargets(s.commonFields.config.Name)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get deploy targets for plugin %s: %v", s.commonFields.config.Name, err)
+	}
+
+	deployTargets := make([]*DeployTarget[DeployTargetConfig], 0, len(dtNames))
+	for _, name := range dtNames {
+		if dt := s.commonFields.config.FindDeployTarget(name); dt != nil {
+			var sdkDt DeployTargetConfig
+			if err := json.Unmarshal(dt.Config, &sdkDt); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to unmarshal deploy target config: %v", err)
+			}
+
+			deployTargets = append(deployTargets, &DeployTarget[DeployTargetConfig]{
+				Name:   name,
+				Labels: dt.Labels,
+				Config: sdkDt,
+			})
+		}
+	}
+
+	// The number of deploy targets should be the same as the number of deploy target names set on the deployment.
+	if len(dtNames) != len(deployTargets) {
+		return nil, status.Errorf(codes.Internal, "the number of deploy targets in the piped plugin config should be the same as the ones set on the deployment")
+	}
+
+	return executeStage(ctx, s.base, &s.config, deployTargets, client, request, s.logger) // TODO: pass the deployTargets
 }
 
 // StagePluginServiceServer is the gRPC server that handles requests from the piped.
--- a/pkg/plugin/sdk/deployment.go
+++ b/pkg/plugin/sdk/deployment.go
@@ -214,18 +214,21 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
 
 	deployTargets := make([]*DeployTarget[DeployTargetConfig], 0, len(dtNames))
 	for _, name := range dtNames {
-		if dt := s.commonFields.config.FindDeployTarget(name); dt != nil {
-			var sdkDt DeployTargetConfig
-			if err := json.Unmarshal(dt.Config, &sdkDt); err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to unmarshal deploy target config: %v", err)
-			}
-
-			deployTargets = append(deployTargets, &DeployTarget[DeployTargetConfig]{
-				Name:   name,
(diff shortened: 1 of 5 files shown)


pkg/plugin/sdk/deployment.go
```diff
@@ -206,7 +206,33 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
 		stageID:       request.GetInput().GetStage().GetId(),
 		logPersister:  lp,
 	}
-	return executeStage(ctx, s.base, &s.config, nil, client, request, s.logger) // TODO: pass the deployTargets
+
+	dtNames, err := request.GetInput().GetDeployment().GetDeployTargets(s.commonFields.config.Name)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get deploy targets for plugin %s: %v", s.commonFields.config.Name, err)
+	}
```
Warashi, line 213: The plugin which doesn't have deploy target configurations such as WAIT will return an error in this if statement, and it's not expected behavior.

ref; https://github.com/pipe-cd/pipecd/blob/4afa3e61937502fd95614889cbd8294dd794fcd6/pkg/model/deployment.go#L217-L224

pkg/plugin/sdk/deployment.go
```diff
@@ -206,7 +206,33 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
...
+			deployTargets = append(deployTargets, &DeployTarget[DeployTargetConfig]{
+				Name:   name,
+				Labels: dt.Labels,
+				Config: sdkDt,
+			})
+		}
+	}
+
+	if len(dtNames) != len(deployTargets) {
+		return nil, status.Errorf(codes.Internal, "the number of deploy targets in the piped plugin config should be the same as the ones set on the deployment")
```
t-kikuc, line 232: [nits]
Let's add `len(dtNames)` and `len(deployTargets)` in the message to make it easier to troubleshoot.

pkg/plugin/sdk/deployment.go
```diff
@@ -206,7 +206,33 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
...
+			if err := json.Unmarshal(dt.Config, &sdkDt); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to unmarshal deploy target config: %v", err)
+			}
+
+			deployTargets = append(deployTargets, &DeployTarget[DeployTargetConfig]{
+				Name:   name,
+				Labels: dt.Labels,
+				Config: sdkDt,
+			})
+		}
```
Warashi, line 228: [IMO]
We can use the early-return pattern with the error containing the deployTarget name that is not found when `dt == nil` .

Warashi (commented):
Thank you.
I think the spec you're trying to implement is good, and I've commented on how you've implemented it.

pkg/plugin/sdk/deployment.go
```diff
@@ -206,7 +206,33 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
 		stageID:       request.GetInput().GetStage().GetId(),
 		logPersister:  lp,
 	}
-	return executeStage(ctx, s.base, &s.config, nil, client, request, s.logger) // TODO: pass the deployTargets
+
+	dtNames, err := request.GetInput().GetDeployment().GetDeployTargets(s.commonFields.config.Name)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get deploy targets for plugin %s: %v", s.commonFields.config.Name, err)
```
t-kikuc, line 212: [nits]
[nit]
I wanna clarify which one we mention,
- the deployTraget was not found in a piped config
- the deployTarget was not set in the request(deployment)

```suggestion
		return nil, status.Errorf(codes.Internal, "failed to get deploy targets from the deployment for plugin %s: %v", s.commonFields.config.Name, err)
```

pkg/plugin/sdk/deployment.go
```diff
@@ -206,7 +206,33 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
...
+	}
+
+	deployTargets := make([]*DeployTarget[DeployTargetConfig], 0, len(dtNames))
+	for _, name := range dtNames {
+		if dt := s.commonFields.config.FindDeployTarget(name); dt != nil {
+			var sdkDt DeployTargetConfig
+			if err := json.Unmarshal(dt.Config, &sdkDt); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to unmarshal deploy target config: %v", err)
+			}
```
t-kikuc, line 221: [IMO] If possible, it's better to cache DeployTargetConfigs in the server instead of repeating unmarshaling.

t-kikuc (commented):
Thank you, I left some nits.

pkg/plugin/sdk/deployment.go
ffjlabo: @Warashi 
Thanks.
I fixed the method logic to allow the case of no deploy target. We can decide whether it is an error or not in the plugin logic from now.
[0e167c0](https://github.com/pipe-cd/pipecd/pull/5616/commits/0e167c0bc0bf7520bb7679e8c322cd6863c04373)

pkg/plugin/sdk/deployment.go
ffjlabo: Fixed on [8173d91](https://github.com/pipe-cd/pipecd/pull/5616/commits/8173d91784e699dcdfdde363e824df4e828f3a5b)

pkg/plugin/sdk/deployment.go
ffjlabo: Fixed on [fbb133e](https://github.com/pipe-cd/pipecd/pull/5616/commits/fbb133e012a69c6715bbced9af4a1bd07174d611)

pkg/plugin/sdk/deployment.go
ffjlabo: Fixed on [fc90b4c](https://github.com/pipe-cd/pipecd/pull/5616/commits/fc90b4c0b03e5e6621c38a38820e46f614b796de)

pkg/plugin/sdk/deployment.go
ffjlabo: @t-kikuc 
Thanks, that sounds nice.
I added a TODO comment for it. I think it would be nice to focus on passing the deploy targets in this PR. Is it OK?
[03519ad](https://github.com/pipe-cd/pipecd/pull/5616/commits/03519addc7d85c8e97a86f4b3c61e42e526e94df)

pkg/plugin/sdk/deployment.go
```diff
@@ -206,7 +206,34 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig]) ExecuteStage
...
 	}
-	return executeStage(ctx, s.base, &s.config, nil, client, request, s.logger) // TODO: pass the deployTargets
+
+	// Get the deploy targets set on the deployment from the piped plugin config.
+	dtNames := request.GetInput().GetDeployment().GetDeployTargets(s.commonFields.config.Name)
+	deployTargets := make([]*DeployTarget[DeployTargetConfig], 0, len(dtNames))
+	for _, name := range dtNames {
+		dt := s.commonFields.config.FindDeployTarget(name)
+		if dt == nil {
+			continue
```
Warashi, line 216: [IMO]
How about returning an error here? If we return here, we can remove the `if len(dtNames) != len(deployTargets)` block after the for loop.

pkg/plugin/sdk/deployment.go
t-kikuc: Thank you! TODO is best for this case!  

We don't have to do that now.

pkg/plugin/sdk/deployment.go
ffjlabo: @Warashi 
Ah, finally I got your point.
Fixed on [1efc476](https://github.com/pipe-cd/pipecd/pull/5616/commits/1efc476f2acab76286b8cfc1354a2ee0c0e4a74f)

Warashi (approved):
LGTM

t-kikuc (approved):

PR #5618, opened 2025-03-04
Add deploy_targets field to GetLivestateRequest message

Warashi:
**What this PR does**:

as title

**Why we need it**:

We need the deploy target information to retrieve live resource states.

Part of #5363 

**Does this PR introduce a user-facing change?**: No.


pkg/plugin/api/v1alpha1/livestate/api.proto
```diff
@@ -30,6 +30,7 @@ service LivestateService {
 message GetLivestateRequest {
     string application_id = 1;
     grpc.plugin.common.v1alpha1.DeploymentSource deploy_source = 2;
+    repeated string deploy_targets = 3;
```
t-kikuc, line 33: [ask] Are these fields required or not?

t-kikuc (commented):
Seems good! I wanna ask you one tiny point!

pkg/plugin/api/v1alpha1/livestate/api.proto
Warashi: I think the application ID seems required, but the others are plugin-dependent.

t-kikuc: Thank you, I got it

t-kikuc (approved):

pkg/plugin/api/v1alpha1/livestate/api.proto
khanhtc1202: IMO, all should be required since this is livestate, so common plugins (like wait, wait approval) would not use this; only platform-based plugins use these fields, so should mark as required, wdyt?

Warashi: I agree with you about the plugins I can imagine now. But I think we should not restrict the possibility of third-party plugins that don't use these fields.

khanhtc1202 (approved):
👌

PR #5621, opened 2025-03-04
Add toolRegistry as a common field for the Client in the SDK

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

We need it to install tools to load manifests and deploy them to the platform for the platform-specific plugin.
(e.g. k8s plugin's [DetermineStrategy](https://github.com/pipe-cd/pipecd/blob/6cf50973ea81bf963cca1f72b7efe0ad0546d875/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go#L114), [DetermineVersion](https://github.com/pipe-cd/pipecd/blob/6cf50973ea81bf963cca1f72b7efe0ad0546d875/pkg/app/pipedv1/plugin/kubernetes/deployment/server.go#L142), [ExecuteStage](https://github.com/pipe-cd/pipecd/blob/6cf50973ea81bf963cca1f72b7efe0ad0546d875/pkg/app/pipedv1/plugin/kubernetes/deployment/sync.go#L97))

Also, we need it for the stage plugin to execute any tool on the generic stage.

It is better to set it on the SDK side because it requires a piped service client.

Part of https://github.com/pipe-cd/pipecd/issues/5530

### DIFF
--- a/pkg/plugin/sdk/client.go
+++ b/pkg/plugin/sdk/client.go
@@ -45,6 +45,7 @@ type Client struct {
 	logPersister StageLogPersister
 
 	// toolRegistry is used to install and get the path of the tools used in the plugin.
+	// TODO: We should consider installing the tools in other way.
 	toolRegistry *toolregistry.ToolRegistry
 }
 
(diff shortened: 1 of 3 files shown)


Warashi (approved):
Let's try

t-kikuc (approved):
Let's try

Issue #5444, opened 2024-12-23
Deployment Tracing feature for better link between CI & CD


khanhtc1202:
**What would you like to be added**:

We want a better way to link CI and CD (using PipeCD). The current PipeCD working context focuses mostly on manifest repo. Users need a way to actually understand which change in the source code repo triggers which deployment(s).

**Why is this needed**:

By introducing Event Context (ref: https://github.com/pipe-cd/pipecd/pull/5295), users now can have a link from deployment to the code changed by visiting the manifest changed commit, with the information provided to PipeCD via pipectl event register command. But it's just one way, from the CD to the CI, for tracking back; it would be better to have a way from CI to CD when devs (who change the source code) know the source code commit and want to find the change reflected deployment(s).

PR #5625, opened 2025-03-05
Add trigger commit hash to deployment model

khanhtc1202:
**What this PR does**:

Add value for `DeploymentTraceCommitHash` to deployment model.

**Why we need it**:

This commit will be used as the key to link between the DeploymentTrace model and the Deployment model. By referring to this key, we can query all deployments, which should be grouped under one DeploymentTrace (aka. Group deployments triggered by the same event)

Part of #5444

### DIFF
--- a/pkg/git/commit_test.go
+++ b/pkg/git/commit_test.go
@@ -78,3 +78,57 @@ This PR was merged by Kapetanios.`,
 	})
 	assert.Equal(t, expected, commits)
 }
+
+func TestGetTrailerValueByKey(t *testing.T) {
+	testcases := []struct {
+		name string
+		body string
+		key  string
+		want string
+	}{
+		{
+			name: "single line trailer",
+			body: "Some commit message\nKey: value",
+			key:  "Key",
+			want: "value",
+		},
+		{
+			name: "multiple line trailers",
+			body: "Some commit message\nKey1: value1\nKey2: value2",
+			key:  "Key2",
+			want: "value2",
+		},
+		{
+			name: "trailer with spaces",
+			body: "Some commit message\nKey:    spaced value    ",
+			key:  "Key",
+			want: "spaced value",
+		},
+		{
+			name: "no matching trailer",
+			body: "Some commit message\nOtherKey: value",
+			key:  "Key",
+			want: "",
+		},
+		{
+			name: "empty body",
+			body: "",
+			key:  "Key",
+			want: "",
+		},
+		{
+			name: "multiline value",
+			body: "Some commit message\nKey: first line\n second line",
+			key:  "Key",
+			want: "first line",
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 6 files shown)


pkg/git/commit_test.go
```diff
@@ -78,3 +78,57 @@ This PR was merged by Kapetanios.`,
 	})
 	assert.Equal(t, expected, commits)
 }
+
+func TestGetTrailerValueByKey(t *testing.T) {
+	testcases := []struct {
```
t-kikuc, line 83: [nit] 
```suggestion
        t.Parallel()
	testcases := []struct {
```

pkg/git/commit_test.go
```diff
@@ -78,3 +78,57 @@ This PR was merged by Kapetanios.`,
...
+			name: "multiline value",
+			body: "Some commit message\nKey: first line\n second line",
+			key:  "Key",
+			want: "first line",
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			c := &Commit{Body: tc.body}
```
t-kikuc, line 129: [nit]
```suggestion
                         t.Parallel()
			c := &Commit{Body: tc.body}
```

pkg/model/deployment_trace.go
```diff
@@ -14,6 +14,12 @@
 
 package model
 
+const (
+	// The key to store the commit hash that triggers the event (in EventWatcher flow).
+	// It will be used as key to store the commit hash as metadata in the commit boy.
```
t-kikuc, line 19: [nit] to simplify + fix typo
```suggestion
	// The key to store the commit hash that triggers the event (in EventWatcher flow) as metadata in the commit body.
```

pkg/model/deployment_trace.go
```diff
@@ -14,6 +14,12 @@
...
+	TraceTriggerCommitHashKey = "Trace-Trigger-Commit-Hash"
```
t-kikuc, line 20: [IMO] Use some prefixes like `pipecd.dev.Trace-Trigger-Commit-Hash` to avoid unintended override. (_Not mandatory for now_)

I think we should do so in other features too.

t-kikuc (commented):
Almost LGTM, I left some nits

Warashi (commented):
Thank you!
I commented on the git trailer key format.

t-kikuc (approved):
Let's try!

pkg/model/deployment_trace.go
```diff
@@ -14,6 +14,11 @@
 
 package model
 
+const (
+	// The key to store the commit hash that triggers the event (in EventWatcher flow) as metadata in the commit body.
+	TraceTriggerCommitHashKey = "pipecd.dev.Trace-Trigger-Commit-Hash"
```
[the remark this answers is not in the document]
khanhtc1202: Thanks, updated 👍

ffjlabo (approved):
It seems that @Warashi 's suggestion is fixed.

PR #5627, opened 2025-03-06
Fix toolregistrytest by using fake

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

It is better to use the same type in both the cases used in actual logic and test like `zaptest.`
https://github.com/uber-go/zap/blob/master/zaptest/logger.go#L77

The users don't need to define an additional interface.

Fixes #


pkg/plugin/toolregistry/toolregistrytest/toolregistrytest.go
```diff
@@ -38,97 +44,76 @@ type templateValues struct {
...
-	return &ToolRegistry{
-		testingT: t,
-		tmpDir:   tmpDir,
-	}, nil
-}
-
-func (r *ToolRegistry) newTmpDir() (string, error) {
-	return os.MkdirTemp(r.tmpDir, "")
+type fakeClient struct {
+	pipedservicetest.MockPluginServiceClient
```
ffjlabo, line 48: Use it to reduce the cost of maintaining fake implementation.

Warashi: [IMO]
Using gomock only for this embedding is a bit complex.
How about embedding service.PluginServiceClient instead of generated mock?
https://github.com/pipe-cd/pipecd/blob/a252de8ea9a09e75978472008d9ac3178ad033f1/pkg/plugin/pipedservice/service_grpc.pb.go#L24

Warashi (commented):
I agree to use fake in the toolregistrytest.
I commented on how we prepare the fake.

pkg/plugin/toolregistry/toolregistrytest/toolregistrytest.go
ffjlabo: Thanks, sounds good!
I thought we needed the actual instance to override the specific method. 🙏 
It works fine.

[1d8bea3](https://github.com/pipe-cd/pipecd/pull/5627/commits/1d8bea39927570ebc7f43ddad7378a052004b1a3)

Warashi (approved):
LGTM

khanhtc1202 (approved):
Nice 👍

PR #5631, opened 2025-03-06
Implement ListDeploymentTrace rpc for Web

khanhtc1202:
**What this PR does**:

Implementation for ListDeploymentTrace rpc for web client.

**Why we need it**:

Part of #5444


pkg/app/server/grpcapi/web_api.go
```diff
@@ -973,59 +978,60 @@ func (a *WebAPI) ListDeploymentTraces(ctx context.Context, req *webservice.ListD
...
-					CommitMessage:   "Message mocked here",
-					CommitTimestamp: 123,
-					Author:          "author",
-				},
-				Deployments: deployments,
-			},
-		},
-		Cursor: "cursor",
+		}
+		deployments, _, err := a.deploymentStore.List(ctx, opts)
```
t-kikuc, line 1022: I understand that this is like an N+1 query, but pageSize will reduce the problem.

t-kikuc (approved):
LGTM! Let's try!

pkg/app/server/grpcapi/web_api.go
khanhtc1202: Yes, it is. I thought of making list deployment based on deployment trace commit hash rpc independently but maybe it's just over killed. So let's make this for now, if we need that when we faced a performance issue or such, we will implement that later 👍

t-kikuc: 100% agree 👍

ffjlabo (approved):
LGTM

PR #5613, opened 2025-02-28
[WaitApproval Plugin] Add NotifyEvent methods 

t-kikuc:
**What this PR does**:

I implemented NotifyEvent methods in both piped and plugin side.

**Why we need it**:

The WaitApproval plugin needs to: 

- notify approvers when the stage starts (`NotificationEventDeploymentWaitApproval`)
- notify when approval is completed  (`NotificationEventDeploymentApproved`)

cf. https://github.com/pipe-cd/pipecd/blob/2ce020e567f97c767675a5769aefe5d9142cdf13/pkg/app/piped/executor/waitapproval/waitapproval.go#L112-L143

**Which issue(s) this PR fixes**:

Part of #5367

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #5632, opened 2025-03-06
[WaitApproval] Notify events of Approved/WaitApproval in Piped side

t-kikuc:
**What this PR does**:

Instead of triggering the events exactly at start/end of the approval, trigger them at start/end of the stage.

- `WaitApproval` Event: Trigger at the same timing as `StageStarted`
- `Approved` Event: Trigger at the same timing as `StageSucceeded`
    - `approver` field will be passed from a plugin **indirectly via metadataStore**

**Why we need it**:

Instead of #5613, this PR **does not add RPCs** for simplicity of the interface.

Part of #5367


pkg/app/pipedv1/controller/scheduler.go
```diff
@@ -813,12 +817,45 @@ func (s *scheduler) notifyStageStartEvent(stage *model.PipelineStage) {
...
 func (s *scheduler) notifyStageEndEvent(stage *model.PipelineStage, result model.StageStatus) {
 	switch result {
 	case model.StageStatus_STAGE_SUCCESS, model.StageStatus_STAGE_EXITED: // Exit stage is treated as success.
+
+		if stage.AvailableOperation == model.ManualOperation_MANUAL_OPERATION_APPROVE {
+			users, groups, err := s.getApplicationNotificationMentions(model.NotificationEventType_EVENT_DEPLOYMENT_APPROVED)
+			if err != nil {
+				s.logger.Error("failed to get the list of users", zap.Error(err))
+			}
+			approver, _ := s.metadataStore.StageGet(stage.Id, "pipecd.dev.Approver") // TODO: Make the key to be a const.
```
t-kikuc, line 847: [help wanted] 
How about not notifying if 'approver' is empty? 
For the situation like the stage ended due to reasons other than approvals.

PR #5636, opened 2025-03-07
fix: Use the specified version of Helm when templating with Kustomize

### LINKED ISSUE #5635: specifying Helm version doesn't seem to be used in plan-preview
**What happened**:
plan-preview currently fails if we use config/templates described in the "How to reproduce it" section.

The error itself is:
```
Reason: failed while planning, unable to run kustomize template: exit status 1: Error: Error: template: datadog/templates/gke_autopilot_allowlist_synchronizer.yaml:1:48: executing "datadog/templates/gke_autopilot_allowlist_synchronizer.yaml" at <include "gke-autopilot-workloadallowlists-enabled" .>: error calling include: template: datadog/templates/_helpers.tpl:74:23: executing "gke-autopilot-workloadallowlists-enabled" at <len $nodes>: error calling len: reflect: call of reflect.Value.Type on zero Value
```

This is caused by older Helm, like v3.8.0. Helm v3.17.0, for instance, treats it without any failure.

The mysterious part of this problem is it still occurs even if we have a config like `helmVersion: 3.17.0` in pipecd.yaml. We thought it was used in plan-preview.

I thought https://github.com/pipe-cd/pipecd/issues/4382 fixed the problem but for some reason our plan preview is still broken. Maybe we've misconfigured something or implementation still needs a tweak.

(note: we haven't checked if this problem only occur on plan-preview or not.)

**What you expected to happen**: should succeed plan-preview (and any other procedure)

**How to reproduce it**:
1. prepare a PipeCD config like:
  ```yaml
kind: KubernetesApp
spec:
  name: datadog-agent
  input:
    kustomizeVersion: 5.6.0
    kustomizeOptions:
      enable-helm: ""
    helmVersion: 3.17.0
  ```
2. prepare a Kustomize config like:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
helmCharts:
  - name: datadog
    releaseName: datadog-agent
    repo: https://helm.datadoghq.com
    version: 3.102.0
```
3. Run `pipectl plan-preview` or trigger the plan-preview action against it

**Environment**:
- `piped` version: v0.50.2
- `control-plane` version: v0.50.2
- Others:
  - datadog-agent chart: v3.98.0 or higher
  - Kustomize: v5.6.0
  - Helm: v3.17.0

JohnTitor:
**What this PR does**: Pass the `--helm-command` flag with the installed Helm path when kustomize-templating.

**Why we need it**: Without this, Kustomize cannot find the specified version of Helm when templating.

Fixes #5635

**Does this PR introduce a user-facing change?**: yes

- **How are users affected by this change**: User can now use the desired version of Helm when Kustomize-templating.


pkg/app/piped/platformprovider/kubernetes/testdata/testkustomize-with-helm/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+# This would fail if it uses the default version (3.8.2).
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+helmCharts:
+  - name: datadog
+    releaseName: datadog-agent
+    repo: https://helm.datadoghq.com
+    version: 3.102.0
```
JohnTitor, line 11: Perhaps we could come up with a minimal testcase and put it on local, would like to leave the decision to the maintainers.

pkg/app/piped/platformprovider/kubernetes/kustomize_test.go
```diff
@@ -36,11 +36,36 @@ func TestKustomizeTemplate(t *testing.T) {
 
 	kustomizePath, _, err := toolregistry.DefaultRegistry().Kustomize(ctx, "")
 	require.NoError(t, err)
+	helmPath, _, err := toolregistry.DefaultRegistry().Helm(ctx, "")
```
Warashi, line 39: Please check the error is nil?
```suggestion
	helmPath, _, err := toolregistry.DefaultRegistry().Helm(ctx, "")
	require.NoError(t, err)
```

Warashi (commented):
Thank you! Almost LGTM.
There is a lint error, so please fix it.

pkg/app/piped/platformprovider/kubernetes/kustomize_test.go
JohnTitor: Ugh, fixed in [`e8860e1` (#5636)](https://github.com/pipe-cd/pipecd/pull/5636/commits/e8860e1df88ab4271cbbfcdbc9ede76456b628af)!

pkg/app/piped/platformprovider/kubernetes/kustomize.go
```diff
@@ -37,12 +37,19 @@ func NewKustomize(version, path string, logger *zap.Logger) *Kustomize {
...
+func (c *Kustomize) Template(ctx context.Context, appName, appDir string, opts map[string]string, helm *Helm) (string, error) {
 	args := []string{
 		"build",
 		".",
 	}
 
+	// Pass the Helm command path to kustomize to use the specified version of Helm.
+	// Unconditionally adding this flag as it's unharmful when Helm is not used.
+	// Note: It's only available on Kustomize v4.1.0 and higher.
+	if c.version >= "4.1.0" && helm != nil {
```
Warashi, line 49: I apologize for overlooking this version comparison issue.
We cannot compare semantic versions by simply comparing strings. For instance, versions greater than or equal to 10 are incorrectly evaluated as less than 9.
Could you please implement a feature to compare semantic versions and include corresponding tests?

Warashi (commented):
Sorry, I overlooked the version comparison issue.

pkg/app/piped/platformprovider/kubernetes/kustomize.go
JohnTitor: Ah, right.
Fixed and added tests in [`84bfe4d` (#5636)](https://github.com/pipe-cd/pipecd/pull/5636/commits/84bfe4db299c1cdae3621608acda94b83da78db5)

Warashi (approved):
Thank you! LGTM

PR #5648, opened 2025-03-12
Fix possible nil error on commit hash assign in case the contexts of event is nil

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

If the event's context is nil, the trailers value will be nil, and leads to nil assignment when we try to add value to TraceTriggerCommitHashKey

Follow #5625 


ffjlabo (approved):
LGTM

PR #5649, opened 2025-03-12
Update pipedv1 trigger to support adding trace commit hash to deployment model

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

We must pass this value to the deployment model to use the Deployment Trace feature. This change is the same as we did in pipedv0 logic (ref: [pipedv0/trigger/deployment](https://github.com/pipe-cd/pipecd/pull/5625/files#diff-9c94e1765f50c81cc1e04535ec50d0f16fd6459fb06402713be9b2eb68a0e3cb))

Follow #5625


Warashi (approved):
LGTM

ffjlabo (approved):

PR #5622, opened 2025-03-04
Create and add deployment trace obj to datastore on event registering flow

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

While registering an event by pipette, we will create a DeploymentTrace object and store it in the datastore. The stored DeploymentTrace object contains information about the commit that triggered the event, which will be used in the DeploymentTrace feature UI.
Also, I updated the Event model to store trigger_commit_hash, which is the commit hash of the commit that triggers this event. This information will be used while handling the event (EventWatcher handling flow)

**Which issue(s) this PR fixes**:

Part of #5444

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #5651, opened 2025-03-12
Add deployment trace model encode/decode logic for MySQL datastore

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

Without this, in case of using MySQL as the datastore of PipeCD control plane, the datastore backend doesn't understand how to encode model object to store to MySQL table 😓 

We will get the below error

```
caller: "mysql/mysql.go:159"
error: "*model.DeploymentTrace is not supported"
kind: "DeploymentTrace"
logger: "pipecd.server"
message: "failed to create entity: failed to encode json data"
```

Follow PR #5622


ffjlabo (approved):
👍

Warashi (approved):
LGTM

PR #5660, opened 2025-03-13
Fix Ui Deployment tracing with real api deployment traces

hongky-1994:
**What this PR does**:
- Fix timestamp of deployment trace need * 1000 before convert.
- Add place holder "No deployment triggered" when no deployment

**Why we need it**:

Part of #5444 


khanhtc1202 (approved):
👌

Warashi (approved):
LGTM

PR #5207, opened 2024-09-11
[Snyk] Security upgrade react-router-dom from 5.3.4 to 6.0.0

caniszczyk:
snyk-top-banner

#### Snyk changed the following file(s):

- `web/package.json`
- `web/yarn.lock`

#### Note for [zero-installs](https://yarnpkg.com/features/zero-installs) users

If you are using the Yarn feature [zero-installs](https://yarnpkg.com/features/zero-installs) that was introduced in Yarn V2, note that this PR does not update the `.yarn/cache/` directory meaning this code cannot be pulled and immediately developed on as one would expect for a zero-install project - you will need to run `yarn` to update the contents of the `./yarn/cache` directory.
If you are not using zero-install you can ignore this as your flow should likely be unchanged.

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
medium severity | Regular Expression Denial of Service (ReDoS) <br/>[SNYK-JS-PATHTOREGEXP-7925106](https://snyk.io/vuln/SNYK-JS-PATHTOREGEXP-7925106) | &nbsp;&nbsp;**738**&nbsp;&nbsp; 

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/cncf-projects-team/project/86c20a19-dd69-4bda-80ba-b0e44400c335?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates)
🛠 [Adjust project settings](https://app.snyk.io/org/cncf-projects-team/project/86c20a19-dd69-4bda-80ba-b0e44400c335?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://support.snyk.io/hc/en-us/articles/360003891078-Snyk-patches-to-fix-vulnerabilities)

---

[...]

PR #5661, opened 2025-03-13
Fix redirect login page when logout, and applications page when login

hongky-1994:
**What this PR does**:
- Redirect to `Login page` when user is `not logged in` and `No routes matched url location`.
- Redirect to `Application page` when user is `logged in` and `No routes matched url location`.Ï

**Why we need it**:
- Fix UX behavior
- Fix bug from migration react-router-dom https://github.com/pipe-cd/pipecd/pull/5207

- None

**Does this PR introduce a user-facing change?**: none


khanhtc1202 (approved):
👌

Warashi (approved):
🚀

PR #5666, opened 2025-03-14
Set limit for gh pr list to 100 to avoid cherrypick limit

khanhtc1202:
**What this PR does**:

**Why we need it**:

The default value for the `gh pr list` command is 30, which is too small in case of 3-week cycle for a patch release now. We set it to 100 to avoid the error cherry-pick failed to apply.

Fixes #


Warashi (approved):
👍🏻

ffjlabo (approved):
👍

PR #5668, opened 2025-03-17
Add more information to the GetLivestateRequest

Warashi:
**What this PR does**:

This PR adds the information below to the GetLivestateRequest
- Piped ID
- Application Name

Additionally, this PR changes the tag number of the protobuf. Since we haven't released any plugins, we can reorder these fields.

**Why we need it**:

We need this information to load the k8s manifest consistently.

Part of #4980 #5363


ffjlabo (approved):
LGTM

pkg/plugin/api/v1alpha1/livestate/api.proto
```diff
@@ -27,13 +27,24 @@ service LivestateService {
...
-    grpc.plugin.common.v1alpha1.DeploymentSource deploy_source = 2;
-    repeated string deploy_targets = 3;
+    // The id of the piped
+    string piped_id = 1;
+    // The id of the application
+    string application_id = 2;
+    // The name of the application
+    string application_name = 3;
+    // The source of the deployment
+    grpc.plugin.common.v1alpha1.DeploymentSource deploy_source = 4;
```
t-kikuc, line 39: [ask] `deploy_source` is required for k8s, right? 
we don't need to use it for other platforms, right?

Warashi: We need deploy_source to calculate the sync state (drift detection feature.)
So, we have to use it for other platforms supporting the drift detection feature.

t-kikuc: I got it!!

The "livestate" itself doesn't need it, DriftDetection needs it.

PR #5682, opened 2025-03-19
Add application livestate version in livestatereporter for pipedv1

ffjlabo:
**What this PR does**:

as title.
It is required for the pipedv0, especially k8s. It is used to update livestate by using the diff from an older version.
Currently, the logic is used only for k8s livestate, and the others are not considered it.

https://github.com/pipe-cd/pipecd/blob/7a07070328e3d36a70c303f3f0c6a9b1f6076cfc/pkg/app/server/applicationlivestatestore/store.go#L93C1-L130C2

**Why we need it**:

Currently, we don't use it after pipedv1, but we use it in the Control Plane to update k8s livestate for apps managed by pipedv0.
So just added to pass the validation.

Fixes #


Warashi (approved):
LGTM

We can choose a method that removes the validation, but I think it's better to leave the timestamp field so we can see how old the live state is.

t-kikuc (approved):

PR #5683, opened 2025-03-19
[K8s Plugin] Add e2e tests for K8S_ROLLBACK stage and fix the bug

Warashi:
**What this PR does**:

- add simple E2E test for K8S_ROLLBACK stage
- add a variant label to manifests when executing the K8S_ROLLBACK stage
    - to make consistent with the K8S_SYNC stage

**Why we need it**:

- to check and keep the k8s plugin's behavior as the test

Part of #4980 


t-kikuc (approved):
LGTM

If needed in the future, let's find a way to split simple unit tests and e2e tests for faster testing.

ffjlabo (approved):
LGTM

PR #5691, opened 2025-03-25
Enable livestate reporter for pipedv1

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

We need it to send the livestate for the app managed by pipedv1.

Part of #5363


Warashi (approved):
Thank you

PR #5693, opened 2025-03-25
Embed k8s deployment and livestate plugins into one structure in main

Warashi:
**What this PR does**:

as title

**Why we need it**:

We don't want to define the same Name and Version method in both the deployment and livestate packages.

Part of #4980

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
@@ -27,8 +27,8 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/plugin/sdk"
 )
 
-// Plugin implements the sdk.DeploymentPlugin interface.
-type Plugin struct {
+// DeploymentPlugin implements the sdk.DeploymentPlugin interface without Name and Version.
+type DeploymentPlugin struct {
 }
 
 type loader interface {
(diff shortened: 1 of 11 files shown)


pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
```diff
@@ -27,47 +27,29 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/plugin/sdk"
 )
 
-// Plugin implements the sdk.DeploymentPlugin interface.
-type Plugin struct {
+// DeploymentPlugin implements the sdk.DeploymentPlugin interface without Name and Version.
+type DeploymentPlugin struct {
```
Warashi, line 31: We cannot embed multiple types with the same name, so I renamed this.

pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
```diff
@@ -27,47 +27,29 @@ import (
...
 type loader interface {
 	// LoadManifests renders and loads all manifests for application.
 	LoadManifests(ctx context.Context, input provider.LoaderInput) ([]provider.Manifest, error)
 }
 
-type toolRegistry interface {
-	Kubectl(ctx context.Context, version string) (string, error)
-	Kustomize(ctx context.Context, version string) (string, error)
-	Helm(ctx context.Context, version string) (string, error)
-}
```
Warashi, line 43: This is unused

pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
```diff
@@ -27,47 +27,29 @@ import (
...
-
-var _ sdk.DeploymentPlugin[sdk.ConfigNone, kubeconfig.KubernetesDeployTargetConfig] = (*Plugin)(nil)
```
Warashi, line 45: The type `DeploymentPlugin` in this package no longer implements `sdk.DeploymentPlugin` because `sdk.DeploymentPlugin` requires `Name` and `Version` methods. So I removed this.

pkg/app/pipedv1/plugin/kubernetes/main.go
```diff
@@ -22,9 +22,23 @@ import (
...
+	livestate.LivestatePlugin
+}
+
+func (p *Plugin) Name() string {
+	return "kubernetes"
+}
+
+func (p *Plugin) Version() string {
+	return "0.0.1" // TODO: update this value
+}
```
Warashi, line 36: Now we can define Name and Version only here.

t-kikuc (approved):

pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
ffjlabo: [IMO] If possible, it would be nice to keep it as is to clarify what should be implemented.  Maybe we need to consider the the interface and SDK.

Warashi: I noticed there is a better SDK interface for registering a plugin.
I'll try to change the SDK.

I imagine something like the one below.
```go
sdk.RegisterPlugin("kubernetes", "0.0.1", sdk.WithDeploymentPlugin(&deployment.Plugin{}), sdk.WithLivestatePlugin(&livestate.Plugin{}))
```

ffjlabo: Thank you! it is worth trying 👍

Warashi: I implemented this in another PR, #5698.
After merging it, I'll rewrite this PR.

Warashi:
The better solution will be done by #5701

pkg/app/pipedv1/plugin/kubernetes/deployment/plugin.go
Warashi: I thought sending another PR to use the new SDK interface would be better, so I closed this.
The new implementation is #5701

PR #5698, opened 2025-03-26
[Plugin SDK] Implement another way to create plugin

Warashi:
**What this PR does**:

as title

**Why we need it**:

In #5693, I wanted to remove the duplicated `Name()` and `Version()` definitions in the Kubernetes plugin.
It's easy to remove the duplication in the k8s plugin only, but it's better to provide a way to implement plugins without requiring the `Name()` and `Version()` methods in multiple interfaces.
I noticed that we can provide the way that requires name and version as positional arguments when creating a plugin.
In this way, the plugin implementations are passed as the Functional Options Pattern to the plugin constructor function.
Additionally, we can implement assertions to the constructor function to provide helpful errors.

Part of #5530 

### DIFF
--- a/pkg/plugin/sdk/plugin.go
+++ b/pkg/plugin/sdk/plugin.go
@@ -1,0 +1,314 @@
+// Copyright 2025 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


ffjlabo (approved):
LGTM

t-kikuc (approved):

PR #5703, opened 2025-03-26
[Plugin SDK] Remove old interfaces and refactor the SDK

Warashi:
**What this PR does**:

- Remove the older SDK interface, as well as the Register and the Run functions.
- move types commonly used in both of the deployment.go and the livestate.go into the plugin.go
- add error logs when something goes wrong

**Why we need it**:

We introduced a new way to implement plugins in #5698 and migrated existing plugins.
So we can remove the old interfaces.

Follows #5698 
Part of #5530 


pkg/plugin/sdk/plugin.go
```diff
@@ -156,27 +195,18 @@ func (p *Plugin_[Config, DeployTargetConfig]) command() *cobra.Command {
...
-	if p.stagePlugin != nil {
-		stagePluginServiceServer = &StagePluginServiceServer[Config, DeployTargetConfig]{base: p.stagePlugin}
-	}
-
-	if p.deploymentPlugin != nil {
-		deploymentPluginServiceServer = &DeploymentPluginServiceServer[Config, DeployTargetConfig]{base: p.deploymentPlugin}
-	}
-
-	if p.livestatePlugin != nil {
-		livestatePluginServiceServer = &LivestatePluginServer[Config, DeployTargetConfig]{base: p.livestatePlugin}
```
Warashi, line 179: I noticed that we can do these in around L273

t-kikuc (approved):

ffjlabo (approved):
👍

PR #5704, opened 2025-03-26
Remove Name and Version methods from plugins

Warashi:
**What this PR does**:

as title

**Why we need it**:

These are requirements of the old SDK interfaces so that we can delete them.

Follows #5703 


khanhtc1202 (approved):
LGTM

t-kikuc (approved):

PR #5716, opened 2025-04-02
add: temporary volume option

na-ga:
**What this PR does**:

Added option to use PVC as temporary volume

**Why we need it**:

I need a workaround for a disk full problem caused by a temporary file generated by piped

Fixes #

- **How are users affected by this change**: Nil
- **Is this breaking change**: Nil


ffjlabo (approved):
Thank you for the PR!
I test it on the kind cluster and checked the PVC are attached to the piped pod.

Warashi (approved):
Thank you! LGTM.

PR #5717, opened 2025-04-02
Fix rolling back deployments are ignored on piped restarted event

khanhtc1202:
**What this PR does**:

Add rolling back deployment as running deployment on deploymentstore sync.

**Why we need it**:

Without this change, when piped is restarted, the deployments at the ROLLING_BACK stage are not listed as running deployments in the deployment store. Thus, restarted piped will ignore these deployments.

Fixes #


Warashi (approved):
Thank you

ffjlabo (approved):
👍

PR #5722, opened 2025-04-03
Do treeless clone for git clone to improve performance

Warashi:
**What this PR does**:

as title

**Why we need it**:

I want to reduce file/network IO.

As before, I implemented git partial cloning in #5412.
This is reverted due to a failure in the event watcher, caused by a failure of git repository cloning from the local cached repository.
This cloning failure occurs because the local cached repository doesn't have the necessary git object because it's cloned "partially."
So, I modified the `CopyToModify` method to clone directly from the original remote repository, such as GitHub, and retry the partial cloning performance tuning.
I tested normal deployment and event watcher behavior on my local, and these ran successfully.

Fixes #

### DIFF
--- a/pkg/git/client.go
+++ b/pkg/git/client.go
@@ -166,7 +166,7 @@ func (c *client) Clone(ctx context.Context, repoID, remote, branch, destination 
 				return nil, err
 			}
 			out, err := retryCommand(3, time.Second, logger, func() ([]byte, error) {
-				args := []string{"clone", "--mirror", remote, repoCachePath}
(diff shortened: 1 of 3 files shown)


ffjlabo (approved):
LGTM

FYI, for the other reviewer, it might be helpful to understand treeless clone.
https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

khanhtc1202 (approved):
Great improvement, thanks 👍

PR #5724, opened 2025-04-04
Change resource name for piped helm chart

ffjlabo:
**What this PR does**:

as title
I fixed the resource name for PVC、Service、ServiceAccount、Secret

**Why we need it**:

We consider the case when users prepare these resources in other ways.

Fixes #


ffjlabo:
We discussed the fix and decided to follow the current spec below (e.g. config map).
```
The name of the ConfigMap to use when create is false.
```

PR #5733, opened 2025-04-11
[K8s Plugin] Refactor to use LoadConfigSpec to load application config

Warashi:
**What this PR does**:

This PR does a refactoring to use sdk.LoadConfigSpec instead of configv1.DecodeYAML

**Why we need it**:

We don't want to import the configv1 package under plugin implementation because it is not a package for plugin implementation.

Part of #5530
Follows #5731 


t-kikuc (approved):

Warashi:
I implemented config unmarshalling before invoking the plugin implementation in #5740, so this is no longer needed.

PR #5734, opened 2025-04-11
Modify volume name attached to deployment

ffjlabo:
**What this PR does**:

Fixed to be able to change volume name when setting it on values.yaml .

**Why we need it**:

Users can't set the volume name as is for now.

Follow #5716


khanhtc1202 (approved):
LGTM 👍 Thanks

Warashi (approved):
LGTM

PR #5737, opened 2025-04-14
[SDK] Add feature to check the pipeline has specific stage

Warashi:
**What this PR does**:

as title

**Why we need it**:

The Kubernetes deployment executor has a feature that switches its behavior depending on whether the pipeline has the `K8S_TRAFFIC_ROUTING` stage or not.
We want to support this feature with the Kubernetes Plugin, so we need the feature to detect whether the pipeline has a specific stage.

ref; https://github.com/pipe-cd/pipecd/blob/22c9cd0acc58c88e19a34bfeec4c71ebc0c4e656/pkg/app/piped/executor/kubernetes/primary.go#L96

Part of #5530 


pkg/plugin/sdk/config.go
```diff
@@ -30,3 +34,34 @@ func LoadConfigSpec[T Spec[RT], RT any](ds DeploymentSource) (T, error) {
 	}
 	return cfg.Spec, nil
 }
+
+// LoadStages loads the pipeline stages from the application config.
+func LoadStages(ds DeploymentSource) (Stages, error) {
```
t-kikuc, line 39: [ASK] Is this `LoadStages()` called from outside?
If not, I just think only `func (s Stages) Has(stage string) bool {` is needed to be public.

Warashi: If we don't make `LoadStages` public, we have to call `config.DecodeYAML` in `Has`.
If we make `LoadStages` public, we can reuse `Stages`.

t-kikuc: Ah, I got it. " multiple times." is the key

t-kikuc (approved):

Warashi:
In #5740, I implemented the unmarshalling config in SDK before invoking the plugin's implementation. I modified this PR to use it.

Warashi:
@t-kikuc Sorry, I rewrote all the codes in this PR, so please review this again.

t-kikuc (approved):

ffjlabo (approved):
LGTM

PR #5748, opened 2025-04-18
Update gen release script

khanhtc1202:
**What this PR does**:

Remove `-i` flag from sed command in script

**Why we need it**:

The script should work on both Linux and Unix, so do not use flags that aren't available on only one side.

Fixes #


Warashi (approved):
LGTM 👍🏻

ffjlabo (approved):

t-kikuc (approved):

PR #5761, opened 2025-04-21
Add ability to push/pull plugin to/from OCI Registry

Warashi:
**What this PR does**:

This PR adds functionality to push/pull OCI artifacts and uses it to pull plugins from the OCI registry.

**Why we need it**:

We have to support pulling plugins with correct platform handling. We can implement this by templating the HTTP URL to download the plugin, but it's a bit complex.
We can provide this feature without templating the URL by using the OCI registry.

Part of #4980 


pkg/oci/options.go
```diff
@@ -0,0 +1,120 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package oci
+
+const (
+	// MediaTypePipedPlugin is the media type for PipeCD Agent plugins.
+	MediaTypePipedPlugin = "application/vnd.pipecd.piped.plugin"
+)
```
ffjlabo, line 20: @Warashi 
[ask] Is `pkg/oci` expected to be used for the other case, except for downloading the plugin?
If not, how about adding it under the `pkg/plugin`?

pkg/oci/push.go
```diff
@@ -0,0 +1,146 @@
...
+	// ArtifactType is the type of the artifact.
+	// e.g. "application/vnd.pipecd.piped.plugin+type"
+	ArtifactType string
+	// FilePaths maps platforms to file paths.
+	FilePaths map[Platform]string
+}
+
+// PushFilesToRegistry pushes files described by the artifact to the target OCI registry URL.
+// It supports options for insecure connections.
+func PushFilesToRegistry(ctx context.Context, workDir string, artifact *Artifact, targetURL string, opts ...PushOption) error {
```
ffjlabo, line 53: [ask] Where is this method for pushing plugin executed? I want to know where you plan to push the plugin if you have idea about it.

ffjlabo (commented):
I feel it is a great enhancement.
I left some ask comments.

pkg/oci/options.go
Warashi: @ffjlabo 
I want to clarify what we should place under the package/plugin. I think we should put the packages used to implement the plugin and only them. If we place features that are not used to implement the plugin, the plugin developers may be confused about which package to use. WDYT?

pkg/oci/push.go
Warashi: I want to implement commands like `pipectl plugin push` or `pipectl push plugin` to push plugins' binary to OCI Registries.

pkg/oci/options.go
ffjlabo: @Warashi 
> I think we should put the packages used to implement the plugin and only them.
I got your point. Sorry, I was told it some time before.

On the other hand, I think it would be nice to minimize the package's scope to avoid using it for an unexpected use case. How about in `pkg/app/pipedv1/plugin`?

pkg/oci/push.go
ffjlabo: Thank you, I got your idea.

pkg/oci/options.go
ffjlabo: @Warashi 
> I want to implement a pipectl subcommand to push plugins to the OCI Registry, so I want to use this in pipectl. So, I don't want to make this scope to pipedv1.

I see, surely that would be strange, package pipectl imports pipedv1's sub-package.
OK, let's keep it as is.

ffjlabo (approved):
Let's try!

t-kikuc (approved):
let's try

PR #5762, opened 2025-04-21
[docs] Add note of checking the SSH key

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

To help troubleshooting.
The app suggestion on UI won't work if a user hasn't set up an SSH key for cloning `git@github.com:pipe-cd/examples.git`.

### DIFF
--- a/cmd/piped/README.md
+++ b/cmd/piped/README.md
@@ -61,4 +61,4 @@ For the full list of available commands, please see the Makefile at the root of 
     make run/piped CONFIG_FILE=piped-config.yaml
     ```
 
-Note: If you have a problem with accessing the example repository, please follow [this guide](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) to register the SSH key or try specifying `remote: https://github.com/pipe-cd/examples.git` instead.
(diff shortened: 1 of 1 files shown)


cmd/piped/README.md
```diff
@@ -60,3 +60,5 @@ For the full list of available commands, please see the Makefile at the root of
     ``` console
     make run/piped CONFIG_FILE=piped-config.yaml
     ```
+
+Note: If you have a problem with accessing the example repository, please follow [this guide](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) to register the SSH key or try specifying `remote: https://github.com/pipe-cd/examples.git` instead in the above `piped-config.yaml`.
```
ffjlabo, line 64: [imo] How about adding the explanation like `Prepare for connecting to github with ssh` on Prerequisites ?
It might be good for developers to know it before trying it.

ffjlabo (commented):
Nice catch! I agree with adding the node for it.
Also left the IMO comment for the placement.

cmd/piped/README.md
t-kikuc: @ffjlabo 
I got your point, and it also seems great.

I think just using HTTPS is enough for some cases, and SSH is not mandatory.

ffjlabo: @t-kikuc 
I got it. It certainly doesn't have to be ssh 👍

ffjlabo (approved):
LGTM

khanhtc1202 (approved):
👍

PR #5765, opened 2025-04-21
Implement k8s multi cluster rollback

ffjlabo:
**What this PR does**:

**Why we need it**:

Fixes #


pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go
```diff
@@ -27,6 +30,73 @@ import (
...
+			lp.Infof("Start rollbacking the deployment to the target %s", tc.deployTarget.Name)
+			status := p.rollback(ctx, input, tc.deployTarget, tc.multiTarget)
+			if status == sdk.StageStatusFailure {
+				return fmt.Errorf("failed to rollback the deployment to the target %s", tc.deployTarget.Name)
+			}
+			return nil
+		})
+	}
+
+	if err := eg.Wait(); err != nil {
```
t-kikuc, line 89: [ASK] What if one rollback fails and others are still rolling back?

Cancel all of them? Or still continue rolback?

ffjlabo: Thank you for pointing that out!
The current implementation is cancelling all.

I forgot to implement to continue rollback. will do it.

BTW, I described that in the rfc 🙏 
https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0014-multi-cluster-deployment-for-k8s.md

ffjlabo: Fixed [362b45e](https://github.com/pipe-cd/pipecd/pull/5765/commits/362b45e18f0f38a88d2c004421e58125e2a1f5b9)

t-kikuc: Thanks, i understood here:

> You can select the following behavior by setting the stage.
Rollback if any app fails
Rollback if all apps fail

pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go
```diff
@@ -73,25 +71,46 @@ func (p *Plugin) executeK8sMultiRollbackStage(ctx context.Context, input *sdk.Ex
...
-		lp.Errorf("Failed while rollbacking the deployment (%v)", err)
-		return sdk.StageStatusFailure
+	wg.Wait()
+
+	finalStatus := sdk.StageStatusFailure
+	for _, result := range results {
+		// success at least one of the rollback succeed
+		if result.status != sdk.StageStatusFailure {
+			finalStatus = sdk.StageStatusSuccess
+		}
```
t-kikuc, line 108: [ASK] Which should we choose here?  (I don't have a strong opinion)
- a. SUCCESS if at least one succeeds
- b. SUCCESS if all succeed

This is difficult... Any choice seems confusing, and I want a status like "Some Success / Some Failure" 😅

> If at least one of the rollback processes succeeds, we consider the rollback successful. This ensures that the rollback is executed for other environments even if one of the deployment environments is inaccessible.

https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0014-multi-cluster-deployment-for-k8s.md#qicksync

ffjlabo: We should choose  a. SUCCESS if at least one succeeds to ensure rollback when one of the cluster has some trouble but want to rollback another cluster.

> This is difficult... Any choice seems confusing, and I want a status like "Some Success / Some Failure" 😅

I know your point, but this spec is only for this plugin.
I plan to add a log to show the each rollback status and the whole status Instead of adding status.

t-kikuc (dismissed):

Warashi (approved):
👍🏻

khanhtc1202 (approved):
👍

PR #5768, opened 2025-04-22
Fix to be able to execute rollback stage for plugin-architectured piped

ffjlabo:
**What this PR does**:

as title.

Currently, the rollback stage doesn't work for two reasons.
1. Find rollback stages by checking only whether the stage name is `ROLLBACK` or `SCRIPT_RUN_ROLLBACK`.
2. baseStageID is used when the stage is for rollback, but it is not inserted into the metadata for plugin.

About 1, I added the new logic to check whether `stage.Rollback` is true.

About 2. Remove the logic because it is the `SCRIPT_RUN` specific spec.

**Why do we need it**:

I want to support executing rollback stage for plugin-architectured piped.

Fixes #


Warashi (approved):
Thank you

t-kikuc (approved):

PR #5769, opened 2025-04-22
remove temp sshKeyFile after use(#2215)

hiep-tk:
**What this PR does**: remove the temporary ssh key file after using it for the config

**Why we need it**: clean up the ssh key folder

Fixes #

### DIFF
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -175,10 +175,12 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 
 	// Configure SSH config if needed.
 	if cfg.Git.ShouldConfigureSSHConfig() {
-		if err := git.AddSSHConfig(cfg.Git); err != nil {
+		tempFile, err := git.AddSSHConfig(cfg.Git)
+		if err != nil {
 			input.Logger.Error("failed to configure ssh-config", zap.Error(err))
 			return err
 		}
+		defer os.Remove(tempFile)
 		input.Logger.Info("successfully configured ssh-config")
 	}
 
--- a/pkg/git/ssh_config.go
+++ b/pkg/git/ssh_config.go
@@ -48,52 +48,61 @@ type sshConfig struct {
 	IdentityFile string
 }
 
-func AddSSHConfig(cfg config.PipedGit) error {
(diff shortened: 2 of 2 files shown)


pkg/app/piped/cmd/piped/piped.go
```diff
@@ -175,10 +175,14 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 
 	// Configure SSH config if needed.
 	if cfg.Git.ShouldConfigureSSHConfig() {
-		if err := git.AddSSHConfig(cfg.Git); err != nil {
+		tempFile, err := git.AddSSHConfig(cfg.Git)
+		if err != nil {
```
khanhtc1202, line 179: This still leaks some cases, if the error is not nil, the AddSSHConfig function returns tempFile, but this block will immediately closes this run function, thus defer at L184 is not triggered 👀

t-kikuc: I agree with @khanhtc1202 , the code might be like this:


```go
tempFile, err := git.AddSSHConfig(cfg.Git)
if len(tempFile) > 0 { // HERE
	defer os.Remove(tempFile)
}
if err != nil {
...
```

hiep-tk: Updated as suggested. @khanhtc1202 @t-kikuc

pkg/git/ssh_config.go
```diff
@@ -48,52 +48,51 @@ type sshConfig struct {
...
 	sshKeyFile, err := os.CreateTemp(sshDir, "piped-ssh-key-*")
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	// TODO: Remove this key file when Piped terminating.
 	if _, err := sshKeyFile.Write(sshKey); err != nil {
-		return err
+		return sshKeyFile.Name(), err
```
khanhtc1202, line 77: How about closing this file in this function if err is not nil, then we can have convention like: `return "", err` if err occurred, while the file is only returned in case there is no issue, and the caller only need to defer close file in case of no error (after err check)? @hiep-tk

hiep-tk: Updated for consistent return format @khanhtc1202

khanhtc1202 (approved):
LNeatTM 👍

Warashi (approved):
LGTM

PR #5773, opened 2025-04-23
Fix to set head commit hash on pipedv1 LivestateReporter

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

I want to clarify the commit hash to track which commit is used.
Always used `HEAD` as the current revision in the current implementation.
PipeCD

Part of #5363


t-kikuc (approved):
Great improvement

Warashi (approved):
👍🏻

Issue #5764, opened 2025-04-21
Implement Kubernetes Plugin


Warashi:
**What would you like to be added**:

Support the Kubernetes platform in the plugin architecture.

**Why is this needed**:

aryasoni98:
Hello @Warashi 

I’d love to be a part of this project and contribute. Looking forward to working together!

ffjlabo:
updated the todos

PR #5774, opened 2025-04-23
Normalize resource key before grouping manifests to convert default namespace to empty

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

When no namespace is set for resources in the `app.pipecd.yaml` or actual resource manifests, there is the diffs like below.

PipeCD

When loading manifests from git, set an empty string as the namespace if no namespace is set.
But when loading manifests from actual clusters, the value `default` is set as the namespace.

So we need to normalize the namespace before comparing the manifests from git and the ones from the actual cluster.
Especially set empty value considering the case when no namespace is set in the manifests from git repo.

Part of #5006 #5764 


t-kikuc (approved):

Warashi (approved):
LGTM

PR #5777, opened 2025-04-24
Fixed to add annotations after calling loader.LoadManifests

ffjlabo:
**What this PR does**:

Move the logic to add the annotation logic into the plugin side.

**Why we need it**:

When creating the commit, the fixes are not related to the target app, there is a diff for `pipecd.dev/commit-hash`.
The current logic detects diffs that are defined in the git repo but do not exist in the actual resource.

So, IMO, we shouldn't include them when loading manifests from the git repo because these annotations are set automatically before applying resources on the piped side.

PipeCD

Part of #5006 #5764 


Warashi (approved):
👍🏻

t-kikuc (approved):

PR #5778, opened 2025-04-24
Implement livestate for k8s multicluster plugin

ffjlabo:
**What this PR does**:

as title. This is a PoC.

PipeCD

**Spec summary**

- Show each diff in one UI
- Show each resource state with a tab

**TODOs**
- decide synstate from multiple sync state
- decide resource state from multiple resource state

**Why we need it**:

We want to support livestate feature for multi cluster plugin.

Part of #5006

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/livestate/plugin.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/livestate/plugin.go
@@ -31,82 +31,145 @@ import (
 
 type Plugin struct{}
 
-// GetLivestate implements sdk.LivestatePlugin.
 func (p Plugin) GetLivestate(ctx context.Context, _ *sdk.ConfigNone, deployTargets []*sdk.DeployTarget[kubeconfig.KubernetesDeployTargetConfig], input *sdk.GetLivestateInput[kubeconfig.KubernetesApplicationSpec]) (*sdk.GetLivestateResponse, error) {
-	if len(deployTargets) != 1 {
-		return nil, fmt.Errorf("only 1 deploy target is allowed but got %d", len(deployTargets))
-	}
-
-	deployTarget := deployTargets[0]
-	deployTargetConfig := deployTarget.Config
-
 	cfg, err := input.Request.DeploymentSource.AppConfig()
 	if err != nil {
 		input.Logger.Error("Failed to load application config", zap.Error(err))
 		return nil, err
 	}
 
+	logger := input.Logger
+
+	type targetConfig struct {
+		deployTarget *sdk.DeployTarget[kubeconfig.KubernetesDeployTargetConfig]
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/kubernetes_multicluster/livestate/plugin.go
```diff
@@ -31,85 +31,148 @@ import (
...
-	// Get the kubectl tool path.
-	kubectlPath, err := toolRegistry.Kubectl(ctx, cmp.Or(cfg.Spec.Input.KubectlVersion, deployTargetConfig.KubectlVersion))
-	if err != nil {
-		input.Logger.Error("Failed to get kubectl tool", zap.Error(err))
-		return nil, err
+	liveStates := make([]sdk.ApplicationLiveState, 0, len(targetConfigs))
+	syncStates := make([]sdk.ApplicationSyncState, 0, len(targetConfigs))
+	for _, tc := range targetConfigs {
+		// Get the kubectl tool path.
+		kubectlPath, err := toolRegistry.Kubectl(ctx, cmp.Or(cfg.Spec.Input.KubectlVersion, tc.deployTarget.Config.KubectlVersion))
```
Warashi, line 88: This may be wrong. I think `tc.deployTarget.Config.KubectlVersion` has a higher priority.
```suggestion
		kubectlPath, err := toolRegistry.Kubectl(ctx, tc.deployTarget.Config.KubectlVersion, cmp.Or(cfg.Spec.Input.KubectlVersion))
```

Warashi (approved):
Let's try

pkg/app/pipedv1/plugin/kubernetes_multicluster/livestate/plugin.go
ffjlabo: @Warashi 
It seems that `cfg.Spec.Input.KubectlVersion` has a higher priority in the old version.
 
> // priority: applicationConfig.KubectlVersion > pipedConfig.KubectlVersion
https://github.com/pipe-cd/pipecd/blob/f53dd3cacc5dab5db3df48332b98f36b53158f17/pkg/app/piped/platformprovider/kubernetes/applier.go#L209C1-L217C2


It is the same as other deployment loginc for k8s plugin and k8s_multicluster.

Warashi: @ffjlabo 
Oops, thank you. I am confused with deployTargets and multiTargets.
ref; https://github.com/pipe-cd/pipecd/blob/f53dd3cacc5dab5db3df48332b98f36b53158f17/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go#L182-L186

khanhtc1202 (approved):
👌

PR #5779, opened 2025-04-25
Add support for OCI registry to push/pull plugins

Warashi:
**What this PR does**:

- add `pipectl plugin push` command to push plugin binary to OCI Registry
    - The auth feature has not been implemented yet
- add `oci` as an allowed scheme to download a plugin

We can use the `pipectl plugin push` as the command below
```sh
$ docker run -d --rm -p 5001:5000 --name registry registry
$ pipectl plugin push --tag=latest --registry=localhost:5001 --repository=piped-kubernetes --files=darwin/arm64=./.artifacts/plugins/kubernetes --insecure
$ pipectl plugin push --tag=latest --registry=localhost:5001 --repository=piped-wait --files=darwin/arm64=./.artifacts/plugins/wait --insecure
```

And we can pull the plugins from OCI Registry with the config below.
```yaml
kind: Piped
spec:
  projectID: ***
  pipedID: ***
  pipedKeyData: ***
  apiAddress: ***
  git: ***
  repositories: ***
  plugins:
    - name: kubernetes
      port: 7001
      url: oci://localhost:5001/piped-kubernetes:latest
      deployTargets:
        - name: local-k8s
          config:
            kubeConfigPath: ***
    - name: wait
      port: 7002
      url: oci://localhost:5001/piped-wait:latest
```

Before trying the config above, we have to patch the below lines to include the option `oci.WithInsecure()` because the local registry hosts the HTTP protocol, not the HTTPS protocol.
https://github.com/pipe-cd/pipecd/blob/9cc23559616e0b5995272739f895c4e9b2b4aafe/pkg/lifecycle/binary.go#L152-L159

**Why we need it**:

We want to support OCI Registry in hosting a piped plugin.

Part of #4980 

### DIFF
--- a/cmd/pipectl/main.go
+++ b/cmd/pipectl/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/migrate"
 	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/piped"
 	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/planpreview"
+	"github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/plugin"
 	"github.com/pipe-cd/pipecd/pkg/cli"
 )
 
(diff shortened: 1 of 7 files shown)


t-kikuc (approved):

ffjlabo (approved):
Let's try

PR #5780, opened 2025-04-25
Fix deployment tracing loading nonStop when loadmore

hongky-1994:
**What this PR does**:

- Fix wrong place for saving minUpdatedAt of deploymentTraces in redux store.
- Reverse sort traces date to show Date in Desc order. 
   CleanShot 2025-04-26 at 01 47 27

**Why we need it**:

- In page Traces, when we scroll down to load more. Api fetch more traces is run non-stop due to wrong saving minUpdatedAt in redux store.

Fixes #


khanhtc1202 (approved):
Thanks :+!

t-kikuc (approved):

PR #5783, opened 2025-04-28
Add RFC for redux replacement feature

hongky-1994:
**What this PR does**:

**Why we need it**:

This will be used as a note for Redux replacement decision.

Fixes #


khanhtc1202 (approved):

khanhtc1202:
Thanks @hongky-1994 
I think the only question remains here is when to release this change, but I think we would not have one big release for all, but release components along with other change in the backend, so no need to specify it here 👍

t-kikuc (approved):

PR #5789, opened 2025-04-30
Fix the type of the Plugins field in the GenericApplicationSpec struct

Warashi:
**What this PR does**:

Change the type of the Plugins field from []string to map[string]struct in the GenericApplicationSpec struct.

**Why we need it**:

We want to put the plugins' config under the `.spec.plugins.${plugin-name}`, so this field becomes map type.

Part of #4980 

**Does this PR introduce a user-facing change?**: No, because the plugin-arch piped has not been released yet.


t-kikuc (approved):

khanhtc1202 (approved):
🚀

PR #5790, opened 2025-04-30
[SDK] Change the config structure of the app.pipecd.yaml

Warashi:
**What this PR does**:

Change the config structure of the app.pipecd.yaml
We put the plugins' configuration under the `.spec.plugins.${plugin-name}`.

**Why we need it**:

Without this change, the users may be confused about which config field belongs to which plugin.

Part of #5530 
Follows #5789 


pkg/app/pipedv1/plugin/registry_test.go
```diff
@@ -130,7 +130,7 @@ func TestPluginRegistry_GetPluginClientsByAppConfig(t *testing.T) {
 
 			pr := tt.setup()
 			plugins, err := pr.GetPluginClientsByAppConfig(tt.cfg)
-			assert.Equal(t, tt.expected, plugins)
+			assert.ElementsMatch(t, tt.expected, plugins)
```
Warashi, line 133: This test became flaky by #5789.
This change fixes that.

t-kikuc (approved):
[ASK] Are we going to add tests in pipedv0 to parse v1 app.pipecd.yaml for the compatibility?

Warashi:
> [ASK] Are we going to add tests in pipedv0 to parse v1 app.pipecd.yaml for the compatibility?

Yes, we will add tests to support the new config format with the current piped.

khanhtc1202 (approved):
LGTM

PR #5799, opened 2025-05-07
Fixed from getPluginClientsByNames to getPluginClientsByPlugins to determine the plugin with app config

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

We want to clarify the role for the method.

Part of #4980
Follow #5789


ffjlabo:
I found that it is no need. So will close

PR #5800, opened 2025-05-07
docs: use Sync instead of AutoSync

ponkio-o:
**What this PR does**: Documentation and WebUI notations were different, so changes were made to align them.

image

**Why we need it**: Because of the confustion

**Which issue(s) this PR fixes**: None


khanhtc1202 (approved):
Thank you! 🙌

ffjlabo (approved):
Thank you!

PR #5801, opened 2025-05-07
Add comment for the plugins field in generic application config

ffjlabo:
**What this PR does**:

as titlw

**Why we need it**:

We want to clarify why defined `spec.plugins` are empty struct.

Follow https://github.com/pipe-cd/pipecd/pull/5789

### DIFF
--- a/pkg/configv1/application.go
+++ b/pkg/configv1/application.go
@@ -58,7 +58,7 @@ type GenericApplicationSpec struct {
 	EventWatcher []EventWatcherConfig `json:"eventWatcher"`
 	// Configuration for drift detection
 	DriftDetection *DriftDetection `json:"driftDetection"`
-	// List of the plugin name
+	// List of the configuration for plugin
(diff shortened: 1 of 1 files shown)


pkg/configv1/application.go
```diff
@@ -59,6 +59,7 @@ type GenericApplicationSpec struct {
 	// Configuration for drift detection
 	DriftDetection *DriftDetection `json:"driftDetection"`
 	// List of the plugin name
+	// This field is plugin-specific, so intentionally restrict the access for the actual value here and decode it on the SDK side.
```
khanhtc1202, line 62: So we should rewrite the above line either. Since it's not just list of plugin name anymore

ffjlabo: Thanks, fixed [2dba71d](https://github.com/pipe-cd/pipecd/pull/5801/commits/2dba71d0adfc3efe0a6fa796193073d1c216816f)

Warashi (approved):
LGTM 👍🏻

t-kikuc (approved):

PR #5806, opened 2025-05-08
Use testing.T.TempDir() instead of tmpDir

Warashi:
**What this PR does**:

Use testing.T.TempDir each time to ensure the InstallTool is concurrent-safe.

**Why we need it**:

This makes the test more stable.
Without this, the test was flaky because concurrent tests read/write to the same directory/file.

Fixes #

### DIFF
--- a/pkg/plugin/toolregistry/toolregistrytest/toolregistrytest.go
+++ b/pkg/plugin/toolregistry/toolregistrytest/toolregistrytest.go
@@ -29,11 +29,6 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/plugin/toolregistry"
 )
 
-type ToolRegistry struct {
-	testingT *testing.T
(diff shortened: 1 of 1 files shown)


pkg/plugin/toolregistry/toolregistrytest/toolregistrytest.go
```diff
@@ -29,11 +29,6 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/plugin/toolregistry"
 )
 
-type ToolRegistry struct {
-	testingT *testing.T
-	tmpDir   string
-}
```
Warashi, line 35: This type is not used anywhere.

khanhtc1202 (approved):
Nice catch 👍

ffjlabo (approved):
Thank you

PR #5811, opened 2025-05-12
Implement comparing bool with boolean string value

### LINKED ISSUE #4790: PipeCD gets diff with bool value in Cloud Run manifets
**What happened**: With Cloud Run application, when we set a bool value for env, PipeCD gets diff the env, like following.
Manifest is
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: xxx
spec:
  template:
    spec:
      containers:
        - name: xxx
          env:
            - name: XXX
              value: false
```

PipeCD console is 
```diff
      spec:
        containers:
          - env:
              -
                #spec.template.spec.containers.0.env.10.value
-               value: false
+               value: <bool Value>
```

**How to reproduce it**:

**Environment**:
- `piped` version: `v0.44.2`
- `control-plane` version: `v0.46.0`
- Others:

Warashi:
**What this PR does**:

- Add support for comparing boolean values with boolean strings

**Why we need it**:

Sometimes, we write a boolean value in the fields whose type is a string. In this case, such fields are converted to strings in the retrieved livestate. This makes non-sense diffs in the drift detection.

Fixes #4790 

Some diffs affected by this bug will disappear.

### DIFF
--- a/pkg/app/piped/driftdetector/cloudrun/detector.go
+++ b/pkg/app/piped/driftdetector/cloudrun/detector.go
@@ -211,6 +211,7 @@ func (d *detector) checkApplication(ctx context.Context, app *model.Application,
 		diff.WithEquateEmpty(),
 		diff.WithIgnoreAddingMapKeys(),
 		diff.WithCompareNumberAndNumericString(),
+		diff.WithCompareBooleanAndBooleanString(),
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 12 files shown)


khanhtc1202 (approved):
LGTM 👍

t-kikuc (approved):

PR #5815, opened 2025-05-13
[K8s plugin] Add health status calculation for Deployment resources

Warashi:
**What this PR does**:

as title

**Why we need it**:

We want to display health statuses in the web UI

Part of #5764 

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/provider/health_status.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/provider/health_status.go
@@ -1,0 +1,68 @@
+// Copyright 2025 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


t-kikuc (approved):

pkg/app/pipedv1/plugin/kubernetes/provider/health_status.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+
+	"github.com/pipe-cd/pipecd/pkg/plugin/sdk"
+)
+
+func (m Manifest) calculateHealthStatus() (sdk.ResourceHealthStatus, string) {
```
ffjlabo, line 25: To distinguish with application's health state.
```suggestion
func (m Manifest) calculateResourceHealthStatus() (sdk.ResourceHealthStatus, string) {
```

Warashi: I think we don't need to distinguish application health status and resource health status within the Manifest method because a single manifest holds a single resource, and a manifest doesn't have the whole application health status.

ffjlabo: IMO, it would be nice to distinguish it because the method is called when creating livestate.

ffjlabo: Wait, sorry It's fine without it after all.
We don't need consider about the application health status.
I forgot to this PR to be merged https://github.com/pipe-cd/pipecd/pull/5829

ffjlabo (approved):
Sorry for the confusion, LGTM

PR #5817, opened 2025-05-14
[k8s] Remove ignoreResourceKeys

ffjlabo:
**What this PR does**:

Removed ignoreResourceKeys to ignore the livestate for specific resources.

**Why we need it**:

The livestates of specific resources are not stored with the hard-coded ignore list by default.
This led to the unexpected result of drift detection.

For example, there is an unexpected diffs when deploying `metrics-server` in the namespace `kube-system`.

PipeCD

There seem to be three types of components in the ignore list, with the understanding that 1 might be deployed with PipeCD, while 2 and 3 do not have such a possibility, so there is no need to save Livestate.
1. Resources that users may install manually or be installed as adon in managed k8s cluster (such as metrics-server, k8s-dashboard)
2. Default components of k8s (such as kube-controller-manager, kube-scheduler)
3. Components presumed to be specific to GKE (such as heapster-gke)

Also, I think it is enough to check only whether it is managed by piped when managing livestate for k8s.

Fixes #

### DIFF
--- a/pkg/app/piped/livestatestore/kubernetes/reflector.go
+++ b/pkg/app/piped/livestatestore/kubernetes/reflector.go
@@ -94,45 +94,6 @@ var (
 		"IngressClass":             {},
 		"Namespace":                {},
 	}
-	ignoreResourceKeys = map[string]struct{}{
-		"v1:Service:default:kubernetes":               {},
-		"v1:Service:kube-system:heapster":             {},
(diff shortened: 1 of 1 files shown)


khanhtc1202 (approved):
Let's try 👌

t-kikuc (approved):

t-kikuc:
let's add `cherry-pick` if needed

PR #5818, opened 2025-05-14
Add test for LoadApplication configv1

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

Fixes #


pkg/configv1/testdata/application/plugins-config.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: pipecd.dev/v1beta1
+kind: Application
+spec:
+  plugins:
+    plugin-1:
+    plugin-2:
+      input:
+        version: 1.0.0
```
khanhtc1202, line 8: This is our expected application plugins config

pkg/configv1/application_test.go
```diff
@@ -643,3 +644,94 @@ func TestGetStageConfigByte(t *testing.T) {
...
+					"plugin-2": {},
+				},
+			},
+			wantErr: false,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := LoadApplication(tc.repoPath, tc.configRelPath)
```
t-kikuc, line 728: add `t.Parallel()`

also in `func TestLoadApplication(t *testing.T) {`

pkg/configv1/application_test.go
```diff
@@ -643,3 +644,94 @@ func TestGetStageConfigByte(t *testing.T) {
...
+						Disabled: newBoolPointer(true),
+					},
+				},
+				Planner: DeploymentPlanner{
+					AutoRollback: newBoolPointer(true),
+				},
+				Pipeline: &DeploymentPipeline{},
+				Plugins: map[string]struct{}{
+					"plugin-1": {},
+					"plugin-2": {},
```
t-kikuc, line 719: [ASK] `LoadApplication()` does not parse `plugin-2.input.version=1.0.0`, and the SDK does instead, right?

t-kikuc (commented):
Almost LGTM

pkg/configv1/application_test.go
khanhtc1202: Yes, it's as expected. We only need the name of the plugin here.

t-kikuc: Thank you!

t-kikuc (approved):
Thank you!

PR #5826, opened 2025-05-15
Enable pipedv0 to parse app config with plugins

khanhtc1202:
**What this PR does**:

**Why we need it**:

we want to make pipedv0 works with applications which registered by pipedv1 - thus contains plugins config in its config

I tried referring to the `configv1/appplication` structure directly from pipedv0, but it's overcomplicated, compared to this way, we just need to keep the configuration surface the same, but the logic for pipedv0 remained using `config/application` structure.

Fixes #

### DIFF
--- a/pkg/config/application.go
+++ b/pkg/config/application.go
@@ -60,7 +60,6 @@ type GenericApplicationSpec struct {
 	DriftDetection *DriftDetection `json:"driftDetection"`
 	// NOTE: Do not use the field, only added here to avoid unmarshalling error.
 	// List of the configuration for plugin
-	// This field is plugin-specific, so intentionally restrict the access for the actual value here and decode it on the SDK side.
 	Plugins map[string]struct{} `json:"plugins"`
 }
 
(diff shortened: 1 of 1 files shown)


pkg/config/application.go
```diff
@@ -58,6 +58,9 @@ type GenericApplicationSpec struct {
 	EventWatcher []EventWatcherConfig `json:"eventWatcher"`
 	// Configuration for drift detection
 	DriftDetection *DriftDetection `json:"driftDetection"`
+	// NOTE: Do not use the field, only added here to avoid unmarshalling error.
```
t-kikuc, line 61: To clarify what's unmarshalling error
```suggestion
	// NOTE: Do not use the field in pipedv0, only added here to avoid unmarshalling error for compatibility with pipedv1.
```

t-kikuc (approved):
Thank you! Important point!

khanhtc1202:
In pipedv1 application config file, `input` moved under `plugins` thus we need to update the pipedv0 reflect that

PR #5831, opened 2025-05-16
update fork-ts-checker-webpack-plugin package

hongky-1994:
**What this PR does**:

- Update version fork-ts-checker-webpack-plugin to ^9.1.0

**Why we need it**:

- Fix bug `Cannot set property mark of #<Object> which has only a getter` when typescipt version is updated.

CleanShot 2025-05-16 at 10 58 28

Fixes #


khanhtc1202 (approved):
👍

PR #5833, opened 2025-05-16
[feat] : Add support for token and certificate authentication in Helm chart registry

### LINKED ISSUE #3743: Support login to Helm chart registry using other method than basic authentication

Currently, we only support login to Helm chart registry using basic authentication, but it would be nice if other methods could be used as well.

**Why is this needed**:

kiratlog:
**What this PR does**:

**Why we need it**:

Fixes #3743

[Reference ](https://helm.sh/docs/helm/helm_registry_login/)


ffjlabo:
@kiratlog 
Thank you for the PR.
At first, we need to clarify the authentication method for the helm chart registry.
Could you investigate it more, other than basic authentication? If possible, it would be nice to have the usage for other methods.

Warashi:
Sorry, we updated the CI and branch protection rules.
I updated the branch of this PR by pushing the `Update branch` button, which is the GitHub feature. 
Please let me know if you see any unexpected behavior.

PR #5836, opened 2025-05-16
added new field errorReason

### LINKED ISSUE #3344: Add a new field to show why command was failed.
https://github.com/pipe-cd/pipecd/blob/48dcb9bb3f580497746d6b6d417bdc29e4bce975/pkg/model/command.pb.go#L163-L166

Anamika1608:
**What this PR does**:
added new field errorReason in the command structure and the services.

**Why we need it**:
To keep record why this command was failed.

~F ixes #3344~
Part of #3344


yarn.lock
t-kikuc, line 1: yarn.lock already exists in web/

Anamika1608: Okay, I'll remove that

Anamika1608:
@khanhtc1202 okay, doing the changes.

PR #5838, opened 2025-05-16
added new field errorReason in command Structure

Anamika1608:
**What this PR does**:
added new field errorReason in the command structure and the services.

**Why we need it**:
To keep record why this command was failed.

Part of #3344


t-kikuc:
@Anamika1608 
Thank you. This PR itself seems good for the first step.

I recommend that you create a ToDo list in https://github.com/pipe-cd/pipecd/issues/3344 at first because I'm afraid of the situation that the `errorReason` field is added but not called.

t-kikuc:
@Anamika1608 
Hi, I'd appreciate your efforts. 
Let's begin with an investigation if you're still interested in this task.

https://github.com/pipe-cd/pipecd/issues/3344#issuecomment-2899742794

Anamika1608:
Hi @t-kikuc, I'm still familiarising with the codebase and figuring out further steps.

Issue #3806, opened 2022-07-08
Request for application registration duplicate check function


jo-jo-1217:
I would like to have a duplicate check function when registering applications with the pipectl add command.

I have many services and projects, and I am having trouble registering applications with the pipectl add command, because the same applications can be registered.

t-kikuc:
Successfully (sadly), I reproduced the situation as below:


### 1. Run a piped
### 2. Execute the following pipectl command once

```sh
pipectl application add \
    --address=xxx \
    --api-key-file=xxx \
    --app-name=duplicate-1 \
    --app-kind=ECS \
    --piped-id=xxx \
    --platform-provider=ecs-dev \
    --repo-id=xxx \
    --app-dir=xxx
```

-> `Successfully added application id = d38a75b1-761a-4b42-b62c-16bd8ab0f3a2`

### 3.  Again, execute the same command as 2.

-> `Successfully added application id = 8fb1d9e0-c81d-487c-b567-6362d5336c86`
Another app with another ID is created.


apps UI:

<img width="1360" alt="Image" src="https://github.com/user-attachments/assets/5bbd0333-5b15-4d20-a814-ccf306b94073" />
deployments for each app were triggered:
<img width="1365" alt="Image" src="https://github.com/user-attachments/assets/1bd07f34-8ee2-447a-a99b-d806a427f32b" />

Tushar240503:
I would like work on this

khanhtc1202:
Hi @Tushar240503 I mentioned how to process this issue via this PR comment 👍 You're on the right track 🙌 
https://github.com/pipe-cd/pipecd/pull/5840#discussion_r2115050840

Tushar240503:
@khanhtc1202 thank you for the inputs

PR #5840, opened 2025-05-16
Add duplicate check when registering a new app for pipectl

Tushar240503:
Fix Issue #3806 
What:
This PR adds a duplicate check to the pipectl application add command, preventing the registration of applications with duplicate names or identical repository paths.

Why:
Previously, users could inadvertently register the same application multiple times, leading to confusion and potential deployment issues. This enhancement ensures the uniqueness of application entries, improving the reliability of the deployment process.

How:

Implements a check that retrieves existing applications via the ListApplications API.

Compares the new application's name and repository path against existing entries.

If a duplicate is detected, the command aborts and returns an appropriate error message.


khanhtc1202:
@Tushar240503 Could you link the issue for this change? 👀

khanhtc1202:
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/grpcapi/api.go#L134

t-kikuc:
> It's better to update the server side grpc handler, 

+1.  I wanted to say the same point 😄

pkg/app/server/grpcapi/api.go
```diff
@@ -157,6 +178,34 @@ func (a *API) AddApplication(ctx context.Context, req *apiservice.AddApplication
 		return nil, err
 	}
 
+	if req.GitPath != nil {
```
khanhtc1202, line 181: This can't be nil, since it's marked as required in request
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/service/apiservice/service.proto#L69

Tushar240503: Updated

pkg/app/server/grpcapi/api.go
```diff
@@ -146,6 +146,27 @@ func (a *API) AddApplication(ctx context.Context, req *apiservice.AddApplication
...
+			},
+		},
+	})
+	if err != nil {
+		return nil, gRPCStoreError(err, "list applications for duplicate check")
+	}
+	if len(existingApps) > 0 {
+		return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("application with name %q already exists in project %q", req.Name, key.ProjectId))
+	}
+
```
Warashi, line 169: IMO, we should allow applications to be registered with the same name because we can distinguish these by their labels.

WDYT? @khanhtc1202 @t-kikuc @ffjlabo

Warashi (commented):
The implementation is nice.
I want to discuss the spec to implement.

pkg/app/server/grpcapi/api.go
```diff
@@ -156,6 +177,33 @@ func (a *API) AddApplication(ctx context.Context, req *apiservice.AddApplication
...
+				Operator: datastore.OperatorEqual,
+				Value:    req.GitPath.Path,
+			},
+			{
+				Field:    "GitPath.ConfigFilename",
+				Operator: datastore.OperatorEqual,
+				Value:    req.GitPath.ConfigFilename,
+			},
+		},
+	})
```
Warashi, line 198: Oops, we should add a ProjectId or PipedId filter here because the GitPath.Repo.Id can be the same across projects.

ref; https://pipecd.dev/docs-v0.51.x/user-guide/managing-piped/adding-a-git-repository/

Warashi (changes requested):
Sorry, we need more filtering conditions to check whether they are duplicated.

pkg/app/server/grpcapi/api.go
t-kikuc: Let's make consistency between pipectl and UI.

(Sorry, I can't find the suggestion logic of UI yet)

Tushar240503: Thanks for bringing this up, @Warashi.

Using labels to distinguish applications with the same name is an interesting idea and could offer more flexibility.

That said, enforcing unique names per project currently helps avoid ambiguity, especially when working with tools like pipectl and in the UI. As @t-kikuc mentioned, we should aim for consistent behavior across all surfaces.

@t-kikuc: No worries about the UI logic—I’ll also take a look and see if I can trace how it currently handles duplicate names.

pkg/app/server/grpcapi/api.go
Tushar240503: @Warashi  updated

pkg/app/server/grpcapi/api.go
Tushar240503: Hey @t-kikuc @Warashi 
I  couldn’t find any logic in the UI handling duplicate application names—addApplication just forwards the request, leaving validation to the backend.

If uniqueness is required, we should add proper validation and user feedback in the UI.

t-kikuc: It seems the duplication is checked by `app.GitPath.GetApplicationConfigFilePath()` .

- piped > appconfigreporter > findUnregisteredApps():   <- Piped reports unregistred apps by this
https://github.com/pipe-cd/pipecd/blob/fd2df12ede6375cc4932857ab1de7923b1d71d7d/pkg/app/piped/appconfigreporter/appconfigreporter.go#L318



- server > ListUnregisteredApplications():   <- UI calls this
https://github.com/pipe-cd/pipecd/blob/fd2df12ede6375cc4932857ab1de7923b1d71d7d/pkg/app/server/grpcapi/web_api.go#L464-L492

- server > ReportUnregisteredApplicationConfigurations():   <- Piped updates apps by this
https://github.com/pipe-cd/pipecd/blob/fd2df12ede6375cc4932857ab1de7923b1d71d7d/pkg/app/server/grpcapi/piped_api.go#L901-L913

Warashi (dismissed):
LGTM

pkg/app/server/grpcapi/api.go
```diff
@@ -156,6 +156,38 @@ func (a *API) AddApplication(ctx context.Context, req *apiservice.AddApplication
...
```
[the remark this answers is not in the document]
Warashi: Sorry, I approved without checking the behavior.
I checked on my local with MySQL datastore and got this error.
```
failed to find entities    {"kind": "Application", "query": "SELECT Data FROM Application WHERE ProjectId = ? AND GitPath.Repo.Id = ? AND GitPath.Path = ? AND GitPath.ConfigFilename = ?", "whereConditionValues": ["hello-pipecd","local-pipecd","simple","app.pipecd.yaml"], "error": "Error 1054: Unknown column 'GitPath.Repo.Id' in 'where clause'"}
```

Warashi:
Sorry, we updated the CI and branch protection rules.
I updated the branch of this PR by pushing the `Update branch` button, which is the GitHub feature. 
Please let me know if you see any unexpected behavior.

pkg/app/server/grpcapi/api.go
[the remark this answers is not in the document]
khanhtc1202: @Tushar240503 @Warashi Thanks for the investigation. PipeCD uses MySQL in the same way we use NoSQL (Firestore): by storing JSON model objects in the field in MySQL directly and using Virtual columns to store specific fields in the model as table columns (such as the ProjectID field). Other columns that don't need to be stored directly in the schema will be generated and stored in indexes.
You can read around the design here: https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0003-sql-datastore.md
For MySQL Schema and Indexes, check files `.sql` in this dir: https://github.com/pipe-cd/pipecd/tree/master/pkg/datastore/mysql/ensurer

Also, for another datastore that PipeCD supported - Firestore, you also need to create indexes corresponding to the field, or GCP Firestore would not accept the queries either. So, check both datastore behavior on new query added, and we can separate PR for datastore indexes preparation. 👍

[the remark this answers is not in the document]
khanhtc1202: Mentioned some hints and explanations on the issue https://github.com/pipe-cd/pipecd/issues/3806#issuecomment-2921097801. You're on the right track, keep going 😄

PR #5843, opened 2025-05-16
delete all application resources when need

### LINKED ISSUE #5413: Way to delete all application resources when need

A way to remove/prune all application resources.

**Why is this needed**:

Say, when I want to delete/remove my application completely, I try to remove all of its manifests from Git and then use PipeCD sync (with the prune option) to remove all of its resources, but it doesn't work.

vaidikcode:
Signed-off-by: Vaidik Bhardwaj <vaidikbhardwaj00@gmail.com>

**What this PR does**:
1. Added special case for when there are no manifests, delete all resources when prune `enables` works.

2. Added a test case for no manifests 

**Why we need it**:
#5413 

The current logic was not explicitly handling, When there were no manifests

Fixes #5413 

- **How are users affected by this change**:No
- **Is this breaking change**:No

### DIFF
--- a/pkg/app/piped/executor/kubernetes/sync.go
+++ b/pkg/app/piped/executor/kubernetes/sync.go
@@ -124,6 +124,15 @@ func (e *deployExecutor) ensureSync(ctx context.Context) model.StageStatus {
 }
 
 func findRemoveResources(manifests []provider.Manifest, liveResources []provider.Manifest) []provider.ResourceKey {
+	// Special case of when no manifests exists
+	if len(manifests) == 0 {
+		// If there are no manifests in Git, all live resources should be removed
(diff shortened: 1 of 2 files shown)


vaidikcode:
I was thinking that we could add a `purgeResource` flag to the `DeleteApplication` endpoint, which would retrieve all resources with application labels or identifiers and then delete those resources directly, regardless of their git state. wdyt?

t-kikuc:
@vaidikcode 

Please try this issue after understanding this: https://github.com/pipe-cd/pipecd/issues/4269
This is really complicated problem.

Besides, please say something on https://github.com/pipe-cd/pipecd/issues/5413 before opening a PR, and clarify what to do.

PR #5844, opened 2025-05-17
Add a button to login into other project from the dropdown

### LINKED ISSUE #5795: Better way to know how to switch between PipeCD projects

If you're using PipeCD to handle deployments for multiple projects, you can switch between them by clicking to the project name on the top left of the PipeCD web console.

Image

We should have a better way to show this feature. An idea is to add `Login to other project` to the list that show in the dialog when click to the project name

**Why is this needed**:

kiratlog:
**What this PR does**:

**Why we need it**:

Fixes #5795
Screenshot from 2025-05-17 09-11-35

### DIFF
--- a/web/src/components/header/index.tsx
+++ b/web/src/components/header/index.tsx
@@ -2,6 +2,7 @@ import {
   AppBar,
   Avatar,
   Button,
+  Divider,
   IconButton,
   Link,
   Menu,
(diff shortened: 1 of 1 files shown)


t-kikuc:
Thank you so much for your great contribution and fast development!

Your code itself seems good, but let's discuss here at first!: 
https://github.com/pipe-cd/pipecd/issues/5795#issuecomment-2889622559

If the conclusion is the same as your current code, let us review again.

PR #5851, opened 2025-05-19
Fix the workflow for non-maintainer

Warashi:
**What this PR does**:

Fix the greeting job to work as expected.

**Why we need it**:

I added a greeting workflow job in #5846, but it doesn't work as expected due to the bash's ` being used as command execution in the heredoc.
https://github.com/pipe-cd/pipecd/actions/runs/15106980054/job/42457646974

Follows #5846 

### DIFF
--- a/.github/workflows/non_maintainer_check.yaml
+++ b/.github/workflows/non_maintainer_check.yaml
@@ -26,8 +26,9 @@ jobs:
         PR_NUMBER: ${{ github.event.pull_request.number }}
         TEMP_DIR: ${{ runner.temp }}
       run: |
-        cat <<EOF > "${TEMP_DIR}/greeting.md"
-        @${AUTHOR}
+        echo "@${AUTHOR}" > "${TEMP_DIR}/greeting.md"
+
+        cat <<'EOF' >> "${TEMP_DIR}/greeting.md"
         Thank you for your contribution! We look forward to seeing more from you.
         Please run the `make check` command to ensure your changes will pass the CI.
         After successfully running the command on your local machine, the instructions will be printed out.
--- a/.github/workflows/non_maintainer_check.yaml
+++ b/.github/workflows/non_maintainer_check.yaml
@@ -47,7 +47,7 @@ jobs:
         Then, please run the `make check` command again to ensure the issues are fixed.
         EOF
 
-        gh pr comment "${PR_NUMBER}" --body-file "${TEMP_DIR}/greeting.md"
+        gh --repo "${REPO_NAME}" pr comment "${PR_NUMBER}" --body-file "${TEMP_DIR}/greeting.md"
 
   check:
     if: github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '/check-commit')
--- a/.github/workflows/non_maintainer_check.yaml
+++ b/.github/workflows/non_maintainer_check.yaml
@@ -89,4 +89,4 @@ jobs:
           Thank you for your contribution!
           EOF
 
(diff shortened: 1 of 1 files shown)


.github/workflows/non_maintainer_check.yaml
```diff
@@ -23,11 +23,22 @@ jobs:
...
-        cat <<EOF > "${TEMP_DIR}/greeting.md"
-        @${AUTHOR}
+        # sometimes the author association of event payload is not correct to handle it
+        # so we fetch the latest author association from the API
+        AUTHOR_ASSOCIATION=$(gh api "/repos/${REPO_NAME}/pulls/${PR_NUMBER}" --jq '.author_association')
+        echo "AUTHOR_ASSOCIATION: ${AUTHOR_ASSOCIATION}"
+        [ "${AUTHOR_ASSOCIATION}" = "COLLABORATOR" ] && exit 0
+        [ "${AUTHOR_ASSOCIATION}" = "MANNEQUIN" ] && exit 0
+        [ "${AUTHOR_ASSOCIATION}" = "MEMBER" ] && exit 0
+        [ "${AUTHOR_ASSOCIATION}" = "OWNER" ] && exit 0
```
Warashi, line 37: I found this job ran on this PR. It's my PR, and I'm a MEMBER, so it's not expected to run.
So I added these double-check lines.
https://github.com/pipe-cd/pipecd/actions/runs/15107247811

Warashi:
/check-commit 2cfb2efc11b4b05226811c72a865fabb77a2e7f75dd7fcc451d3af62982bf9a7

Warashi:
The check job also fails 😇 
I'll fix it in this PR
https://github.com/pipe-cd/pipecd/actions/runs/15107661232/job/42459725663

.github/workflows/non_maintainer_check.yaml
```diff
@@ -78,4 +89,4 @@ jobs:
           Thank you for your contribution!
           EOF
 
-          gh pr comment "${PR_NUMBER}" --body-file "${TEMP_DIR}/comment.md"
+          gh --repo "${REPO_NAME}" pr comment "${PR_NUMBER}" --body-file "${TEMP_DIR}/comment.md"
```
Warashi, line 92: I checked the gh command behavior outside the git repository and found that we need the `--repo` option.

Warashi:
@ffjlabo 
Thank you for your suggestion.

> IMO, how about re-checking whether to add a comment in this PR by commenting out if statement temporarily?

It sounds nice if we can do so, but I think we cannot.
This workflow runs on the pull_request_target event.
It runs on the base branch of the pull request, not the head.
This means we cannot run the job with the modification in this PR.

.github/workflows/non_maintainer_check.yaml
```diff
@@ -36,7 +47,7 @@ jobs:
         Then, please run the `make check` command again to ensure the issues are fixed.
         EOF
 
-        gh pr comment "${PR_NUMBER}" --body-file "${TEMP_DIR}/greeting.md"
+        gh --repo "${REPO_NAME}" pr comment "${PR_NUMBER}" --body-file "${TEMP_DIR}/greeting.md"
```
Warashi, line 50: I noticed we need -the `--repo` option here, too.

ffjlabo:
@Warashi 
> It runs on the base branch of the pull request, not the head.
This means we cannot run the job with the modification in this PR.

I got it, thank you.

t-kikuc (approved):

Issue #5615, opened 2025-03-03
[epic] Documentation tasks of UserGuide for the plugin arch


t-kikuc:
**What would you like to be added**:

This issue tracks tasks of documentation of the user guide to the pipedv1.
When a new task or PR emerges, please add/link it to this PR.

**Tasks for the plugin developer guide should be managed in another PR**

**Migration docs tasks should be managed in https://github.com/pipe-cd/pipecd/issues/5542**

**Why is this needed**:

- To correct info (e.g. **app config**, **piped config**, 'kind' deprecated, UI, etc.)
- To optimize the structure for searchability (e.g. the current [user-guide/configuration-reference](https://pipecd.dev/docs-v0.50.x/user-guide/configuration-reference/) contains too many configs including plugins specific ones)

---

# Tasks

- [ ] Installation
- [ ] UserGuide/
    - [ ] Managing Application/
    - [ ] Managing Piped/
    - [ ] [Examples](https://pipecd.dev/docs-v0.50.x/user-guide/examples/)
- [ ] [Examples Repo](https://pipecd.dev/docs-v0.50.x/examples/)

- [ ] Feature Status

ubermensch19:
@t-kikuc , is it completed?

t-kikuc:
no

Ayushmore1214:
@t-kikuc I would like to take this one , can you let me know which one will be suitable for me ?

PR #5853, opened 2025-05-19
Add README for pipedv1

ffjlabo:
**What this PR does**:

as title
It is almost based on [the one for pipedv0](https://github.com/pipe-cd/pipecd/blob/master/cmd/piped/README.md), but I changed the piped config.

**Why we need it**:

We want to describe how to run pipedv1.

Part of #5615

### DIFF
--- a/cmd/pipedv1/README.md
+++ b/cmd/pipedv1/README.md
@@ -1,0 +1,77 @@
+# Piped Agent for plugin architecture
+
+See [Overview of the Plan for Pluginnable PipeCD](https://pipecd.dev/blog/2024/11/28/overview-of-the-plan-for-pluginnable-pipecd/) to understand what's PipeCD plugin.
+
+
+## Prerequisites
+
+- [Go 1.24 or later](https://go.dev/)
+
+## Repositories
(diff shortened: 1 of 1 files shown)


cmd/pipedv1/README.md
```diff
@@ -0,0 +1,77 @@
...
+        url: file:///path/to/.piped/plugins/kubernetes # It's OK using any value for now because it's a dummy. We will implement it later.
+        deployTargets: # This field is alternative for platform providers
+        - name: kubernetes-dev
+          config:
+            masterURL: https://127.0.0.1:61337 # FIXME: shown by kubectl cluster-info
+            # FIXME: Replace here with your kubeconfig absolute file path.
+            kubeConfigPath: /path/to/.kube/config
+    ```
+
+3. Ensure that your `kube-context` is connecting to the right kubernetes cluster
```
t-kikuc, line 69: TODO:  This line should be modified or removed because it is valid only when the Control Plane runs as a K8s cluster.
(pipedv0 too)

t-kikuc (approved):

PR #5862, opened 2025-05-22
Set install path for setup-envtest

Warashi:
**What this PR does**:

as title

**Why we need it**:

Without this PR, setup-envtest is installed under the `$HOME/go/bin` or user-defined GOBIN directory.
The tests fail when the user does not include it in the PATH variable.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,9 +270,9 @@ kind-down:
 	kind delete cluster --name pipecd
 
 .PHONY: setup-envtest
-setup-envtest: export GOBIN ?= ${PWD}/.dev/bin # Where to install the setup-envtest binary
-setup-envtest: ## Download setup-envtest locally if necessary.
-	test -s $(GOBIN)/setup-envtest || go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+setup-envtest: export GOBIN ?= ${PWD}/.dev/bin
+setup-envtest: ## Download setup-envtest locally if necessary. The setup-envtest binary is installed under the ${PWD}/.dev/bin directory.
+	test -x $(GOBIN)/setup-envtest || go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 # Check commands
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,8 +93,11 @@ test/go: COVERAGE ?= false
 test/go: COVERAGE_OPTS ?= -covermode=atomic
 test/go: COVERAGE_OUTPUT ?= coverage.out
 test/go: setup-envtest
-test/go: ENVTEST_BIN ?= ${PWD}/.dev/bin # We need an absolute path for setup-envtest
-test/go: KUBEBUILDER_ASSETS ?= "$(shell setup-envtest use --bin-dir $(ENVTEST_BIN) -p path)"
+# Where to find the setup-envtest binary
+test/go: GOBIN ?= ${PWD}/.dev/bin
+# We need an absolute path for setup-envtest
+test/go: ENVTEST_BIN ?= ${PWD}/.dev/bin
```
Warashi, line 99: If we put the comments on the same line with the variable definition, the variable has trailing space with its value 😇

Makefile
```diff
@@ -269,8 +272,10 @@ kind-down:
 	kind delete cluster --name pipecd
 
 .PHONY: setup-envtest
+# Where to install the setup-envtest binary
+setup-envtest: export GOBIN ?= ${PWD}/.dev/bin
 setup-envtest: ## Download setup-envtest locally if necessary.
-	test -s $(GOBIN)/setup-envtest || go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	test -x $(GOBIN)/setup-envtest || go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
Warashi, line 278: I changed the test option to check whether the binary is executable.

khanhtc1202 (approved):
Nice catch 👍

ffjlabo (approved):
Thank you!

Issue #5867, opened 2025-05-23
Make SDK as a separate go module


Warashi:
**What would you like to be added**:
 
Make SDK as a separate go module from pipe-cd/pipecd.

**Why is this needed**:

We want to do versioning SDK separately from pipecd itself

t-kikuc:
## Tasks: 

- [ ] 1. Copy `toolregistry`, `signalhandler`, `logpersister`, `diff`, and `pipedapi` under `pkg/plugin/sdk` (not move)
- [ ] 2. Place the contents of `pkg/plugin/sdk/pipedapi` directly under `sdk` and merge `pipedapi.Client` into `sdk.Client`
- [ ] 3. Run `go mod init github.com/pipe-cd/pipecd/pkg/plugin/sdk` in `pkg/plugin/sdk`
- [ ] 4. Run `go mod tidy` in `pkg/plugin/sdk`
- [ ] 5. Create a PR at this point and merge it into master
- [ ] 6. Run `GOPROXY=direct go get github.com/pipe-cd/pipecd/pkg/plugin/sdk@master` at the root of the repository
- [ ] 7. Modify the import path in the plugins
    - The plugins imports only under `pkg/plugin/sdk`.
- [ ] 8. After all modifications are complete, delete `toolregistry`, `signalhandler`, `logpersister`, `diff`, and `pipedapi` from `pkg/plugin`.
    - Only `api`, `sdk`, and `pipedservice` remain under `pkg/plugin`.

PR #5876, opened 2025-05-26
Split SDK and plugins as separate module

Warashi:
**What this PR does**:

as title

**Why we need it**:

To version them separately from PipeCD main module.

Part of #5867 


ffjlabo (approved):
👍

PR #5885, opened 2025-05-28
[SDK] Use new import paths and remove unused

Warashi:
**What this PR does**:

- Use new import paths for the SDK in the SDK itself
- Remove toolregistry, signalhandler, logpersister, diff, and pipedapi from pkg/plugin.

**Why we need it**:

- To make separate SDK from main module

Part of #5867 


khanhtc1202 (approved):
Great 👍

Issue #5868, opened 2025-05-23
Replace web redux with context and tanStack Query


hongky-1994:
**What would you like to be added**:

TODO:
- [x] Install Tanstack/query https://github.com/pipe-cd/pipecd/pull/5893
- [x] Replace Redux in pages
   - [x] Login pages https://github.com/pipe-cd/pipecd/pull/5893
   - [x] Aplication pages
      - [x] Application list page https://github.com/pipe-cd/pipecd/pull/6034
      - [x] Application detail page https://github.com/pipe-cd/pipecd/pull/6025
   - [x] Deployment pages
      - [x] Deployment list page https://github.com/pipe-cd/pipecd/pull/6012
      - [x] Deployment detail page https://github.com/pipe-cd/pipecd/pull/5972
   - [x] Deployment Trace Page https://github.com/pipe-cd/pipecd/pull/5955
   - [x] Setting pages
      - [x] Setting pipes https://github.com/pipe-cd/pipecd/pull/5954
      - [x] Setting project https://github.com/pipe-cd/pipecd/pull/5941
      - [x] Setting api-keys https://github.com/pipe-cd/pipecd/pull/5933
   - [x] Insights page https://github.com/pipe-cd/pipecd/pull/5915
   - [x] Events page https://github.com/pipe-cd/pipecd/pull/5928
- [x] Remove redux package https://github.com/pipe-cd/pipecd/pull/6046

**Why is this needed**:
- Redux make web app complexity and heavily boilerplate. We should simplify the code base by apply modern alternatives like React Context, hooks, and [React-query](https://tanstack.com/query/latest).
- RFCS: [docs/rfcs/0019-redux-replacement.md](https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0019-redux-replacement.md)

khanhtc1202:
Done, thanks @hongky-1994 for the works on this 💯

PR #5893, opened 2025-05-28
Install react query v4 and add auth context

hongky-1994:
**What this PR does**:

- Install react query v4
- Add Auth Context 

**Why we need it**:
- AuthContext to replace module/me in redux storage.
- We use react query v4 instead of v5 due to dependency conflict. React-Query v5 only support react 18+, whereas v4 is support for react 16.8+ which is more suitable for our current setting (react 17).  

Partof #5868


t-kikuc (approved):
Migrated from me to contexts/, I got it.

Warashi (approved):
LGTM
I checked that this works on my local.

PR #5895, opened 2025-05-29
Rename completed job to avoid confusing behavior

Warashi:
**What this PR does**:

Rename GitHub Action Job

**Why we need it**:

We set `go-completed` as the required check, but this condition passes at one of the lint/test passes and the other is not running yet.
This behavior is not expected; we want to set both of the lint/test as required.

Fixes #


t-kikuc (approved):

khanhtc1202 (approved):
👌

PR #5896, opened 2025-05-29
[Plugins] Remove replace directive from go.mod

Warashi:
**What this PR does**:

Remove the replace directive from the go.mod of each plugin.

**Why we need it**:

To use published SDK

Part of #5867 


t-kikuc (approved):
I got it

This commit: 
https://github.com/pipe-cd/piped-plugin-sdk-go/commit/8bd42026e370511fc1064b92352e54ff123eb8d9

PR #5900, opened 2025-05-29
Fix removal of existing cloudrun revision tags when deploying

### LINKED ISSUE #4954: Tags Assigned to Existing Revisions are Removed when Deploying CloudRun with PipeCD
**What happened**:

When deploying a new revision of CloudRun with PipeCD, tags assigned to existing revisions are removed. Specifically, when tags are allocated to an existing revision of CloudRun and a new deployment is carried out via PipeCD, the tags that were previously assigned to the existing revision are removed.
I added the following code to the `spec.pipeline.stages` in `app.pipecd.yaml` and redeployed. As a result, the traffic to the revision deployed by PipeCD was 0%, but similar to the previous case, the tags assigned before the deployment were removed from the existing revision.

```yaml
pipeline:
  stages:
    - name: CLOUDRUN_PROMOTE
      with:
        percent: 0
```

Deploying CloudRun with PipeCD should not affect the tags assigned to existing revisions.

**How to reproduce it**:

1. Assign tags to the existing revision of the target CloudRun.
2. Deploy a new revision of the target CloudRun using PipeCD.
3. Observe that the previously assigned tags are removed from the existing revision.

**Environment**:
- `piped` version: v0.47.2
- `control-plane` version:
- Others:

niladrix719:
I have investigated about the problem, https://github.com/pipe-cd/pipecd/blob/c1ca195128301d0b12ad69971c28b8c702e7aacb/pkg/app/piped/platformprovider/cloudrun/client.go#L94-L96

The issue was that when calling  `svc.ReplaceService(name, svcCfg)`, the tags associated with revisions in the traffic configuration were being lost.

To fix this, we can retrieve the current service configuration, extracts the existing revision tags from its traffic configuration, and then copy these tags to the corresponding revisions in the new service configuration before calling ReplaceService.

But still, the tags of outdated revisions would be removed, as we are only keeping track of the new revision and the last deployed revision.

https://github.com/pipe-cd/pipecd/blob/c1ca195128301d0b12ad69971c28b8c702e7aacb/pkg/app/piped/executor/cloudrun/deploy.go#L188-L197

To solve this, we can mark the outdated revisions that have tags with 0% traffic, so it can replicate the behavior seen in Cloud Run

**What this PR does**:

- Fixes the removal of existing Cloud Run revision tags during deployment
- Adds tests for the new functionality

Screenshot 2025-05-29 at 7 19 43 PM

**Why we need it**:

Revision tags created by the user should not be removed

Fixes #4954 


niladrix719:
/check-commit f944fdff91e0f94d15656571aa3dc07df9b40949c4e8e8850fbf09d3a301d105

t-kikuc:
@niladrix719 
Hi, Thank you so much.

At first, I recommend you split into functions and add tests.

niladrix719:
Test cases and splitting function are done. Let me know if there is anything else

khanhtc1202:
Let's me hold this one and try add this while implementing the CloudRun plugin 👀

Issue #5901, opened 2025-05-30
Implement `SCRIPT_RUN` Stage Plugin


t-kikuc:
** What **

- Implement the SCRIPT_RUN stage plugin like the WaitStage plugin for pipedv1

** Why **

- To support ScriptRun in pipedv1 too

hiep-tk:
Todo list:

- [x] [update sdk deployment model to get access to fields required by scriptrun](https://github.com/pipe-cd/pipecd/pull/5902)
- [x] [update sdk execute stage request model, allow scriptrun and its rollback stage to communicate and match](https://github.com/pipe-cd/pipecd/pull/5910)
- [ ] port scriptrun to plugin, add rollback

hiep-tk:
Scriptrun Rollback Implementation
### Proposal 1: One rollback stage for each scriptrun stage
1. When planning, for each scriptrun stage, add one rollback stage with same index (sharing ```stageConfig```)
2. When a scriptrun stage run, mark its index (pseudo id) in persistence storage with ```PutDeploymentPluginMetadata```
3. When a rollback happens, each rollback stage will check whether its index exists in the persistence storage with ```GetDeploymentPluginMetadata``` before starting.
- Pro: better visualization when drawing the deployment flow graph
- Con:  
    - need to update sdk for more stage info exposure as stage requires index when running. 
    - async communication between piped and plugin for several stages.
### Proposal 2: Single rollback stage
1. When planning, only add one rollback stage.
2. When a scriptrun stage run, append its (env config, rollback command) to the persistence storage.
3. When a rollback happens, the rollback stage will read all the needed (env config, rollback command) and run them in a single go.
- Pro: centralize rollback ops
- Con: may make debug harder (individual rollback can fail,)

Warashi:
IMO, we can use the `StageConfig` field of [ExecuteStageRequest](https://pkg.go.dev/github.com/pipe-cd/piped-plugin-sdk-go#ExecuteStageRequest) in Proposal 1; it may hold configs from the stage index, which is the same as the "normal" stage. The stage index of the rollback stage will be the same as the "normal" stage so that we can retrieve config from this field.

https://github.com/pipe-cd/pipecd/blob/1ac2006f28fde90f53cc6817edd37dc188e597a5/pkg/app/pipedv1/controller/scheduler.go#L539-L547

t-kikuc:
I think [Proposal 1] is better because that is the same as the current spec.
[...]

PR #5902, opened 2025-05-30
add more fields from deployment proto to deployment model for scriptrun plugin

hiep-tk:
**What this PR does**: add repositoryURL, summary, and labels available from deployment proto to deployment model

**Why we need it**: scriptrun env tagging utilize this. need to be a separate commit because we need to have a new concrete piped-plugin-sdk-go version to point to from the plugin

**Which issue(s) this PR fixes**: Part of #5901


t-kikuc (approved):
Go!

ffjlabo (approved):
Sorry to be late. LGTM

PR #5908, opened 2025-06-02
Revert "Do treeless clone for git clone to improve performance"

Warashi:
Reverts pipe-cd/pipecd#5722

The eventwatcher calls `repo.CopyToModify` every time, regardless of whether the event is registered.
This causes git clone from remote, e.g., GitHub, every time and costs network I/O.

The solution might be to call it `CopyToModify` only when the modification occurs.
For now, I revert the treeless clone because I don't have enough time to do so.

https://github.com/pipe-cd/pipecd/blob/24a573d32640e2dc78835af3ec161023cd07017d/pkg/app/piped/eventwatcher/eventwatcher.go#L322-L336


ffjlabo (approved):
Thank you

PR #5910, opened 2025-06-02
expose stage info created by buildpipeline phase to executestage phase. add null check for getDeploymentPluginMetadata

hiep-tk:
**What this PR does**: as title

**Why we need it**: currently the script_run stages need more context to publish their stage already run status to the script_run_rollback so multiple rollbacks can be differentiated and selectively run

**Which issue(s) this PR fixes**: Part of #5901 

Fixes #


pkg/plugin/sdk/client.go
```diff
@@ -150,13 +150,16 @@ func (c *Client) PutStageMetadataMulti(ctx context.Context, metadata map[string]
...
 		DeploymentId: c.deploymentID,
 		PluginName:   c.pluginName,
 		Key:          key,
 	})
-	return resp.Value, err
+	if err != nil {
+		return "", false, err
+	}
+	return resp.Value, resp.Found, err
 }
```
t-kikuc: Please split the PR since this does not seem to be directly related to SCRIPT_RUN.

And please make consistent with other GetXxxMetadata() funcs.

t-kikuc (changes requested):
Please describe your design on the issue in advance and submit the next draft PR because I cannot imagine what you are going to do.

pkg/plugin/sdk/client.go
hiep-tk: currently this getxxxMetadata function's error checking is missing, which would cause nil pointer deref panic when the grpc call response with error. About the return signature of this format, it currently only extracts value property and not the found one, which would make it hard to differentiate, for example when the metadata is stored with empty value. Do you think the signature should stay as before, or as updated? @t-kikuc

t-kikuc:
@hiep-tk 
Let's restart as a new PR!

PR #5915, opened 2025-06-03
Remove Redux in Insight page

hongky-1994:
**What this PR does**:
- This MR depends on [!5893](https://github.com/pipe-cd/pipecd/pull/5893)
-   remove redux modules/insight
-   remove redux modules/deployment-change-failure-rate
-   remove redux modules/deployment-frequency
-   add query for insight
-   add query for app list , piped list
-   add toast messages context
-   update query provider with error handler show toast

**Why we need it**:

- remove redux module from insight page

Partof #5868


khanhtc1202 (dismissed):
Here you go 👍

ffjlabo:
@hongky-1994 
I have two questions about this PR. Could you check them?

**1. Applications by piped show the different values for the two cases below. Is this related to the fix in this PR?**

- The view on the local with the current PR branch
PipeCD

- The view on https://demo.pipecd.dev/insights?project=dev
PipeCD


**2. On my local check, I found the icon like the image below. Is it for development mode?**
PipeCD

When clicking the icon, we can see something like the image below.
PipeCD

hongky-1994:
Thanks @ffjlabo 
1. **Applications by piped show the different values for the two cases below. Is this related to the fix in this PR?**
- My mistake, I have updated the code to not count deleted apps. The applications by piped should not count deleted app.

2. **On my local check, I found the icon like the image below. Is it for development mode?**
 Yes it is only on development. It is the [react-query devtools](https://tanstack.com/query/v4/docs/framework/react/devtools) support view cache and query state of react-query. In production mode it will be hidden.
 
CleanShot 2025-06-28 at 09 02 32

ffjlabo:
@hongky-1994 Thank you for checking, LGTM!

ffjlabo (approved):

ffjlabo:
@khanhtc1202 Sorry, plz re-approve

PR #5922, opened 2025-06-06
Fix nil pointer dereference error on spec.plugins config on app config not set

Warashi:
**What this PR does**:

as title

**Why we need it**:

for safe execution


t-kikuc (approved):

khanhtc1202 (approved):
Nice catch 👍

ffjlabo (approved):

PR #5926, opened 2025-06-06
Fix the build quick sync stages method to return an empty response

Warashi:
**What this PR does**:

as title

**Why we need it**:

The stage plugins don't support quick sync stages, so it's enough to just return an empty response.

Fixes #

### DIFF
--- a/pkg/plugin/sdk/deployment.go
+++ b/pkg/plugin/sdk/deployment.go
@@ -263,7 +263,7 @@ func (s *StagePluginServiceServer[Config, DeployTargetConfig, ApplicationConfigS
 	return buildPipelineSyncStages(ctx, s.name, s.base, &s.config, client, request, s.logger)
 }
 func (s *StagePluginServiceServer[Config, DeployTargetConfig, ApplicationConfigSpec]) BuildQuickSyncStages(context.Context, *deployment.BuildQuickSyncStagesRequest) (*deployment.BuildQuickSyncStagesResponse, error) {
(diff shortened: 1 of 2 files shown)


ffjlabo (approved):

pkg/plugin/sdk/deployment.go
```diff
@@ -263,7 +263,8 @@ func (s *StagePluginServiceServer[Config, DeployTargetConfig, ApplicationConfigS
 	return buildPipelineSyncStages(ctx, s.name, s.base, &s.config, client, request, s.logger)
 }
 func (s *StagePluginServiceServer[Config, DeployTargetConfig, ApplicationConfigSpec]) BuildQuickSyncStages(context.Context, *deployment.BuildQuickSyncStagesRequest) (*deployment.BuildQuickSyncStagesResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method BuildQuickSyncStages not implemented")
+	// There is no stage, so just return an empty response.
```
khanhtc1202:
```suggestion
	// Return an empty response in case the plugin does not support the QuickSync strategy. 
```
nits

t-kikuc (approved):

pkg/plugin/sdk/deployment.go
Warashi: Thank you, I applied the suggestion.

t-kikuc (approved):

khanhtc1202 (approved):

PR #5929, opened 2025-06-09
Add default response of DetermineStrategy()

t-kikuc:
**What this PR does**:

- `DetermineStrategy()` returns PipelineSync by default
- Describe the spec of `DetermineStrategy()`

cf. pipedv0 uses pipelineSync by default.

https://github.com/pipe-cd/pipecd/blob/a19732a292b9407c8ac03963d71e39a0347329d8/pkg/app/piped/planner/terraform/terraform.go#L103-L115

**Why we need it**:

- Prevent plugin developers from getting confused.  They can just return (nil, nil) for DetermineStrategy().
   - Plugins other than the k8s plugin probably do not have the specific logic.

Part of #5530 

### DIFF
--- a/pkg/plugin/sdk/deployment.go
+++ b/pkg/plugin/sdk/deployment.go
@@ -39,6 +39,8 @@ type DeploymentPlugin[Config, DeployTargetConfig, ApplicationConfigSpec any] int
 	// DetermineVersions determines the versions of the resources that will be deployed.
 	DetermineVersions(context.Context, *Config, *DetermineVersionsInput[ApplicationConfigSpec]) (*DetermineVersionsResponse, error)
 	// DetermineStrategy determines the strategy to deploy the resources.
+	// This is called when the strategy was not determined by common logic, including judging by the pipeline length, whether it is the first deployment, and so on.
+	// It should return (nil, nil) if the plugin does not have specific logic for DetermineStrategy.
 	DetermineStrategy(context.Context, *Config, *DetermineStrategyInput[ApplicationConfigSpec]) (*DetermineStrategyResponse, error)
 	// BuildQuickSyncStages builds the stages that will be executed during the quick sync process.
 	BuildQuickSyncStages(context.Context, *Config, *BuildQuickSyncStagesInput) (*BuildQuickSyncStagesResponse, error)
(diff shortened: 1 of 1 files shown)


ffjlabo:
@t-kikuc 
I think we need to modify the caller logic to handle nil.
This might cause panic because of nil pointer dereference.
https://github.com/pipe-cd/pipecd/blob/a19732a292b9407c8ac03963d71e39a0347329d8/pkg/app/pipedv1/controller/planner.go#L375C1-L388C3

t-kikuc:
@ffjlabo 
I think that should be done in another PR as a controller bug.

In this PR, even if `DetermineStrategy()` of a plugin returns nil, the SDK will not return nil to piped.
(Now, if a plugin returns nil, the SDK will panic.)

ffjlabo:
@t-kikuc 
Sorry, I misunderstood. 🙏  I got your point.

ffjlabo (approved):
👍

pkg/plugin/sdk/deployment.go
```diff
@@ -153,6 +155,13 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig, ApplicationCo
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to determine strategy: %v", err)
 	}
+	if response == nil {
+		// when a plugin does not have specific logic.
```
khanhtc1202:
```suggestion
		// If the plugin does not have specific logic to determine strategy
		// returns PipelineSync as default.
```
nits, the comment should be clear, not half-phase, to reduce confusion.

t-kikuc: @khanhtc1202 
Thanks, I updated!
[c5f6d75](https://github.com/pipe-cd/pipecd/pull/5929/commits/c5f6d759ad71b04fb924312b1818538dab342644)

khanhtc1202 (approved):
Thanks 👍

t-kikuc:
@ffjlablo 
Sorry, please approve again 🙏 
And I opened a PR for the issue you mentioned: https://github.com/pipe-cd/pipecd/pull/5930

PR #5930, opened 2025-06-09
Ignore when DetermineStrategy() is unsupported or returns nil

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

- To handle `Unsupported` flag for plugins like StagePlugin
- To prevent nil panic when the plugin returns nil (unintentionally?)

Part of #4980

### DIFF
--- a/pkg/app/pipedv1/controller/planner.go
+++ b/pkg/app/pipedv1/controller/planner.go
@@ -379,6 +379,10 @@ func (p *planner) buildPlan(ctx context.Context, runningDS, targetDS *common.Dep
 			p.logger.Warn("Unable to determine strategy using current plugin", zap.Error(err))
 			continue
 		}
+		if res == nil {
+			p.logger.Warn("plugin returned nil for DetermineStrategy()")
+			continue
+		}
 		// If the plugin does not support DetermineStrategy(), then ignore.
 		if res.Unsupported {
 			continue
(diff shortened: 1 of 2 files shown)


ffjlabo (approved):
Thank you! LGTM

pkg/app/pipedv1/controller/planner.go
```diff
@@ -379,6 +379,14 @@ func (p *planner) buildPlan(ctx context.Context, runningDS, targetDS *common.Dep
 			p.logger.Warn("Unable to determine strategy using current plugin", zap.Error(err))
 			continue
 		}
+		if res == nil {
+			p.logger.Warn("plugin returned nil for DetermineStrategy()")
+			continue
+		}
```
khanhtc1202: It's probably better to have a res value in all cases (the default value is used if no res is given by the plugin), wdyt?
We can handle that on by sdk.

t-kikuc: It's just for safety.  

Although all plugins SHOULD return a value (I agree with you), we cannot control all plugins.
For example, some plugins might not depend on the SDK.

khanhtc1202: In such case, we have to check nil for all responses from the plugins, which is complicated and costly. It's not good to make this check only available here but not for the others.

khanhtc1202: If the panic caused by the next line derefference Unsupported field from a nil response, then if condition like below could help

```golang
if res == nil || res.Unsupported {
    continue
}
```

t-kikuc: @khanhtc1202 Thank you, I removed the nil check. [0d8ee43](https://github.com/pipe-cd/pipecd/pull/5930/commits/0d8ee435ea7f9c434513a57e521b0127eb41fb55)

Let's do it as a gRPC interceptor 👍

PR #5931, opened 2025-06-09
Pass plugin name from Piped to plugin

t-kikuc:
**What this PR does**:

- Deleted the `name` arg from `sdk.NewPlugin()`
- Instead, pass the name from the piped plugin config

After merging this PR, we need to update the plugins.

**Why we need it**:

- The uniqueness scope of `name` is confusing for plugin developers 
- Currently, users need to get plugin names from each plugin code since they must be matched
    - Now, deploymentSource is looked up by the name defined in the plugin.
      https://github.com/pipe-cd/pipecd/blob/8dbe938c3e948dba12031a56c2473562df5e2254/pkg/plugin/sdk/deployment_source.go#L42-L51

Part of #5530 

This is a breaking change of the SDK. Not a user-facing change.

### DIFF
--- a/pkg/plugin/sdk/plugin.go
+++ b/pkg/plugin/sdk/plugin.go
@@ -102,9 +102,11 @@ func WithLivestatePlugin[Config, DeployTargetConfig, ApplicationConfigSpec any](
 // Plugin is a wrapper for the plugin.
 // It provides a way to run the plugin with the given config and deploy target config.
 type Plugin[Config, DeployTargetConfig, ApplicationConfigSpec any] struct {
+
 	// plugin info
-	name    string
 	version string
+	// name is the name of the plugin defined in the piped plugin config.
+	name string
 
 	// plugin implementations
 	stagePlugin      StagePlugin[Config, DeployTargetConfig, ApplicationConfigSpec]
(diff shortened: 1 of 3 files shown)


pkg/plugin/sdk/plugin.go
```diff
@@ -153,7 +154,7 @@ func NewPlugin[Config, DeployTargetConfig, ApplicationConfigSpec any](name, vers
 // Run runs the plugin.
 func (p *Plugin[Config, DeployTargetConfig, ApplicationConfigSpec]) Run() error {
 	app := cli.NewApp(
-		fmt.Sprintf("pipecd-plugin-%s", p.name),
```
t-kikuc, line 156: This name is not important. And `p.name` is empty when showing this message.

pkg/plugin/sdk/plugin.go
```diff
@@ -172,10 +173,12 @@ func (p *Plugin[Config, DeployTargetConfig, ApplicationConfigSpec]) Run() error
 func (p *Plugin[Config, DeployTargetConfig, ApplicationConfigSpec]) command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "start",
-		Short: fmt.Sprintf("Start running a %s plugin.", p.name),
```
t-kikuc, line 175: This name is not important. And `p.name` is empty when showing this message.

ffjlabo (approved):
👍

khanhtc1202 (approved):
Great catch, thank you 👍

PR #5928, opened 2025-06-09
Redux replace events page

hongky-1994:
**What this PR does**:

- This MR depends on https://github.com/pipe-cd/pipecd/pull/5915
- remove redux modules/event
- add query for get event infinites
- Update test

**Why we need it**:

- remove redux module from event page
 
**Which issue(s) this PR fixes**:

Partof #5868

**Does this PR introduce a user-facing change?**: no

- **How are users affected by this change**: no
- **Is this breaking change**: no
- **How to migrate (if breaking change)**: no

PR #5933, opened 2025-06-09
Redux replace setting api keys page

hongky-1994:
**What this PR does**:

- This MR depends on https://github.com/pipe-cd/pipecd/pull/5928
- remove redux modules/api-keys
- add query hook for get api keys
- add mutation hook for generate key, disabled key
- Update test

**Why we need it**:

- remove redux module from settings/api-keys page

Partof #5868


khanhtc1202 (approved):
Thanks 👍

khanhtc1202:
Confirmed the behaviour as same as current on dev env 👀

t-kikuc (approved):

PR #5935, opened 2025-06-10
[Plugin] Remove plugin name from the arg when calling `sdk.NewPlugin`

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

We need to support `--name` option for each plugin to pipedv1 start them from #5931.

Follow #5931

### DIFF
--- a/pkg/app/pipedv1/plugin/example/go.mod
+++ b/pkg/app/pipedv1/plugin/example/go.mod
@@ -2,7 +2,7 @@ module github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/example
 
 go 1.24.1
 
-require github.com/pipe-cd/piped-plugin-sdk-go v0.0.0-20250610110220-8b7d2c785ded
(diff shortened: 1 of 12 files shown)


t-kikuc (dismissed):
Thank you! it's so helpful

pkg/app/pipedv1/cmd/piped/piped.go
```diff
@@ -663,7 +663,6 @@ func (p *piped) runPlugins(ctx context.Context, pluginsCfg []config.PipedPlugin,
 			return nil, fmt.Errorf("failed to prepare plugin %s config: %w", pCfg.Name, err)
 		}
 		args = append(args, "--config", string(b))
-		args = append(args, "--name", pCfg.Name)
```
t-kikuc: It might be better to `git pull origin master`??
I guess this diff will be removed.

Warashi (approved):
LGTM

PR #5936, opened 2025-06-11
Use PipedPlugin.Name instead of passing --name in the SDK

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

`--name` is not necessary.

This PR is simpler than #5931

Follows #5931

### DIFF
--- a/pkg/plugin/sdk/plugin.go
+++ b/pkg/plugin/sdk/plugin.go
@@ -177,8 +177,6 @@ func (p *Plugin[Config, DeployTargetConfig, ApplicationConfigSpec]) command() *c
 		RunE:  cli.WithContext(p.run),
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/plugin/sdk/plugin.go
```diff
@@ -204,7 +201,6 @@ func (p *Plugin[Config, DeployTargetConfig, ApplicationConfigSpec]) run(ctx cont
 		// When this happens, it means that there is a bug in the SDK, because these are private fields.
 		input.Logger.Error(
 			"something went wrong in the SDK, please report this issue to the developers",
-			zap.String("name", p.name),
```
t-kikuc, line 207: p.name is empty here.

ffjlabo (approved):

PR #5937, opened 2025-06-11
[SDK] Allow users to set nothing config for the plugin in the app.pipecd.yaml

ffjlabo:
**What this PR does**:

as ttile

**Why we need it**:

There is a case when users deploy without any plugin config in app.pipecd.yaml.

Follow #5922

### DIFF
--- a/pkg/plugin/sdk/deployment_source.go
+++ b/pkg/plugin/sdk/deployment_source.go
@@ -111,15 +111,15 @@ func (c *ApplicationConfig[Spec]) parsePluginConfig(pluginName string) error {
 		c.pluginConfigs = nil
 	}()
 
-	if c.pluginConfigs == nil || c.pluginConfigs[pluginName] == nil {
-		// No config is set for this plugin.
-		// Set the Spec to the zero value of the spec type to avoid nil pointer dereference.
-		c.Spec = new(Spec)
-		return nil
+	// It is necessary to prepare config with default value when users doesn't set any config.
+	data := []byte("{}")
+
+	if c.pluginConfigs != nil && c.pluginConfigs[pluginName] != nil {
+		data = c.pluginConfigs[pluginName]
 	}
 
 	var spec Spec
-	if err := json.Unmarshal(c.pluginConfigs[pluginName], &spec); err != nil {
+	if err := json.Unmarshal(data, &spec); err != nil {
 		return fmt.Errorf("failed to unmarshal application config: plugin spec: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/plugin/sdk/deployment_source.go
```diff
@@ -111,15 +111,15 @@ func (c *ApplicationConfig[Spec]) parsePluginConfig(pluginName string) error {
 		c.pluginConfigs = nil
 	}()
 
-	if c.pluginConfigs == nil || c.pluginConfigs[pluginName] == nil {
-		// No config is set for this plugin.
-		// Set the Spec to the zero value of the spec type to avoid nil pointer dereference.
-		c.Spec = new(Spec)
-		return nil
+	// It is necessary to prepare config with default value when users doesn't set any config.
```
Warashi: It's not only default config but also custom unmarshalling logic.
The point is to call `json.Unmarshal` to invoke custom unmarshalling logic.

pkg/plugin/sdk/deployment_source.go
```diff
@@ -128,7 +128,7 @@ func (c *ApplicationConfig[Spec]) parsePluginConfig(pluginName string) error {
 	}
 
 	// Validate the spec if it implements the Validate method.
-	if v, ok := any(spec).(interface{ Validate() error }); ok {
+	if v, ok := any(&spec).(interface{ Validate() error }); ok {
```
Warashi: Is this the same as just below?
`parsePluginConfig` checks both `Spec` and `*Spec` are implement `Validate() error`.

ffjlabo: It's a mistake. Thanks, fixed on [b5c2fa2](https://github.com/pipe-cd/pipecd/pull/5937/commits/b5c2fa2ee16cc8e496712f3df02fb323d7ec0add)

pkg/plugin/sdk/deployment_source.go
ffjlabo: @Warashi 
Thank you for the commnet. I fixed the comment. 
Is it ok?
[01b8d51](https://github.com/pipe-cd/pipecd/pull/5937/commits/01b8d517e7322fee785f6b38d9944008d44b06ea)

pkg/plugin/sdk/deployment_source.go
```diff
@@ -111,15 +111,15 @@ func (c *ApplicationConfig[Spec]) parsePluginConfig(pluginName string) error {
 		c.pluginConfigs = nil
 	}()
 
-	if c.pluginConfigs == nil || c.pluginConfigs[pluginName] == nil {
-		// No config is set for this plugin.
-		// Set the Spec to the zero value of the spec type to avoid nil pointer dereference.
-		c.Spec = new(Spec)
-		return nil
+	// It is necessary to prepare config with default value when users doesn't set any config, or when developers implements custom unmarshalling logic.
```
t-kikuc: [nit] to clarify
```suggestion
	// It is necessary to prepare config with default value when users don't set any config, or when plugin developers implement custom unmarshalling logic.
```

ffjlabo: Thanks, fixed [5f2f453](https://github.com/pipe-cd/pipecd/pull/5937/commits/5f2f4535fff4a90d76b6b31d4d844db3353962be)

t-kikuc (approved):
Thank you

Warashi (approved):
LGTM

PR #5939, opened 2025-06-11
Deprecate `ArtifactVersion.kind`

t-kikuc:
**What this PR does**:

- removed `sdk.ArtifactVersion.Kind`
- deprecate `model.ArtifactVersion.kind` (not deleted)

After merging this PR, we need to update the official plugins.

Let's delete `model.ArtifactVersion.kind` later. For now, we need it for backward compatibility.

**Why we need it**:

- to clean unused `kind` from the SDK

Fixes https://github.com/pipe-cd/piped-plugin-sdk-go/issues/3


khanhtc1202 (approved):
LGTM 👌

PR #5940, opened 2025-06-11
Get only livestate-supported plugin clients when flushing the livestate

ffjlabo:
**What this PR does**:

This is the fix for pluginRegistry's logic.

- Fixed to check whether the plugin server supports `LivestateService` in advance, by using gRPC reflection.
- Get only livestate-supported plugin clients when flushing the livestate.

**Why we need it**:

Currently, the livestate reporter fails to report when there are some plugins that don't support the livestate feature. (e.g. wait stage).

Livestate reporter calls `GetLivestate` for each plugin when flushing an app.
But currently, it fails when one of them has an error, including an unimplemented status.

Fixes #


ffjlabo:
I found the simple alternative way. So close this PR.

PR #5942, opened 2025-06-12
Ignore unsupported plugin when reporting livestate

ffjlabo:
**What this PR does**:

Get only livestate-supported plugin clients when flushing the livestate.

Currently, the livestate reporter fails to report when there are some plugins that don't support the livestate feature. (e.g. wait stage).

Livestate reporter calls GetLivestate for each plugin when flushing an app.
But currently, it fails when one of them has an error, including an unimplemented status.

**Why we need it**:

Follow https://github.com/pipe-cd/pipecd/pull/5940
Part of #5363

### DIFF
--- a/pkg/app/pipedv1/livestatereporter/livestatereporter.go
+++ b/pkg/app/pipedv1/livestatereporter/livestatereporter.go
@@ -27,6 +27,8 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/deploysource"
 	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin"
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/livestatereporter/livestatereporter.go
```diff
@@ -230,6 +232,13 @@ func (r *reporter) flush(ctx context.Context, app *model.Application, repo git.R
 			DeployTargets:   app.GetDeployTargets(),
 		})
 		if err != nil {
+			st, ok := status.FromError(err)
+			if ok && st.Code() == codes.Unimplemented {
+				// TODO: show plugin name
+				r.logger.Info("plugin does not support livestate feature")
+				continue
+			}
```
t-kikuc, line 240: When needed in the future (not now),
let's remove such plugins from `pluginClis, err := r.pluginRegistry.GetPluginClientsByAppConfig(cfg.Spec)` by like `r.pluginRegistry.markPluginAsLiveStateUnsupported(xxx)` to reduce unnecessary gRPC calls.

Warashi (approved):
👍🏻

Issue #5259, opened 2024-10-07
PipeCD plugin architecture


khanhtc1202:
**What would you like to be added**:

This issue is the meta (epic) issue that links all issues related to adopting plugin architecture for PipeCD. 
- Piped plugin architecture (v1.0 for piped) supports applications' deployments ref: #4980 
- Make control plane support both piped v0 and v1 (plugin-arched piped) ref: #5252 
- Livestate feature in plugin architecture ref: #5363
- Support stages as plugins https://github.com/pipe-cd/pipecd/issues/5367
- Migrate from pipedv0 -> pipedv1 https://github.com/pipe-cd/pipecd/issues/5542

Related issues:
- Remove deprecated features/codebase from pipedv1  https://github.com/pipe-cd/pipecd/issues/5956

**Why is this needed**:

PR #5946, opened 2025-06-12
Add index validation of buildPipelineSyncStages in planner

t-kikuc:
**What this PR does**:

- Add index validations for duplication and range.
- Add the API spec comments in api.proto and the SDK

**Why we need it**:

- To prevent errors in the [scheduler](https://github.com/pipe-cd/pipecd/blob/d522ec41f5893c4dd1108c7024e06e84933443e8/pkg/app/pipedv1/controller/scheduler.go#L539-L547).
- To make the rollback order determinate
- To keep the rollback spec simple

Part of #5259 


khanhtc1202 (dismissed):
Thanks 👍

ffjlabo (approved):
LGTM

PR #5947, opened 2025-06-12
feat(piped): make planpreview handler configurable via piped config

### LINKED ISSUE #5916: Add a way to configure plan-preview handling

The way to configure plan-preview handling behavior.

**Why is this needed**:

We have some options in the `planpreview` package, but they are not used or are only used in the tests.

https://github.com/pipe-cd/pipecd/blob/d667a1c43fa245519079d106a717fe1e0125021a/pkg/app/piped/planpreview/handler.go#L52-L74

kadai0308:
**What this PR does**:
This PR makes the planpreview handler behavior configurable via the piped.yaml config file. This allows operators to tune performance and resource usage based on deployment scale or workload.

Changes
- Added the following fields to PipedSpec:

- handlerWorkerNum

- handlerCommandQueueBufferSize

- handlerCommandCheckInterval

- handlerCommandHandleTimeout

Passed these values into the planpreview handler using new functional options:

- WithWorkerNum

- WithCommandQueueBufferSize

- WithCommandCheckInterval

- WithCommandHandleTimeout

**Why we need it**:
We have some options in the planpreview package, but they are not used or are only used in the tests.

https://github.com/pipe-cd/pipecd/blob/d667a1c43fa245519079d106a717fe1e0125021a/pkg/app/piped/planpreview/handler.go#L52-L74

Fixes #5916

**Does this PR introduce a user-facing change?** & **How are users affected by this change**: 
User now can add the handler configs.


pkg/config/piped.go
```diff
@@ -80,6 +80,14 @@ type PipedSpec struct {
 	EventWatcher PipedEventWatcher `json:"eventWatcher"`
 	// List of labels to filter all applications this piped will handle.
 	AppSelector map[string]string `json:"appSelector,omitempty"`
+	// worker num for handling commands.
```
t-kikuc, line 83: These fields should be under the new `PlanPreview` struct to simplify the `Piped` struct.

pkg/config/piped_test.go
```diff
@@ -361,6 +361,10 @@ func TestPipedConfig(t *testing.T) {
 						},
 					},
 				},
+				HandlerWorkerNum:              3,
+				HandlerCommandQueueBufferSize: 10,
+				HandlerCommandCheckInterval:   Duration(5 * time.Second),
+				HandlerCommandHandleTimeout:   Duration(5 * time.Minute),
```
t-kikuc, line 367: please add them to the test file too.
it's better than default values.

pkg/config/piped.go
```diff
@@ -80,6 +80,14 @@ type PipedSpec struct {
...
+	HandlerWorkerNum int `json:"handlerWorkerNum,omitempty" default:"3"`
+	// The maximum number of commands that can be queued for handling.
+	HandlerCommandQueueBufferSize int `json:"handlerCommandQueueBufferSize,omitempty" default:"10"`
+	// How often to check whether there are commands to handle.
+	HandlerCommandCheckInterval Duration `json:"handlerCommandCheckInterval,omitempty" default:"5s"`
+	// How long to wait for a command to be handled.
+	HandlerCommandHandleTimeout Duration `json:"handlerCommandHandleTimeout,omitempty" default:"5m"`
```
t-kikuc, line 90: I think the default values should be injected in `planpreview/` package, not piped.go, to put planpreview-related logics in one place.

t-kikuc (commented):
thank you.
Basically it's good.

pkg/config/piped.go
t-kikuc: @kadai0308 Hi, any update on this PR?

kadai0308: @t-kikuc Sorry, I forgot to follow up this PR.

What's the path of `planpreview/` package? 
Is `pkg/app/piped/planpreview` ?

pkg/config/piped_test.go
kadai0308: Do you mean to add these value to `pkg/config/testdata/piped/piped-config.yaml` ?

pkg/config/piped.go
t-kikuc: @kadai0308 
> pkg/app/piped/planpreview
yes!

PR #5949, opened 2025-06-13
Add --match opion for `git describe` for `build/chart`

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

When creating the tag whose format is not semantic version, `build/chart` fails because helm chart supports only semantic versioning.
So we decided to allow only semantic versioned one.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ build/web:
 	yarn --cwd web build
 
 .PHONY: build/chart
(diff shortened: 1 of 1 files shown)


Warashi (approved):
👍🏻

t-kikuc (approved):
Nice!!!

PR #5950, opened 2025-06-13
Fix error stages are not shown for plugin-arch piped created pipeline

khanhtc1202:
**What this PR does**:

**Why we need it**:

Fixes #

### DIFF
--- a/web/src/components/deployments-detail-page/pipeline/index.tsx
+++ b/web/src/components/deployments-detail-page/pipeline/index.tsx
@@ -61,8 +61,25 @@ const createStagesForRendering = (
     return [];
   }
 
+
+  console.log("[DEBUG]: deployment.stagesList");
+  deployment.stagesList.forEach((stage) => {
+    console.log(stage.name, stage.visible, stage.rollback, stage.requiresList);
+    // stage.rollback = !stage.visible;
+  });
+
   const stages: Stage[][] = [];
-  const visibleStages = deployment.stagesList.filter((stage) => stage.visible);
+  const notRollbackCount = deployment.stagesList.filter((stage) => !stage.rollback).length;
+  let visibleStages: Stage[] = [];
+  if (notRollbackCount === deployment.stagesList.length) {
+    visibleStages = deployment.stagesList.filter((stage) => stage.visible);
+  } else {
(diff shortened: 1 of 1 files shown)


web/src/components/deployments-detail-page/pipeline/index.tsx
```diff
@@ -54,16 +54,27 @@ const findDefaultActiveStage = (
...
   deployment: Deployment.AsObject | undefined
 ): Stage[][] => {
   if (!deployment) {
     return [];
   }
 
-  const stages: Stage[][] = [];
-  const visibleStages = deployment.stagesList.filter((stage) => stage.visible);
+  let visibleStages: Stage[] = [];
+  if (deployment.deployTargetsByPluginMap?.length === 0) {
```
Warashi: I'm not sure about the optional chaining feature in TypeScript, but I think that the `deployment.deployTargetsByPluginMap?.length` will be `undefined` when the `deployment.deployTargetsByPluginMap` is null/undefined. And `undefined === 0` will be `false`.

Warashi (commented):
I left a comment about the optional chaining feature of TypeScript.

web/src/components/deployments-detail-page/pipeline/index.tsx
khanhtc1202: Got your point 👀 Should I change this to be as same as this one https://github.com/pipe-cd/pipecd/blob/master/web/src/components/deployments-detail-page/deployment-detail/index.tsx#L80
cc @hongky-1994

hongky-1994: Yep, I think you can change like that. 

```
 if (deployment?.deployTargetsByPluginMap?.length) return PIPED_VERSION.V1;
 ```
 
 this would reject both "undefined" and "length === 0"

khanhtc1202: Thanks both, addressed by https://github.com/pipe-cd/pipecd/pull/5950/commits/1460183ce018092d177f69fbacf326c4bbaee51b 🙏

hongky-1994 (approved):
LGTM 👍

PR #5959, opened 2025-06-19
[SDK] add unit types to use in config types

Warashi:
**What this PR does**:

as title

**Why we need it**:

They are used generally in the platform provider's specific configs in pipedv0, and are useful to define config types.
And I want to use them in the Kubernetes plugin.

In the current implementation of the Kubernetes plugin, we use `pkg/configv1`, but it's not a good way to import it.
https://github.com/pipe-cd/pipecd/blob/f39124f9ff935a5bdbb7ee9489a7ba74a4d00ceb/pkg/app/pipedv1/plugin/kubernetes/config/baseline.go#L17-L25

Part of #5530 

### DIFF
--- a/pkg/plugin/sdk/unit/duration.go
+++ b/pkg/plugin/sdk/unit/duration.go
@@ -1,0 +1,52 @@
+// Copyright 2025 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


khanhtc1202:
It's a bit hard to decide whether we should include this in the SDK or keep these plugins separate and accept the fact that they could be redundant and copied around.

I think what is placed in the SDK should be something more like piped care about, not something plugins need. Since we can say it's common, but it may block/confuse other plugins that don't want to use the SDK-defined config structure. WDYT?

Warashi:
Yes, I also think it's hard to decide whether we should put these in the SDK. I noticed we already have a " diff " package; it's not something piped cares about, but a package we thought was commonly used across plugins.
IMO, we can propose some best practices for plugin developers about the config structure and include the use of this unit package.

khanhtc1202:
@Warashi I got your point, then for now I only concern for the name of package, but it's not too strong. Let's keep it as is for now 👍

khanhtc1202 (approved):
LGTM 👍

PR #5960, opened 2025-06-21
feat(ui): add confirmation dialog for generating new piped key

### LINKED ISSUE #5951: Add confirmation dialog before adding a new piped key

When adding a new key to any existing Piped, there should be a confirmation dialog before actually generating the new key.

The message might be like: "This piped has one key. Are you sure you want to generate a new key?"
(+optional) "After adding it and choosing 'delete old key',  the current key will be unavailable."

**Why is this needed**:

Currently, a new key will be generated right after clicking `Add new Key` in the pipeds page.
That causes:
- When unintentionally adding a new key, a user is recommended to delete the existing key. That might stop the existing piped.

**Note of the current spec**

- One piped can have at most two piped keys.
- A user cannot add a new key if a piped already has two keys.
- When a piped has two keys and a user clicks 'delete old key', the older key will be deleted.
- When a piped has two keys and a user clicks 'add new key', then no-op.
Image

niladrix719:
**What this PR does**:

Adds confirmation dialog for generating new piped key

Screenshot 2025-06-21 at 1 22 03 PM

**Why we need it**:

To prevent accidental key generation that may lead to disruption of active piped connections

Fixes #5951 

### DIFF
--- a/web/src/components/settings-page/piped/components/piped-table-row.tsx
+++ b/web/src/components/settings-page/piped/components/piped-table-row.tsx
@@ -21,6 +21,7 @@ import { Highlight, themes } from "prism-react-renderer";
 import * as React from "react";
 import { FC, memo, useCallback, useState } from "react";
 import { CopyIconButton } from "~/components/copy-icon-button";
+import DialogConfirm from "~/components/dialog-confirm";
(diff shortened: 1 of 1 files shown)


t-kikuc:
thanks, LGTM

@hongky-1994 please check clearness of front-end code 👍

t-kikuc (approved):
Thank you!

PR #5965, opened 2025-06-24
Remove pipelineStage ID config from SDK logic

khanhtc1202:
**What this PR does**:

Remove the usage of pipelineStage ID config from SDK logic.

**Why we need it**:

The pipelineStage ID will be created by the piped, plugins will only use it, so I remove logic around that part on SDK.

Follow #5964


Warashi (approved):
👍🏻

ffjlabo (approved):

PR #5967, opened 2025-06-24
[Stage Plugins] Report stage commands as handled after the stage finished

t-kikuc:
**What this PR does**:

- as title

**Why we need it**:

- To suppress a too large response in `apiClient.ListUnhandledCommands()`

Part of #5367 


khanhtc1202 (approved):
LGTM 👌

PR #5969, opened 2025-06-25
[Plugins] Replace configv1 unit types with SDK unit types

Warashi:
**What this PR does**:

as title

**Why we need it**:

We don't want to import `pkg/configv1` from plugins

Follows #5959 


khanhtc1202 (approved):
👍

Issue #5956, opened 2025-06-19
Remove long time deprecated features from codebase (prepare for pipecd-v1 release)


khanhtc1202:
**What would you like to be added**:

As preparation for PipeCD v1 release (scheduled in the end of this year), I want to do clean up the codebase. As the result of this issue, features  marked as deprecated would be:
- Deleted from the codebase (both v0 and preparing v1)
- Has a clear decision that whether we bring that feature to the v1 codebase or not

This issue will be used to track the progress of deprecated features removal.

**Why is this needed**:

khanhtc1202:
Features/Configurations marked as deprecated:
- [ ] FileDB (https://github.com/pipe-cd/pipecd/issues/5652)
- [x] EnableDefaultKubernetesCloudProvider config in piped (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/cmd/piped/piped.go#L148)
- [ ] EventWatcher config via `.pipe/` directory (ref: https://pipecd.dev/docs-v0.52.x/user-guide/event-watcher/#use-the-pipe-directory)
- [x] SaveDeploymentMetadata in piped/service.proto (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/server/service/pipedservice/service.proto#L90C9-L90C31)
- [x] Platform(XXX)Input.AutoRollback (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/config/application_ecs.go#L67-L68)
- [ ] App config NotificationMention.slack (ref: https://pipecd.dev/docs-v0.52.x/user-guide/configuration-reference/#notificationmention)
- [x] Piped config CloudProviders/PlatformProvider  (except for deployment metrics collector)
- [ ] Piped config Notifications.Receivers.Slack.OAuthToken (ref: https://pipecd.dev/docs-v0.52.x/user-guide/managing-piped/configuration-reference/#notificationreceiverslack)
- [ ] All KindPlatform in the application config other than KindApplication
- [x] `pipectl init` command (ref: https://github.com/pipe-cd/pipecd/issues/4677) -> https://github.com/pipe-cd/pipecd/pull/5958
- [ ] TBD

Model / Internal:
- [ ] Model ArtifactVersion.Kind (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/model/common.proto#L95-L96)
[...]

PR #5971, opened 2025-06-25
[pipedv1] Remove visible field from controller logic

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

Pipedv1 logic would not based on visible field.

Follow PR #5950
Part of #5956


Warashi (approved):
👍🏻

ffjlabo (approved):
👍

PR #5974, opened 2025-06-27
Rename live demo to playground

khanhtc1202:
**What this PR does**:

Rename demo related docs to playground to make it less confusing.

**Why we need it**:

We have play.pipecd.dev mentioned as live demo, while demo.pipecd.dev used as dev. Should rename to make it less confusing 👀 

Fixes #


ffjlabo (approved):
👍

t-kikuc (approved):

PR #5978, opened 2025-06-30
Fix scheduler status handling

t-kikuc:
**What this PR does**:

[1] Use `defer` to ensure reporting the status even if the stage is cancelled
[2] Handle Skipped/Exited before `executeStage` when the stage was already completed by a previous scheduler.

**Why we need it**:

[1] Without this PR, the following problems will occur:

A. The stage is marked as 'running' forever. (it should be 'cancelled')

image

B. The stage is marked as 'failure' even if I cancelled. (it should be 'cancelled')

image

[2] To simplify status handling in executeStage.

Part of #5259 

### DIFF
--- a/pkg/app/pipedv1/controller/scheduler.go
+++ b/pkg/app/pipedv1/controller/scheduler.go
@@ -303,8 +303,8 @@ func (s *scheduler) Run(ctx context.Context) error {
 			break
 		}
 		if ps.Status == model.StageStatus_STAGE_SKIPPED {
-			statusReason = fmt.Sprintf("Stage %s was skipped", ps.Id)
-			continue // TODO
+			statusReason = fmt.Sprintf("Stage %s has been already skipped", ps.Id)
+			continue
 		}
 		if ps.Status == model.StageStatus_STAGE_EXITED {
 			deploymentStatus = model.DeploymentStatus_DEPLOYMENT_SUCCESS
@@ -489,6 +489,13 @@ func (s *scheduler) executeStage(sig StopSignal, ps *model.PipelineStage) (final
 		originalStatus = ps.Status
 	)
 
+	defer func() {
+		// Ensure reporting the status even if the stage is cancelled.
+		if err := s.reportStageStatus(context.Background(), ps.Id, finalStatus, ps.Requires); err != nil {
+			s.logger.Error("failed to report stage status", zap.Error(err))
+		}
+	}()
+
 	tds, err := s.targetDSP.Get(ctx, io.Discard)
 	if err != nil {
 		s.logger.Error("failed to get target deployment source", zap.String("stage-name", ps.Name), zap.Error(err))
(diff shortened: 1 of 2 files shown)


pkg/app/pipedv1/controller/scheduler_test.go
```diff
@@ -338,5 +338,5 @@ func TestExecuteStage_SignalCancelled(t *testing.T) {
 
 	handler.Cancel()
 	finalStatus := s.executeStage(sig, s.deployment.Stages[0])
-	assert.Equal(t, model.StageStatus_STAGE_FAILURE, finalStatus)
+	assert.Equal(t, model.StageStatus_STAGE_CANCELLED, finalStatus)
```
t-kikuc, line 341: due to this change:

```go
if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_RUNNING, ps.Requires); err != nil {
			if sig.Signal() == StopSignalCancel {
				return model.StageStatus_STAGE_CANCELLED
			}
		}
```

pkg/app/pipedv1/controller/scheduler.go
```diff
@@ -302,6 +302,15 @@ func (s *scheduler) Run(ctx context.Context) error {
...
 		}
+		if ps.Status == model.StageStatus_STAGE_SKIPPED {
+			statusReason = fmt.Sprintf("Stage %s has been already skipped", ps.Id)
+			continue
+		}
+		if ps.Status == model.StageStatus_STAGE_EXITED {
+			deploymentStatus = model.DeploymentStatus_DEPLOYMENT_SUCCESS
+			statusReason = fmt.Sprintf("Deployment was exited before stage %s", ps.Id)
+			break
+		}
```
t-kikuc, line 313: These are [2].  
[1] is not related to here.

pkg/app/pipedv1/controller/scheduler.go
```diff
@@ -523,6 +539,9 @@ func (s *scheduler) executeStage(sig StopSignal, ps *model.PipelineStage) (final
 	// Update stage status to RUNNING if needed.
 	if model.CanUpdateStageStatus(ps.Status, model.StageStatus_STAGE_RUNNING) {
 		if err := s.reportStageStatus(ctx, ps.Id, model.StageStatus_STAGE_RUNNING, ps.Requires); err != nil {
+			if sig.Signal() == StopSignalCancel {
+				return model.StageStatus_STAGE_CANCELLED
+			}
```
t-kikuc, line 544: This solves [1]-B.

pkg/app/pipedv1/controller/scheduler.go
```diff
@@ -480,6 +489,13 @@ func (s *scheduler) executeStage(sig StopSignal, ps *model.PipelineStage) (final
 		originalStatus = ps.Status
 	)
 
+	defer func() {
+		// Ensure reporting the status even if the stage is cancelled.
+		if err := s.reportStageStatus(context.Background(), ps.Id, finalStatus, ps.Requires); err != nil {
+			s.logger.Error("failed to report stage status", zap.Error(err))
+		}
+	}()
+
```
t-kikuc, line 498: The cause of [1]-A was that `reportStageStatus` was not called when the stage is cancelled and returns some status before `reportStageStatus` at the bottom of this method.

PR #5980, opened 2025-07-01
[Stage Plugins] Use `METADATA_STAGE_DISPLAY_KEY` instead of SkippedBy/ApprovedBy

### LINKED ISSUE #5367: [epic][plugin] Implement Stage Plugins

t-kikuc:
**What this PR does**:

- Show the stage metadata by the common key `MetadataKey_StageDisplay` instead of `SkippedBy` and `ApprovedBy`
  - This will be used in k8s plugin too for the trafficPercentage.

**Why we need it**:

- To suppress hard coding of `SkippedBy` and `ApprovedBy`.
    - It reduces tightly-coupled code of web/server/piped/plugin

**TODO:**

- In the SDK, add `pkg/plugin/sdk/constants/constants.go` with `MetadataKey_StageDisplay`
usage:
```go
import (
   constants "github.com/pipe-cd/piped-plugin-sdk-go/constants"
)

func executeStage(...){
 ...
 client.PutStageMetadata(ctx, constants.MetadataKey_StageDisplay, "Skipped by: user-A")
 ...
}
```

Fixes #5367


pkg/constants/constants.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package constants
+
+const (
+	// MetadataKeyStageDisplay is the key of the metadata to be displayed on the deployment detail UI.
+	MetadataKeyStageDisplay = "pipecd/stage-display-metadata"
+)
```
ffjlabo: [IMO] How about defining the const value in the `model` package to clarify when to use it by grouping related things.
For example, Maybe in `pkg/model/stage.go`?

BTW, I referred the current implementation.
e.g. 
https://github.com/pipe-cd/pipecd/blob/master/pkg/model/deployment.go#L23-L25

t-kikuc: it seems good.

Now we don't need to create constants.go.
(I guess `MetadataKeyDeploymentNotification` will not be used in pipedv1)

tbh, previously the file contained a const for `Approvers`, which will be used in server too.

t-kikuc: @ffjlabo 
I moved:  [831dee3](https://github.com/pipe-cd/pipecd/pull/5980/commits/831dee3be91ff36147f06214de52dabfac686605)

ffjlabo: @t-kikuc 
I got it, thanks! 

If this is the stage-specific knowledge, then we might put it in `pkg/model/stage.go`.
WDYT?

t-kikuc: @ffjlabo Oh, I missed `stage.go`.

But I noticed that `stage.go` does not seem to be used in pipedv1.
WDYT?

ffjlabo: OK, so let's keep it as is.

ffjlabo (approved):
LGTM

Issue #5985, opened 2025-07-02
PlanPreview feature for the plugin-arch piped


t-kikuc:
**What would you like to be added**:

TODO:

- [ ] Define API between piped and plugin
- [ ] Implement Piped side

And implement in each plugin.

I think pipectl and actions-plan-preview won't change.

**Why is this needed**:

PR #5986, opened 2025-07-02
Define PlanPreviewService proto

t-kikuc:
**What this PR does**:

as title

**Note**

- The API is based on here: https://github.com/pipe-cd/pipecd/blob/b7555ad0355a98c9e15ee65ce760d34515328ef9/pkg/app/piped/planpreview/builder.go#L245-L260
- `GetPlanPreviewResponse.details` is `bytes` for faster communication. SDK will serve a buffered writer to plugin developers.
- Please add new fields of application to `GetPlanPreviewRequest` while implementing the k8s plugin. Others will only use `application_id`.
- At first, we won't pass RunningDeploySource because it's better to compare TargetDeploySource and LiveState.
    - In the future, we might add RunningDeploySource to `GetPlanPreviewRequest` or add a new rpc.

**Why we need it**:

To start developing Piped-side and the SDK of PlanPreview

Part of #5985 


t-kikuc:
For now, we'll focus on comparing with LiveState, not with RunningDeploymentSource.
If they're needed in the future, let's define a new v2 interface of the PlanPreviewService.

ffjlabo (approved):
LGTM

Warashi (approved):
LGTM

PR #5987, opened 2025-07-02
Pass ApplicationConfigFiilename when creating deploysource.DeploySource

Warashi:
**What this PR does**:

as title

**Why we need it**:

We have to pass it because plugins need this.

Part of #4980 

### DIFF
--- a/pkg/app/pipedv1/deploysource/deploysource.go
+++ b/pkg/app/pipedv1/deploysource/deploysource.go
@@ -186,10 +186,11 @@ func (p *provider) prepare(ctx context.Context, lw io.Writer) (*DeploySource, er
 	}
 
 	return &DeploySource{
-		RepoDir:           repoDir,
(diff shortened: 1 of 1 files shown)


t-kikuc (approved):
[ASK] Where is `ApplicationConfigFiilename` used?

Warashi:
@t-kikuc 
In the Kubernetes plugin, we use `ApplicationConfigFiilename` to avoid loading app.pipecd.yaml as a Kubernetes manifest.
The `configFilename` in the below line comes from `ApplicationConfigFiilename`.
https://github.com/pipe-cd/pipecd/blob/73b52db56400b695b9c7a1b7044236949ce21341/pkg/app/pipedv1/plugin/kubernetes/provider/loader.go#L212-L214

t-kikuc:
@Warashi Thank you, I got it.

Issue #5992, opened 2025-07-03
Implement Terraform plugin


t-kikuc:
**What would you like to be added**:

**Why is this needed**:

PR #5996, opened 2025-07-03
Fix go version of Terraform plugin to 1.24.1

t-kikuc:
**What this PR does**:

fix by `go mod edit -go=1.24.1 `

**Why we need it**:

To use the same version as other plugins.

Part of #5992 


t-kikuc:
it's better to update other plugins!

PR #5998, opened 2025-07-04
[K8s Plugin] Remove unnecessary func and if-else

Warashi:
**What this PR does**:

as title

**Why we need it**:

to hardening codebase

Part of #5764 


pkg/app/pipedv1/plugin/kubernetes/provider/liveresources.go
```diff
@@ -46,49 +43,3 @@ func GetLiveResources(ctx context.Context, kubectl *Kubectl, kubeconfig string,
 
 	return namespacedLiveResources, clusterScopedLiveResources, nil
 }
-
-// BuildApplicationLiveState builds the live state of the application from the given manifests.
-func BuildApplicationLiveState(deploytarget string, manifests []Manifest, now time.Time) *model.ApplicationLiveState {
```
Warashi, line 51: This is an old implementation that was implemented before using the SDK.

pkg/app/pipedv1/plugin/kubernetes/provider/loader.go
```diff
@@ -205,10 +204,6 @@ func LoadPlainYAMLManifests(dir string, names []string, configFilename string) (
 			if ext := filepath.Ext(d.Name()); ext != ".yaml" && ext != ".yml" && ext != ".json" {
 				return nil
 			}
-			if model.IsApplicationConfigFile(d.Name()) {
-				// MEMO: can we remove this check because we have configFilename?
-				return nil
-			}
```
Warashi, line 211: Before #5987, the `configFilename` was an empty string, so this is needed.
But now, we can remove this if clause.

t-kikuc (approved):

PR #6001, opened 2025-07-04
Remove 'replace' of the SDK in k8s and wait plugins

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

- Make the plugins depend on the published SDK, not the local one.
   - I think this `replace` was a temporary solution when splitting the go.mod.
   - I guess this is the cause of https://github.com/pipe-cd/pipecd/actions/runs/16065019520/job/45337800981?pr=6000.
- To make consistency with other plugins

Part of #5867 


khanhtc1202 (approved):
👌

PR #6002, opened 2025-07-04
Mark kind field in application and deployment model as deprecated

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

The field Kind in application/deployment model should not be used in pipedv1 (plugin-arch piped) to build logic.

Related to #5956


t-kikuc (approved):

ffjlabo (approved):
LGTM

PR #6009, opened 2025-07-04
Remove deployment kind value from pipedv1 logic

khanhtc1202:
**What this PR does**:

- Remove deployment kind value from Slack notifier (pipedv1)
- Remove deployment kind value from tracer log for planner and scheduler (pipedv1)
- Remove deployment kind value from trigger logic (pipedv1)

**Why we need it**:

Pipedv1 logic should not depended on deployment kind value

Part of #5252 


ffjlabo (approved):
LGTM

PR #5972, opened 2025-06-25
Redux replace deployment detail page

hongky-1994:
**What this PR does**:

- This MR depends on https://github.com/pipe-cd/pipecd/pull/5955
- Add commandContext for tracing command
- split util function from modules/deployment
- add query for get deployment by id, stage logs, command
- add mutation for cacnel deployment, approve stage, skip stage

**Why we need it**:

- remove redux module from deployment-detail pages

**Which issue(s) this PR fixes**:

Partof #5868

**Does this PR introduce a user-facing change?**: no

- **How are users affected by this change**: no
- **Is this breaking change**: no
- **How to migrate (if breaking change)**: no

PR #6012, opened 2025-07-07
Redux replace deployment list

hongky-1994:
**What this PR does**:

- This MR depends on https://github.com/pipe-cd/pipecd/pull/5972
- Remove modules/deployment
- Remove modules/stage-logs
- Add query for deployment list infinite
- add component label-auto-complete
- fix test application filter

**Why we need it**:

- remove redux module from deployment-list pages

Partof #5868


ffjlabo (approved):
👍

PR #6020, opened 2025-07-11
Rename Helm Chart from pipecd to control-plane

### LINKED ISSUE #5858: Rename pipecd helm chart and container image to be control plane

**Why is this needed**:

niladrix719:
**What this PR does:**
Renames the Helm chart from `pipecd` to `control-plane`

**Why we need it:**
To align the chart name with its intended purpose and improve clarity

Fixes #5858

Screenshot 2025-07-11 at 5 28 42 PM

**Does this PR introduce a user-facing change?:**
Yes

**How are users affected by this change:**
Users need to reference `control-plane` instead of `pipecd` when deploying the Helm chart

**Is this breaking change:**
Yes

**How to migrate (if breaking change):**
Update all Helm chart references from `pipecd` to `control-plane`


niladrix719:
I have preserved the deprecated `manifests/pipecd` and `cmd/pipecd`. Hey @khanhtc1202, can you please confirm if this is how we want this

khanhtc1202:
Hi @niladrix719, Sorry for the unclear explanation 🙏 

Here is the way to do what we want 👀 https://github.com/pipe-cd/pipecd/commit/f70912a15908266f1d263fff03528dad8f6c8dee

Basically, if you add a new control-plane directory that duplicates the pipecd directory like this, if we need to update the control plane source code, we need to do it for both directories, which is complicated. What we want is to make the same Dockerfile (under the control-plane dir—currently named pipecd) be used to build both the `control-plane` and `pipecd` container image and chart.

You can
- [ ] Rename the directory `cmd/pipecd` to `cmd/control-plane`
- [ ] Update the CI for building container image and chart (as I did here: https://github.com/pipe-cd/pipecd/commit/f70912a15908266f1d263fff03528dad8f6c8dee) - maybe you need to recheck other place
- [ ] Update the makefile (for the makefile, basically for local use so we can completely replace pipecd with control-plane)
- [ ] TBD

I also recommend separating PRs, like you can update CI and docs by separate PRs, so it's easier to understand/review. Thanks!

niladrix719:
@khanhtc1202 I've separated the docs changes in this PR (#6073), and both PR are ready for review

Makefile
```diff
@@ -291,11 +291,11 @@ release/docs:
 
 .PHONY: kind-up
 kind-up:
-	./hack/create-kind-cluster.sh pipecd
+	./hack/create-kind-cluster.sh control-plane
```
khanhtc1202: the cluster should named `pipecd`, control-plane is the name of the "application" that run on the pipecd cluster.

Makefile
```diff
@@ -142,39 +142,39 @@ test/integration:
...
+		--set server.image.repository=localhost:5001/control-plane \
+		--set ops.image.repository=localhost:5001/control-plane \
 		--values $(CONTROL_PLANE_VALUES)
 
-.PHONY: stop/pipecd
-stop/pipecd:
-	helm -n pipecd uninstall pipecd
+.PHONY: stop/control-plane
+stop/control-plane:
+	helm -n control-plane uninstall control-plane
```
khanhtc1202: The namespace name should be `pipecd`, aka the command should be

> helm -n pipecd uninstall control-plane

Same with other places with namespace values

niladrix719:
I've renamed the namespace and cluster back to `pipecd`

khanhtc1202:
Hi @niladrix719, Since the change in the PR touches a lot of places at once, I recommend separating it into a few PRs
- [x] Rename cmd/pipecd -> cmd/control-plane (and update all places that refer to it in Makefile and CI)
- [x] Rename container image from pipecd -> control-plane (check Makefile, CI, and Manifests that refer to the pipecd container image)
- [ ] Update the helm chart (make sure pipecd and control-plane helm charts exist at the same time)

We can't allow a PR that touches that many places at once since it's too risky, conflicts potential, and hard to make sure we handle all the case we need

niladrix719:
@khanhtc1202 Now port-forward will work `kubectl port-forward -n pipecd svc/control-plane 8080`

PR #6022, opened 2025-07-14
Delete unused flag 'Retries' from Terraform config

t-kikuc:
**What this PR does**:

- Delete 'Retries' from TerraformSyncStageOptions and TerraformApplyStageOptions
- Delete 'Retries' from all docs

**Why we need it**:

'Retries' has never been used.  
It's a trap for users.

cf.
v0.42.0 does not use 'Retries'
https://github.com/pipe-cd/pipecd/blob/v0.42.0/pkg/config/application_terraform.go#L58-L74


Warashi (approved):
👍🏻

t-kikuc:
Let me rerun the lint/go.

> 🐶 Installing reviewdog ... https://github.com/reviewdog/reviewdog
> Error: Error: socket hang up

PR #6024, opened 2025-07-14
[K8s Plugin] Implement K8S_TRAFFIC_ROUTING with Pod Selector

Warashi:
**What this PR does**:

Implement K8S_TRAFFIC_ROUTING with pod selector.

**Why we need it**:

This is to support blue/green deployment with a pod selector.

Part of #5764 


pkg/app/pipedv1/plugin/kubernetes/deployment/traffic.go
```diff
@@ -43,14 +48,144 @@ func (p *Plugin) executeK8sTrafficRoutingStage(ctx context.Context, input *sdk.E
...
+	}
+
+	var targetVariant string
+	switch {
+	case primaryPercent == 100:
+		targetVariant = cfg.Spec.VariantLabel.PrimaryValue
+	case canaryPercent == 100:
+		targetVariant = cfg.Spec.VariantLabel.CanaryValue
+	default:
+		lp.Errorf("PodSelector requires either primary or canary to be 100%% (primary=%d, canary=%d)",
```
khanhtc1202, line 81:
```suggestion
		lp.Errorf("PodSelector requires either primary or canary to be 100% (primary=%d, canary=%d)",
```
typo

t-kikuc: i guess %% will be '%' in Errorf.

https://pkg.go.dev/fmt

Warashi: As t-kikuc said, `%%` means `%` because `%` has special meaning in the `Errorf` message format string, so it must be escaped.

khanhtc1202: Got it, I thought the only the `\` is for escape but til about `%` as escape for `%` 👌

khanhtc1202 (approved):
Here you go 👌

PR #6029, opened 2025-07-14
restore NewNotifier() in pipedv1

t-kikuc:
**What this PR does**:

Uncomment to be the same as v0:
https://github.com/pipe-cd/pipecd/blob/efbe9c96961948a4a7db951d9cf4aaf1a66307ff/pkg/app/piped/notifier/notifier.go#L59-L86

**Why we need it**:

Although they are commented out in https://github.com/pipe-cd/pipecd/pull/5392, they are used in pipedv1 without the v0 config.

Part of #5259 


khanhtc1202 (approved):
Thanks 👍

PR #6031, opened 2025-07-14
Add check for divergent branches error

### LINKED ISSUE #6016: Application doesn't Sync when last commit is edited using --amend flag
**What happened**:

When we make an amend commit to a repo that is tracked by piped and force push it to the remote repository, piped is not able to pull the latest changes due to `divergent branches` and errors. On the control plane, if I manually try to run sync, then it keeps on loading indefinitely until the page is refreshed and no changes are applied to the cluster. Because of this, other applications from the same repo also hang up with the updates.

Piped should try to pull the changes, and if it detects divergent branches, it should pull the changes using a different strategy like `git reset --hard origin/<branch>` 

**How to reproduce it**:

Add a new application and sync it to the remote repo, then edit the last commit using `--amend` and push it to the remote. Piped will raise the error `failed to update git repository tutorial-repo to latest`, and if we try to sync using UI manually, the sync button will buffer until refreshed.

**Environment**:

- `piped` version: v0.52.1
- `control-plane` version: v0.50.1
- Others: K3s cluster running on Ubuntu 24.04

kunalvirwal:
**What this PR does**:

Fixes the error in piped when the last commit is amended and force pushed which used to cause a divergent branches error. 

**Why we need it**:

Because otherwise whenever an emergency amend commit is made to the tracked repo, piped would raise an error and the control plane would hang while syncing.

Fixes #6016 

No

### DIFF
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -293,10 +293,26 @@ func (r *repo) CheckoutPullRequest(ctx context.Context, number int, branch strin
 // Pull fetches from and integrate with a local branch.
 func (r *repo) Pull(ctx context.Context, branch string) error {
 	out, err := r.runGitCommand(ctx, "pull", r.remote, branch)
-	if err != nil {
-		return formatCommandError(err, out)
+
+	if err == nil {
+		return nil
 	}
-	return nil
+
+	if strings.Contains(string(out), "divergent branches") {
+		// need to use --force to move the remote-ref to the new commit
+		// when the update is a non-fast-forward one like divergent branches
+		out, err := r.runGitCommand(ctx, "fetch", r.remote, branch, "--force")
+		if err != nil {
+			return formatCommandError(err, out)
+		}
+		// because we want our local to exactly replicate the remote
+		out, err = r.runGitCommand(ctx, "reset", "--hard", "FETCH_HEAD")
+		if err != nil {
+			return formatCommandError(err, out)
+		}
+		return nil
+	}
+	return formatCommandError(err, out)
 }
 
 // MergeRemoteBranch merges all commits until the given one
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -292,13 +292,13 @@ func (r *repo) CheckoutPullRequest(ctx context.Context, number int, branch strin
 
 // Pull fetches from and integrate with a local branch.
 func (r *repo) Pull(ctx context.Context, branch string) error {
-	out, err := r.runGitCommand(ctx, "pull", r.remote, branch)
+	out, err := r.runGitCommand(ctx, "pull", "--ff-only", r.remote, branch)
 
 	if err == nil {
 		return nil
 	}
-
-	if strings.Contains(string(out), "divergent branches") {
+	// Handle "divergent branches" error, caused by amending the remote
+	if strings.Contains(string(out), "Diverging branches") {
 		// need to use --force to move the remote-ref to the new commit
 		// when the update is a non-fast-forward one like divergent branches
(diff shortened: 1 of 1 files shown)


pkg/git/repo.go
```diff
@@ -293,10 +293,26 @@ func (r *repo) CheckoutPullRequest(ctx context.Context, number int, branch strin
...
 	out, err := r.runGitCommand(ctx, "pull", r.remote, branch)
-	if err != nil {
-		return formatCommandError(err, out)
+
+	if err == nil {
+		return nil
 	}
-	return nil
+
+	if strings.Contains(string(out), "divergent branches") {
```
t-kikuc: [ask] is this message stable?
which git version introduced this message?

t-kikuc (commented):
thank you. 
i guess this issue is not so prioritized since amending the remote commit is bad.

pkg/git/repo.go
kunalvirwal: This message was introduced in 2.27.0, which was around 5 years back. These keywords have since been used in the Git CLI and many SDKs. So it is pretty stable to be used for error detection. I also tried to look for error codes, but unfortunately, Git does not have any predefined error codes, and it raises code 128 in many exceptions, including this one.

kunalvirwal:
Hi @t-kikuc , thanks a lot for taking a look!

I encountered this scenario while testing with piped, and I agree that amending the remote is not a practice to be followed, but it's just that when such an error is encountered, syncs stop as piped keeps on trying to pull the remote repo, and the control plane is thus left hanging until piped is restarted so I added this edge case for it.
Please do let me know if you would want me to proceed in some other way.
Thank You!

pkg/git/repo.go
khanhtc1202:
```suggestion
	// Handle "divergent branches" error, caused by amending the remote
	if strings.Contains(string(out), "divergent branches") {
```
nits

pkg/git/repo.go
```diff
@@ -293,10 +293,26 @@ func (r *repo) CheckoutPullRequest(ctx context.Context, number int, branch strin
 // Pull fetches from and integrate with a local branch.
 func (r *repo) Pull(ctx context.Context, branch string) error {
 	out, err := r.runGitCommand(ctx, "pull", r.remote, branch)
```
khanhtc1202: We should add the `--ff-only` flag here to prevent git from trying to merge in case a divergent branch event occurs. Only allow fast forward pulls here, and if they fail, try to resolve the divergent branch issue using the logic below.
thanks @Warashi 👌

khanhtc1202: /ping @kunalvirwal

kunalvirwal: Okk sure. There will be a change in the yielded error after using that flag. I will make sure everything works fine and push the changes ASAP. Thanks

khanhtc1202 (approved):
Nice catch, thank you 👌

Warashi (commented):
I checked the behavior of `git pull --ff-only` and noticed `Diverging branches` output can be disabled by config.
IMO, we can use `Not possible to fast-forward` because it's not a hint, so we cannot disable this output.
```
$ git pull --ff-only
hint: Diverging branches can't be fast-forwarded, you need to either:
hint:
hint:   git merge --no-ff
hint:
hint: or:
hint:
hint:   git rebase
hint:
hint: Disable this message with "git config set advice.diverging false"
fatal: Not possible to fast-forward, aborting.
```

kunalvirwal:
Okk...Ig that will be a better option. I'll update the PR. Thanks

Warashi (approved):
Thank you!

PR #6032, opened 2025-07-15
[State Plugins] Added NotifyWaitApproval and NotifyApproved APIs

t-kikuc:
**What this PR does**:

Defined two APIs and implemented the piped side of:
- NotifyWaitApproval
- NotifyApproved

Later, I will implement the SDK side to add `NotifyWaitApproval()` and `NotifyApproved()` methods.

**Why we need it**:

To notify approvers when the approval stage starts.

**Note**

I thinks the old PR https://github.com/pipe-cd/pipecd/pull/5632 is not good for extensibility.

Part of #5367 


t-kikuc:
Let me create a better one based on https://github.com/pipe-cd/pipecd/pull/5632/files

PR #6036, opened 2025-07-16
[Analysis Plugin] Add SharedConfigDirectory to deploySource message for analysisTemplate

t-kikuc:
**What this PR does**:

as title.

I think passing `RepoDir` is not necessary yet.  Let's add it when needed.

TODO: I will add `SharedConfigDirectory` to the SDK deploySource.

**Why we need it**:

To pass analysisTemplates from piped to plugins like Analysis Stage.

Part of #5367 


khanhtc1202 (approved):
I am still concerned whether we should allow making this configurable or just allow only `.pipe` as the shared directory for sharing config files

Warashi (approved):
LGTM

PR #6044, opened 2025-07-18
Chore: Add local Codegen scripts

Okabe-Junya:
**What this PR does**:

Add script and build targets to build and run code generation locally using a local Docker image.

**Why we need it**:

Code generation currently requires pulling the image from GHCR and doesn't reflect local changes to codegen tools immediately. This creates friction for local development.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,13 @@ build/plugin:
 	done
 	@echo "Plugins are built and copied to $(PLUGINS_BIN_DIR)"
 
+.PHONY: build/codegen
+build/codegen: CODEGEN_IMAGE_TAG ?= pipecd-local-codegen:latest
+build/codegen:
+	@echo "Building local codegen Docker image..."
+	docker build -t $(CODEGEN_IMAGE_TAG) ./tool/codegen
+	@echo "Successfully built $(CODEGEN_IMAGE_TAG)"
+
 .PHONY: push
 push/chart: BUCKET ?= charts.pipecd.dev
 push/chart: VERSION ?= $(shell git describe --tags --always --dirty --abbrev=7)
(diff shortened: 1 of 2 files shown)


Okabe-Junya:
@khanhtc1202 

Just a stale notifications, any ideas about this PR?
Thanks!

hack/local-codegen.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 The PipeCD Authors.
```
khanhtc1202, line 3:
```suggestion
# Copyright 2025 The PipeCD Authors.
```

Makefile
```diff
@@ -260,6 +267,12 @@ gen/code:
 	# NOTE: Keep this container image as same as defined in .github/workflows/gen.yml
 	docker run --rm -v ${PWD}:/repo -it --entrypoint ./tool/codegen/codegen.sh ghcr.io/pipe-cd/codegen@sha256:831f2dda2f56b1d12e90f88c0cb4168f51aa4eb5907b468e74bc42670939fff2 /repo # v0.50.0-215-g3f6a738
 
+.PHONY: gen/code-local
+gen/code-local: CODEGEN_IMAGE_TAG ?= pipecd-local-codegen:latest
```
khanhtc1202, line 271: Better to ensure run `build/codegen` before this step. WDYT?

khanhtc1202:
@Okabe-Junya Thanks for the idea. But I think it's a bit odd 🤔 Even after we can gen local code using the latest built codegen image, the CI will still be failed if codegen image used in CI step older than the one built in your local. 

I know we have something like `2 steps` update each time we want to change the generated code which needs the codegen image update and we want to make it less pain, but I don't think this is the answer for that 🤔 

If you just want to control the codegen image version on your local, the maybe you can add a flag to the makefile command

```Makefile
.PHONY: gen/code
gen/code: CODEGEN_IMAGE ?= ghcr.io/pipe-cd/codegen@sha256:831f2dda2f56b1d12e90f88c0cb4168f51aa4eb5907b468e74bc42670939fff2 # keep the default value same as CI image
gen/code: 
  ...
```

PR #6033, opened 2025-07-15
Notify WAIT_APPROVAL and APPROVED in piped side

t-kikuc:
**What this PR does**:

Notify the following two events in the scheduler.
- `WAIT_APPROVAL` on starting the stage
- `APPROVED` on ending the stage

TODO: Add `MetadataKeyStageApprovers` const to the SDK like this: https://github.com/pipe-cd/pipecd/blob/fd781978f811bef837af0b209e15cef267157e9f/pkg/plugin/sdk/client.go#L32-L34

**Why we need it**:

To notify the two events.

**Which issue(s) this PR fixes**:

Part of #5367 

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #6049, opened 2025-07-22
[stage plugins] SDK: add MetadataKeyStageApprovedUsers and DeploymentSource.SharedConfigDirectory

t-kikuc:
**What this PR does**:

- Added `MetadataKeyStageApprovedUsers` const
- Added `DeploymentSource.SharedConfigDirectory`

**Why we need it**:

- To persist the approved users
- To pass the analysisTemplate in `.pipe/` to a. plugin

Part of #5367 
Follows #6036 (analysis), #6033 (notifier)

### DIFF
--- a/pkg/plugin/sdk/client.go
+++ b/pkg/plugin/sdk/client.go
@@ -31,6 +31,10 @@ import (
 const (
 	// MetadataKeyStageDisplay is the key of the stage metadata to be displayed on the deployment detail UI.
 	MetadataKeyStageDisplay = model.MetadataKeyStageDisplay
+	// MetadataKeyStageApprovedUsers is the key of the metadata of who approved the stage.
+	// It will be displayed in the DEPLOYMENT_APPROVED notification.
+	// e.g. user-1,user-2
+	MetadataKeyStageApprovedUsers = "pipecd/stage-approved-users"
 
 	listStageCommandsInterval = 5 * time.Second
 )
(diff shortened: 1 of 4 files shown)


pkg/plugin/sdk/client.go
```diff
@@ -31,6 +31,10 @@ import (
 const (
 	// MetadataKeyStageDisplay is the key of the stage metadata to be displayed on the deployment detail UI.
 	MetadataKeyStageDisplay = model.MetadataKeyStageDisplay
+	// MetadataKeyStageApprovedUsers is the key of the metadata of who approved the stage.
+	// It will be displayed in the DEPLOYMENT_APPROVED notification.
+	// e.g. user-1,user-2
+	MetadataKeyStageApprovedUsers = "pipecd/stage-approved-users"
```
Warashi: How about using `model.MetadataKeyStageApprovedUsers` in the same way as `MetadataKeyStageDisplay`

https://github.com/pipe-cd/pipecd/blob/5722fabb80ce401bdc1a101561a3fd5ad8351a75/pkg/model/deployment.go#L28-L31

t-kikuc: ohh sorry, 

I fixed: [210c786](https://github.com/pipe-cd/pipecd/pull/6049/commits/210c78699f2a24ed5ddb8e3aa04a780699f7f84a)

khanhtc1202 (approved):
👍

PR #6054, opened 2025-07-22
Rollback and Index for visible stage in piped v1 - deployment page

hongky-1994:
**What this PR does**:

- Update filter logic for visible stage depend on piped v0 or v1 in deployment page.
- Piped v1 deployment stages:
   - stage shown when is a forward stage (rollback = false) or stage is started
   - order by: forwardStage > rollBackStage and field `stage.index`
- Update function findDefaultActiveStage depend on version of piped

**Why we need it**:

- We should support visible stage logic for both deployment of pipedv0 and pipedv1
- For decide which stage and in what order to display.
  - In pipedv0, the fields `visible` and `requiresList` should be used.
  - In pipesv1, idealy the fields `rollback` and `index` should be used.
     - Currently, sorting by `index` is not yet available, so we keep using requiresList for the ordering stage. 

Partof https://github.com/pipe-cd/pipecd/issues/5252#issuecomment-3007679251

### DIFF
--- a/web/src/components/deployments-detail-page/pipeline/index.tsx
+++ b/web/src/components/deployments-detail-page/pipeline/index.tsx
@@ -28,6 +28,11 @@ import { ApprovalStage } from "./approval-stage";
 import { PipelineStage } from "./pipeline-stage";
 import { ManualOperation } from "~~/model/deployment_pb";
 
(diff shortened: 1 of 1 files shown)


khanhtc1202 (approved):
Thank you 💯

PR #6040, opened 2025-07-18
[Analysis Plugin] Defined GetApplicationSharedObject() and PutApplicationSharedObject() for storing analysisResult

t-kikuc:
**What this PR does**:

add `GetApplicationSharedObject()` and `PutApplicationSharedObject()` in pipedservices(server<-piped, piped<-plugin) 

TODO: 
- server:  store the data to filestore
- piped:  app object store like metadataStore
- SDK:  wrap the two methods
- migration way of AnalysisResultStore -> new store

**Why we need it**:

- Currently, Analysis Stage needs to put the last analysis result for the `PREVIOUS` strategy.
- The current AnalysisResultStore is not flexible for plugins.
    - AnalysisResultStore will be deprecated and the new store 'ApplicationSharedObject' will be created.

**Which issue(s) this PR fixes**:

Part of #5367 

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #6055, opened 2025-07-22
[Analysis plugin] piped side for analysisResultStore

t-kikuc:
**What this PR does**:

Implemented the piped side of `GetApplicationSharedObject()` and `PutApplicationSharedObject()`.

I did not prepare the cache layer in piped because the appSharedObject will not be read frequently, unlike the deployment metadataStore, 

**Why we need it**:

To persist the application shared object from a plugin to the server.

Part of https://github.com/pipe-cd/pipecd/issues/5367
Follows https://github.com/pipe-cd/pipecd/pull/6040


khanhtc1202 (approved):
👌

Issue #5652, opened 2025-03-12
Deprecate FileDB as PipeCD datastore


khanhtc1202:
**What would you like to be added**:

We should remove the FileDB as PipeCD datastore due to the overheating of its maintenance.

**Why is this needed**:

Previously, we thought that making the filestore a datastore was a good idea to simplify the PipeCD control plane operator tasks. Still, due to the lack of maintaining resources of FileDB, it might block a lot of our new potential features. 
Also, the current implementation of PipeCD control plane (both quickstart and production mode) is small and easy enough to get in, so making filestore as a datastore brings less value than its cost.

Okabe-Junya:
Hello folks,

I have created https://github.com/pipe-cd/pipecd/pull/5786. This is a codebase cleanup, and I believe the ultimate goal is to merge this PR.

To move forward with this issue, we need to confirm two things:

1. How should users who are already using FileDB perform their migration?  
2. From which release will FileDB no longer be available?

t-kikuc:
Thank you!
It'd be best to publish an official blog on [1] how to migrate and [2] when the deadline is.

> 1. How should users who are already using FileDB perform their migration?

1. Update their Control Plane config
2. Migrate data from FileDB to another store (tough???)

We need to prepare some detailed guidance...


> 2. From which release will FileDB no longer be available?

There's no rule for deprecation span yet.

IMO, there should be some minor releases until the deprecation.  (What do you think, the maintainers, developers, users?)

_(although the feature status is Alpha (described at https://github.com/pipe-cd/pipecd/pull/4809), which means `No guarantees on backward compatibility.`...)_
https://pipecd.dev/docs-v0.51.x/feature-status/

khanhtc1202:
Thanks @Okabe-Junya for take over this issue. I will assign you to this issue 👍

khanhtc1202:
TODO list
- [x] Remove commanders (in each store) <- https://github.com/pipe-cd/pipecd/pull/6043
- [x] Remove shard encoder/decoder interface (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/datastore.go#L73-L87)
- [x] Remove shard storable interface (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/datastore.go#L91-L106)
- [x] Remove SetUpdatedAt in models
- [x] Remove FileDB codebase <- https://github.com/pipe-cd/pipecd/pull/5786
[...]

PR #6069, opened 2025-07-25
Remove commander from protoc gen auth template

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

We will remove FileDB support, so the commander is not needed.

Part of #5652


Warashi (approved):
🆗

t-kikuc (approved):

PR #6073, opened 2025-07-26
update control-plane docs for renamed Helm chart and container image

niladrix719:
What this PR does:
Updates the control-plane documentation to reflect the recently renamed Helm chart and container image.

Why we need it:
The documentation needs to refer to updated chart and image after this PR (https://github.com/pipe-cd/pipecd/pull/6020)  gets merged

Which issue(s) this PR fixes:#5858


khanhtc1202:
Let's hold this until helm chart control-plane is available

PR #6074, opened 2025-07-28
add stageIndex and correct logPersisterTest write method

hiep-tk:
**What this PR does**: 
- add stageIndex per https://github.com/pipe-cd/pipecd/issues/5901 discussion
- make ```TestLogPersister``` Write method behaves the same as ```stageLogPersister```

**Why we need it**:
- the exec.Command will wire its output and error to the log persister. This will break when testing with the current implementation of ```TestLogPersister```

Fixes # part of https://github.com/pipe-cd/pipecd/issues/5901

### DIFF
--- a/pkg/plugin/sdk/logpersister/logpersistertest/logpersister.go
+++ b/pkg/plugin/sdk/logpersister/logpersistertest/logpersister.go
@@ -38,7 +38,7 @@ func (lp TestLogPersister) StageLogPersister(deploymentID, stageID string) logpe
 func (lp TestLogPersister) Write(log []byte) (int, error) {
 	// Write the log to the test logger
 	lp.t.Log(string(log))
-	return 0, nil
+	return len(log), nil
 }
 func (lp TestLogPersister) Info(log string) {
 	lp.t.Log("INFO", log)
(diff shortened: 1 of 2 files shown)


pkg/plugin/sdk/logpersister/logpersistertest/logpersister.go
```diff
@@ -38,7 +38,7 @@ func (lp TestLogPersister) StageLogPersister(deploymentID, stageID string) logpe
 func (lp TestLogPersister) Write(log []byte) (int, error) {
 	// Write the log to the test logger
 	lp.t.Log(string(log))
-	return 0, nil
+	return len(log), nil
```
t-kikuc, line 41: Sorry, I cannot understand the background yet.

Please let me know the concrete case of the problem.

Warashi: Sorry, it's my bad for the first implementation.

`io.Writer` requires returning `len(log)` for the first return value when the write succeeded.
ref; https://pkg.go.dev/io#Writer

> Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. Write must return a non-nil error if it returns n < len(p). Write must not modify the slice data, even temporarily.

Warashi (approved):
LGTM

pkg/plugin/sdk/logpersister/logpersistertest/logpersister.go
hiep-tk: currently the ```stageLogPersister``` write method return ```len(log)``` whereas the ```TestLogPersister``` method does not even though it could. internally it just passes the log to testing.T.Log method.
This return value is necessary for exec.Command internal. If it needs to write something through this method and the return value is 0, it would think the write method is broken and err out. This would only be the issue when do the testing

t-kikuc: @Warashi @hiep-tk Thank you!

PR #6078, opened 2025-07-29
[K8s Plugin] Add logger for loadManifests input

Warashi:
**What this PR does**:

as title

**Why we need it**:

We use it in loader.LoadManifests, but it's nil for now.

Part of #5764 


t-kikuc (approved):
i got it, here:  
https://github.com/pipe-cd/pipecd/blob/9246c5b8f09d8db02b8649f02fcd2c7da56e49be/pkg/app/pipedv1/plugin/kubernetes/provider/loader.go#L68-L69

PR #6079, opened 2025-07-29
Add default attributes to the pluign logger

Warashi:
**What this PR does**:

This PR adds default attributes `plugin-name` and `plugin-version` to the plugin logger.

**Why we need it**:

It's convenient to determine the log from which plugin.

Part of #5530 


t-kikuc (approved):
LGTM

i like logger.with()

khanhtc1202 (approved):
Nice improvement 👌

PR #6080, opened 2025-07-29
[K8s Plugin] Implement K8S_TRAFFIC_ROUTING with Istio VirtualService

Warashi:
**What this PR does**:

as title

**Why we need it**:

to support traffic routing with Istio VirtualService

Part of #5764 


ffjlabo:
Will merge because it affects only pipedv1.

PR #6086, opened 2025-07-29
[Terraform plugin] Implemented Apply and Rollback stages

t-kikuc:
**What this PR does**:

as title

I referred to pipedv0:
- apply: https://github.com/pipe-cd/pipecd/blob/c3edb9fa6cd2ae53e12b53e7b2fc80247d01dea7/pkg/app/piped/executor/terraform/deploy.go#L185-L219
- rollback: https://github.com/pipe-cd/pipecd/blob/c3edb9fa6cd2ae53e12b53e7b2fc80247d01dea7/pkg/app/piped/executor/terraform/rollback.go#L48-L115

**Why we need it**:

to support the stages in pipedv1 too.

Part of #5992 


khanhtc1202 (approved):
Here you go 👌

PR #6089, opened 2025-07-30
Rename Control plane cmd to `control-plane` from `pipecd`

### LINKED ISSUE #5858: Rename pipecd helm chart and container image to be control plane

niladrix719:
**What this PR does**:

Renames the `cmd/pipecd` -> `cmd/control-plane`

**Why we need it**:

To align the cmd name with its intended purpose and improve clarity

Fixes #5858


.github/workflows/publish_image_chart.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
           - image: launcher-okd
             dockerfile: cmd/launcher/Dockerfile-okd
           - image: pipecd
-            dockerfile: cmd/pipecd/Dockerfile
+            dockerfile: cmd/control-plane/Dockerfile
```
khanhtc1202, line 42:
```suggestion
            dockerfile: cmd/control-plane/Dockerfile
          - image: control-plane
            dockerfile: cmd/control-plane/Dockerfile
```
nits

cmd/control-plane/README.md
```diff
@@ -0,0 +1,61 @@
...
+2. Install the web dependencies module
+
+    ``` console
+    make update/web-deps
+    ```
+
+3. Install Control Plane into the local cluster
+
+    ``` console
+    make run/pipecd
```
khanhtc1202: Should be renamed to `make run/control-plane`. Also need to update the Makefile

khanhtc1202: From line: https://github.com/pipe-cd/pipecd/pull/6089/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R145

.github/workflows/publish_image_chart.yaml
khanhtc1202: Aka. Build both pipecd image and control-plane image based on `cmd/control-plane/Dockerfile`

cmd/control-plane/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+ARG TARGETARCH
+
+WORKDIR /app
+
+COPY go.* ./
+RUN go mod download
+
+COPY . ./
+
+RUN make build/go MOD=pipecd BUILD_OS=${TARGETOS} BUILD_ARCH=${TARGETARCH}
```
khanhtc1202:
```suggestion
RUN make build/go MOD=control-plane BUILD_OS=${TARGETOS} BUILD_ARCH=${TARGETARCH}
```
It has to be like this or the build would fail (after update the Makefile)

cmd/control-plane/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    ./google-cloud-sdk/install.sh --quiet
+
+ENV PATH="/google-cloud-sdk/bin:${PATH}"
+
+RUN mkdir -p web/static
+
+COPY --from=builder /app/.artifacts/pipecd /usr/local/bin/pipecd
+COPY --from=web /app/.artifacts/web-static web/static
+
+ENTRYPOINT ["pipecd"]
```
khanhtc1202: Since the binary updated here (ref: [cmd/control-plane/main.go#L25](https://github.com/pipe-cd/pipecd/pull/6089/files#diff-fc5de60c2f59bda238cc9502e249e54d655a9f28ad24122c1363dbed9b294b1bR25), the binary used as entrypoint name may changed to `control-plane`, please check the built image behavior.

cmd/control-plane/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz && \
+    tar -zxvf ./google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz && \
+    rm ./google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz && \
+    ./google-cloud-sdk/install.sh --quiet
+
+ENV PATH="/google-cloud-sdk/bin:${PATH}"
+
+RUN mkdir -p web/static
+
+COPY --from=builder /app/.artifacts/pipecd /usr/local/bin/pipecd
```
khanhtc1202:
```suggestion
COPY --from=builder /app/.artifacts/control-plane /usr/local/bin/control-plane
```
I think the binary name changed, isn't it?

khanhtc1202: Same reason as https://github.com/pipe-cd/pipecd/pull/6089/files#r2242682328

khanhtc1202:
Hi @niladrix719, I have a question: Why should we not remove the `cmd/pipecd` directory? Since the PR name has been renamed, I think the dir `cmd/pipecd` should be completely removed after this is merged, doesn't it?

.github/workflows/publish_image_chart.yaml
niladrix719: This changes are already present in the #6090, should we also re-do it here?

niladrix719:
> Hi @niladrix719, I have a question: Why should we not remove the `cmd/pipecd` directory? Since the PR name has been renamed, I think the dir `cmd/pipecd` should be completely removed after this is merged, doesn't it?

Oh I think I missed that one, updated

.github/workflows/publish_image_chart.yaml
khanhtc1202: So, your idea is to separate the image renaming to another PR. Then it can be like that, thanks for clarifying 🙏

cmd/control-plane/README.md
niladrix719: Readme has been updated in the docs PR https://github.com/pipe-cd/pipecd/pull/6073/files#diff-5208b50232ce527cfc67c836312b6b69efd8aea87b11347ddc08d1456f28088f

Makefile has been updated in container image PR https://github.com/pipe-cd/pipecd/pull/6090/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

cmd/control-plane/Dockerfile
niladrix719: For the Dockerfile I have updated it here - https://github.com/pipe-cd/pipecd/pull/6090/files#diff-1a80c9bfc80d6a53e2406f32a9cd61fe2a4c096c8bf2ad646c8bfc24774d931d, so this can be imported after merging it, there will be naming conflicts(as we have renamed the cmd/control-plane folder here) but I think we can resolve it

cmd/control-plane/Dockerfile
niladrix719: same as above

cmd/control-plane/Dockerfile
niladrix719: Yeah but in this PR its still pipecd, so if we rename now then it will fail, I've updated this part in https://github.com/pipe-cd/pipecd/pull/6090/files#diff-1a80c9bfc80d6a53e2406f32a9cd61fe2a4c096c8bf2ad646c8bfc24774d931d

.github/workflows/publish_image_chart.yaml
niladrix719: Yeah right, I have divided them into four parts - https://github.com/pipe-cd/pipecd/issues/5858#issuecomment-3136268990

niladrix719:
On a second thought, I think instead of separate PRs for cmd and container image we can combine both of them into one PR, that will be simpler to merge

niladrix719:
Changes moved to - https://github.com/pipe-cd/pipecd/pull/6090

PR #6092, opened 2025-07-31
[Terraform Plugin] implemented GetLiveState()

t-kikuc:
**What this PR does**:

- moved `initialize.go` to the `provider` package to call it from not only `deployment`.
- implemented livestateplugin.  cf.:

    https://github.com/pipe-cd/pipecd/blob/5be3da35d9dda6b2f8c30c8d05c1076166c67d57/pkg/app/piped/driftdetector/terraform/detector.go#L255-L320

LiveState is not supported as pipedv0

**Why we need it**:

to support driftdetection for terraform in pipedv1 too.

Part of #5992 


Warashi (approved):
LGTM

khanhtc1202 (approved):
Here you go 👌

PR #6098, opened 2025-08-01
docs(user-guide): update secret management documentation to reflect new UI flow

### LINKED ISSUE #6039: Add the "Encrypt Secret" button on the applications page

Add `Encrypt Secret` button next to the `+ADD` button on the applications page.

It will open a right-side dialog like adding an application and encrypting a secret.

Image

Image

**Why is this needed**:

- It's unclear that secrets belong to a piped, not an application. Actually, users can use the same secret for multiple applications.
- Enable encryption of a secret before adding an application.
- `pipectl encrypt` requires an API key, but users want to avoid it for security.

The current form: 
Image

niladrix719:
**What this PR does**:

Updates secret management documentation

Screenshot 2025-08-01 at 9 49 57 PM

**Why we need it**:

Need to reflect new UI flow

Fixes #6039


khanhtc1202 (approved):
👌

PR #6102, opened 2025-08-04
feat(plugin): initialize waitapproval plugin for pipedv1

### LINKED ISSUE #6067: Implement WaitApproval Stage Plugin

Implement the WAIT_APPROVAL stage plugin like the WaitStage plugin for pipedv1

**Why is this needed**:

To support WaitApproval in pipedv1 too.

### Development Note

- Use `ListStageCommands()` to fetch the commands. See https://github.com/pipe-cd/pipecd/pull/5639 including the example usage.
- Use `sdk.MetadataKeyStageDisplay` and `AuthorizedOperators` instead of `Approvers` and `ApprovedBy` metadatas. See https://github.com/pipe-cd/pipecd/pull/6000
- We don't need to handle `Notifier` in this plugin. See https://github.com/pipe-cd/pipecd/pull/6033

niladrix719:
**What this PR does**:

 initialize waitapproval plugin

**Why we need it**:

waitapproval stage migration to plugin

Fixes Parts of #6067


pkg/app/pipedv1/plugin/waitapproval/options.go
```diff
@@ -0,0 +1,51 @@
...
+
+// decode decodes the raw JSON data and validates it.
+func decode(data json.RawMessage) (waitApprovalStageOptions, error) {
+	var opts waitApprovalStageOptions
+	if err := json.Unmarshal(data, &opts); err != nil {
+		return waitApprovalStageOptions{}, fmt.Errorf("failed to unmarshal the config: %w", err)
+	}
+	if opts.MinApproverNum <= 0 {
+		opts.MinApproverNum = 1
+	}
```
t-kikuc: Let's move to validate() like pipedv0 for compatibility.
```
func (w *WaitApprovalStageOptions) Validate() error {
	if w.MinApproverNum < 1 {
		return fmt.Errorf("minApproverNum %d should be greater than 0", w.MinApproverNum)
	}
	return nil
}
```

t-kikuc (changes requested):
thank you!
basically LGTM

pkg/app/pipedv1/plugin/waitapproval/options.go
niladrix719: Okay I see so we want to keep the same behavior of returning error here also

niladrix719: and do we keep the `o.Approvers` validations along with it or just keep the same Validate func as pipedv0?

t-kikuc: @niladrix719 
oh, let's keep like this:

```go
	if len(o.Approvers) == 0 {
		return fmt.Errorf("approvers must be set")
	}
	if o.MinApproverNum < 1 {
		return fmt.Errorf("minApproverNum %d should be greater than 0", o.MinApproverNum)
	}
        // DELETE. (this validation is really great, but it might cause a breaking change)
	// if o.MinApproverNum > len(o.Approvers) {
	// 	return fmt.Errorf("minApproverNum must be less than or equal to the number of approvers")
	// }
```

cf. 
https://pipecd.dev/docs-v0.52.x/user-guide/configuration-reference/#waitapprovalstageoptions

t-kikuc: @niladrix719 
(my code example was unclear.) I meant:

```go
	if len(o.Approvers) == 0 {
		return fmt.Errorf("approvers must be set")
	}
	if o.MinApproverNum < 1 {
		return fmt.Errorf("minApproverNum %d should be greater than 0", o.MinApproverNum)
	}
```

`if o.MinApproverNum > len(o.Approvers)` is great, but let's ignore it for now.

niladrix719: sure updated 👍  thank you!

t-kikuc (approved):
Great, thank you!!!!

PR #6110, opened 2025-08-08
feat: implement server side pagination for ListNotCompletedDeployments

### LINKED ISSUE #1937: Call ListNotCompletedDeployments itervally until all required deployments are fetched.
https://github.com/pipe-cd/pipe/blob/0489f896fab115263a6f6a110d82431a48eb5a24/pkg/app/piped/apistore/deploymentstore/store.go#L112-L115

priyanshuharshbodhi1:
**What this PR does**:
Moves pagination logic to the server for ListNotCompletedDeployments

**Why we need it**:

Fixes #1937

No

- **How are users affected by this change**: Not affected, piped/pipedv1 still make a single RPC call.
- **Is this breaking change**:No

### DIFF
--- a/pkg/app/server/grpcapi/piped_api.go
+++ b/pkg/app/server/grpcapi/piped_api.go
@@ -356,33 +356,63 @@ func (a *PipedAPI) ListNotCompletedDeployments(ctx context.Context, req *pipedse
 		return nil, err
 	}
 
-	opts := datastore.ListOptions{
-		Filters: []datastore.ListFilter{
-			{
-				Field:    "PipedId",
-				Operator: datastore.OperatorEqual,
-				Value:    pipedID,
-			},
-			// TODO: Change to simple conditional clause without using OR clause for portability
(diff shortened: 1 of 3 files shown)


ubermensch19:
@khanhtc1202  PTAL!

pkg/app/server/grpcapi/piped_api.go
```diff
@@ -356,33 +356,63 @@ func (a *PipedAPI) ListNotCompletedDeployments(ctx context.Context, req *pipedse
...
+    orders := []datastore.Order{
+        {
+            Field:     "UpdatedAt",
+            Direction: datastore.Desc,
+        },
+        {
+            Field:     "Id",
+            Direction: datastore.Asc,
+        },
+    }
```
khanhtc1202, line 386: Could you test this on your local control-plane. Since we're using virtual column MySQL or FireStore as the database for controlplane, both require indexes to make queries work.

PR #6111, opened 2025-08-08
feat(kubernetes): support --force-conflicts for kubectl Server-Side Apply via annotation

### LINKED ISSUE #4639: Support `--force-conflicts` option for kubectl Server-Side Apply
**What would you like to be added**: Support `--force-conflicts` option for `kubectl` Server-Side Apply.

**Why is this needed**: When use a KubernetesApp, with [PipeCD Server-Side Apply](https://github.com/pipe-cd/pipecd/blob/90559a36795b648149a63f2139016f2a5cb7afba/pkg/app/piped/platformprovider/kubernetes/kubernetes.go#L34), a piped gets the error, `Failed to apply manifest`, for a resource that has applied with Client-Side Apply and has been modified.

```
Failed to apply manifest: name="XXX", kind="XXX", namespace="XXX", apiVersion="XXX" (%!w(*fmt.wrapError=&{failed to apply: error: Operation cannot be fulfilled on XXX "XXX": the object has been modified; please apply your changes to the latest version and try again
Please review the fields above--they currently have other managers. Here
are the ways you can resolve this warning:
* If you intend to manage all of these fields, please re-run the apply
  command with the `--force-conflicts` flag.
* If you do not intend to manage all of the fields, please edit your
  manifest to remove references to the fields that should keep their
  current managers.
* You may co-own fields by updating your manifest to match the existing
  value; in this case, you'll become the manager if the other manager(s)
  stop managing the field (remove it from their configuration).
See http://k8s.io/docs/reference/using-api/api-concepts/#conflicts
 (exit status 1) 0xc0260e7160}))
```

priyanshuharshbodhi1:
**What this PR does**:
Adds a new annotation `pipecd.dev/force-conflicts: "true"`.
When used with `pipecd.dev/server-side-apply: "true"`, kubectl apply includes `--force-conflicts`.

**Why we need it**:
SSA can fail on resources previously managed by Client-Side Apply due to field ownership conflicts. `--force-conflicts` resolves conflicts by taking ownership, as recommended by kubectl’s error guidance.

Fixes #4639

- **How are users affected by this change**: New opt-in annotation allows forcing conflicts with SSA.

### DIFF
--- a/docs/content/en/docs-dev/user-guide/configuration-reference.md
+++ b/docs/content/en/docs-dev/user-guide/configuration-reference.md
@@ -48,6 +48,7 @@ Kubernetes resources can be managed by some annotations provided by PipeCD.
 |-|-|-|-|
 | `pipecd.dev/ignore-drift-detection` | any | "true" | Whether the drift detection should ignore this resource. |
 | `pipecd.dev/server-side-apply` | any | "true" | Use server side apply instead of client side apply. |
+| `pipecd.dev/force-conflicts` | any | "true" | Force conflicts when using server side apply. Appends `--force-conflicts` to kubectl apply (effective only when `pipecd.dev/server-side-apply` is "true"). |
 | `pipecd.dev/sync-by-replace` | any | "enabled" | Use `replace` instead of `apply`. |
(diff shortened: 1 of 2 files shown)


ubermensch19:
@khanhtc1202 PTAL!

PR #6116, opened 2025-08-13
[SDK]: Update ListStageCommands to accept sdk CommandType

niladrix719:
**What this PR does**:

Updates `ListStageCommands` to accept `CommandType` as the second parameter instead of `model.Command_Type`

**Why we need it**:

Plugins using the `ListStageCommands()` should not depend on `model` package for `Command_Type`

Instead we should be able to use it as:
```
ListStageCommands(ctx, sdk.CommandTypeApproveStage)
```


pkg/plugin/sdk/client.go
```diff
@@ -243,10 +243,19 @@ func (c *Client) ToolRegistry() *toolregistry.ToolRegistry {
...
 // ListStageCommands returns the list of stage commands of the given command types.
-func (c Client) ListStageCommands(ctx context.Context, commandTypes ...model.Command_Type) iter.Seq2[*StageCommand, error] {
+func (c Client) ListStageCommands(ctx context.Context, commandTypes ...CommandType) iter.Seq2[*StageCommand, error] {
 	return func(yield func(*StageCommand, error) bool) {
 		returned := map[string]struct{}{}
 
+		modelCommandTypes := make([]model.Command_Type, 0, len(commandTypes))
+		for _, cmdType := range commandTypes {
+			modelType, err := cmdType.toModelEnum()
+			if err != nil {
```
t-kikuc, line 253: This should return err to prevent unsupported commandTypes and for easier debugging

ffjlabo (approved):
LGTM

t-kikuc (approved):
Tahnk you!

PR #6122, opened 2025-08-15
Support all kustomization files name

khanhtc1202:
**What this PR does**:

Recognize all official kustomization file names, other than just `kustomization.yaml`

**Why we need it**:

Current pipecd logic only recognizes `kustomization.yaml` file as the config file for kustomize, while the official kustomize mentions that we should support three file names: `kustomization.yaml, kustomization.yml, Kustomization`

ref: https://github.com/kubernetes-sigs/kustomize/blob/7c04cbb23791c0675cfaa3806ef9b99d59698a98/api/konfig/general.go#L10-L16

Fixes #

### DIFF
--- a/pkg/app/piped/platformprovider/kubernetes/kubernetes.go
+++ b/pkg/app/piped/platformprovider/kubernetes/kubernetes.go
@@ -40,6 +40,4 @@ const (
 	IgnoreDriftDetectionTrue = "true"
 	UseReplaceEnabled        = "enabled"
 	UseServerSideApply       = "true"
-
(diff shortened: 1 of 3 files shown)


ffjlabo:
@khanhtc1202 
Could you implement the logic in the k8s plugin and k8s multicluster side?

khanhtc1202:
@ffjlabo Sure, I'm creating a separated PR for that

ffjlabo (approved):
LGTM

t-kikuc (approved):

PR #6131, opened 2025-08-17
Improve make file

linhdangduy:
**What this PR does**:

1. **Remove kind-registry when make kind-down**
2. **Check if registry is connected to kind network before connect**

**Why we need it**:

1. Remove kind-registry when make kind-down
kind-registry is not removed as describing in make kind-down [here](https://github.com/pipe-cd/pipecd/blob/4d42ed8a3b0aa79e402725144e276029ef557e65/CONTRIBUTING.md?plain=1#L181)
> When cleaning up, run `make kind-down` to stop and delete the registery and the cluster.

2. Check if registry is connected to kind network before connect
This change for avoiding duplicated executing of network connect. Make following kind tutorials [here](https://kind.sigs.k8s.io/docs/user/local-registry/)

Fixes #

- **How are users affected by this change**: No affected
- **How to migrate (if breaking change)**: It is not breaking change


khanhtc1202:
Nice catch, thanks @linhdangduy 👍 Maybe we can separate the command to prepare the local registry and the kind-cluster. Something like
```makefile
.PHONY: local-cluster
local-cluster: local-registry kind-cluster
```
WDYT?

linhdangduy:
@khanhtc1202 Not sure I understand the idea. Does it mean defining separated .PHONY for `local-registry`, `kind-cluster` then executing them together with `make local-cluster`? But then we have to have local-cluster `up`, and `down` separately as well 🤔 

The current structure is, having `kind-up` which create kind cluster and the local-registry at the same time. So this corresponding `kind-down` is for demolishing them.

khanhtc1202:
@linhdangduy Yes, that's what I mean 👍 tbh, I don't use kind cluster for the local K8s cluster; I'm using docker4mac k8s cluster instead. The local registry preparation is required regardless of the cluster you choose, and it is a one-time preparation afair.
So, it could be better if we separated the commands for preparing the local registry and the kind-cluster, and users could run the command they wanted. Also, we have the make local-cluster command that runs both commands as recommended for the local dev environment.
For the `kind-down`, the current implementation of that command doesn't remove the local-registry container (afair), so we can rename it as `down/kind-cluster`. Thus, we should have something like

```makefile
.PHONY: up/local-cluster
up/local-cluster: up/local-registry up/kind-cluster

.PHONY: down/kind-cluster
```

linhdangduy:
@khanhtc1202 Thanks for the clarification. Understood that the local registry is necessary for any local k8s cluster. 

IMHO, the current Makefile structure (having only cluster up and down) is more consistent with existing with what we have: only having doc for kind cluster, and there is no document / process to have the local-registry up individually to use with other local k8s cluster. So I'm more prone to not separating them. WDYT? if you see that separation is better, I can update the PR to align with it.

---

Incase we have the name change from `kind-up` -> `up/local-cluster`, I should change the README / CONTRIBUTING page as well. But how about pipecd doc of historical versions like [this one](https://github.com/linhdangduy/pipecd/blob/76056f86455f66ea6fc5bcc46d4fa5667cf57fc9/docs/content/en/docs-v0.39.x/contribution-guidelines/development.md?plain=1#L41) and [this one](https://github.com/linhdangduy/pipecd/blob/76056f86455f66ea6fc5bcc46d4fa5667cf57fc9/docs/content/en/docs-v0.40.x/contribution-guidelines/development.md?plain=1#L41)? - I think it is unnecessary

khanhtc1202:
@linhdangduy I think it could be better to separate the command, and yes, we have to update the docs (README/CONTRIBUTING.md) in that case.

People who don't want to specify a local cluster other than kind can use `make up/local-cluster` (prepare both registry and kind cluster), while others can use `make up/local-registry` and use their own cluster without redundantly creating a kind cluster on their local.

linhdangduy:
@khanhtc1202 I have separated local registry and kind, and prepared the up/local-cluster and down/local-cluster .PHONY

For the documentation, I have updated all except the historical versions, for example: [this one](https://github.com/linhdangduy/pipecd/blob/76056f86455f66ea6fc5bcc46d4fa5667cf57fc9/docs/content/en/docs-v0.39.x/contribution-guidelines/development.md?plain=1#L41) and [this one](https://github.com/linhdangduy/pipecd/blob/76056f86455f66ea6fc5bcc46d4fa5667cf57fc9/docs/content/en/docs-v0.40.x/contribution-guidelines/development.md?plain=1#L41)

Issue #6133, opened 2025-08-18
Server returns runtime error index out of range for ListDeprecatedNotes api


linhdangduy:
**What happened**:
On ListDeprecatedNotes api, when getting oldest version of piped [here](https://github.com/pipe-cd/pipecd/blob/2cdfb5285218c392df64a5217fdb6a2c08843ecb/pkg/app/server/grpcapi/web_api.go#L1996-L2005), the runtime error index out of range is thrown

Reasons: on [semver.IsValid(piped.Version)](https://github.com/pipe-cd/pipecd/blob/2cdfb5285218c392df64a5217fdb6a2c08843ecb/pkg/app/server/grpcapi/web_api.go#L1998) check, there is no piped has valid Version, we end up with empty `versions` list, then `versions[0]` raises index out of range exception.

**What you expected to happen**:
We should check if `versions` is empty before the `versions[0]` access 

**How to reproduce it**:
I'm try to following CONTRIBUTING.md to getting familiar with this project.
After creating my first new piped [following this](https://github.com/pipe-cd/pipecd/blame/2cdfb5285218c392df64a5217fdb6a2c08843ecb/CONTRIBUTING.md#L204-L206), click somewhere on the UI (Project / Piped) to load page and wait for ~5 seconds, the frontend's auto request to ListDeprecatedNotes get response with `503 service unavailable`

what user see|network
---|---
<img width="787" height="843" alt="Image" src="https://github.com/user-attachments/assets/c46181c6-0d21-41fc-af56-4b19125c0267" />|<img width="956" height="549" alt="Image" src="https://github.com/user-attachments/assets/1c455aed-980f-456b-8b66-35bdea36ab4f" />

By executing `kubectl logs -f <pipecd-server-name> -n pipecd` to see server logs on local, I was able to see the error as follow image

<img width="1147" height="106" alt="Image" src="https://github.com/user-attachments/assets/b6a90dce-4c34-42ef-aa4e-758c26e142a8" />

**Environment**:
- `piped` version: 
- `control-plane` version: v0.52.0
- Others:

**Solution**
You can refer here: https://github.com/pipe-cd/pipecd/pull/6135

Some thought: I see that piped Version is optional field, when it is created, Version is just empty, that why the piped in the check has invalid Version. Should we improve the behavior around this by having the Version from beginning? Thank you.

PR #6135, opened 2025-08-18
Improve get piped's oldestVersion in ListDeprecatedNotes api

linhdangduy:
**What this PR does**:
- Get the oldest piped version by using semver.Compare instead of semver.Sort
- With this we don't have to create versions list, and also avoid index out of range error

**Why we need it**:
- Simple way to get oldest piped version. Avoid unexpected error.

Fixes https://github.com/pipe-cd/pipecd/issues/6133

result: no more error
image

- **How are users affected by this change**: No more 503 Service Available error as described in the issue


khanhtc1202 (approved):
Nice catch, thanks 💯

Issue #6068, opened 2025-07-25
Implement Analysis Stage Plugin


t-kikuc:
**What would you like to be added**:

- Implement the ANALYSIS stage plugin like the WaitStage plugin for pipedv1

**Why is this needed**:

To support the ANALYSIS stage in pipedv1 too.

### Development Note

- Use `ListStageCommands()` to fetch the commands. See https://github.com/pipe-cd/pipecd/pull/5639 including the example usage.
- Use `sdk.MetadataKeyStageDisplay` instead of `Skippers`. See https://github.com/pipe-cd/pipecd/pull/6000
- Use `sdk.GetApplicationSharedObject` and `sdk.PutApplicationSharedObject` instead of the analysisResultStore. 
See #6053

### TODO

- [x] init plugin 
- [x] import configs #6153
  - [x] stage config
  - [x] application config
  - [x] piped config
  - [x] analysisTemplate
- [ ] import analysisProvider #6155
- [ ] impl executeStage (cf. executor) #6159

mohammedfirdouss:
@t-kikuc Can I give this a try?

t-kikuc:
@mohammedfirdouss 
Thank you! I assigned you since you have experience in the Analysis: https://github.com/pipe-cd/pipecd/pull/6010

mohammedfirdouss:
@t-kikuc Thanks 👍🏼

t-kikuc:
tasks: 
- [x] First PR:  Init the plugin and Impl `FetchDefinedStages()` and `BuildPipelineSyncStages()` with tests
   - cf. https://github.com/pipe-cd/pipecd/pull/5990
   - create an empty stageOptions struct
- [ ] Impl `ExecuteStage()`
   - [ ] Second PR: Import the current `pkg/app/piped/analysisprovider`  package (modify the copyright year)
   - [ ] next PRs:  Import and MODIFY the current `pkg/app/piped/executor/analysis` 
      - See the "Deployment Note" of this issue.

mohammedfirdouss:
@t-kikuc I was just about to show you what I could deduce from the issue. do you mind to see?

t-kikuc:
@mohammedfirdouss yes please 👍 

it's better to check now than in PRs

mohammedfirdouss:
@t-kikuc Hey!! The first PR is open, i want you to look at it to see if we are on track. Arigato!

t-kikuc:
[...]

PR #6157, opened 2025-08-25
[Analysis Plugin] Add Found flag to `GetApplicationSharedObject()` to distinguish NotFound

t-kikuc:
**What this PR does**:

as title

**Why we need it**:

To distinguish 
- when the result does not exist yet
- when failed to get 

https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/apistore/analysisresultstore/store.go#L56-L71

Part of #6068 


khanhtc1202 (approved):
Got the point, LGTM 🙆‍♂️

Warashi (approved):
LGTM

PR #6161, opened 2025-08-25
[k8s plugin] Implement helm registry login

ffjlabo:
**What this PR does**:

Implement the feature to login to the OCI helm registry.

**Why we need it**:

We want to support the feature to use remote helm chart.

Part of https://github.com/pipe-cd/pipecd/issues/5764
Follow https://github.com/pipe-cd/pipecd/pull/6158


khanhtc1202 (approved):
👌

PR #6165, opened 2025-08-26
Make build/plugin supports single plugin build

khanhtc1202:
**What this PR does**:

as title

**Why we need it**:

I want to run `make build/plugin` for a specific plugin

Fixes #


khanhtc1202:
I missed, we can do that by using `make build/plugin PLUGINS=xxx`

PR #6167, opened 2025-08-26
[k8s plugin] Add custom unmarshal logic for k8s plugin configs

ffjlabo:
**What this PR does**:

SSIA

**Why we need it**:

We want to set the default values for configs.

Part of #5764


khanhtc1202 (dismissed):
👌

PR #6170, opened 2025-08-26
Remove helm remote git chart feature from kubernetes plugin

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

Remove unused config for Helm GitRemoveChart feature. Pipedv0 supports that feature, but since Helm registry is more recommended now, we decided not to bring this feature to the k8s plugin.

Follow #6162


ffjlabo (approved):
LGTM

Issue #5542, opened 2025-02-06
[epic] Migration Tasks for Plugin-arched PipeCD


t-kikuc:
**What would you like to be added**:

This issue tracks ALL tasks for migrating to the pipedv1.
When a new task or PR emerges, please add/link it to this PR.

**Why is this needed**:

- To make existing applications available in pipedv1
- To complete the migration without omission
- To make it easier to migrate for users (both Control Plane owners and Piped managers)

# Tasks

## DB in Control Plane

- [x] migrate Application's PlatformProvider to the set of plugins and deployTargets (ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/model/application.proto#L55)
  - [x] Application.Kind -> deploy_targets_by_plugin key
  - [x] Application.PlatformProvider -> deploy_targets_by_plugin value[0]

- [x] ~(Filestore) migrate from analysisResultStore to the new applicationSharedObjectStore https://github.com/pipe-cd/pipecd/pull/6040~
    - -> [ ] Ignore this pattern. Add to the guide

## Config files

- [ ] Piped Config:  **Manual edit.** TODO: Add to guide 
    - [x] analysisProvider:  Move to under `plugins` of Analysis
    - [x] convert PlatformProvider to deployTargets
        - PlatformProvider.type → plugin name
        - PlatformProvider.name → deployTarget name

- [x] Application Config
    - [x] skipOn:  Update indent of `skipOn`
    - [x] timeout: Update indent of `timeout` in `stages`

## Docs

- [ ] TBA

## Recommended migration flow

1. Update app.pipecd.yaml for v1 (commands will be provided) <- Perform `Config files` change (Application Config migate task)
2. Execute DB Migration commands <- Perform DB in ControlPlane task
3. Update piped version

t-kikuc:
There are some tasks AFTER the migration to clean the codebase.

e.g. 
- https://github.com/pipe-cd/pipecd/pull/5968 (Remove stageName condition)
    - This will affect all deployments that are made while the server does not inject `availableOperation`

t-kikuc:
[Copied to the head thread]
+1 task.

we need to migrate from `analysisResultStore` the new `applicationSharedObjectStore`

ref. https://github.com/pipe-cd/pipecd/pull/6040

t-kikuc:
[Copied to the head thread]
+1 task.

[...]

PR #6175, opened 2025-08-26
Add pipectl migrate application-config command

Warashi:
**What this PR does**:

SSIA

**Why we need it**:

We have to provide a way to migrate application config files.

Part of #5542


ffjlabo (approved):
This is the great step. Let's try!

PR #6182, opened 2025-08-27
Update migrate app-config command to support migrate skipOn and timeout value in stage config

khanhtc1202:
**What this PR does**:

Copy STAGE `timeout` and `skipOn` config under `pipeline.stages[].with` to `pipeline.stages[]` (Application config)

**Why we need it**:

We decided to make `timeout` and `skipOn` options available as stage config for all stages from pipedv1.

We keep the original `timeout` and `skipOn` config under `pipeline.stages[].with` for backward compatibility, in case users want to downgrade pipedv1 to pipedv0.
`pipeline.stages[].with.{timeout, skipOn}` will be marked as deprecated in v1.

Part of #5542 
Follow #6175 

### DIFF
--- a/pkg/app/pipectl/cmd/migrate/application_config.go
+++ b/pkg/app/pipectl/cmd/migrate/application_config.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"maps"
 	"os"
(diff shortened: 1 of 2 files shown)


ffjlabo (approved):
👍

PR #6183, opened 2025-08-27
Handling stage timeout for plugin-arch piped

khanhtc1202:
**What this PR does**:

ssia

**Why we need it**:

For the new pipedv1, the stage timeout (config via `pipeline.stages[].timeout`) will be handled by piped, not plugins. This configuration will be available for all stages.

Tested on dev with WAIT stage config duration set to 1h, while stage config timeout to 30s

```yaml
kind: Application
spec:
  name: timeout-test
  pipeline:
    stages:
      - name: WAIT
        timeout: 30s
        with:
          duration: 1h
```

Screenshot 2025-08-27 at 20 15 03

Part of #4980 


ffjlabo (approved):
LGTM

Warashi (approved):
LGTM

PR #6184, opened 2025-08-27
Improve `make run/pipecd` execution speed

### LINKED ISSUE #6152: Improve developer experience on local environment
## Problem

I'm getting familiar and exploring the code base of the project. I build and run the project by the guidance in https://github.com/pipe-cd/pipecd/blob/master/CONTRIBUTING.md page.
Every time I make change to frontend (`web/`) or backend source code (`pkg`), I have to execute `make run/pipecd` again to bring the changes to the local kind cluster to see the change on running program.

The problem is `make run/pipecd` takes very longgg time to be completed. It hinders the development process and I could not do many experiments as I wished.

## How should the development process be?

1. For frontend development, developer usually want to see the changes reflected as soon as possible.
Webpack provide Hot Module Replacement (HRM) [by default](https://webpack.js.org/guides/hot-module-replacement/#enabling-hmr), where change in source code is reflected immediately, so we don't have to execute `make run/pipecd` again. We should utilize this feature to speed up frontend development. -> https://github.com/pipe-cd/pipecd/pull/6226

2. For backend development, I haven't yet come up with the proper way to apply hot reloading for the Go web server with current local development structure.
But there could be several improvements for `make run/pipecd` so that it can be completed faster, helping us to see the result of code changes faster on pipecd program (like add dockerignore file, decrease the build step if possible): https://github.com/pipe-cd/pipecd/pull/6184/

3. For backend, I have created piped properly, have the applications run, all good...but after I shutdown the kind cluster (for example: `make stop/pipecd`), all local data is gone and I have to create it all over again -> we need to persisted the local data to local folder, so developer does not have to create piped, application... again when they execute `make stop/pipecd` or `make kind-down` even restart her computer: https://github.com/pipe-cd/pipecd/pull/6192

## Question

Do you have the same experience as the problem I described above?
- If yes, I'm happy to raise the PR. I did PoC and verified behavior on my local.
- If you have your own setup to speed it up, please share in this issue so we can further improve the local development.

## Some thoughts / The approaches that I have explored

On frontend development, originally, I tried mounting the local web/static folder into the [web/static folder used by pipecd web service](https://github.com/pipe-cd/pipecd/blob/8810ce2da4724a93d636c0737c2a59c890f39fe0/cmd/pipecd/server.go#L107), so after making some changes, we can just execute `make build/web` to reflect the change and don't need `make run/pipecd`. But I found that we can try HRM as described above.

Beside that, I did try to understand the deployment structures / infra we are using for local. The nice point is the whole setup (kind / helm chart / gateway ...) seems similar with what we use for running on stg / prod environment, but I feel that somehow it is quite complicate for local development. Then I tried to simplify the setup on local so we can have more straightforward code hot reloading for both frontend and backend (using [Air](https://github.com/air-verse/air/) for example!?), but I have not come up with the any promising solution. 
(Now I'm ok to use the current local deployment structure, it is helpful for understanding how the whole service is run.)

linhdangduy:
**What this PR does**:

1. Optimize dockerfile:
- COPY only what is necessary for each RUN command:
  - only `web/package.json` and `web/yarn.lock` are necessary for `yarn install`
  - only `.git`, `web/`, `Makefile` are necessary for `make build/web`
  - only `pkg/` and `cmd/` are necessary for `make go/build`

- Add `.dockerignore`: to ignore unnecessary file in Dockerfile COPY. After this change. Because the files to be COPYed already limited so only `web/node_modules` (and some `web/...`) are relevant, but I still keep this file (based on gitignore content + `web/node_modules`, `examples`, `docs`)

2. Remove local build in `make run/pipecd`

**Why we need it**: 🐰 We have much faster `make run/pipecd` execution time during development 🐰 

1. Optimize dockerfile:

- Utilize the docker image layer cache -> speed up build process
  - Before: if any file changes, either in web or go, all the build command: `update/web-deps`, `go mod download`, `build/web`, `build/go` be executed all over again -> this really takes time
  - After: if the dependency files are not changes, then the build command is cached and don't have executed again. For example:

|when to execute `make run/pipecd`|what will happen|
|---|---|
|first time|Image|
|subsequent build when: some web files change|only trigger `build/web`, other builds already cached<br><br>Image|
|subsequent build when:: only go files change|only trigger `build/go`, other builds already cached Image|

- It also speed up docker build time by not copy unnecessary files `web/node_modules`.

Copy consumed time Before >90 sec|Copy consumed time After: almost ~0 sec
---|---
image|image

2. Remove local build in `make run/pipecd`

- There is no usage for local build here because the [go build](https://github.com/pipe-cd/pipecd/blob/e7ec5c4fe30ed3acc63fad108d6bde4835939bbe/cmd/pipecd/Dockerfile#L28) and [web build](https://github.com/pipe-cd/pipecd/blob/e7ec5c4fe30ed3acc63fad108d6bde4835939bbe/cmd/pipecd/Dockerfile#L12-L13) already done in Dockerfile.
  - `go build`: execute very fast. We can keep here because it can fail fast if go source is valid or not (for example if we import a package but not use), but I decided to remove it as well because we can use other command to confirm on local like `make build/go`.
  - `web build`: it take around ~13 seconds in my local
    - image

It is one of the approaches improving local development

Fixes #6152 

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,60 @@
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+
+# Test binary, build with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+.DS_Store
+
+# Binaries
+.artifacts
+
+# docs
+docs
+
+examples
+
+# Go
+/vendor
+
+# nodejs
+node_modules
+
+# web
+web/node_modules
+web/dist
+web/.env
+web/.cache
+web/coverage
+pkg/app/web/.env
+
+# IDE config files
+.ijwb
+.idea
+
+.dev
+
+# Terraform workspace
+.terraform
+.terraform-credentials
+
+.rendered-manifests
+
+# manfiests
+manifests/*
+
+# gomock generated reflect files
+gomock_reflect_*/
+
+# Cache
+.cache
+
+# hack
+hack/*
--- a/.dockerignore
+++ b/.dockerignore
@@ -47,7 +47,7 @@ pkg/app/web/.env
 
 .rendered-manifests
 
-# manfiests
+# manifests
 manifests/*
 
 # gomock generated reflect files
--- a/Makefile
+++ b/Makefile
@@ -154,12 +154,6 @@ run/pipecd: BUILD_LDFLAGS_PREFIX := -X github.com/pipe-cd/pipecd/pkg/version
 run/pipecd: BUILD_OPTS ?= -ldflags "$(BUILD_LDFLAGS_PREFIX).version=$(BUILD_VERSION) $(BUILD_LDFLAGS_PREFIX).gitCommit=$(BUILD_COMMIT) $(BUILD_LDFLAGS_PREFIX).buildDate=$(BUILD_DATE) -w"
 run/pipecd: CONTROL_PLANE_VALUES ?= ./quickstart/control-plane-values.yaml
 run/pipecd:
-	@echo "Building go binary of Control Plane..."
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 $(BUILD_ENV) go build $(BUILD_OPTS) -o ./.artifacts/pipecd ./cmd/pipecd
-
-	@echo "Building web static files..."
-	yarn --cwd web build
-
 	@echo "Building docker image and pushing it to local registry..."
 	docker build -f cmd/pipecd/Dockerfile -t localhost:5001/pipecd:$(BUILD_VERSION) .
 	docker push localhost:5001/pipecd:$(BUILD_VERSION)
--- a/cmd/pipecd/Dockerfile
+++ b/cmd/pipecd/Dockerfile
@@ -10,7 +10,9 @@ RUN apk add --no-cache make git
 COPY web/package.json web/yarn.lock ./web/
 RUN yarn --cwd web install --prefer-offline
 
-COPY . .
+COPY .git ./.git
+COPY web ./web
+COPY Makefile .
 RUN make build/web
 
 # pipecd builder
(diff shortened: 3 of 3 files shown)


.dockerignore
```diff
@@ -0,0 +1,60 @@
...
+.dev
+
+# Terraform workspace
+.terraform
+.terraform-credentials
+
+.rendered-manifests
+
+# manfiests
+manifests/*
```
Okabe-Junya:
```suggestion
# manifests
manifests/*
```

typo

Okabe-Junya (commented):
Thanks!
(unofficial review but,) LGTM except one minor typo.

.dockerignore
linhdangduy: Thank a lot!
Applied the fix.

khanhtc1202 (approved):
LNeatTM 🙆

Warashi (approved):
Thank you!

ffjlabo:
@linhdangduy  cc @khanhtc1202 @Warashi 

We are sorry but we will revert this PR because of the problem below. 🙏 
We want to find another way to improve the speed.

Currently, we didn't expect to exist the suffix `-dirty` for the version for the release after v0.54.0.

PipeCD

This version is determined when running `yarn --cwd web build` in `make build/web`.
https://github.com/pipe-cd/pipecd/blob/453583fff0e8e4c7c161ff8ddc2c11e1b925b01f/web/package.json#L8


```
PIPECD_VERSION=`git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*'`
```

When there is a `--dirty` option, it means some uncommitted changes exist in the local repo.
https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---dirtymark
> Describe the state of the working tree. When the working tree matches HEAD, the output is the same as "git describe HEAD". If the working tree has local modification "-dirty" is appended to it. If a repository is corrupt and Git cannot determine if there is local modification, Git will error out, unless ‘--broken’ is given, which appends the suffix "-broken" instead.

Also, from the perspective of build reproducibility, it is not desirable to build with uncommitted local differences.

PR #6190, opened 2025-08-29
Add release plugin workflow

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

We want to have a way to release plugins independently

Part of #4980


Warashi (approved):
Let's try 👍🏻

PR #6192, opened 2025-08-29
Keep pipecd server data even when developer shut down pipecd or kind on local computer

linhdangduy:
**What this PR does**:

- Create docker volume `pipecd-data`.
- Mount above volume to local kind cluster, and use it for mysql, minio, cache deployment.

**Why we need it**:

On local, every time `make stop/pipecd` is executed, all the local data that the developer has "worked hard" to create on the local pipecd server also got lost, then she have to create them all over again on next pipecd run.
The root cause is: the mysql, minio, cache storage deployments got restart, and their data also lost.

This PR is for mounting the local volume to storage data in kind cluster, so the data will be kept even when the mysql, minio, cache storage deployments are shut down. When the developer executes `make stop/pipecd`, or `make kind-down` or even restarts her computer, she would still have all data available on the next run.

Solve Item 3 in #6152 

**Test**

To test this PR's change, you can do the following step:
1. turn off current kind cluster if there is: `make kind-down`
2. `make kind-up` -> `make run/pipecd` -> port forward ... then access to pipecd web
3. Create piped, or new api key
4. `make stop/pipecd` -> `make kind-down`: Turn off pipecd/kind
5. Run kind, pipecd again like step (2), then go to piped or api key manage page. We can see that the created ones on step (3) still be available.

To confirm the pipecd-data volume content:
```sh
 docker run -v $(docker volume inspect pipecd-data --format '{{ .Mountpoint }}'):/mount alpine ls /mount

# minio
# mysql
# redis
```


hack/create-kind-cluster.sh
```diff
@@ -51,6 +51,11 @@ REG_CONFIG_DIR="/etc/containerd/certs.d"
 cat <<EOF | kind create cluster --name ${CLUSTER} --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+  - hostPath: $(pwd)/.pipecd-data
```
Okabe-Junya: When I checked out this branch on my laptop (M1 Macbook Pro; launcher desktop) and verified it, the following error occurred.

```
➜  pipecd git:(kind-mount-volume-to-keep-storage-data-on-local) make kind-up
./hack/create-kind-cluster.sh pipecd
Creating local registry container...
...
Creating cluster "pipecd" ...
 ✓ Ensuring node image (kindest/node:v1.34.0) 🖼 
 ✗ Preparing nodes 📦  
Deleted nodes: ["pipecd-control-plane"]
ERROR: failed to create cluster: command "docker run --name pipecd-control-plane --hostname pipecd-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --label io.x-k8s.kind.cluster=pipecd --net kind --restart=on-failure:1 --init=false --cgroupns=private --volume=/Users/junyaokabe/src/github.com/Okabe-Junya/pipecd/.pipecd-data:/tmp/pipecd-data --publish=127.0.0.1:58123:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.34.0@sha256:7416a61b42b1662ca6ca89f02028ac133a309a2a30ba309614e8ec94d976dc5a" failed with error: exit status 126
Command Output: 06917ad86e097ff38166a92cd0d235e966634b1c5a96a6f5acbdc8518efabe7b
docker: Error response from daemon: error while creating mount source path '/Users/junyaokabe/src/github.com/Okabe-Junya/pipecd/.pipecd-data': chown /Users/junyaokabe/src/github.com/Okabe-Junya/pipecd/.pipecd-data: permission denied.
```

I think the cause of this error is that `.pipecd-data` directory doesn't exist. 
It looks that we need to run `mkdir -p .pipecd-data` before in this block

Okabe-Junya (commented):
Really nice improvement!
left just a minor comment, PTAL :)

hack/create-kind-cluster.sh
linhdangduy: @Okabe-Junya Thank you for trying it out. I also tried to reproduce above error on my local.

TL;DR: I have updated it with better way of creating volume. Please try this PR's change again! 🚀 

### The behavior
docker does not have the permission to using `/Users` directory on your local to mounting the volume. So even if `.pipecd-data` folder exist, it could not use the folder for mounting.

### Docker desktop behavior - root cause
There is the `File sharing` setting (image 1: example for my setting), which we can config the eligible directory path for mounting.
By default, it already include `/Users` directory, so there should be no error -> [reference](https://docs.docker.com/desktop/settings-and-maintenance/settings/#file-sharing:~:text=By%20default%20the%20/Users%2C%20/Volumes%2C%20/private%2C%20/tmp%20and%20/var/folders%20directory%20are%20shared)
But somehow, on your local computer, probably this config does not include `/Users` directory.

So, to resolve this problem, we have to make sure  `/Users` or at least `/Users/junyaokabe/src/github.com/Okabe-Junya/pipecd/` is mountable. On my local, docker shows error explicitly as you see below image

when|behavior
---|---
Default setting|Pasted_Image_2025_09_10__2_36
I reproduce the error by removing `/Users` from config|Pasted_Image_2025_09_10__2_39

### How to overcome the problem
I see that it is quite an overhead for developers if they have similar config like your local, so I tried to find another way to create the folder seamlessly and developers don't have to configure anything 😌 

The solution is creating docker volume `pipecd-data`, then mount it to local kind cluster! https://github.com/pipe-cd/pipecd/pull/6192/commits/afe54b27279db765a2fcb0e7cd41d6de19161523
Now we have natural volume and still making sure minio, mysql, cache data persisted even when we stop pipecd/kind. And actually it is better than previous way 🎉 

Please try the change again when you have time 🙇🏼

Okabe-Junya: Thanks, I think my error is due to a known bug occurring in the launcher desktop - https://github.com/rancher-sandbox/rancher-desktop/issues/1209. 
Anyway, I think afe54b2 is a good improvement. Thanks

Okabe-Junya (dismissed):
Thanks, LGTM!

linhdangduy:
@Okabe-Junya Thank you. I have just resolve a conflict with master branch. Please check it again 🙇🏼

Warashi (approved):
LGTM 👍🏻

hack/create-kind-cluster.sh
```diff
@@ -36,11 +36,24 @@ CLUSTER=$1
...
 # Create a cluster with the local registry enabled in containerd
 REG_CONFIG_DIR="/etc/containerd/certs.d"
 cat <<EOF | kind create cluster --name ${CLUSTER} --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+  - hostPath: ${VOLUME_MOUNT_POINT}
+    containerPath: /tmp/pipecd-data
```
khanhtc1202, line 56: Just realized this only auto creates volumes for Kind, and I'm using Docker4Mac k8s cluster on my local, so I have to do it myself 😂 cc @linhdangduy

PR #6201, opened 2025-09-03
Enable ssh configration for pipedv1

ffjlabo:
**What this PR does**:

as title
To realize this, we decided to define a type alias for `configv1.PipedGit` in `pkg/config`.

Currently, it's ok because `configv1.PipedGit` and `config.PipedGit` are the same definition.

```go
// in pkg/config
type PipedGit = configv1.PipedGit
```

⚠️ We need to consider the backward capability for v0 when fixing `configv1.PipedGit`. 
We decided to do it to fix the config easily in the future, and we will support only pipedv1.

**Why we need it**:

The current pipedv1 can't clone git repo with ssh.

Part of #5259

### DIFF
--- a/pkg/app/pipedv1/cmd/piped/piped.go
+++ b/pkg/app/pipedv1/cmd/piped/piped.go
@@ -167,7 +167,7 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 	// Register all metrics.
 	registry := registerMetrics(cfg.PipedID, cfg.ProjectID, p.launcherVersion)
 
-	// // Configure SSH config if needed.
+	// Configure SSH config if needed.
 	if cfg.Git.ShouldConfigureSSHConfig() {
 		tempFile, err := git.AddSSHConfig(cfg.Git)
 		if err != nil {
--- a/pkg/app/pipedv1/cmd/piped/piped.go
+++ b/pkg/app/pipedv1/cmd/piped/piped.go
@@ -169,10 +169,12 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 
 	// // Configure SSH config if needed.
 	if cfg.Git.ShouldConfigureSSHConfig() {
-		if _, err := git.AddSSHConfig(cfg.Git); err != nil {
+		tempFile, err := git.AddSSHConfig(cfg.Git)
+		if err != nil {
 			input.Logger.Error("failed to configure ssh-config", zap.Error(err))
 			return err
 		}
+		defer os.Remove(tempFile)
 		input.Logger.Info("successfully configured ssh-config")
 	}
 
--- a/pkg/app/pipedv1/cmd/piped/piped.go
+++ b/pkg/app/pipedv1/cmd/piped/piped.go
@@ -168,13 +168,13 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 	registry := registerMetrics(cfg.PipedID, cfg.ProjectID, p.launcherVersion)
 
 	// // Configure SSH config if needed.
-	// if cfg.Git.ShouldConfigureSSHConfig() {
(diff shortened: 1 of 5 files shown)


pkg/app/pipedv1/cmd/piped/piped.go
```diff
@@ -168,13 +168,13 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
...
-	// 	}
-	// 	input.Logger.Info("successfully configured ssh-config")
-	// }
+	if cfg.Git.ShouldConfigureSSHConfig() {
+		if _, err := git.AddSSHConfig(cfg.Git); err != nil {
+			input.Logger.Error("failed to configure ssh-config", zap.Error(err))
+			return err
+		}
+		input.Logger.Info("successfully configured ssh-config")
+	}
```
khanhtc1202: Should update to remove ssh key file as we do for pipedv0
ref: https://github.com/pipe-cd/pipecd/blob/341526906dfbb3b93f6feff52ccee3157cc1eb3d/pkg/app/piped/cmd/piped/piped.go#L180-L189

pkg/git/ssh_config.go
```diff
@@ -22,7 +22,7 @@ import (
 	"path/filepath"
 	"text/template"
 
-	"github.com/pipe-cd/pipecd/pkg/config"
+	configv1 "github.com/pipe-cd/pipecd/pkg/configv1"
```
khanhtc1202, line 25:
```suggestion
	"github.com/pipe-cd/pipecd/pkg/configv1"
```
Or how about making this like we did in pipedv1 entry
ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/app/pipedv1/cmd/piped/piped.go#L72

pkg/app/pipedv1/cmd/piped/piped.go
ffjlabo: Thank you, will fix it.
Note: It is supported in the PR https://github.com/pipe-cd/pipecd/pull/5769

pkg/git/ssh_config.go
ffjlabo: @khanhtc1202 
This package is for both pipedv0 and pipedv1, so it would be nice to distinguish config and configv1 to avoid misinterpreting them.
So I think it is good as is.
WDYT?

pkg/app/pipedv1/cmd/piped/piped.go
ffjlabo: fixed on [01304fe](https://github.com/pipe-cd/pipecd/pull/6201/commits/01304fe667549747b17d187f670cf7914140bf96)

pkg/git/ssh_config.go
khanhtc1202: Got your point 🙆‍♂️

pkg/app/pipedv1/cmd/piped/piped.go
```diff
@@ -168,13 +168,15 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 	registry := registerMetrics(cfg.PipedID, cfg.ProjectID, p.launcherVersion)
 
 	// // Configure SSH config if needed.
```
khanhtc1202:
```suggestion
	// Configure SSH config if needed.
```
nits

khanhtc1202 (dismissed):
LGTM, letf a nits 🙆‍♂️

pkg/app/pipedv1/cmd/piped/piped.go
ffjlabo: fixed [05ee84e](https://github.com/pipe-cd/pipecd/pull/6201/commits/05ee84e2cc8dca9f1d6584c6d87f571d756f951f)

khanhtc1202 (approved):
LGTM, thanks 🙆‍♂️

PR #6208, opened 2025-09-03
Add migration to pipedv1 blog

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

The guideline for users to migrate from piped to pipedv1.

Part of #5542


khanhtc1202:
Close, we will prepare docs for migration instead of blog

PR #6216, opened 2025-09-12
Refine local environment commands in makefile

khanhtc1202:
**What this PR does**:

ssia

**Why we need it**:

Follow PR #6131

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@
 #   gen:     execute code or docs generation
 #   release: commands used in release flow
 #   push:    push artifacts such as helm chart
+#   up:      prepare local environment (registry, kind cluster)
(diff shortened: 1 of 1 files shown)


khanhtc1202:
@linhdangduy PTAL when you have time 👀

Makefile
```diff
@@ -290,11 +291,19 @@ up/kind-cluster:
...
 up/local-cluster: up/local-registry up/kind-cluster
 
-.PHONY: down/local-cluster
-down/local-cluster:
+.PHONY: down/kind-cluster
+down/kind-cluster:
 	kind delete cluster --name pipecd
+
+.PHONY: down/local-registry
+down/local-registry:
```
khanhtc1202, line 299: Better to separate this from `down/kind-cluster` to keep the built container images on local.

Issue #4845, opened 2024-03-21
"Error: Invalid Semantic Version" running `make run/pipecd`


dgannon991:
**What happened**: Attempting to install pipeCD in my local cluster, it threw an error about invalid semantic versioning

**What you expected to happen**: It installed successfully.

**How to reproduce it**: Follow the local running steps on a linux laptop. 

**Environment**:   
- `piped` version: https://github.com/pipe-cd/pipecd/commit/13c3ca06b6669af999750636091233005b967338
- `control-plane` version: N/A
- Others: 
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"

Helm: version.BuildInfo{Version:"v3.14.3", GitCommit:"f03cc04caaa8f6d7c3e67cf918929150cf6f3f12", GitTreeState:"clean", GoVersion:"go1.21.7"}

ffjlabo:
@dgannon991 Thank you for submit this bug noticed at the contribute fest's one :)
If possible, can I add the logs to the comment?

ffjlabo:
📝 It happens when executing `make run/pipecd`

dgannon991:
Hi Guys,

The logs are

`david@Davids-Laptop:~/Projects/contribfest/pipecd-pipecd$ make run/pipecd
Building go binary of Control Plane...
GOOS=linux GOARCH=amd64 CGO_ENABLED=0  go build -ldflags "-X github.com/pipe-cd/pipecd/pkg/version.version="db771fe-1711281878" -X github.com/pipe-cd/pipecd/pkg/version.gitCommit=db771fe61f5f68d944869b7624759a2c91bdbf68 -X github.com/pipe-cd/pipecd/pkg/version.buildDate=20240324-120438 -w" -o ./.artifacts/pipecd ./cmd/pipecd
Building web static files...
yarn --cwd web build
yarn run v1.22.22
$ PIPECD_VERSION=`git describe --tags --always --dirty --abbrev=7` webpack build --mode production --env htmlTemplate=./base.html --config ./webpack.config.js
assets by status 4.38 MiB [cached] 14 assets
assets by path . 15.7 KiB
  asset favicon.ico 15 KiB [compared for emit]
  asset index.html 635 bytes [compared for emit]
[...]

PR #6239, opened 2025-09-17
web generate pipecd version based only on web folder

linhdangduy:
- **Revert "Revert "Improve `make run/pipecd` execution speed (#6184)" (#6235)"**
- **get PIPECD_VERSION before docker build**

**What this PR does**:

**Why we need it**:

Fixes #


linhdangduy:
@Warashi Can you try execute `make stop/pipecd` once before `make run/pipecd`, and see the result?

As evidence share in the PR's description, I see that the version is available at the top header, and also in the menu tab
image

khanhtc1202:
@linhdangduy Thanks for the patch 👍 
I also found another place that we needed to check, which is `make build/go MOD=piped`. This is the bug I found with v0.54.0 (the always `-dirty` version)

```bash
❯ docker run --rm -it --entrypoint /bin/sh ghcr.io/pipe-cd/piped:v0.54.0
/ $ piped version
Version: v0.54.0-dirty, GitCommit: 025ac2ec2d7b7edcc774a3cc548819e9c9fc9c6d, BuildDate: 20250916-063057
```

I think it's caused by the Dockerignore, which makes Docker ignore files that are watched by git.

Warashi:
@linhdangduy 

I retried and got the same result.

As far as I know, the `docker build` doesn't carry environment variables to container image. So it's not the correct way that we pass the PIPECD_VERSION as the environment variable when running the `docker build`.
Alternatively, we can use `ARG` in the Dockerfile and pass argument with `--build-arg` option.
ref; https://docs.docker.com/reference/dockerfile/#arg

linhdangduy:
@khanhtc1202 @Warashi 
Thank you for the suggestion! 🙇🏼 

Understood the root cause that `piped` (and builds that are not pipecd) build also produce `-dirty` tag is we have `.dockerignore` exclude the files watch by git. The viable option is just have `.dockerignore` content same with `.gitignore`, so there should be no `-dirty`
And for pipecd build, both `.dockerignore` and selective copy (only copy some of the local content to docker image) cause the issue.

Let's me check how can the issue be resolved.

web/package.json
```diff
@@ -5,8 +5,8 @@
   "private": true,
   "license": "Apache-2.0",
   "scripts": {
-    "build": "PIPECD_VERSION=`git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*'` webpack build --mode production --env htmlTemplate=./base.html --config ./webpack.config.js",
-    "dev": "PIPECD_VERSION=`git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*'` webpack serve --env htmlTemplate=./base.html --config ./webpack.config.dev.js",
+    "build": "PIPECD_VERSION=${PIPECD_VERSION:-`git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*'`} webpack build --mode production --env htmlTemplate=./base.html --config ./webpack.config.js",
+    "dev": "PIPECD_VERSION=${PIPECD_VERSION:-`git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*'`} webpack serve --env htmlTemplate=./base.html --config ./webpack.config.dev.js",
```
linhdangduy: confirmed on local

when|evidence
---|---
directly set customized PIPECD_VERSION value at build/web<br>`PIPECD_VERSION=v1-customized_web_version yarn --cwd web build`|image
remove PIPECD_VERSION at build/web<br>`yarn --cwd web build`<br>note: `-dirty` appeared because Makefile is changed|image

linhdangduy:
@Warashi @khanhtc1202 @ffjlabo 
Sorry for the trouble on the previous #6184 PR.
I have pushed the fix of the root cause of the error.
The solution is getting BUILD_VERSION first, then passing it to the docker build.
The reason I still would like to this is we will have fast build time for pipecd, accelerate the local dev experience for dev.

Please also refer to the PR's description and review when you have time 🙇🏼

.github/workflows/publish_image_chart.yaml
```diff
@@ -92,6 +92,8 @@ jobs:
           context: .
           file: ${{ matrix.dockerfile }}
           platforms: linux/amd64,linux/arm64
+          build-args: |
+            BUILD_VERSION=${{ env.PIPECD_VERSION }}
```
linhdangduy, line 96: confirmed the bug that happened on github action for the previous v0.54.0 build. For example

build|image
----|---
[piped](https://github.com/pipe-cd/pipecd/actions/runs/17756788285/job/50460923301)|image
[launcher](https://github.com/pipe-cd/pipecd/actions/runs/17756788285/job/50460923304)|image

Expected that all of them will be fixed by this arg passing

Makefile
```diff
@@ -155,15 +156,10 @@ run/pipecd: BUILD_DATE ?= $(shell date -u '+%Y%m%d-%H%M%S')
 run/pipecd: BUILD_LDFLAGS_PREFIX := -X github.com/pipe-cd/pipecd/pkg/version
 run/pipecd: BUILD_OPTS ?= -ldflags "$(BUILD_LDFLAGS_PREFIX).version=$(BUILD_VERSION) $(BUILD_LDFLAGS_PREFIX).gitCommit=$(BUILD_COMMIT) $(BUILD_LDFLAGS_PREFIX).buildDate=$(BUILD_DATE) -w"
 run/pipecd: CONTROL_PLANE_VALUES ?= ./quickstart/control-plane-values.yaml
+run/pipecd: PIPECD_VERSION ?= $(shell git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*')
```
Warashi: I found the note below on L150-152. I think we can use `BUILD_VERSION` as `--build-arg BUILD_VERSION=$(BUILD_VERSION)` instead of defining `PIPECD_VERSION`

> NOTE: previously `git describe --tags` was used to determine the version for running locally
> However, this does not work on a forked branch, so the decision was made to hardcode at version 0.0.0
> see: https://github.com/pipe-cd/pipecd/issues/4845

linhdangduy: @Warashi Thanks for your comment.

> I found the note below on L150-152. I think we can use BUILD_VERSION as --build-arg BUILD_VERSION=$(BUILD_VERSION) instead of defining PIPECD_VERSION

yeah, that is the reason I add PIPECD_VERSION here. Using the local BUILD_VERSION to pass to docker file is fine for me. I will fix it

> There are arguments/variables named BUILD_VERSION and PIPECD_VERSION with the similar meaning. It's a bit confusing, so I want to use only either of them.

Understand your point. IMO, I prefer to use BUILD_VERSION than PIPECD_VERSION because: go backend code use BUILD_VERSION internally, and it is more general and can pass to any go build command (for pipectl / pided...)

But PIPECD_VERSION is also written in many places like:
- web [build](https://github.com/pipe-cd/pipecd/blob/2fd31dabaa784cfe0b1b5793d88be43d96d915ef/web/package.json#L8) / [source code](https://github.com/search?q=repo%3Apipe-cd%2Fpipecd+process.env.PIPECD_VERSION&type=code) 
- [github action build](https://github.com/pipe-cd/pipecd/blob/2fd31dabaa784cfe0b1b5793d88be43d96d915ef/.github/workflows/publish_image_chart.yaml#L63-L64)

So for now, there are 2 approaches: 
1. for the new code in the scope of PR, use the BUILD_VERSION variable, do not touch the existing place that have PIPECD_VERSION defined (and work on renaming later)
2. OR change PIPECD_VERSION -> BUILD_VERSION as well

I'm thinking of going for first approach.
What do you think?

Warashi: @linhdangduy 
Thank you for explaining.

I think it's good to going the first approach. Please doing so?

linhdangduy: Sure. Sorry for late. I have unified the new code to BUILD_VERSION 🙇🏼 

https://github.com/pipe-cd/pipecd/pull/6239/commits/a5709928cd9a0f0f63c1498ed4a8469a3c4d7d74
and https://github.com/pipe-cd/pipecd/pull/6239/commits/ba6eddabe49459f29b16f63f4b42a045a2e2d9b7

linhdangduy: local web access via http://localhost:8080 would become like this 
image

PR #6254, opened 2025-09-19
Rename log persister to stage log persister sdk

khanhtc1202:
**What this PR does**:

ssia

**Why we need it**:

Better to name it as what it is, reduce the risk of using this as a general log persister for plugins.

Fixes #


ffjlabo (approved):
LGTM

PR #6256, opened 2025-09-19
Fix failed tests on action planpreview

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

Fix tests reflect the change on the planpreview result via #6248 

Follow PR #6248


Warashi (approved):
OK, I reproduced the issue on my local machine.
LGTM 👍🏻

PR #6257, opened 2025-09-20
Change test/lint workflow to correctly report completion status

Warashi:
**What this PR does**:

as title

**Why we need it**:

to use it in branch protection rule

Fixes #

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -50,10 +50,18 @@ jobs:
   # It is used to set as required check for the branch protection rules
   go-lint-completed:
     runs-on: ubuntu-24.04
+    if: always()
     needs: go
     steps:
-      - run: |
-          echo completed
+      - name: Check if all go lint jobs succeeded
+        # if jobs in the 'go' job matrix failed or were cancelled, this job will fail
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -50,10 +50,18 @@ jobs:
...
     needs: go
     steps:
-      - run: |
-          echo completed
+      - name: Check if all go lint jobs succeeded
+        # if jobs in the 'go' job matrix failed or were cancelled, this job will fail
+        # otherwise this job is marked as successful because all steps are skipped
+        run: exit 1
+        if: >-
+          ${{
```
khanhtc1202, line 61: Thanks for the patch 👍 
I'm not sure this could help, but I checked the CI flow that was supposed to fail due to the test change. It passed anyway because it was not actually running but returned from cache. So maybe even this check doesn't change the issue, wdyt?
https://github.com/pipe-cd/pipecd/actions/runs/17848010213/job/50780173450#step:4:12

Warashi: #6253 was merged without passing the tests, and then the CI on the master branch failed.
This PR can fix like this case.

khanhtc1202 (approved):
LGTM 👍

PR #6259, opened 2025-09-22
[terraform plugin] Comment out stage config unmarshaling logic for TERRAFORM_APPLY

ffjlabo:
**What this PR does**:

as title

**Why we need it**:

Currently, the `TERRAFORM_APPLY` stage doesn't have any options.
https://github.com/pipe-cd/pipecd/blob/96c15d8ad7499d96f80471a899e1305d8f08bcbd/pkg/app/pipedv1/plugin/terraform/config/config.go#L61-L63

So we don't need the logic for now. Also, it led the failure to execute `TERRAFORM_APPLY` as QuickSync 

Fixes #

### DIFF
--- a/pkg/app/pipedv1/plugin/terraform/deployment/apply.go
+++ b/pkg/app/pipedv1/plugin/terraform/deployment/apply.go
@@ -32,13 +32,6 @@ func (p *Plugin) executeApplyStage(ctx context.Context, input *sdk.ExecuteStageI
 		return sdk.StageStatusFailure
 	}
 
-	// TODO: use stageConfig if this stage has any options in the future.
-	// stageConfig := config.TerraformApplyStageOptions{}
-	// if err := json.Unmarshal(input.Request.StageConfig, &stageConfig); err != nil {
-	// 	lp.Errorf("Failed to unmarshal stage config (%v)", err)
-	// 	return sdk.StageStatusFailure
-	// }
-
 	if err = cmd.Apply(ctx, lp); err != nil {
 		lp.Errorf("Failed to apply changes (%v)", err)
 		return sdk.StageStatusFailure
--- a/pkg/app/pipedv1/plugin/terraform/deployment/apply.go
+++ b/pkg/app/pipedv1/plugin/terraform/deployment/apply.go
@@ -16,7 +16,6 @@ package deployment
 
 import (
 	"context"
-	"encoding/json"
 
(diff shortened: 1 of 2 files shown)


t-kikuc (approved):
thank you

pkg/app/pipedv1/plugin/terraform/deployment/apply.go
```diff
@@ -33,11 +32,12 @@ func (p *Plugin) executeApplyStage(ctx context.Context, input *sdk.ExecuteStageI
 		return sdk.StageStatusFailure
 	}
 
-	stageConfig := config.TerraformApplyStageOptions{}
-	if err := json.Unmarshal(input.Request.StageConfig, &stageConfig); err != nil {
-		lp.Errorf("Failed to unmarshal stage config (%v)", err)
-		return sdk.StageStatusFailure
-	}
+	// TODO: use stageConfig if this stage has any options in the future.
+	// stageConfig := config.TerraformApplyStageOptions{}
```
khanhtc1202: @ffjlabo Should we remove the `config.TerraformApplyStageOptions` structure and this commented code block? If it's not being used, better to remove it 🤔

khanhtc1202: Also, the root cause of this issue is more like the `input.Request.StageConfig` is nil, so it's better to have a nil check for `input.Request.StageConfig` before unmarshalling, wdyt?

ffjlabo: @khanhtc1202 
As you said, this option is not used for now. I will remove this logic and struct.
https://github.com/pipe-cd/pipecd/blob/af58b81d1d576b11ae90908aee6ad6f7632dd89a/pkg/config/application_terraform.go#L60-L73

> Also, the root cause of this issue is more like the input.Request.StageConfig is nil, so it's better to have a nil check for input.Request.StageConfig before unmarshalling, wdyt?

It is good for the work around. But we might solve this problem in the piped or SDK.
But I will remove this logic.

khanhtc1202: Thanks, please remove the code block instead of comment out.

khanhtc1202 (approved):
Thank you! 🙏 
Note: We will make another issue for updating the SDK to ensure the stage config is not nil.

PR #6274, opened 2025-09-26
Update deployment trigger sync strategy based on planned step output

### LINKED ISSUE #6207: Ensure deployment trigger sync strategy be reflected after planning step
**What happened**:

When clicking `SYNC` on the application detailed page, the deployment triggered by this action will have a `trigger.sync_strategy` value of `AUTO`. After the deployment planning step, the in-model value remains 'AUTO' even if the strategy is determined as `QUICK_SYNC` or `PIPELINE_SYNC`. This makes logic that depends on the `deployment.trigger.sync_strategy` value wouldn't work as expected

Should update the in-model `deployment.trigger.sync_strategy` value to the type determined by the planning step.

ref: https://github.com/pipe-cd/pipecd/blob/bc292f2a13f2c665c10dbd0c6ddb2076d168dfb9/pkg/app/pipedv1/controller/planner.go#L248

NOTE: This change will only be reflected in pipedv1

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

If users trigger deployment under the AUTO strategy, the value stored in `model.deployment.trigger.sync_strategy` remains as is while executing the deployment. That makes later logic depend on the deployment's strategy value work unexpectedly, since we only support QUICK_SYNC and PIPELINE_SYNC. The AUTO strategy will be inferred to one of QUICK or PIPELINE sync by the planning step; thus, the value in the deployment model should be updated to reflect the inferred output.

Fixes #6207


pkg/app/pipedv1/controller/planner.go
```diff
@@ -573,6 +573,7 @@ func (p *planner) reportDeploymentPlanned(ctx context.Context, out *plannerOutpu
 
 	req := &pipedservice.ReportDeploymentPlannedRequest{
 		DeploymentId:              p.deployment.Id,
+		SyncStrategy:              out.SyncStrategy,
```
t-kikuc, line 576: [ASK] I think pipedv0 needs the same logic, is that correct?

khanhtc1202: Not exactly. Pipedv0 can also have this logic, but the current pipedv0 stage executing steps logic doesn't depend on the deployment sync strategy value, so the same issue we faced on pipedv1 has not occurred.

khanhtc1202: Aka. issue #6207 only occurred on the pipedv1 application deployment.

t-kikuc: ok

PR #6275, opened 2025-09-26
Fix panic on cancel stage while executing

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

To handle stage cancel, the current logic doesn't cover the case when the `plugin.ExecuteStage` gets a cancel signal from the context. That causesan  error as below

```bash
 command CancelDeployment was forwarded to its scheduler       {}
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c90416]

goroutine 2202 [running]:
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.deferwrap1()
        /Users/s12228/workspace/pipecd/pipecd/vendor/go.opentelemetry.io/otel/sdk/trace/span.go:398 +0x25
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End(0xc0001bda00, {0x0, 0x0, 0xc0009108c0?})
        /Users/s12228/workspace/pipecd/pipecd/vendor/go.opentelemetry.io/otel/sdk/trace/span.go:436 +0xa4f
panic({0x21721c0?, 0x2cc7a70?})
        /usr/local/go/src/runtime/panic.go:783 +0x132
github.com/pipe-cd/pipecd/pkg/app/pipedv1/controller.(*scheduler).executeStage(0xc0006eac60, {0x2392468, 0xc000ca3ec0}, 0xc0007be1c0)
        /Users/s12228/workspace/pipecd/pipecd/pkg/app/pipedv1/controller/scheduler.go:615 +0xe36
github.com/pipe-cd/pipecd/pkg/app/pipedv1/controller.(*scheduler).Run.func3()
        /Users/s12228/workspace/pipecd/pipecd/pkg/app/pipedv1/controller/scheduler.go:346 +0x747
created by github.com/pipe-cd/pipecd/pkg/app/pipedv1/controller.(*scheduler).Run in goroutine 2180
        /Users/s12228/workspace/pipecd/pipecd/pkg/app/pipedv1/controller/scheduler.go:335 +0x1637
exit status 2
make: *** [run/piped] Error 1
```

The codebase referred to `res.Status`, which is a nil reference since res is nil in case the plugin execution stage is canceled, causing a panic.

Before (piped paniked, need to restart to keep working on running deployment)
Screenshot 2025-09-26 at 17 42 11

After (go through as expected)
Screenshot 2025-09-26 at 17 41 47

Follow PR #5597


pkg/app/pipedv1/controller/scheduler.go
```diff
@@ -603,14 +603,25 @@ func (s *scheduler) executeStage(sig StopSignal, ps *model.PipelineStage) (final
...
-	// otherwise, return the error.
+
+	// Return stage failure status if the plugin execution failed and the context is not cancelled.
+	// The cause probably is the plugin binary is not available.
 	if err != nil && ctx.Err() == nil {
 		s.logger.Error("failed to execute stage", zap.String("stage-name", ps.Name), zap.Error(err))
 		return model.StageStatus_STAGE_FAILURE
 	}
 
+	// Handle case where res is nil.
```
Warashi: We should handle the case `err != nil && ctx.Err() == nil`, which is not handled in the current implementation, instead of the case `res == nil` because this is gRPC invocation, its response will nil only when its error is not nil. WDYT?

khanhtc1202: Hi @Warashi 👋 
Just want to confirm, the case `err != nil && ctx.Err() == nil` is handled by L609. Or do you mean a different case 👀

khanhtc1202: I think the missing case is `err != nil && ctx.Err() != nil`, isn't it?

Warashi: @khanhtc1202 Hi 👋
oops, it's right. I did a typo.

khanhtc1202: @Warashi thanks 👍 I address your point by https://github.com/pipe-cd/pipecd/pull/6275/commits/95c78a441067a2556e8a1e68cc3c8a2426b3d17f

Issue #6260, opened 2025-09-22
Error Unauthenticated in login page.


hongky-1994:
**What happened**:
When user in login page, there is an error message show Unauthenticated. 

<img width="1502" height="904" alt="Image" src="https://github.com/user-attachments/assets/4d9284f1-9305-4695-95a9-0267c492d27b" />

**What you expected to happen**:
- Should not have this message show because user is not logged in

**How to reproduce it**:
- Go to https://demo.pipecd.dev/login, 
- Logout if needed.

**Environment**:
- `piped` version: 
- `control-plane` version: v0.54.0-20-gf5b5d81
- Others:

PR #6281, opened 2025-10-02
fix prevent login path can access when already logged in

hongky-1994:
**What this PR does**:

- Trigger getMe Api in login page and prevent throw error message Unauthorized in login page.

**Why we need it**:

- when logged in , user still can navigate to login page by type /login in url address. Therefore, 
   - api getMe should be call in /login page to check token valid and redirect to "/"
   - should not show error message when api req failed.

Partof #6260
Update PR #6261 


khanhtc1202 (approved):
LGTM 👍

PR #6292, opened 2025-10-06
Update event comment reflect change on dev environment

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

I'm going to migrate the dev environment control plane to another cluster and use another piped (v1) to manage it. The internal change is already WIP, and these event files here will only change pipecd.dev website.

Fixes #


nghialv (approved):
🐬

ffjlabo (approved):
👍

PR #6305, opened 2025-10-10
 Fix pipedv1 binary showing wrong version

### LINKED ISSUE #6297: Ensure version for pipedv1 is set correctly
**What happened**:

The built pipedv1 binary is currently showing the wrong version. This affects not only the piped binary version command but also the version value shown in the settings/piped list page.

Should show the pipedv1 version correctly on both the binary version command output and the web UI.

The current version of pipedv1 binary is being set in this release action
https://github.com/pipe-cd/pipecd/actions/runs/18114585135/workflow#L31

**How to reproduce it**:

Download the pipedv1 binary and check for the version.

```bash
$ curl -Lo pipedv1 https://... <- down load binary from release
$ ./pipedv1 version
Version: v0.54.0-32-ge4f1a5d, GitCommit: e4f1a5da6124dd14ed702d381c8f7f6b5735b0dc, BuildDate: 20250930-002615
```

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

rishabh998186:
**Problem:** Built pipedv1 binary shows `v0.54.0-32-ge4f1a5d` instead of correct pipedv1 version.

**Root Cause:** Makefile uses `--match 'v[0-9]*.*'` which only matches piped tags, not pipedv1 tags.

**Solution:**
- Updated workflow to explicitly pass `BUILD_VERSION`

fixes #6297


Makefile
```diff
@@ -21,7 +21,12 @@
 build: build/go build/web
 
 .PHONY: build/go
-build/go: BUILD_VERSION ?= $(shell git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*')
+build/go: BUILD_VERSION ?= $(shell \
+	if [ "$(MOD)" = "pipedv1" ]; then \
+		git describe --tags --always --dirty --abbrev=7 --match 'pipedv1*'; \
+	else \
+		git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*'; \
+	fi)
```
khanhtc1202: I think the version string for pipedv1 also follows semantic version, which means its value will be like `v1.0.0-rc1`, so `v[0-9]*.*` should match that as well. 
ref: https://github.com/pipe-cd/pipecd/actions/runs/18369985533/workflow#L7

khanhtc1202: For example, this is what I tested on my local with the make command

```bash
$ make build/go MOD=pipedv1 BUILD_VERSION=v1.0.0-rc6
GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 go build -ldflags "-X github.com/pipe-cd/pipecd/pkg/version.version=v1.0.0-rc6 -X github.com/pipe-cd/pipecd/pkg/version.gitCommit=a9c8387a4a06295d083beaeaa642ea1f37f96fd5 -X github.com/pipe-cd/pipecd/pkg/version.buildDate=20251010-061017 -w" -trimpath -o ./.artifacts/pipedv1 ./cmd/pipedv1
$ ./.artifacts/pipedv1 version
Version: v1.0.0-rc6, GitCommit: a9c8387a4a06295d083beaeaa642ea1f37f96fd5, BuildDate: 20251010-061017
```

rishabh998186: @khanhtc1202  i removed the Makefile conditional logic i removed the Makefile conditional logic  it was unnecessary since pipedv1 uses the same semantic versioning pattern (v1.0.0-rc*) as piped so the tag matching approach wouldn't work anyway.



I tested your approach locally and confirmed it works:

```
$ make build/go MOD=pipedv1 BUILD_VERSION=v1.0.0-rc-test
$ ./.artifacts/pipedv1 version
Version: v1.0.0-rc-test, GitCommit: 17da42bef..., BuildDate: 20251010-063016   

```
Without BUILD_VERSION, it falls back to git describe (showing commit hash or wrong tags). The PR now only modifies the workflow to pass BUILD_VERSION=${{ inputs.version }} explicitly. Since the Makefile uses BUILD_VERSION ?=, the explicit parameter overrides the git describe default without requiring any Makefile changes.

khanhtc1202: Thanks, one the pipedv1 is officially released, we will rebuild the CI pipeline for releasing, so until then, let keep using this actions for pipedv1 release 👍

PR #6307, opened 2025-10-10
Fix plugin version showing unspecified

### LINKED ISSUE #6306: Ensure version of plugins binary showing correctly

Ensure the version tag shown by piped's plugins is correct. Currently, all plugins do not show their version

```bash
$ curl -Lo wait_plugin https://github.com/pipe-cd/pipecd/releases/download/pkg%2Fapp%2Fpipedv1%2Fplugin%2Fwait%2Fv0.1.1/wait_v0.1.1_darwin_arm64
$ ./wait_plugin version
Version: unspecified, GitCommit: unspecified, BuildDate: unspecified
```

The cause is that we do not set the version while building the plugin binary
ref: https://github.com/pipe-cd/pipecd/blob/master/.github/workflows/plugin_release.yaml#L34

Related issue #6305 

**Why is this needed**:

rishabh998186:
Downloaded plugin binaries currently show unspecified version information:

```

$ curl -Lo wait_plugin https://github.com/pipe-cd/pipecd/releases/download/pkg%2Fapp%2Fpipedv1%2Fplugin%2Fwait%2Fv0.1.1/wait_v0.1.1_darwin_arm64
$ ./wait_plugin version
Version: unspecified, GitCommit: unspecified, BuildDate: unspecified

```
This makes it impossible to verify which version of a plugin is deployed.

Root Cause
The plugin_release workflow wasn't passing version information when building plugins, and the Makefile's build/plugin target didn't inject version details via ldflags like the build/go target does.

Changes
1 - Updated Makefile 
 Modified the build/plugin target to inject version/commit/buildDate via ldflags, using the same pattern as build/go:

- Added BUILD_VERSION, BUILD_COMMIT, BUILD_DATE variables

- Updated BUILD_OPTS to include version ldflags

2 - Updated workflow 
Modified .github/workflows/plugin_release.yaml to pass BUILD_VERSION=${{ inputs.version }} to all plugin build commands (all 4 platform targets)

Test 
Verified the fix works locally:

```
$ make build/plugin PLUGINS=wait BUILD_VERSION=v0.1.1-test
$ ./.artifacts/plugins/wait version
Version: v0.1.1-test, GitCommit: 03df4a5a6d7be22a238c88fe3d741bbf3fb27e20, BuildDate: 20251010-074520

```
Also tested with the kubernetes plugin to confirm the fix works generically across all plugins:

```
$ make build/plugin PLUGINS=kubernetes BUILD_VERSION=v0.2.0-test
$ ./.artifacts/plugins/kubernetes version
Version: v0.2.0-test, GitCommit: 03df4a5a6d7be22a238c88fe3d741bbf3fb27e20, BuildDate: 20251010-074748
 
```

fixes #6306


Makefile
```diff
@@ -63,7 +63,11 @@ build/plugin: PLUGINS_BIN_DIR ?= ~/.piped/plugins
 build/plugin: PLUGINS_SRC_DIR ?= ./pkg/app/pipedv1/plugin
 build/plugin: PLUGINS_OUT_DIR ?= ${PWD}/.artifacts/plugins
 build/plugin: PLUGINS ?= $(shell find $(PLUGINS_SRC_DIR) -mindepth 1 -maxdepth 1 -type d | while read -r dir; do basename "$$dir"; done | paste -sd, -) # comma separated list of plugins. eg: PLUGINS=kubernetes,ecs,lambda
-build/plugin: BUILD_OPTS ?= -ldflags "-s -w" -trimpath
+build/plugin: BUILD_VERSION ?= $(shell git describe --tags --always --dirty --abbrev=7)
```
khanhtc1202:
```suggestion
build/plugin: BUILD_VERSION ?= $(shell git describe --tags --always --dirty --abbrev=7 --match 'v[0-9]*.*')
```
should be like this to ensure the value is a semantic version string

rishabh998186: @khanhtc1202  as you say i Added `--match 'v[0-9]*.*'` to ensure semantic version matching. Tested locally and it works correctly.

PR #6308, opened 2025-10-10
fix: piped plugins display correct version information

GuruOrGoru:
## What this PR does

This PR fixes an issue where piped plugins were showing "unspecified" for version, git commit, and build date information when running the `version` command.

## Problem

Currently, when downloading and running a plugin binary, the version command shows:

```bash
$ curl -Lo wait_plugin https://github.com/pipe-cd/pipecd/releases/download/pkg%2Fapp%2Fpipedv1%2Fplugin%2Fwait%2Fv0.1.1/wait_v0.1.1_darwin_arm64
$ ./wait_plugin version
Version: unspecified, GitCommit: unspecified, BuildDate: unspecified
```

This happened because the plugin build process in the Makefile was not setting the necessary ldflags to inject version information during compilation.

## Solution

Modified the build/plugin target in the Makefile to include version ldflags similar to how the main build/go target works. The changes add:

-> ```BUILD_VERSION```, ```BUILD_COMMIT```, ```BUILD_DATE``` variables

-> ```BUILD_LDFLAGS_PREFIX``` for the version package

-> Updated ```BUILD_OPTS``` to include ldflags that inject version information

## After the fix

Now plugin binaries will correctly display version information:

```bash
$ ./wait_plugin version
Version: v0.1.1, GitCommit: abc123def456, BuildDate: 20231010-123456
```
## Testing

-> Built the wait plugin locally and verified the version command now shows proper information.

-> The fix maintains consistency with how main PipeCD binaries are built.

## Related Issues

Fixes [#6306](https://github.com/pipe-cd/pipecd/issues/6306)

## Changes Made

* Makefile: Added version ```ldflags``` to the ```build/plugin``` target to inject version, git commit, and build date information during plugin compilation.

## Note: Existing plugin binaries remain unaffected; only new builds include version info.

### DIFF
--- a/.github/workflows/plugin_release.yaml
+++ b/.github/workflows/plugin_release.yaml
@@ -31,10 +31,10 @@ jobs:
         run: echo "PLUGIN_NAME=$(basename ${{ inputs.path }})" >> $GITHUB_ENV
       - name: Build binary artifacts
         run: |
-          make build/plugin PLUGINS=$PLUGIN_NAME BUILD_OS=linux BUILD_ARCH=amd64 BIN_SUFFIX=_${{ inputs.version }}_linux_amd64
-          make build/plugin PLUGINS=$PLUGIN_NAME BUILD_OS=linux BUILD_ARCH=arm64 BIN_SUFFIX=_${{ inputs.version }}_linux_arm64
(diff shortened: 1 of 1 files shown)


GuruOrGoru:
I noticed this PR is almost identical to #6307. I’m happy to close mine if you prefer, or collaborate on a single PR.

khanhtc1202:
Thanks @GuruOrGoru , let's keep #6307 since it came first. Thank you!

PR #6320, opened 2025-10-16
K8s plugin health status implementation

khanhtc1202:
**What this PR does**:

Implement health status check for resources:
- DaemonSet
- Pod

**Why we need it**:

Part of https://github.com/pipe-cd/pipecd/issues/5764
Follow https://github.com/pipe-cd/pipecd/pull/6303


Warashi (approved):
LGTM

reference implementations are https://github.com/pipe-cd/pipecd/blob/9d15e3556e800f07ca2807ab9bac6bcdf7ce1206/pkg/app/piped/platformprovider/kubernetes/state.go#L248 and https://github.com/pipe-cd/pipecd/blob/9d15e3556e800f07ca2807ab9bac6bcdf7ce1206/pkg/app/piped/platformprovider/kubernetes/state.go#L378

PR #6331, opened 2025-10-28
Support ECS capacity provider strategy

### LINKED ISSUE #6330: Support CapacityProviderStrategy for ECS workload

SSIA

**Why is this needed**:

Current piped only supports ECS workload of launchType (EC2 or FARGATE), while users want to use CapacityProviderStrategy to leverage the newly arrived AWS Managed Instance (ref: https://aws.amazon.com/ecs/managed-instances/)

khanhtc1202:
**What this PR does**:

SSIA

**Why we need it**:

Users want to leverage ECS Managed Instance while creating ECS services. For now we only support LaunchType fargate/ec2 pattern.

Fixes #6330

### DIFF
--- a/pkg/app/piped/platformprovider/ecs/client.go
+++ b/pkg/app/piped/platformprovider/ecs/client.go
@@ -87,6 +87,9 @@ func (c *client) CreateService(ctx context.Context, service types.Service) (*typ
 	if service.DeploymentController == nil || service.DeploymentController.Type != types.DeploymentControllerTypeExternal {
 		return nil, fmt.Errorf("failed to create ECS service %s: deployment controller of type EXTERNAL is required", *service.ServiceName)
 	}
+	if service.LaunchType != "" && service.CapacityProviderStrategy != nil {
+		return nil, fmt.Errorf("failed to create ECS service %s: launch type and capacity provider strategy cannot be specified together", *service.ServiceName)
+	}
 	input := &ecs.CreateServiceInput{
 		Cluster:                       service.ClusterArn,
 		ServiceName:                   service.ServiceName,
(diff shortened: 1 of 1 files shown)


pkg/app/piped/platformprovider/ecs/client.go
```diff
@@ -87,6 +87,9 @@ func (c *client) CreateService(ctx context.Context, service types.Service) (*typ
 	if service.DeploymentController == nil || service.DeploymentController.Type != types.DeploymentControllerTypeExternal {
 		return nil, fmt.Errorf("failed to create ECS service %s: deployment controller of type EXTERNAL is required", *service.ServiceName)
 	}
+	if service.LaunchType != "" && service.CapacityProviderStrategy != nil {
```
t-kikuc, line 90: these kinds of validation should be executed by AWS API, not on PipeCD side. the condition might be changed in the future

khanhtc1202: I understand your point, but I want to keep this logic to make the CreateService the same as UpdateService ([L148](https://github.com/pipe-cd/pipecd/pull/6331/files/5bca56257841fc6c0f15e6d9d713899e36c1c155#diff-d72df5132a62e1bc9918e8313b516291c54b87a84acd3686bf5233adf4775ccR148)).

As mentioned in the code comment (L121), we need to know the LaunchType for CreateTaskSetInput (if it presents), but we cannot get that value because the DeploymentController is of type EXTERNAL. This validation ensures we don't run into L125 - assigning LaunchType to the output service object. WDYT?

t-kikuc: `DeploymentController=EXTERNAL` is checked because the whole ECS code is based on it.

This modification is caused the AWS update. Such updates should be modified automatically by just updating the SDK version for maintainability.

If you can afford to maintain them, I will approve this PR.

khanhtc1202: > If you can afford to maintain them, I will approve this PR.

@t-kikuc  Please 👍 

To be noted, I have some idea about how to fetch the service fields value that is not available as returned from Create or Update service of type deployment controller EXTERNAL, but it's out of context for this PR. I will do it later by separate PRs. Thank you!

t-kikuc (approved):
try

PR #6336, opened 2025-11-06
Support ecs quicksync force new deployment config

khanhtc1202:
**What this PR does**:

Add ECS QuickSync forceNewDeployment config. The configuration would be like:

```yaml
kind: ECSApp
spec:
  quickSync:
    forceNewDeployment: true
```

**Why we need it**:

It requires forcing a new deployment when updating the ECS service configuration from LaunchType to Capacity provider strategy. So, my plan is to support switching from launch type to capacity provider strategy on existing service by enabling force new deployment on the updated service in case of ECS quick sync deployment triggered.

Error message
> InvalidParameterException: When switching from launch type to capacity provider strategy on an existing service, or making a change to a capacity provider strategy on a service that is already using one, you must force a new deployment.

Follow PR #6331 

### DIFF
--- a/pkg/app/piped/executor/ecs/deploy.go
+++ b/pkg/app/piped/executor/ecs/deploy.go
@@ -113,7 +113,7 @@ func (e *deployExecutor) ensureSync(ctx context.Context) model.StageStatus {
 	forceNewDeployment := e.appCfg.QuickSync.ForceNewDeployment
 
 	// Store force new deployment flag to metadata store.
-	e.Input.MetadataStore.Shared().Put(ctx, "force-new-deployment", strconv.FormatBool(forceNewDeployment))
+	e.Input.MetadataStore.Shared().Put(ctx, forceNewDeploymentKey, strconv.FormatBool(forceNewDeployment))
 
 	if !sync(ctx, &e.Input, e.platformProviderName, e.platformProviderCfg, recreate, forceNewDeployment, taskDefinition, servicedefinition, primary) {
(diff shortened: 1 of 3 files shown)


pkg/app/piped/executor/ecs/deploy.go
```diff
@@ -109,7 +110,12 @@ func (e *deployExecutor) ensureSync(ctx context.Context) model.StageStatus {
 	}
 
 	recreate := e.appCfg.QuickSync.Recreate
-	if !sync(ctx, &e.Input, e.platformProviderName, e.platformProviderCfg, recreate, taskDefinition, servicedefinition, primary) {
+	forceNewDeployment := e.appCfg.QuickSync.ForceNewDeployment
+
+	// Store force new deployment flag to metadata store.
+	e.Input.MetadataStore.Shared().Put(ctx, "force-new-deployment", strconv.FormatBool(forceNewDeployment))
```
t-kikuc: nit: use a const for the key

t-kikuc (dismissed):
LGTM, let's try

pkg/app/piped/executor/ecs/deploy.go
khanhtc1202: @t-kikuc Thanks, fixed by https://github.com/pipe-cd/pipecd/pull/6336/commits/025460c1bd14fc362094d0c61c602ec944d1a555. Please re-approve 🙏

PR #6363, opened 2025-11-26
feat: Implement project enabling/disabling functionality

### LINKED ISSUE #4754: Add ability to disable project
Add ability to disable projects on WebUI and so on.

**Why is this needed**:
Currently, we can't do a soft delete of projects. We need to manage projects when they leave.

niladrix719:
**What this PR does**:

- Adds Disable option in project settings (no new logins or API calls are allowed until it is re-enabled)
- Adds a Disable and Enable Field in Owner page

Screenshot 2025-11-26 at 9 41 28 PM

Screenshot 2025-11-26 at 6 56 37 PM

Screenshot 2025-11-26 at 7 59 50 PM

Screenshot 2025-11-27 at 2 00 19 AM

**Why we need it**:

Fixes #4754


niladrix719:
I think we can still merge this PR, since it will trigger publish_tool.yaml to publish a new version of the codegen image. We can then update the Makefile and gen.yaml image in a later PR. What do you think, @khanhtc1202?

Or, we could update the image and use the new one in this PR, but I don’t think I have write access to push a new image version.

khanhtc1202:
@niladrix719 Since the codegen check is a required step in the CI, I think it's better to fix the broken CI first 👀

niladrix719:
okay can you help me publish the new image, so I can use it here

khanhtc1202:
@niladrix719 sure, sorry for the missed 🙏

pkg/jwt/jwt.go
```diff
@@ -39,8 +39,12 @@ type Claims struct {
 }
 
 // NewClaims creates a new claims for a given github user.
-func NewClaims(githubUserID, avatarURL string, ttl time.Duration, role model.Role) *Claims {
+func NewClaims(githubUserID, avatarURL string, ttl time.Duration, role *model.Role) *Claims {
```
khanhtc1202: Why do we need to change this to pointer?

tool/codegen/protoc-gen-auth/file.go
```diff
@@ -98,6 +99,13 @@ func (a *authorizer) getAllProjectRBACRoles(ctx context.Context, projectID strin
 		return nil, err
 	}
 
+	if p.Disabled {
+		a.logger.Warn("project is disabled",
+			zap.String("project", projectID),
+		)
+		return nil, fmt.Errorf("project %s is disabled", projectID)
+	}
```
khanhtc1202: We can remove this file change and do it by another PR to resolve the codegen diff 👀

niladrix719: sure

pkg/jwt/jwt.go
niladrix719: just thought it would be better, should I revert this

khanhtc1202: Then it should be a separated PR for that, to reduce the change we have in each PR 🙏

niladrix719:
Hey @khanhtc1202 can you check now

khanhtc1202:
We may reboot this later when we found proper resource for it

PR #6368, opened 2025-11-28
docs: clarify that HelmChart version field is only for chart repository

### LINKED ISSUE #6366: Documentation: Clarify that version field in HelmChart is ignored when using gitRemote
### What would you like to be added
The HelmChart configuration reference should clarify that the version field is only applicable when using a Helm chart repository (repository field), and is ignored when using a git remote chart (gitRemote field).
Current documentation
https://pipecd.dev/docs-v0.55.x/user-guide/configuration-reference/#helmchart
Currently, the HelmChart section describes the fields as follows:

Field | Type | Description | Required
-- | -- | -- | --
gitRemote | string | Git remote address where the chart is placing. Empty means the same repository. | No
ref | string | The commit SHA or tag value. Only valid when gitRemote is not empty. | No
path | string | Relative path from the repository root to the chart directory. | No
repository | string | The name of a registered Helm Chart Repository. | No
name | string | The chart name. | No
version | string | The chart version. | No

The ref field has a note saying "Only valid when gitRemote is not empty", but the version field lacks a similar clarification.
Why this is confusing
When using a git remote chart (with gitRemote), users specify the version via the ref field (e.g., a tag like v0.5.0). The version field in this case is not used. However, the current documentation does not make this clear, which can lead to confusion.
For example, looking at the Kubernetes app configuration examples:

```yaml
# Remote git chart - uses `ref` for version
helmChart:
  gitRemote: git@github.com:pipe-cd/manifests.git
  ref: v0.5.0
  path: manifests/helloworld

# Remote chart repository - uses `version`
helmChart:
  repository: pipecd
  name: helloworld
  version: v0.5.0
```
### Suggested change
Update the version field description to clarify when it is applicable:

Field | Type | Description | Required
-- | -- | -- | --
version | string | The chart version. Only valid when using a chart repository (repository field). | No

tani-yu:
**What this PR does**:
Adds clarification to the `version` field description in HelmChart configuration reference, noting that it is only valid when using a chart repository.

**Why we need it**:
Currently, the `ref` field has a note saying "Only valid when gitRemote is not empty", but the `version` field lacks a similar clarification. This can confuse users about which field to use for specifying chart versions in different scenarios (git remote vs chart repository).

Fixes #6366

- **How are users affected by this change**: Documentation only, no functional change

**Note**
I only updated docs-dev. Let me know if other versioned docs (e.g., docs-v0.55.x) should also be updated.

### DIFF
--- a/docs/content/en/docs-v0.55.x/user-guide/configuration-reference.md
+++ b/docs/content/en/docs-v0.55.x/user-guide/configuration-reference.md
@@ -329,7 +329,7 @@ One of `yamlField` or `regex` is required.
 | path | string | Relative path from the repository root to the chart directory. | No |
 | repository | string | The name of a registered Helm Chart Repository. | No |
 | name | string | The chart name. | No |
-| version | string | The chart version. | No |
+| version | string | The chart version. Only valid when using a chart repository (repository field). | No |
 
 ### HelmOptions
 


khanhtc1202:
Please also update this v0.55.x docs (same change in docs-dev)
https://github.com/pipe-cd/pipecd/blob/master/docs/content/en/docs-v0.55.x/user-guide/configuration-reference.md

tani-yu:
Done, updated v0.55.x as well.

khanhtc1202 (approved):
LGTM

PR #6386, opened 2025-12-23
update CONTRIBUTING.md

eeshaanSA:
**What this PR does**:

Adds info on TLS conecction in local dev

**Why we need it**:

Fixes #

### DIFF
--- a/docs/content/en/docs-v1.0.x/user-guide/managing-application/manifest-attachment.md
+++ b/docs/content/en/docs-v1.0.x/user-guide/managing-application/manifest-attachment.md
@@ -1,0 +1,63 @@
+---
+title: "Manifest attachment"
+linkTitle: "Manifest attachment"
+weight: 10
+description: >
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -261,6 +261,8 @@ CONFIG_FILE=path/to/piped-config.yaml \
 INSECURE=true
 ```
 
+where the `CONFIG_FILE` is the path to your piped confiuration file and the `INSECURE` flag is set to `true` to allow `piped` to connect to the control plane without TLS verification.
```
khanhtc1202, line 264: The part about `CONFIG_FILE` is duplicated with the line below, isn't it?

eeshaanSA: wdym? I think this info is good to be added.

eeshaanSA:
@khanhtc1202, what are we doing with this?

khanhtc1202 (approved):

PR #6403, opened 2026-01-05
Refactor Docker image build workflows to use native ARM/AMD runners instead of QEMU

Okabe-Junya:
**What this PR does**:

Refactor Docker image build workflows to use native ARM/AMD64 runners in parallel instead of QEMU emulation

- Split single QEMU-based build jobs into `build` → `merge` pipeline (for publish workflows)
- Use `ubuntu-24.04` for AMD64 and `ubuntu-24.04-arm` for ARM64 native builds
- Build images by digest, then merge into multi-arch manifest using `docker buildx imagetools create`
- Remove `docker/setup-qemu-action` dependency

**Why we need it**:

QEMU-based ARM64 emulation on x86 runners is significantly slower than native builds. By using GHA ARM runners, we can build both architectures in parallel with native performance, reducing overall CI time while producing identical multi-arch images.

Fixes #

### DIFF
--- a/.github/workflows/build_tool.yaml
+++ b/.github/workflows/build_tool.yaml
@@ -12,11 +12,16 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
-  tool:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 15
+  # Build images for each platform in parallel using native runners
+  build:
     strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


Okabe-Junya:
FYI: Tested in https://github.com/pipe-cd/pipecd/pull/6428, and looks working well

PR #6421, opened 2026-01-10
Auto-assign PR author on pull request creation

MUFFANUJ:
**What this PR does**:
Added .github/workflows/auto-assign-pr.yaml workflow that automatically assigns the PR author when a pull request is opened. 

**Why we need it**:
Improves PR tracking by ensuring the author is assigned from the start, making it easier to identify PR owners and manage reviews.

Fixes #

**Does this PR introduce a user-facing change?**: None
- **Is this breaking change**: None
- **How to migrate (if breaking change)**: None

### DIFF
--- a/.github/workflows/auto-assign-pr.yaml
+++ b/.github/workflows/auto-assign-pr.yaml
@@ -1,0 +1,26 @@
+name: Auto-assign PR author
+
+on:
+  pull_request:
+    types: [opened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  assign:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Assign PR author
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const prAuthor = context.payload.pull_request.user.login;
+            const prNumber = context.payload.pull_request.number;
+
+            await github.rest.issues.addAssignees({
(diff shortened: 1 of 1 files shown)


Okabe-Junya:
(I'm not a official reviewer but...)

I have no preference about this, but GitHub PRs can be queried by Author.  e.g., https://github.com/pipe-cd/pipecd/pulls/MUFFANUJ

So I don't think traceability improves significantly; it just makes https://github.com/pipe-cd/pipecd/pulls slightly easier to read.

MUFFANUJ:
Hi @Okabe-Junya, you’re right that author filtering shows who opened the PR. The difference is that assignees represent ownership, which many maintainer tools and workflows may rely on for triage, review tracking, and accountability.
Auto-assigning ensures every PR has a clear owner from creation without any manual step. It’s a small UX improvement, not a workflow requirement.

Okabe-Junya:
If you think the author should have ownership, wouldn't an author filter suffice? 
If that isn't always correct, I think different ownership would be assigned automatically (and this just confuses others

Automation like Prow in Kubernetes community - chatops command to assign reviewer/approver might be necessary. But in PipeCD org, automatic reviewer assignment w/ GitHub teams is enabled, so I'm not sure if this is even needed.

PR #6427, opened 2026-01-11
Sync Description 

### LINKED ISSUE #3426: Enable to fetch application description information on adding app

Tushar240503:
his PR fixes a bug where the application description from app.pipecd.yaml was not being reflected in the control plane immediately after adding a new application. It modifies the appconfigreporter component to bypass the commit-cache optimization when the application's description is empty. This forces an immediate re-scan and sync of the configuration, ensuring the description is populated without requiring a subsequent git commit.

Why we need it:

Currently, when an application is added (especially via suggestion or without manual description input), the Description field in the Control Plane remains empty. The Piped agent skips syncing because the git commit hash hasn't changed relative to the previous state (or lack thereof). Users see an empty description in the UI until they push a new commit, which is confusing and inconsistent.

Which issue(s) this PR fixes:

Fixes #3426

Does this PR introduce a user-facing change?:

How are users affected by this change: The application description will now appear automatically in the Web UI shortly after adding the application, matching the value in app.pipecd.yaml.
Is this breaking change: No
How to migrate (if breaking change): N/A

### DIFF
--- a/.github/workflows/build_tool.yaml
+++ b/.github/workflows/build_tool.yaml
@@ -12,11 +12,16 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
-  tool:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 15
+  # Build images for each platform in parallel using native runners
+  build:
     strategy:
+      fail-fast: false
       matrix:
+        platform:
+          - runner: ubuntu-24.04
+            arch: amd64
+          - runner: ubuntu-24.04-arm
+            arch: arm64
         image:
           - actions-gh-release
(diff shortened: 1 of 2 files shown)


eeshaanSA:
cc/ @khanhtc1202 @t-kikuc

PR #4745, opened 2024-01-01
Update copyright

khanhtc1202:
**What this PR does / why we need it**:

We can use `make update/copyright` to update all our copyrights in files.
Note: File `service.pb.auth.go`, which is generated by our protoc-gen-auth needs to be regenerated later using the updated `tool/codegen` image published after this PR is merged in order to pass the action/code check (I know it's quite meta 😂), will do it later by separated PR.

Happy New Year 🐲 

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**:
- **Is this breaking change**:
- **How to migrate (if breaking change)**:

PR #6451, opened 2026-01-22
fix: update-copyright script for linux compatibility

Parship12:
**What this PR does**:
Fixed the `update-copyright.sh` to work correctly: 
- Removing `-Z` from grep (using newline-separated output)  
- Adding cross-platform for `sed -i` (mac vs Linux)

**Why we need it**: The script `grep -rlZ` which causing all filenames to concatenate into one string. Also, the script used macOS `sed -i ''` which is not working on linux.

Before:
Screenshot 2026-01-25 152804

After:
Screenshot 2026-01-25 153250

- **How are users affected by this change**: NA
- **Is this breaking change**: NO
- **How to migrate (if breaking change)**: NA

### DIFF
--- a/hack/update-copyright.sh
+++ b/hack/update-copyright.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2026 The PipeCD Authors.
+# Copyright 2024 The PipeCD Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


ghost:
@eeshaanSA Yeah sorry for that, actually I have done 2 things in the PR:
1. Update the script file `update-copyright.sh` -> This was my main aim
2. Ran the script: And after running the script all the files are the just updated by 2025 -> 2026.

Okay, I will create a different PR for the changes done in the script.

Ayushmore1214:
@eeshaanSA @Parship12 quick question , doesn't the copyright year usually represent the creation year?

ghost:
> @eeshaanSA @Parship12 quick question , doesn't the copyright year usually represent the creation year?

I thought the same, but then I found this PR: #4745

eeshaanSA:
> @eeshaanSA @Parship12 quick question , doesn't the copyright year usually represent the creation year?

New code added in 2025 and 2026 has to reflect the changes. That is why, it is recommended, to have the current year, or the range, for example 2021-2026.

eeshaanSA:
> > @eeshaanSA @Parship12 quick question , doesn't the copyright year usually represent the creation year?
> 
> 
> 
> I thought the same, but then I found this PR: #4745 

So, the motive of this PR is that the .hack/update-copyright.sh file fails?

ghost:
> **Why we need it**: The `update-copyright.sh` script was failing on linux with errors like "sed: can't read". The script `grep -rlZ` which causing all filenames to concatenate into one string. Also, the script used macOS `sed -i ''` which is not working on linux.

Yes, it was failing on Linux and also the file names were concatenate.

eeshaanSA:
Understood, thanks! 

> > **Why we need it**: The `update-copyright.sh` script was failing on linux with errors like "sed: can't read". The script `grep -rlZ` which causing all filenames to concatenate into one string. Also, the script used macOS `sed -i ''` which is not working on linux.
> 
> Yes, it was failing on Linux and also the file names were concatenate.

ghost:
@eeshaanSA Sure, I will do it by EOD. Do I need to open an issue and link it?

ghost:
@eeshaanSA @khanhtc1202 PTAL

PR #6465, opened 2026-01-27
Doc right section fix

### LINKED ISSUE #6464: The right nav should highlighting along with scrolling of the content
It would be be better if the sections highlight along with scrolling of the content.

Image

antedotee:
**What this PR does**:
This PR adds the functionality of highlighting of the right section along with scrolling of the main content.

**Why we need it**:
To enhance the user experience

Fixes #6464

### DIFF
--- a/docs/assets/js/toc-scroll-spy.js
+++ b/docs/assets/js/toc-scroll-spy.js
@@ -60,15 +60,11 @@
       // Remove active class from all links
       sections.forEach(function (section) {
         section.link.classList.remove('active');
-        section.link.style.fontWeight = '';
-        section.link.style.color = '';
       });
 
       // Add active class to current link
       if (activeSection) {
         activeSection.link.classList.add('active');
-        activeSection.link.style.fontWeight = 'bold';
-        activeSection.link.style.color = '#007bff'; // Bootstrap primary color
       }
     }
 
--- a/docs/assets/scss/_styles_project.scss
+++ b/docs/assets/scss/_styles_project.scss
@@ -78,13 +78,14 @@
 
 .td-sidebar-toc {
     a {
-        transition: all 0.2s ease-in-out;
+        transition: all 0.3s ease-in-out;
+        border-left: 3px solid transparent;
+        padding-left: 8px;
         
         &.active {
             font-weight: 700 !important;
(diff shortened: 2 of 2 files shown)


antedotee:
@eeshaanSA Please take a look.

docs/assets/js/toc-scroll-spy.js
```diff
@@ -0,0 +1,97 @@
...
+        section.link.classList.remove('active');
+        section.link.style.fontWeight = '';
+        section.link.style.color = '';
+      });
+
+      // Add active class to current link
+      if (activeSection) {
+        activeSection.link.classList.add('active');
+        activeSection.link.style.fontWeight = 'bold';
+        activeSection.link.style.color = '#007bff'; // Bootstrap primary color
```
rahulshendre: @antedotee, wouldn't it be better to let JS only toggle the active class and keep all styling in SCSS to avoid duplication and overrides, what do you think?

rahulshendre:
https://github.com/user-attachments/assets/09a98914-8206-4417-bc68-57e082459dc9

@eeshaanSA As mentioned in the previous meeting, you suggested that anyone with JS experience to take a look at this. I have reviewed and tested it locally.

Overall the behavior looks good. A couple of suggestions:
1. The transition speed feels a bit fast, increasing it to 0.3s for a smoother look would be better (currently 0.2s).
2. There's a minor UX issue (pointed out in the video as well), when text becomes "active", the width increases by 11px (3px border + 8px padding) causing text to wrap and create a visual disturbance. @antedotee you can consider reserving this space in the inactive state to prevent layout shift.

Please address the review comment about removing inline styles from JS. Once those are fixed, this LGTM.
Thanks

antedotee:
@rahulshendre I have implemented the suggested changes. Please have a look and let me know if there are any further improvements to be worked upon.

rahulshendre:
@antedotee I pulled the latest changes and tested them locally.

The current behaviour still feels off in practice.
From a JS perspective, one thing that stands out is that the scroll handler updates DOM state on every scroll event, even when the active section hasn’t changed, which can lead to unnecessary DOM manipulation during scrolling.

PR #6480, opened 2026-01-29
Fix: Hide version selector on Home page and make v1 default

### LINKED ISSUE #6478: Version selector has inconsistent version on homepage
**What happened**:

The version selector on the homepage shows version (**v1.0.x**). Happening because the template hardcodes a specific index from the versions list. When navigating to the Documentation section, the site redirects to the latest docs version (currently, v0.55.x), and the selector updates correctly there. The homepage and documentation pages display different versions, causing inconsistency. 

Two possible approaches:
1. Show the latest documentation version on the homepage instead of using a hardcoded index.
2. Hide the version selector entirely when on home page.

The version selector should be consistent with the actual documentation version, either showing the latest version on the homepage or not appearing there at all.

**How to reproduce it**:

1. Open the homepage.
2. Observe the version selector, it shows **v1.0.x**.
3. Click on Documentation.
4. The selector changes to the latest version.

**Environment**:
- Docs site [pipecd.dev](https://pipecd.dev/)

rahulshendre:
**What this PR does**:

- Hide version selector on the homepage.
- Set docs default to latest, `docs-v1.0`

**Why we need it**:
Homepage was showing a different version than the docs due to hardcoded selector and as v1 is going to be the latest version, we need it to be the default.

Fixes #6478

### DIFF
--- a/docs/content/en/docs-dev/faq/_index.md
+++ b/docs/content/en/docs-dev/faq/_index.md
@@ -60,6 +60,8 @@ You can create a new Piped key. Go to the `Piped` tab at `Settings` page, and cl
 
 Yes, PipeCD is fully open source project with APACHE LICENSE, VERSION 2.0!!
 
+From May 2023, PipeCD joined CNCF as a [Sandbox project](https://www.cncf.io/projects/pipecd/).
+
 ### 9. How should I investigate high CPU usage or memory usage in piped, or when OOM occurs?
 
 If you're noticing high CPU usage, memory usage, or facing OOM issues in Piped, you can use the built-in support for `pprof`, a tool for visualization and analysis of profiling data.  
--- a/docs/content/en/docs-v0.55.x/faq/_index.md
+++ b/docs/content/en/docs-v0.55.x/faq/_index.md
@@ -60,6 +60,8 @@ You can create a new Piped key. Go to the `Piped` tab at `Settings` page, and cl
 
 Yes, PipeCD is fully open source project with APACHE LICENSE, VERSION 2.0!!
 
+From May 2023, PipeCD joined CNCF as a [Sandbox project](https://www.cncf.io/projects/pipecd/).
+
 ### 9. How should I investigate high CPU usage or memory usage in piped, or when OOM occurs?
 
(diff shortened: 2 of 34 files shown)


rahulshendre:
@eeshaanSA, PTAL when you have time : )

docs/main.go
```diff
@@ -28,7 +28,7 @@ import (
 const dir = "/public"
 
 // Don't update here manually. /hack/gen-release-docs.sh does.
```
eeshaanSA, line 30: Pretty self explanatory, isnt it? 👀

rahulshendre: I did see the comment, but noticed that latestPath was manually updated in the v0.55.0 release (commit bc3d5e0 in docs/main.go #6298 by @khanhtc1202 ) when changing from v0.54.x to v0.55.x.

Since v1.0.x already exists in config.toml, the script exits early and doesn't update latestPath. So manual edit seems to be the standard approach for updating the latest version when it already exists.

eeshaanSA: I see, I will have to check on that. But your theory makes sense.

rahulshendre:
@eeshaanSA, there is one more thing I noticed. On smaller screens (phones), the version selector is not visible, so users currently have no way to change the version.

Earlier users landed on v0.55, but after this PR they will land on v1 docs by default. However, on mobile they still won’t be able to switch versions since the selector is hidden.

After we get this PR is merged, I’ll open a separate issue for this.

docs/main.go
rahulshendre: Great, thanks

eeshaanSA (changes requested):
@rahulshendre, while workflows are being run, can you provide a detailed screen recording of the change? Sorry for the trouble, but we do not have a live preview.

rahulshendre:
> @rahulshendre, while workflows are being run, can you provide a detailed screen recording of the change? Sorry for the trouble, but we do not have a live preview.

Sure @eeshaanSA
Here you go. Screen recording below.

https://github.com/user-attachments/assets/e6d25b57-5473-42e8-add0-ccaa460a2dfc

eeshaanSA (approved):
LGTM

PR #6490, opened 2026-01-31
added template for documentation issue

### LINKED ISSUE #6484: There is no default issue template for documentation
As the docs site gains traction, I believe it will be a good idea to create a documentation template for contributors to open up documentation issues. Also, it will additionally helpful to add a tag like `kind/documentation` whenever someone tries to open up a documentation issue.

**Why is this needed**:
To enhance the contributor experience.

antedotee:
**What this PR does**:
This PR adds feature to specifically submit documentation issues. Additionally, a`kind/documentation` label will be added along with opening the documentation issue.

Fixes #6484

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation.md
+++ b/.github/ISSUE_TEMPLATE/documentation.md
@@ -1,0 +1,11 @@
+---
+name: Documentation Issue
+about: Report an issue with the documentation
+labels: kind/documentation
+
+---
+<!-- Please only use this template for submitting documentation issues -->
+
+**Describe the issue**:
+
+**Relevant URLs**:


antedotee:
@eeshaanSA PTAL.

Ayushmore1214 (commented):
since we are setting this up fresh, I'd suggest using GitHub YAML Issue Forms instead of a Markdown templates

antedotee:
@Ayushmore1214 Yes, I was thinking of initially that only but I didn't used that as rest of the templates doesn't use that. It is upto for Eeshan to decide upon which approach we should go with.

.github/ISSUE_TEMPLATE/documentation.md
eeshaanSA, line 1: We can also add area/docs label. And also, we can do better, I mean, have something more than just description and URL. 👀

eeshaanSA (changes requested):

PR #6497, opened 2026-02-02
added the workflow and script to pin commit hash

### LINKED ISSUE #6492: There should be a commit hash check that ensures nothing uses tag and uses only commit hash
I have noticed most of the workflows uses commit hash while some doesn't. It would be good if we use a commit hash check that ensures everything uses commit hash. Also, ideally, there will be a script and a file containing centralized commit hashes.

antedotee:
**What this PR does**:
This PR introduces a workflow and a script which will ensure every tag is pinned to the commit hash. If anything uses tag and not a commit hash, the workflow will fail and will tell you to run `hack/gha-reversemap.sh apply-reversemap` locally. I have also added `.gha-reversemap.yml` which will act as a single source of truth for approved hashes.

**Why we need it**:
For security purposes, makes the supply chain auditable

Fixes #6492

### DIFF
--- a/.gha-reversemap.yml
+++ b/.gha-reversemap.yml
@@ -1,0 +1,153 @@
+# Approved commit hashes for GitHub Actions used in .github/workflows.
+# Every workflow must reference these actions by this SHA (not by tag).
+# See hack/gha-reversemap.sh for verify/apply/update commands.
+
+actions/cache:
+  sha: 0057852bfaa89a56745cba8c7296529d2fc39830
+  sha-url: https://github.com/actions/cache/commit/0057852bfaa89a56745cba8c7296529d2fc39830
+  tag: v4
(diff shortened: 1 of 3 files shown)


antedotee:
@khanhtc1202 Please take a look. Instead of raising PR solely for the workflow, the workflow was referencing the local script, so I added the local script here for end to end setup. I have run the script locally and it is working perfectly. If this gets merged, I think I can raise next PR for changing all the tags to commit hashes, and another PR for referencing the changes made in the documentation so that contributors can reference it to know what is going on.

Warashi:
I think we should enforce pinning with the native GitHub feature, not implementing the custom solution.
https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

WDYT? @khanhtc1202

Issue #6174, opened 2025-08-26
Implement setting default values in plugins


Warashi:
**What would you like to be added**:

Implement setting default values in the application config when loading in all plugins.

**Why is this needed**:

In #6168, we stopped setting default values when loading application config in the SDK.
So we need to implement setting default values in plugins.

t-kikuc:
### TODO

#### Deployment plugins

- [ ] k8s
    - desc: 
- [ ] k8s multicluster
    - desc: 
- [ ] terraform
    - desc: 
- [ ] cloudrun services
    - desc: 

####  Stage plugins

- [ ] wait
    - desc: 
- [ ] waitApproval
    - desc: 
- [ ] analysis
    - desc: 
- [x] scriptRun
    - desc: it already has `defaults.Set()` 
       https://github.com/pipe-cd/pipecd/blob/45a82f79be938544e374655ff4698d8702b2e2fc/pkg/app/pipedv1/plugin/scriptrun/options.go#L40-L54

#### community-plugins

- [ ] notify to developers

rawadhossain:
Hi @Warashi, I’ve been going through the issue and the existing plugins to understand how defaults were previously handled and what patterns already exist.

This is the approach I'm thinking of:
For stage plugins (`wait, analysis`), I’m applying `defaults.Set()` in their local decode logic, since that’s already how those plugins work.

For deployment plugins (`terraform, k8s, cloudrun`), I’m using `UnmarshalJSON` where configs are unmarshaled directly by SDK, to stay consistent with the existing setup there.

I’m keeping the changes minimal and focused only on default handling, without changing any other behavior.

To keep reviews manageable, I’m planning to submit incremental PRs per plugin or small plugin groups instead of one large PR.
Please let me know if this direction sounds reasonable. Thanks.

rawadhossain:
@t-kikuc @eeshaanSA  can I get this issue assigned?

PR #6513, opened 2026-02-06
Apply defaults in wait and analysis Stage Plugins

rawadhossain:
**What this PR does**:
Applies `defaults.Set()` when decoding stage configurations in the `wait` and `analysis` stage plugins, ensuring default values are initialized correctly during config loading.

**Why we need it**:
- Stage plugins need to apply defaults during config decoding.
- `wait` and `analysis` were unmarshalled without calling `defaults.Set()`.
- Hence config defaults were not being applied.

### **Part of Issue #6174**

No
- **How are users affected by this change**: Stage configurations now correctly apply default values when fields are omitted, as expected behavior.


rawadhossain:
Thanks, all good. Happy to contribute.

PR #6516, opened 2026-02-07
Apply default values in Terraform and CloudRun Deployment plugin

rawadhossain:
**What this PR does**:
Implements setting default values in the `CloudRun` and `Terraform deployment` plugins by adding `UnmarshalJSON` methods that call `defaults.Set()`.

**Why we need it**:
- CloudRun’s `ServiceManifestFile` was documented to default to `service.yaml`, but the default was never applied.
- Terraform’s `DriftDetectionEnabled` had a default `true` tag but the value was never set during unmarshaling.

### **Part of Issue #6174**

  - CloudRun configs that omit `serviceManifestFile` will now correctly default to `service.yaml`
  - Terraform deploy target configs that omit `driftDetectionEnabled` will now correctly default to `true`


rawadhossain:
Hi @Warashi, just checking in since it was marked stale and may auto close soon.
Would appreciate your review or any feedback when you have time. I'm happy to make changes if needed. Thanks.

PR #6520, opened 2026-02-07
added additional flag for base64 encoding under pipectl encrypt

### LINKED ISSUE #6509: Document --use-base64-encoding for pipectl encrypt
**What happened**:
The docs for pipectl encrypt in docs-dev only mention stdin and --input-file, but the CLI supports --use-base64-encoding. Users won’t discover this option from the docs, even though it changes input handling semantics and may be required for certain payloads.

Docs should list --use-base64-encoding as an optional flag for pipectl encrypt, with a short explanation of when to use it.

Evidence:
`pipectl encrypt --help` shows:
```
--use-base64-encoding   Whether the plaintext should be base64 encoded before encrypting or not. (default false)

```

antedotee:
**What this PR does**:
This PR adds a short note that pipectl encrypt supports an optional --use-base64-encoding flag (base64-encodes plaintext before encrypting).

**Why we need it**:
This fixes the docs gap where the flag existed but wasn’t mentioned in the CLI guide, which is why users wouldn’t know to use it.

Fixes #6509

### DIFF
--- a/docs/content/en/docs-dev/user-guide/command-line-tool.md
+++ b/docs/content/en/docs-dev/user-guide/command-line-tool.md
@@ -355,6 +355,9 @@ You can encrypt it the same way you do [from the web](../managing-application/se
       --input-file={PATH_TO_SECRET_FILE}
   ```
 
+Optional encoding:
+- `--use-base64-encoding` to base64-encode the plaintext before encrypting.
+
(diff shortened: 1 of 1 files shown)


antedotee:
CC @eeshaanSA .PTAL.

PR #6551, opened 2026-02-28
fix: honor explicit pipeline config when no workloads exist in Kubernetes app

### LINKED ISSUE #4799: K8s App kind with only ConfigMap can only trigger QUICK_SYNC
**What happened**:

Kubernetes application with only ConfigMap can only trigger QUICK_SYNC even if we set pipeline in `app.pipecd.yaml`.

**How to reproduce it**:

**Environment**:
- `piped` version:
- `control-plane` version:
- Others:

mihir-dixit2k27:
**What this PR does**:

Updates the `decideStrategy` function in the Kubernetes planner to accept a `pipelineDefined` parameter. When a pipeline is explicitly configured in `app.pipecd.yaml`, the planner now respects that configuration and triggers a progressive sync — even when the application contains no workload manifests (e.g., ConfigMap-only apps).

**Why we need it**:

Previously, the planner determined sync strategy purely based on workload detection. If no workloads (Deployments, etc.) were found, it unconditionally fell through to QUICK_SYNC, silently ignoring any pipeline defined by the user in `app.pipecd.yaml`. This made it impossible to use pipeline-based deployments for apps that only manage ConfigMaps or similar non-workload resources.

Fixes #4799

Yes.

- **How are users affected by this change**: Kubernetes applications that contain only ConfigMaps (or other non-workload resources) and have a `pipeline` defined in `app.pipecd.yaml` will now correctly trigger a progressive sync instead of always falling back to QUICK_SYNC.
- **Is this breaking change**: No. Existing apps without a pipeline configured retain identical behavior. All prior `decideStrategy` logic is preserved — the `pipelineDefined` check only activates when the user has explicitly defined pipeline stages.


eeshaanSA:
@khanhtc1202

Issue #6443, opened 2026-01-20
[LFX Mentorship T-1 2026] Implement Amazon ECS Plugin for pipedv1


khanhtc1202:
## Overview

This issue tracks the LFX Mentorship project for Term 1 2026, focused on implementing the Amazon ECS plugin for the PipeCD v1 plugin architecture. This project aims to bring full ECS deployment support to PipeCD v1.

## Background

PipeCD v1 introduced a plugin-based architecture that decouples deployment logic from the core piped agent. While the legacy piped (v0) supports ECS, we need to bring this functionality to v1 as well.

Amazon ECS is a fully managed container orchestration service that powers mission-critical workloads across enterprises. Supporting ECS in pipedv1 will enable teams to leverage PipeCD's unified GitOps experience for their AWS container infrastructure.

---

## Ideal Mentee

This project is highly technical. We are looking for mentees with:

* **Go programming skills**
* **Understanding of AWS ECS** (task definitions, services, target groups)
* **Familiarity with container orchestration** concepts
* **Understanding of CD strategies** (Canary, Blue/Green)
* **Familiarity with gRPC** and plugin architectures
* **Familiarity with open source workflows** and version control

## Mentors

* Khanh Tran [@khanhtc1202](https://github.com/khanhtc1202)
* Shinnosuke Sawada-Dazai [@Warashi](https://github.com/Warashi)

## Timeline

In accordance with the LFX Mentorship Term 1 2026 timeline.

## Other Resources

* **Existing ECS Implementation (v0):** [ECS platform provider](https://github.com/pipe-cd/pipecd/tree/master/pkg/app/piped/platformprovider/ecs)
* **Kubernetes Plugin (reference):** [Kubernetes plugin for pipedv1](https://github.com/pipe-cd/pipecd/tree/master/pkg/app/pipedv1/plugin/kubernetes)
* **Plugin Architecture Blog:** [Plugin architecture Piped alpha version](https://pipecd.dev/blog/2025/06/16/plugin-architecture-piped-alpha-version-has-been-released/)

* **CNCF Slack:** Join the PipeCD Slack Channel [here](https://communityinviter.com/apps/cloud-native/cncf).

**Interested in contributing as an LFX Mentee?** Submit your application to the [LFX Mentorship portal](https://mentorship.lfx.linuxfoundation.org/#projects_all) before the deadline!

armistcxy:
I’d be happy to work on this issue if it’s available

[...]

PR #6559, opened 2026-03-06
[ECS-Plugin] Implement ECS_ SYNC stage

armistcxy:
**What this PR does**:
Implementing `ECS_SYNC` stage. The quick sync and sync option will use this stage by default

**Why we need it**:

**Which issue(s) this PR fixes**: Part of #6443

Fixes #

### DIFF
--- a/pkg/app/pipedv1/plugin/ecs/deployment/sync.go
+++ b/pkg/app/pipedv1/plugin/ecs/deployment/sync.go
@@ -62,21 +62,15 @@ func (p *ECSPlugin) executeECSSyncStage(
 		return sdk.StageStatusSuccess
 	}
 
-	serviceDef, err := provider.LoadServiceDefinition(input.Request.TargetDeploymentSource.ApplicationDirectory, cfg.Spec.Input.ServiceDefinitionFile)
+	serviceDef, err := provider.LoadServiceDefinition(
+		input.Request.TargetDeploymentSource.ApplicationDirectory,
+		cfg.Spec.Input.ServiceDefinitionFile,
+		input,
+	)
 	if err != nil {
 		lp.Errorf("Failed to load service definition: %v", err)
 		return sdk.StageStatusFailure
 	}
-
-	// Add PipeCD-managed tags
-	serviceDef.Tags = append(serviceDef.Tags,
-		provider.MakeTags(map[string]string{
-			provider.LabelManagedBy:   provider.ManagedByECSPlugin,
-			provider.LabelPiped:       input.Request.Deployment.PipedID,
-			provider.LabelApplication: input.Request.Deployment.ApplicationID,
-			provider.LabelCommitHash:  input.Request.TargetDeploymentSource.CommitHash,
-		})...,
-	)
 
 	var primary *types.LoadBalancer
 	// When the services is not accessed via ELB, the target group is not used
--- a/pkg/app/pipedv1/plugin/ecs/provider/ecs.go
+++ b/pkg/app/pipedv1/plugin/ecs/provider/ecs.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/ecs/config"
+	sdk "github.com/pipe-cd/piped-plugin-sdk-go"
 	"golang.org/x/sync/singleflight"
 )
 
@@ -63,9 +64,22 @@ func LoadTaskDefinition(appDir, taskDefinition string) (types.TaskDefinition, er
 }
 
 // LoadServiceDefinition returns Service object from a given service definition file.
-func LoadServiceDefinition(appDir, serviceDefinition string) (types.Service, error) {
+func LoadServiceDefinition(appDir, serviceDefinition string, input *sdk.ExecuteStageInput[config.ECSApplicationSpec]) (types.Service, error) {
 	path := filepath.Join(appDir, serviceDefinition)
-	return loadServiceDefinition(path)
+	service, err := loadServiceDefinition(path)
+	if err != nil {
+		return types.Service{}, err
+	}
+	// Add PipeCD-managed tags
+	service.Tags = append(service.Tags,
+		MakeTags(map[string]string{
+			LabelManagedBy:   ManagedByECSPlugin,
+			LabelPiped:       input.Request.Deployment.PipedID,
+			LabelApplication: input.Request.Deployment.ApplicationID,
+			LabelCommitHash:  input.Request.TargetDeploymentSource.CommitHash,
+		})...,
+	)
+	return service, nil
 }
 
 // LoadTargetGroups returns primary & canary target groups according to the defined in pipe definition file.


armistcxy:
There some works still needed to do before merging, I 've just created this PR so you (@khanhtc1202) can have a quick glance and give early feedbacks ^ ^ 

To make it easy when reviewing PR, you can refer to these resources:
- [DeepWiki: ECS_SYNC stage](https://deepwiki.com/pipe-cd/pipecd/7.2-ecs-executor#ecs_sync-stage)
- [ECS Executor Code (V0)](https://github.com/pipe-cd/pipecd/tree/1c045f7eed559a4fbd03dcecde1b25e0beac386b/pkg/app/piped/executor/ecs) 

Some changes have been made (different from v0)
1. No longer check PipCD managed tags for service: [Change](https://github.com/pipe-cd/pipecd/blob/1c045f7eed559a4fbd03dcecde1b25e0beac386b/pkg/app/piped/executor/ecs/ecs.go#L195C1-L197C4)
2. Check service's health status before update service: [Check](https://github.com/pipe-cd/pipecd/pull/6559/changes/c75587548d68d08beea0c9b6e9a00bbf8542dfc2#diff-d20e1dae48a2ed16be795941fdb960b03158733abb47419bf3132dc80986180cR193-R200)
3. Remove redundant slice copy in `client.ListTags`
```go
func (c *client) ListTags(ctx context.Context, resourceArn string) ([]types.Tag, error) {
	input := &ecs.ListTagsForResourceInput{
		ResourceArn: aws.String(resourceArn),
	}

	output, err := c.ecsClient.ListTagsForResource(ctx, input)
	if err != nil {
		return nil, err
	}

	tags := make([]types.Tag, 0, len(output.Tags))
	for _, t := range output.Tags {
		tags = append(tags, types.Tag{
			Key:   t.Key,
			Value: t.Value,
		})
	}
	return tags, nil
}
```

TODO (in works):
- [x] Figure out how to implement test for sync stage -> use mock for testing (logic-focus only, not real integrated) [add unit test for sync function](https://github.com/pipe-cd/pipecd/pull/6559/changes/5f46f2e92bc20111a8339c50c3ed9634c828c8aa)
- [x] Figure out how to support other properties/fields for tasks and services (currently only supports v0-compatible props)

armistcxy:
About the testing part, I've already made something like this to allow real integration test. But this is not safe due to potential waste if not handle cleaning up resources properly after test, maybe mocking will be better here

```go
	var (
		clusterARN   = os.Getenv("ECS_CLUSTER_ARN")
		subnetIDsStr = os.Getenv("SUBNET_IDS")
		sgIDsStr     = os.Getenv("SECURITY_GROUP_IDS")

		subnetIDs []string
		sgIDs     []string
	)

	if clusterARN == "" {
		t.Skip("ECS_CLUSTER_ARN is not set, skipping ECS sync stage test")
	}
	if subnetIDsStr == "" {
		t.Skip("SUBNET_IDS is not set, skipping ECS sync stage test")
	}
	for _, id := range strings.Split(subnetIDsStr, ",") {
		subnetIDs = append(subnetIDs, strings.TrimSpace(id))
	}
	if len(subnetIDs) == 0 {
		t.Skip("SUBNET_IDS is empty, skipping ECS sync stage test")
	}
	if sgIDsStr == "" {
		t.Skip("SECURITY_GROUP_IDS is not set, skipping ECS sync stage test")
	}
```

khanhtc1202:
For DetermineVersion function, I think you can ignore it in this PR, we can implement it by a separated PR

khanhtc1202:
Also, your commits had shown as `Unverified` currently, could you check this docs and update Github setting to fix that
https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits
Probably, this is caused by Github SSH Key is not being used

pkg/app/pipedv1/plugin/ecs/deployment/sync.go
```diff
@@ -0,0 +1,323 @@
...
+
+	// Add PipeCD-managed tags
+	serviceDef.Tags = append(serviceDef.Tags,
+		provider.MakeTags(map[string]string{
+			provider.LabelManagedBy:   provider.ManagedByECSPlugin,
+			provider.LabelPiped:       input.Request.Deployment.PipedID,
+			provider.LabelApplication: input.Request.Deployment.ApplicationID,
+			provider.LabelCommitHash:  input.Request.TargetDeploymentSource.CommitHash,
+		})...,
+	)
```
khanhtc1202: How about to move in inside LoadServiceDefinition function? Since we may need to do the same when implement rollout stage.

pkg/app/pipedv1/plugin/ecs/deployment/sync.go
```diff
@@ -0,0 +1,323 @@
...
+	lp := input.Client.LogPersister()
+	lp.Info("Running standalone task")
+
+	lp.Info("Start applying the ECS task definition")
+	tags := provider.MakeTags(map[string]string{
+		provider.LabelManagedBy:   provider.ManagedByECSPlugin,
+		provider.LabelPiped:       input.Request.Deployment.PipedID,
+		provider.LabelApplication: input.Request.Deployment.ApplicationID,
+		provider.LabelCommitHash:  input.Request.TargetDeploymentSource.CommitHash,
+	})
```
khanhtc1202: This `tags` is not being used until L183, so should move this block to right before that line (keep variable closed to it being used place to avoid unused var)

pkg/app/pipedv1/plugin/ecs/config/sync.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+// ECSSyncStageOptions contains options for the ECS sync stage.
+type ECSSyncStageOptions struct {
+	// Recreate indicates whether to recreate the service.
```
khanhtc1202, line 19:
```suggestion
	// Recreate indicates whether to recreate the service.
	// Enable this will ensure stop all running tasks before creating new task set.
```
nits, let add expected behavior of configuration for better understanding of its functionality.

armistcxy: Resolved with [chore: add expected behavior of configuration](https://github.com/pipe-cd/pipecd/pull/6559/changes/b05fcb52c91f8ea995aba90b01178140b4ef9471)

pkg/app/pipedv1/plugin/ecs/deployment/sync.go
armistcxy: Resolved with [chore: move tags to nearest used place](https://github.com/pipe-cd/pipecd/pull/6559/changes/983f35f3435a734bea73c99e832021fc712b06be)

pkg/app/pipedv1/plugin/ecs/deployment/sync.go
armistcxy: I think` LoadServiecDefinition` purpose is parsing manifest into `types.Service`, adding tags here I just feel not the right place to do that

armistcxy: IMO, adding tags explicitly in execute stage function will avoid confusion for the maintain work in the future

Also, `LoadServiceDefinition` current input consists of 2 params: `appDir, serviceDefinition string`, if moving adding tags to this function, we need to add `pipeId`, `applicationId`, `commitHash`, which will make the function bloat and drift away from its function

khanhtc1202: Okay, let re-check the service tags value of service object fetched from API when we update service. If it contains the tags as well, we can safely move this tag assign logic to load service definition, ensure service object in data flow will always have the tags. 👍 
For now, let's keep the implementation as your 👍

armistcxy: The service returned from UpdateService API doesn't contain tags but all the stage use `applyServiceDefinition` and inside that function the tags are re assigned

```go
// Re-assign tags to service object because UpdateService API doesn't return tags.
service.Tags = serviceDefinition.Tags
```

And my bad, the v0 `loadServiceDefinition` already add the tags inside
```go
func loadServiceDefinition(in *executor.Input, serviceDefinitionFile string, ds *deploysource.DeploySource) (types.Service, bool) {
	in.LogPersister.Infof("Loading service manifest at commit %s", ds.Revision)

	serviceDefinition, err := provider.LoadServiceDefinition(ds.AppDir, serviceDefinitionFile)
	if err != nil {
		in.LogPersister.Errorf("Failed to load ECS service definition (%v)", err)
		return types.Service{}, false
	}

	serviceDefinition.Tags = append(
		serviceDefinition.Tags,
		provider.MakeTags(map[string]string{
			provider.LabelManagedBy:   provider.ManagedByPiped,
			provider.LabelPiped:       in.PipedConfig.PipedID,
			provider.LabelApplication: in.Deployment.ApplicationId,
			provider.LabelCommitHash:  in.Deployment.CommitHash(),
		})...,
	)

	in.LogPersister.Infof("Successfully loaded the ECS service definition at commit %s", ds.Revision)
	return serviceDefinition, true
}
```

Update:
```go
// LoadServiceDefinition returns Service object from a given service definition file.
func LoadServiceDefinition(appDir, serviceDefinition string, input *sdk.ExecuteStageInput[config.ECSApplicationSpec]) (types.Service, error) {
	path := filepath.Join(appDir, serviceDefinition)
	service, err := loadServiceDefinition(path)
	if err != nil {
		return types.Service{}, err
	}
	// Add PipeCD-managed tags
	service.Tags = append(service.Tags,
		MakeTags(map[string]string{
			LabelManagedBy:   ManagedByECSPlugin,
			LabelPiped:       input.Request.Deployment.PipedID,
			LabelApplication: input.Request.Deployment.ApplicationID,
			LabelCommitHash:  input.Request.TargetDeploymentSource.CommitHash,
		})...,
	)
	return service, nil
}
```

armistcxy: Resolved with [chore: move add pipecd-related tags to load service function](https://github.com/pipe-cd/pipecd/pull/6559/changes/49dfb32e2df519dda9dff7ce02a5d64d468f49ab)

PR #6562, opened 2026-03-07
fix(cloudrun): guard against nil panics in manifest parsing, client options, and cache type assertion

mihir-dixit2k27:
## What this PR does

Fixes four nil-panic risks in `pkg/app/piped/platformprovider/cloudrun/`:

- **servicemanifest.go**: `ParseServiceManifest` — `sigs.k8s.io/yaml` 
  unmarshals empty input without error but leaves `obj.Object` nil, 
  panicking on any subsequent field access. Added nil Object guard.
- **revisionmanifest.go**: `ParseRevisionManifest` — same issue, same fix.
- **client.go**: `List`/`ListRevisions` dereference `options.Limit` 
  immediately — nil options panics. Default to empty struct instead.
- **cache.go**: Bare `item.(ServiceManifest)` assertion panics on type 
  mismatch. Replaced with comma-ok pattern + structured error log.

## Why we need it

All four are silent — only surface in production under specific inputs 
(empty config file, nil options, cache collision). Fixes are minimal 
and non-breaking; valid inputs follow identical code paths as before.

## Does this PR introduce a user-facing change?

- **How are users affected**: Empty `service.yaml` now returns a clear 
  error instead of panicking. All other behavior unchanged.
- **Is this a breaking change**: No.
- **How to migrate**: No migration needed.


mohammedfirdouss:
@khanhtc1202 this is waiting your review.

Issue #6446, opened 2026-01-20
[LFX Mentorship T-1 2026] Kubernetes Multi-Cluster Plugin #6446


mohammedfirdouss:
## Overview
This issue tracks the LFX Mentorship project for Term 1 2026, focused on enhancing the PipeCD v1 plugin architecture. Specifically, this project aims to complete the multi-cluster deployment support for Kubernetes applications as outlined in **RFC-0014**.

## Background
PipeCD v1 introduced a plugin-based architecture, and while a `kubernetes_multicluster` plugin exists, it currently only supports basic QuickSync and Rollback functions. 

To support enterprise-grade reliability (multi-region availability, backup systems, and localized data residency), PipeCD needs to bridge the gap between single-cluster feature parity and multi-cluster flexibility. Currently, advanced deployment strategies like Canary and Baseline rollouts are missing for multi-cluster environments.

### Current Status
| Feature | Status | Notes |
|---------|--------|-------|
| QuickSync (`K8S_MULTI_SYNC`) | Done | Deploys to multiple clusters at once |
| Rollback (`K8S_MULTI_ROLLBACK`) | Done | Handles partial failures |
| Aggregated Livestate | Partial | Functional, but has TODOs for ignore configs |
| Pipeline Sync stages | Not started | Canary, baseline, primary rollout missing |
| Traffic routing | Not started | Cross-cluster traffic management missing |

### Pipeline Sync Enhancements
* **Implement missing stages from RFC-0014:** `K8S_CANARY_ROLLOUT`, `K8S_CANARY_CLEAN`, `K8S_BASELINE_ROLLOUT`, `K8S_BASELINE_CLEAN`, and `K8S_PRIMARY_ROLLOUT`.
* **Per-stage multi-target support:** Allow different clusters to be targeted at specific stages of the pipeline.
* **Failure isolation:** Ensure that a deployment failure in one cluster does not unnecessarily halt the process for others.

### Reliability & Drift Detection
* **Livestate Improvements:** Implement missing ignore configs for more accurate state reporting.
* **Logic Refinement:** Resolve `DetermineStrategy` TODOs in the codebase to allow for intelligent strategy selection across multiple targets.

### Documentation & Testing
* **User Docs:** Update documentation to reflect the new multi-cluster capabilities and configuration options.
* **Testing:** Create comprehensive integration tests and example applications in the `example/` directory.

## Expected Outcomes
[...]

PR #6518, opened 2026-02-07
Apply defaults for Kubernetes Rollout and Multicluster Plugins

rawadhossain:
**What this PR does**:
Applies default values in Kubernetes-related plugins by adding `UnmarshalJSON` methods that call `defaults.Set()` where configs are unmarshaled without default handling.

**Why we need it**: 
Kubernetes rollout and multicluster configs had default tags that were not applied during unmarshaling.

### **Part of Issue #6174**

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**: 
  - Kubernetes rollout stages now correctly apply default suffixes (`primary`, `canary`, `baseline`) when omitted
  - Kubernetes multicluster applications now correctly apply default variant labels when not specified
- **Is this breaking change**: No
- **How to migrate (if breaking change)**: N/A

PR #6583, opened 2026-03-12
[Multi_K8s-Plugin] Canary Rollout

mohammedfirdouss:
**What this PR does**: Implements the K8S_CANARY_ROLLOUT stage for the kubernetes_multicluster plugin. When this stage runs it deploys the new version as a small canary replica set (configurable by count or percentage) to all targeted clusters in parallel, while leaving the primary workload untouched.

**Why we need it**: The multicluster plugin previously only supported K8S_MULTI_SYNC (all-or-nothing rollout) and K8S_MULTI_ROLLBACK. There was no way to progressively test a new version across clusters before promoting it. K8S_CANARY_ROLLOUT fills that gap and is the first step toward a full canary/baseline/primary pipeline strategy for multi-cluster deployments.

**Which issue(s) this PR fixes**: Part of #6446 

Yes — users can now add a K8S_CANARY_ROLLOUT stage to their app.pipecd.yaml pipeline when using the kubernetes_multicluster plugin.

- **How are users affected by this change**: Users who do not use a pipeline are unaffected. Users who configure a pipeline can now add K8S_CANARY_ROLLOUT with replicas, suffix, createService, and patches options.

**Canary Rollout with service**

Screenshot Capture - 2026-04-06 - 00-54-10
Screenshot Capture - 2026-04-06 - 00-57-07
Screenshot Capture - 2026-04-06 - 01-00-01
Screenshot Capture - 2026-04-06 - 01-00-12
Screenshot Capture - 2026-04-06 - 01-00-18

**Canary Rollout without service**

Screenshot Capture - 2026-04-06 - 00-49-42
Screenshot Capture - 2026-04-06 - 00-50-54
Screenshot Capture - 2026-04-06 - 00-51-28
Screenshot Capture - 2026-04-06 - 00-52-00
Screenshot Capture - 2026-04-06 - 00-52-20

### DIFF
--- a/.github/labeler.yaml
+++ b/.github/labeler.yaml
@@ -18,6 +18,10 @@ area/build:
 area/manifests:
 - manifests/**/*
 
+area/docs:
+- docs/*
+- docs/**/*
+
 area/examples:
 - examples/*
 - examples/**/*
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,3 +6,5 @@
 
 # Docs dir
 /docs/ @pipe-cd/pipecd-approvers-docs @pipe-cd/pipecd-approvers
+CONTRIBUTING.md @pipe-cd/pipecd-approvers-docs @pipe-cd/pipecd-approvers
+README.md @pipe-cd/pipecd-approvers-docs @pipe-cd/pipecd-approvers
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 [![Documentation](https://img.shields.io/badge/Documentation-pipecd-informational.svg)](https://pipecd.dev/docs/)
 [![Slack](https://img.shields.io/badge/Slack-%23pipecd-informational.svg)](https://app.slack.com/client/T08PSQ7BQ/C01B27F9T0X)
 [![Twitter](https://img.shields.io/twitter/url/https/twitter.com/pipecd_dev.svg?style=social&label=Follow%20%40pipecd_dev)](https://twitter.com/pipecd_dev)
+[![CNCF Status](https://img.shields.io/badge/CNCF%20Status-Sandbox-lightgrey.svg)](https://www.cncf.io/projects/pipecd/)
 
 <p align="center">
   <img src="https://github.com/pipe-cd/pipecd/blob/master/docs/static/images/logo.png" width="180"/>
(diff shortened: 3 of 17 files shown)


mohammedfirdouss:
@khanhtc1202 I have a question, am I meant to create example yml files for the pipecd-config, service and deployment to test fully? or it depends on how the user wants to setup their application.

Warashi:
Sorry, I merged #6518, which caused a merge conflict with this PR.

> I have a question, am I meant to create example yml files for the pipecd-config, service and deployment to test fully? or it depends on how the user wants to setup their application.

You need to create configs if you want to test on your local piped.
On the other hand, you don't need to do so if you think the test code is enough to check behavior.
It's better to check the behavior on your local, but it's not mandatory, I think. It's best to be able to check the expected behavior with the reproducible test code.

From another point of view, it would be better to demonstrate how to use this plugin by including example configs in the repository.

mohammedfirdouss:
> Sorry, I merged #6518, which caused a merge conflict with this PR.
> 
> > I have a question, am I meant to create example yml files for the pipecd-config, service and deployment to test fully? or it depends on how the user wants to setup their application.
> 
> You need to create configs if you want to test on your local piped. On the other hand, you don't need to do so if you think the test code is enough to check behavior. It's better to check the behavior on your local, but it's not mandatory, I think. It's best to be able to check the expected behavior with the reproducible test code.
> 
> From another point of view, it would be better to demonstrate how to use this plugin by including example configs in the repository.

@Warashi Thanks for the feedback! I've added example configs mirroring the structure used in the single-cluster plugin. I also added tests for both paths — one that verifies the canary Service is created with the variant selector injected, and one that verifies no Service is created when createService is false. Will resolve the conflicts now.

mohammedfirdouss:
@Warashi I intend to create seperate individual branches for all deployment/development stages of this plugin. What best practice would you give here, eg. Should Canary have both rollout and clean stages in one branch or seperate branches etc.

Warashi:
> @Warashi I intend to create seperate individual branches for all deployment/development stages of this plugin. What best practice would you give here, eg. Should Canary have both rollout and clean stages in one branch or seperate branches etc.

Yes, I want each stages implementations to be separate PRs, i.e. separate branches.
Thank you for your confirmination.

mohammedfirdouss:
> > @Warashi I intend to create seperate individual branches for all deployment/development stages of this plugin. What best practice would you give here, eg. Should Canary have both rollout and clean stages in one branch or seperate branches etc.
> 
> Yes, I want each stages implementations to be separate PRs, i.e. separate branches. Thank you for your confirmination.

Alright no problem, it is most likely going to cause a bit conflicts, but no issues.

mohammedfirdouss:
@Warashi @t-kikuc I have added pics showing how the canary rollout affects the GitOps cycle with and without service.

PR #6585, opened 2026-03-13
[ECS-Plugin] Setting default value for config

armistcxy:
**What this PR does**: 
- Setting default value for config
- Remove default value of ServiceDefinitionFile (no need for default value of this field because it will affect StandAloneTask mode) 

**Why we need it**:
- For example, if launch type is not specified in manifest, the value of launch type will be empty, ECS will try to use EC2 instead of fargate mode 

Fixes #

### DIFF
--- a/pkg/app/pipedv1/plugin/ecs/config/application.go
+++ b/pkg/app/pipedv1/plugin/ecs/config/application.go
@@ -35,8 +35,7 @@ type ECSDeploymentInput struct {
 
 	// ServiceDefinitionFile is the name of service definition file placing in application directory
 	// e.g., "servicedef.json" or "ecs/servicedef.yaml"
-	// Default: servicedef.json
-	ServiceDefinitionFile string `json:"serviceDefinitionFile,omitempty" default:"servicedef.json"`
+	ServiceDefinitionFile string `json:"serviceDefinitionFile,omitempty"`
 
 	// RunStandaloneTask indicates whether to run the task as a standalone task without creating/updating an ECS service.
 	// If true, the plugin will run the task directly without managing it through an ECS service.


pkg/app/pipedv1/plugin/ecs/config/application.go
```diff
@@ -29,8 +35,7 @@ type ECSDeploymentInput struct {
 
 	// ServiceDefinitionFile is the name of service definition file placing in application directory
 	// e.g., "servicedef.json" or "ecs/servicedef.yaml"
-	// Default: servicedef.json
-	ServiceDefinitionFile string `json:"serviceDefinitionFile,omitempty" default:"servicedef.json"`
+	ServiceDefinitionFile string `json:"serviceDefinitionFile,omitempty"`
```
Warashi, line 38: This requires users to set the serviceDefinitionFile when using a non-standalone task.
It would be better to set servicedef.json as the default for non-standalone tasks because it is more compatible with the piped v0 configuration.

@khanhtc1202 WDYT?

armistcxy: If @khanhtc1202 agree to set default value for service path, I will move check flag runStandAlone before load service file as the same way v0 implemented

```go
	if ecsInput.IsStandaloneTask() {
		if !runStandaloneTask(ctx, &e.Input, e.platformProviderName, e.platformProviderCfg, taskDefinition, &ecsInput) {
		}
	}

	servicedefinition, ok := loadServiceDefinition(&e.Input, ecsInput.ServiceDefinitionFile, e.deploySource)
	if !ok {
	}
```

armistcxy: Actually the old version did not have default tag, sorry for the confusion

```go
// pkg/config/application_ecs.go

// The name of service definition file placing in application directory.
ServiceDefinitionFile string `json:"serviceDefinitionFile"`
```

PR #6620, opened 2026-03-27
[ECS-Plugin] Implement Livestate Plugin

armistcxy:
**What this PR does**: Implements the livestate plugin for the ECS plugin, enabling PipeCD to report the live state of ECS applications to the control plane

**Why we need it**:

**Which issue(s) this PR fixes**: Part of #6443

- **How are users affected by this change**: Without this, PipeCD has no visibility into the live state of ECS applications after deployment. Users cannot tell whether their ECS serviceis healthy, how many tasks are running vs desired, or whether the cluster is drifted from the declared Git state.
- 

### DIFF
--- a/pkg/app/pipedv1/plugin/ecs/deployment/test_helper.go
+++ b/pkg/app/pipedv1/plugin/ecs/deployment/test_helper.go
@@ -41,11 +41,13 @@ func (f *fakeLogPersister) Complete(time.Duration) error      { return nil }
 type mockECSClient struct {
 	CreateServiceFunc               func(ctx context.Context, service types.Service) (*types.Service, error)
 	UpdateServiceFunc               func(ctx context.Context, service types.Service) (*types.Service, error)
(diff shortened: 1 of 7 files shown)


armistcxy:
Some screenshots as proof:

### Overview
image

### Resource detail

image

image


### Note
There's an issue with the UI (pipecd/web) that seems like currently support only K8s (my guess because I see "Kind", "Namespace", "API Version")

image

I quick fix the issue with Claude Code just to verify the livestate plugin implementation, have a look here
https://github.com/pipe-cd/pipecd/blob/223a97230bb4b96ef5b24034d9fc0789aabd82a0/web/src/components/application-detail-page/application-state-view/live-state-view/resource-detail/index.tsx#L19-L75.

will create a PR soon ^^ 
Edit: Fixed with #6803

khanhtc1202 (approved):
LGTM, thanks

PR #6646, opened 2026-04-05
Add support for configurable Plan Preview behavior in piped

rawadhossain:
**What this PR does**:
Adds a configurable `planPreview` section to piped config (`PipedSpec` in both `pkg/config` and `pkg/configv1`).

These values are validated and then passed to `planpreview.NewHandler` using existing options:
- `WithWorkerNum`
- `WithCommandQueueBufferSize`
- `WithCommandCheckInterval`
- `WithCommandHandleTimeout`

The configuration is applied in both piped entrypoints (`pkg/app/piped/cmd/piped` and `pkg/app/pipedv1/cmd/piped`).

Also includes tests to ensure:
- YAML parsing works correctly
- Invalid (negative) values fail validation

**Why we need it**: 
Currently plan preview behavior can only be tuned internally (mainly in tests).
This change allows to configure things like worker count, queue size, and timing directly from piped config.

### **Fixes Issue #5916**

  - Users can optionally add a `spec.planPreview` block in piped YAML:
    - `workerNum`
    - `commandQueueBufferSize`
    - `commandCheckInterval`
    - `commandHandleTimeout` (duration format like `5s`, `10m`)
  - If not provided, existing default behavior remains unchanged

### DIFF
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -535,6 +535,10 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			decrypter,
 			appManifestsCache,
 			cfg,
+			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
+			planpreview.WithCommandQueueBufferSize(cfg.PlanPreview.CommandQueueBufferSize),
+			planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()),
+			planpreview.WithCommandHandleTimeout(cfg.PlanPreview.CommandHandleTimeout.Duration()),
 			planpreview.WithLogger(input.Logger),
 		)
 		group.Go(func() error {
--- a/pkg/app/piped/cmd/piped/piped.go
+++ b/pkg/app/piped/cmd/piped/piped.go
@@ -526,6 +526,13 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			input.Logger.Info("successfully cleaned gitClient for plan-preview")
 		}()
 
+		ppOpts := []planpreview.Option{
+			planpreview.WithLogger(input.Logger),
+			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
+			planpreview.WithCommandQueueBufferSize(cfg.PlanPreview.CommandQueueBufferSize),
+			planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()),
+			planpreview.WithCommandHandleTimeout(cfg.PlanPreview.CommandHandleTimeout.Duration()),
+		}
 		h := planpreview.NewHandler(
 			gc,
 			apiClient,
@@ -535,11 +542,7 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			decrypter,
 			appManifestsCache,
 			cfg,
-			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
-			planpreview.WithCommandQueueBufferSize(cfg.PlanPreview.CommandQueueBufferSize),
-			planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()),
-			planpreview.WithCommandHandleTimeout(cfg.PlanPreview.CommandHandleTimeout.Duration()),
-			planpreview.WithLogger(input.Logger),
+			ppOpts...,
 		)
 		group.Go(func() error {
 			return h.Run(ctx)
--- a/pkg/app/pipedv1/cmd/piped/piped.go
+++ b/pkg/app/pipedv1/cmd/piped/piped.go
@@ -500,6 +500,10 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			decrypter,
 			cfg,
 			pluginRegistry,
+			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
+			planpreview.WithCommandQueueBufferSize(cfg.PlanPreview.CommandQueueBufferSize),
+			planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()),
+			planpreview.WithCommandHandleTimeout(cfg.PlanPreview.CommandHandleTimeout.Duration()),
 			planpreview.WithLogger(input.Logger),
 		)
 		group.Go(func() error {
--- a/pkg/app/pipedv1/cmd/piped/piped.go
+++ b/pkg/app/pipedv1/cmd/piped/piped.go
@@ -491,6 +491,13 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			input.Logger.Info("successfully cleaned gitClient for plan-preview")
 		}()
 
+		ppOpts := []planpreview.Option{
+			planpreview.WithLogger(input.Logger),
+			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
+			planpreview.WithCommandQueueBufferSize(cfg.PlanPreview.CommandQueueBufferSize),
+			planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()),
+			planpreview.WithCommandHandleTimeout(cfg.PlanPreview.CommandHandleTimeout.Duration()),
+		}
 		h := planpreview.NewHandler(
 			gc,
 			apiClient,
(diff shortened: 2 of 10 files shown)


armistcxy:
Hi @rawadhossain, let me help you review this PR ^^

pkg/config/piped.go
```diff
@@ -1211,3 +1216,30 @@ type PipedEventWatcherGitRepo struct {
...
+}
+
+func (p *PipedPlanPreview) Validate() error {
+	if p.WorkerNum < 0 {
+		return errors.New("planPreview.workerNum must be greater than or equal to 0")
+	}
+	if p.CommandQueueBufferSize < 0 {
+		return errors.New("planPreview.commandQueueBufferSize must be greater than or equal to 0")
+	}
+	if p.CommandCheckInterval < 0 {
```
armistcxy, line 1238: `CommandCheckInterval` will later be used for creating `time.Ticker`

```go
commandTicker := time.NewTicker(h.options.commandCheckInterval)
```
This will cause panic if `h.options.commandCheckInterval = 0`

pkg/configv1/piped.go
```diff
@@ -632,6 +637,33 @@ type PipedEventWatcherGitRepo struct {
...
+}
+
+func (p *PipedPlanPreview) Validate() error {
+	if p.WorkerNum < 0 {
+		return errors.New("planPreview.workerNum must be greater than or equal to 0")
+	}
+	if p.CommandQueueBufferSize < 0 {
+		return errors.New("planPreview.commandQueueBufferSize must be greater than or equal to 0")
+	}
+	if p.CommandCheckInterval < 0 {
```
armistcxy, line 658: same reason with v0, commandCheckInterval = 0 would cause panic

pkg/app/piped/cmd/piped/piped.go
```diff
@@ -535,6 +535,10 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			decrypter,
 			appManifestsCache,
 			cfg,
+			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
```
armistcxy: IMO, this is dangerous, I would think a safer approach is checking whether these values from `PlanPreview` options are not zero value first: if it != 0, then add the option

For example

```go
if cfg.PlanPreview.WorkerNum > 0 {
    planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
}
```

pkg/app/pipedv1/cmd/piped/piped.go
```diff
@@ -500,6 +500,10 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			decrypter,
 			cfg,
 			pluginRegistry,
+			planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum),
```
armistcxy: Same reason as v0

pkg/config/piped.go
```diff
@@ -1211,3 +1216,30 @@ type PipedEventWatcherGitRepo struct {
...
+	// Channel buffer size used for plan-preview commands.
+	CommandQueueBufferSize int `json:"commandQueueBufferSize,omitempty"`
+	// Interval to fetch plan-preview commands.
+	CommandCheckInterval Duration `json:"commandCheckInterval,omitempty"`
+	// Timeout to handle each plan-preview command.
+	CommandHandleTimeout Duration `json:"commandHandleTimeout,omitempty"`
+}
+
+func (p *PipedPlanPreview) Validate() error {
+	if p.WorkerNum < 0 {
```
armistcxy, line 1232: You allow WorkerNum = 0, I scare that commands can be block by this allowance, please verify the behavior to find the right condition for validating

rawadhossain:
Sure @armistcxy, please take a look.

pkg/app/piped/cmd/piped/piped.go
rawadhossain: ahh, good one. Yeah, how about I frame it like this, does it seem right now:
``` go
ppOpts := []planpreview.Option{
			planpreview.WithLogger(input.Logger),
		}
		if cfg.PlanPreview.WorkerNum > 0 {
			ppOpts = append(ppOpts, planpreview.WithWorkerNum(cfg.PlanPreview.WorkerNum))
		}
		if cfg.PlanPreview.CommandQueueBufferSize > 0 {
			ppOpts = append(ppOpts, planpreview.WithCommandQueueBufferSize(cfg.PlanPreview.CommandQueueBufferSize))
		}
		if cfg.PlanPreview.CommandCheckInterval > 0 {
			ppOpts = append(ppOpts, planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()))
		}
		if cfg.PlanPreview.CommandHandleTimeout > 0 {
			ppOpts = append(ppOpts, planpreview.WithCommandHandleTimeout(cfg.PlanPreview.CommandHandleTimeout.Duration()))
		}

h := planpreview.NewHandler(
			//....
			ppOpts...,
		)
```

pkg/config/piped.go
rawadhossain: Good point, I checked the handler behavior here in `piped/planpreview/handler.go`:
``` go
for i := 0; i < h.options.workerNum; i++ {
		go startWorker(ctx, h.commandCh)
	}
```
From the handler logic, `workerNum = 0` would result in no workers being started, so commands wouldn’t be processed if it reached there as is.

In the current flow, we guard this at the callsite (`> 0` check in `piped.go`), so a zero value isn’t passed to `WithWorkerNum` and the handler falls back to its default worker count.

I’m happy to enforce `> 0` in `Validate()` as well if you think that’s preferable, though it may also affect cases where the field is omitted and defaults are expected.

pkg/config/piped.go
rawadhossain: right, `time.NewTicker` will panic for non positive durations, so this definitely needs to be guarded.

We handle this the same way as above in `piped.go` by passing the option when the value is `> 0`, so zero isn’t forwarded and the handler uses its default interval.
``` go
if cfg.PlanPreview.CommandCheckInterval > 0 {
    ppOpts = append(ppOpts, planpreview.WithCommandCheckInterval(cfg.PlanPreview.CommandCheckInterval.Duration()))
}
```

rawadhossain:
@armistcxy heya I responded to your comments, PTAL. 
Will commit the changes if this seems right.

pkg/config/piped.go
armistcxy: I'm not sure the right thing to do here, but isn't that we are using defensive programming in a lot of place, I don't think this is a good idea

maybe just concentrate on validating config at one place, what do you think ?

pkg/app/piped/cmd/piped/piped.go
armistcxy: it seems like a good approach now, i think you should verify whether `CommandQueueBufferSize = 0` is valid. Go has buffer size = 0 for channel right ^ ^, maybe this case can be similar

rawadhossain: Yeah I checked this, since we do `make(chan ..., commandQueueBufferSize)` in the handler, `make(chan T, 0)` is valid in Go (unbuffered), so `CommandQueueBufferSize = 0` itself wouldn’t break anything.

in `pkg/app/piped/planpreview/handler.go`:
```go
h := &Handler{
		gitClient:     gc,
		commandLister: cl,
		commandCh:     make(chan model.ReportableCommand, opt.commandQueueBufferSize),
		prevCommands:  map[string]struct{}{},
		options:       opt,
		logger:        opt.logger.Named("plan-preview-handler"),
	}
```

The tricky part is config: since it’s a plain `int`, omitted and `0` both become `0` after decode. So if we pass `0` through, even “not set” would behave like unbuffered instead of using the default (10), which might be unexpected.

Right now with the `> 0` guard, both cases just fall back to the default, which I felt safer.

If we want to support unbuffered explicitly, we’d probably need a way to distinguish omitted vs 0.

What do you think, should we support that, or keep default-only for now?

pkg/config/piped.go
rawadhossain: yeah, I think that would be cleaner. I’ve moved the `> 0` checks into the `With*` option functions in the handler, so the “0 = keep default” logic is handled in one place now.

`Validate()` is just handling invalid cases (like negatives). Since with plain `int`/`Duration`, omitted and `0` both decode the same after YAML parsing, we can’t really reject `0` there without breaking configs where users just omit the `planPreview` block.

Let me know your thoughts on this.

rawadhossain:
hey @armistcxy sorry couldn't respond earlier, was caught up with exams. I made some changes and replied to your follow-ups. PTAL.

rawadhossain:
Hi @Warashi, can you please have a look into the pr?

Warashi (approved):
Sorry for late.
LGTM, Thanks!

rawadhossain:
> Sorry for late. LGTM, Thanks!

no worries at all, glad to contribute✌️

PR #6668, opened 2026-04-14
bug(lambda): retain repository name for unknown git providers

### LINKED ISSUE #6666: bug(lambda): Retain repository name when parsing unknown Git providers for artifact versions
**What happened**:

In `pkg/app/piped/platformprovider/lambda/function.go`, the `FindArtifactVersions` function looks at the Git URL to generate a commit hyperlink. If it encounters a Git host it doesn't explicitly recognize (like AWS CodeCommit or a self-hosted Git server), it falls back to a `default:` case where it zeroes out both the `gitURL` **and** the `repoPath` (`repoPath = ""`). This results in `ArtifactVersion.Name` being completely empty.

While it makes sense for the `gitURL` hyperlink to be empty if the provider's web-viewer format is unknown, the parsed `repoPath` should NOT be zeroed out. Discarding it causes a completely blank artifact name/repository to be displayed in the PipeCD UI.

**How to reproduce it**:

Define a lambda `source` manifest where the `git` URL points to an unrecognized host (e.g., `git: https://git.mycompany.com/my/repo.git`). When viewing the deployment on the control-plane UI, the source artifact name displays as blank.

**Environment**:

- `piped` version: `master`
- `control-plane` version: `master`
- Others: N/A

I am intending to submit a PR to safely retain the `repoPath` name for this :)

zyzzmohit:
**What this PR does**:

In `pkg/app/piped/platformprovider/lambda/function.go`, the `FindArtifactVersions` function attempts to parse the Git URL to generate a commit hyperlink for Lambda Source Code deployments. If it encounters a Git host it doesn't instantly recognize (like AWS CodeCommit or an internal Git server), it formerly fell into a `default` case that completely wiped out both the `gitURL` hyperlink AND the `repoPath`. This PR removes the line zeroing out `repoPath`, safely retaining the parsed repository path as the artifact name fallback when the Git provider's hyperlink format is visually unknown.

**Why we need it**:

Zeroing out the `repoPath` caused the `ArtifactVersion.Name` to become empty, resulting in completely blank artifact layers surfacing within the PipeCD web UI. By safely preserving the `repoPath`, PipeCD can correctly identify and display the source repository path during deployments even if it can't hyperlink the commit hash itself. 

Fixes #6666 

Yes. 

Users loading AWS Lambda configurations backed by unknown / self-hosted Git providers will now see their repository correctly displayed in the PipeCD UI.

### DIFF
--- a/pkg/app/piped/platformprovider/lambda/function.go
+++ b/pkg/app/piped/platformprovider/lambda/function.go
@@ -270,7 +270,6 @@ func FindArtifactVersions(fm FunctionManifest) ([]*model.ArtifactVersion, error)
 		default:
 			// TODO: Show repo name with commit link for other git provider
 			gitURL = ""
-			repoPath = ""
 		}
 
 		return []*model.ArtifactVersion{
--- a/pkg/app/piped/platformprovider/lambda/function_test.go
+++ b/pkg/app/piped/platformprovider/lambda/function_test.go
@@ -512,7 +512,7 @@ spec:
 				{
 					Kind:    model.ArtifactVersion_GIT_SOURCE,
 					Version: "dede7cdea5bbd3fdbcc4674bfcd2b2f9e0579603",
-					Name:    "",
+					Name:    "username/lambda-function-code",
 					Url:     "",
 				},
 			},


khanhtc1202 (dismissed):
Nice catch, thanks 👍

rahulshendre:
I think we are good to go on this one
cc @khanhtc1202 @eeshaanSA

PR #6671, opened 2026-04-15
[ECS-Plugin]: Determine strategy for deployment

armistcxy:
**What this PR does**:  Implements `DetermineStrategy` for the ECS plugin

**Why we need it**: compares the container images between the running and target task definitions to choose the appropriate sync strategy:
- Any image added, removed, or changed: PipelineSync (canary/progressive deployment)
- No image difference: QuickSync

**Which issue(s) this PR fixes**: Previously DetermineStrategy always returned QuickSync, which caused the pipeline configured by users to be silently bypassed on every deployment

Fixes # Part of #6443 

### DIFF
--- a/pkg/app/pipedv1/plugin/ecs/deployment/determine.go
+++ b/pkg/app/pipedv1/plugin/ecs/deployment/determine.go
@@ -15,6 +15,8 @@
 package deployment
 
 import (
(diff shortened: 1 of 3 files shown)


khanhtc1202 (approved):
LGTM

PR #6672, opened 2026-04-15
[Multi_K8s-Plugin] Fix DetermineStrategy to be multi-target aware

mohammedfirdouss:
**What this PR does**:
Makes `DetermineStrategy` and `DetermineVersions` in the `kubernetes_multicluster` plugin aware of per-cluster manifest overrides (`multiTargets`). Previously both functions passed an empty `KubernetesMultiTarget{}` when loading manifests, so per-target manifest paths were silently ignored when deciding whether a deployment required progressive delivery.

**Why we need it**:
When a user configures different manifest paths per cluster (e.g. `./cluster-us` and `./cluster-eu`), a change only to one cluster's `deployment.yaml` would not trigger `PipelineSync` — the strategy check was loading from the app-level manifest path and missing the per-target change entirely. This could cause a workload update to be deployed via QuickSync, bypassing canary and baseline stages completely.

Fixes #6446

- **How are users affected by this change**: `DetermineStrategy` now correctly triggers `PipelineSync` when a workload change is made to any configured `multiTarget`, even if the app-level manifests are unchanged.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/plugin.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/plugin.go
@@ -150,15 +150,41 @@ func (p *Plugin) DetermineVersions(ctx context.Context, _ *sdk.ConfigNone, input
 		return nil, err
 	}
 
-	// TODO: consider multiTarget later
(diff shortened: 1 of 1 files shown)


mohammedfirdouss:
@khanhtc1202 @Warashi This is ready for review.

khanhtc1202 (approved):
LGTM

PR #6673, opened 2026-04-15
[Multi_K8s-Plugin] Fix livestate drift detection and resource filtering

mohammedfirdouss:
**What this PR does**:
Implements two missing livestate features in the `kubernetes_multicluster` plugin, porting them from the single-cluster `kubernetes` plugin:

1. **Drift detection ignore**: Resources annotated with `pipecd.dev/ignore-drift-detection: "true"` are now excluded from drift comparison, matching the behaviour of the single-cluster plugin.
2. **Include/exclude resource filtering**: The deploy target config now supports `appStateInformer.includeResources` and `appStateInformer.excludeResources` to control which resource types are included in the livestate drift check.

**Why we need it**:
Without drift detection ignore, applications with resources that are expected to drift at runtime (e.g. HPAs, Jobs managed by external controllers) would always show as `OutOfSync` in the PipeCD UI, even when everything is working correctly. Without include/exclude filtering, there was no way to scope the drift check to the resource types that matter for a given deploy target.

Fixes #6446

- **How are users affected by this change**: Users can now annotate resources with `pipecd.dev/ignore-drift-detection: "true"` to exclude them from drift detection. Users can also configure `appStateInformer.includeResources` and `appStateInformer.excludeResources` per deploy target to scope the livestate comparison.


mohammedfirdouss:
@khanhtc1202 @Warashi This is ready for review.

khanhtc1202 (approved):
LGTM, thanks

Issue #6691, opened 2026-04-17
Way to transfer PipeCD data across control plane


khanhtc1202:
**What would you like to be added**:

Have a way to transfer PipeCD data across the PipeCD Control Plane. The user can keep using PipeCD with minimal config updates or settings.

**Why is this needed**:

Same as issue #5311 but across ControlPlane too. I think support this issue will also resolve #5311

lavanyayadawad30-lab:
@eeshaanSA, can I work on this issue?

PR #6692, opened 2026-04-17
Introduce pipectl transfer command

khanhtc1202:
**What this PR does**:

Introduce `pipectl transfer` command, this command will be used to migrate pipecd applications and pipeds from a source controlplane to a target controlplane. Useful for migrating control plane without manually recreating all applications/pipelines, everything will keep working as is after migration.

This transfer command contains 2 subcommands

- pipectl transfer backup: export all current piped and applications of the current project to JSON file
- pipectl transfer restore {piped,application}: add exported piped and applications to the target controlplane

Note:
- The implementation relies on the current cli API of the pipecd controlplane to reduce the complexity of supporting all current pipecd controlplane datastores (MySQL and Firestore)
- Since the current cli API does not support adding deployments, all previous applications' deployments are not backed up (reason we should do it as follows)
- After migrating the piped, it will trigger SYNC for all applications it's managing, since there are no previous deployments, triggered deployments are all QUICK_SYNC, thus basically no change/no failure as workload unchanged in users' cluster.
- Docs for transfer flow will be prepared by a separate PR.

**Why we need it**:

Part of #6691


pkg/app/pipectl/cmd/transfer/restore_piped.go
```diff
@@ -0,0 +1,159 @@
...
+	pipedIDMap := make(map[string]string, len(pipeds))
+
+	for _, p := range pipeds {
+		resp, err := cli.RegisterPiped(ctx, &apiservice.RegisterPipedRequest{
+			Name: p.Name,
+			Desc: p.Desc,
+		})
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to register piped %q (%s): %w", p.Name, p.Id, err)
+		}
```
armistcxy, line 122: Not bug but why registering pipeds exit when an error occured
while registering applications keeps adding and return number of success and fail restoration

armistcxy: For example if there are 10 pipeds need to be restored, if it fail early then we need to add remain ones manually
P1
P2
P3: fail
P4: need manual adding
....
P10: need manual adding

I know this problem can be easily solved by deleting P1, P2, P3 from backup file and reapply the restore command but I still think that continuing even fail is better

pkg/app/pipectl/cmd/transfer/restore_application.go
```diff
@@ -0,0 +1,181 @@
...
+		if err != nil {
+			logger.Warn("Failed to restore application",
+				zap.String("app-name", app.Name),
+				zap.String("app-id", app.Id),
+				zap.Error(err),
+			)
+			failed++
+			continue
+		}
+
```
armistcxy, line 153: Restored application does not have deploy target to deploy

Add application does not have field deployTargets, but we can workaround with `UpdateApplicationDeployTargets`

I think you can update like this
```go
		if len(app.DeployTargetsByPlugin) > 0 {
			if _, err := cli.UpdateApplicationDeployTargets(ctx, &apiservice.UpdateApplicationDeployTargetsRequest{
				ApplicationId:         resp.ApplicationId,
				DeployTargetsByPlugin: app.DeployTargetsByPlugin,
			}); err != nil {
				logger.Warn("Application restored but failed to set deploy targets",
					zap.String("app-name", app.Name),
					zap.String("new-id", resp.ApplicationId),
					zap.Error(err),
				)
			}
		}
```

pkg/app/pipectl/cmd/transfer/restore_application.go
```diff
@@ -0,0 +1,181 @@
...
+				)
+			}
+		} else {
+			logger.Info("Restored application",
+				zap.String("name", app.Name),
+				zap.String("old-id", app.Id),
+				zap.String("new-id", resp.ApplicationId),
+			)
+		}
+		restored++
```
armistcxy, line 178: not bug but success restore still count if a disabled application (original) -> enabled application (restored). what do you think about this ?

armistcxy: should we count this as fail ? failed++

armistcxy (commented):
Overall LGTM but there're some concerns

pkg/app/pipectl/cmd/transfer/restore_piped.go
```diff
@@ -0,0 +1,159 @@
...
+
+		mappings = append(mappings, PipedMapping{
+			OldPipedID: p.Id,
+			NewPipedID: resp.Id,
+			NewKey:     resp.Key,
+			PipedName:  p.Name,
+		})
```
armistcxy, line 129: oh, I almost forget to mention that should we return Base64 encoded key here for easy copy paste

when testing this feature, I need to run "echo <key> | base64" and copy, I would love to return the base64 encode as the UI return

pkg/app/pipectl/cmd/transfer/restore_application.go
khanhtc1202: Yes, since this pipectl transfer command supports only pipedv0, no deploy targets feel would available at the time application be transferred.

pkg/app/pipectl/cmd/transfer/restore_application.go
khanhtc1202: That's because I think the application that successfully be pushed to the new datastore should be counted as restored, the warning is to warn the users that they should manually disable that application later via UI. If we count it as failed, users may try to manually add/register that application again, which leads to duplication.

pkg/app/pipectl/cmd/transfer/restore_piped.go
khanhtc1202: Good point, let me address that 👍

pkg/app/pipectl/cmd/transfer/restore_piped.go
khanhtc1202: @armistcxy The number of piped is always smaller compare to the number of applications, thus I prefer to retry whole things (after delete P1->P3 on UI)

Another reason is the output of the pipectl transfer restore piped command is  a piped mapping file, which will be used by the restore application step. I don't want to force users to manually update/add piped mapping to that file in order to make restore application step work. So, fail fast and delete P1->P3, then re-do from start is recommended.

khanhtc1202:
Note: Checking the labels for applications

pkg/app/pipectl/cmd/transfer/restore_piped.go
khanhtc1202: Addressed by https://github.com/pipe-cd/pipecd/pull/6692/commits/e748084a526e5d17d1f2ae64acda91948ccb4ff2

khanhtc1202:
> Note: Checking the labels for applications

Need to update the pipectl API to support labels for application on registration (restore app), let me address it by a separated PR, thanks @armistcxy

eeshaanSA:
600+ plus lines. 😿 

I will give it a look tonight, if you need a reviewing hand. 😃

pkg/app/pipectl/cmd/transfer/restore_application.go
```diff
@@ -0,0 +1,181 @@
...
+				zap.String("app-id", app.Id),
+				zap.Error(err),
+			)
+			failed++
+			continue
+		}
+
+		// Preserve disabled status — AddApplication always creates apps as enabled.
+		if app.Disabled {
+			if _, err := cli.DisableApplication(ctx, &apiservice.DisableApplicationRequest{
```
Warashi, line 156: I haven't verified this yet, but as far as I know, AddApplication triggers the initial deployment. I'm concerned that adding an application as "enabled" (for migration purposes) will trigger a deployment before we disable it.

Warashi (commented):
I haven't run this yet, but the code itself seems good.
I commented on triggering the initial deployment.

pkg/app/pipectl/cmd/transfer/restore_application.go
khanhtc1202: Hi @Warashi sorry for the late reply, just came back from my vacation :smile: 
Users can stop piped while importing/registering applications to ensure no applications' deployment be triggered while transfering, does this solve your concern :eyes:

Warashi: Yes, it solved my concern :+1:
LGTM

Warashi (approved):
LGTM

PR #6697, opened 2026-04-20
[Multi_K8s-Plugin] Config hash for StatefulSet and DaemonSet

mohammedfirdouss:
**What this PR does**:
Extends the config hash annotation feature in the `kubernetes_multicluster` plugin to cover StatefulSet and DaemonSet workloads, in addition to the existing Deployment support.

When a ConfigMap or Secret referenced by a workload changes, PipeCD annotates the workload's pod template with a hash of the config content triggering a rollout automatically. Before this change, only Deployments received this annotation. StatefulSets and DaemonSets were silently skipped (a TODO comment in `annotate.go` line 55 noted this gap).

**Why we need it**:
StatefulSets and DaemonSets can reference ConfigMaps and Secrets just like Deployments. Without this fix, changes to those config resources would not trigger a restart of StatefulSet or DaemonSet pods, leaving them running stale configuration silently.

Related to #6446

- **How are users affected by this change**: StatefulSet and DaemonSet pods now automatically restart when a referenced ConfigMap or Secret changes, matching the existing behaviour for Deployments.


khanhtc1202 (approved):
Thanks 👍

PR #6704, opened 2026-04-22
Update NodeJS version to v20.19.0 for CI and Docs

khanhtc1202:
**What this PR does**:

Update NodeJS version to v20.19.0 for CI and Docs

**Why we need it**:

Keep NodeJS version consistency across the project. This version is already being used in the controlplane Dockerfile, so it should not affect our users.

NOTE: The site is using NodeJS v24, which is required by Hugo, and this doesn't relate to the PipeCD codebase itself.

Fixes #


t-kikuc (approved):
thank you!

PR #6710, opened 2026-04-23
[Pipedv1] Scan preinstalled tools

### LINKED ISSUE #6709: pipedv1 tool registry does not scan pre-installed tools
**What would you like to be added**: Scan the tools directory on registry initialization so already-installed tools are recognized without re-downloading (currently piped download tools and overwrite tool even exist) 

**Why is this needed**: Quicker start piped + save network bandwith

armistcxy:
**What this PR does**: Adds scanning of the tools directory on registry initialization in pipedv1, so already-installed tools are recognized without re-downloading.

Also cleans up the outDir after a successful tool install

**Why we need it**: Previously, newToolRegistry initialized installedTools as an empty map, causing piped to re-download and overwrite all tool binaries on every restart even if the binaries already existed on disk

Fixes #6709


khanhtc1202 (approved):
LGTM

PR #6718, opened 2026-04-27
[Multi_K8s-Plugin] Support per-target kustomizeDir

mohammedfirdouss:
**What this PR does**:
Implements support for the ``kustomizeDir`` field on ``KubernetesMultiTarget``, which was defined in the config struct and documented in ``RFC-0014`` but never wired up to the manifest loader.

Before this change, the ``kustomizeDir`` field was silently ignored and users with multi-target Kustomize overlays in per-cluster subdirectories would never have their kustomization files detected or used.

**Why we need it**:
Without this fix, any multi-target app relying on per-target Kustomize overlays (e.g. ``cluster-us/kustomization.yaml``) would fall back to plain YAML loading silently, producing incorrect manifests with no error.

Part of #6446

- **How are users affected by this change**: Multi-target apps with ``kustomizeDir`` set per target now correctly detect and run kustomize from that subdirectory instead of the root app directory.


mohammedfirdouss:
@Warashi Please look when you have time for review.

Warashi (approved):
LGTM

PR #6721, opened 2026-04-27
fix: propagate json.Unmarshal error in mysql rowDataConverter

### LINKED ISSUE #6699: Missing error handling on json.Unmarshal inside MySQL Row Iteration
### Description
In the MySQL Datastore implementation, rowDataConverter.Data() blindly ignores the error returned by json.Unmarshal(jsonRaw, &obj).

If the datastore retrieves a malformed JSON slice or truncated bytes from MySQL, json.Unmarshal fails silently. Consequently, the function returns an empty or partially populated map. This leads to silent logic errors and "ghost bugs" down the pipeline instead of throwing an explicit error.

### Code Context
https://github.com/pipe-cd/pipecd/blob/master/pkg/datastore/mysql/iterator.go#L86

### Proposed Fix
Update the compiling interface (dataConverter) to return (map[string]interface{}, error) and bubble the failure up to the Iterator.Next() trace. I am raising a PR for this shortly.

vedant21-oss:
Resolves #6699

The rowDataConverter in the MySQL datastore now returns an error that bubbles up through Iterator.Next() to handle silent json.Unmarshal failures.

### DIFF
--- a/pkg/datastore/mysql/iterator.go
+++ b/pkg/datastore/mysql/iterator.go
@@ -24,7 +24,7 @@ import (
 )
 
 type dataConverter interface {
-	Data() map[string]interface{}
+	Data() (map[string]interface{}, error)
 }
 
 // Iterator for MySQL result set
@@ -59,7 +59,10 @@ func (it *Iterator) Cursor() (string, error) {
 		return "", datastore.ErrInvalidCursor
 	}
 
-	lastObjData := it.last.Data()
+	lastObjData, err := it.last.Data()
+	if err != nil {
+		return "", err
+	}
 
 	cursor := make(map[string]interface{}, len(it.orders))
 	for _, o := range it.orders {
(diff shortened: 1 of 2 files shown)


eeshaanSA (commented):
Hey @vedant21-oss, thanks for this. Do you have any logs that can support this? 

Did anyone try reproducing?

@rahulshendre @Ayushmore1214

vedant21-oss:
Hey @eeshaanSA, thanks for the review!

I just updated the PR to include a unit test that explicitly reproduces this behavior in `pkg/datastore/mysql/iterator_test.go`.

Without the fix, the JSON unmarshaling silently failed, and an empty/partial object was returned. With this fix, running the tests explicitly captures the error. Here is the log output from the new testcase simulating invalid JSON data from the database:

`=== RUN   TestData/invalid_json_data`
`    iterator_test.go:152: Expected error caught: unexpected end of JSON input`
`--- PASS: TestData/invalid_json_data (0.00s)`

The unmarshal error is now properly propagated up to the caller instead of being swallowed. I've force-pushed the test coverage alongside the fix.

rahulshendre (commented):
Reproduced locally. All tests pass, invalid_json_data catches the error correctly.
One note: 
Cursor() error path (lines 62-65 of iterator.go) isn't covered by any test case, it returns nil.
@vedant21-oss i think we should add a case there too

rahulshendre:
hey @vedant21-oss, any updates on this one 👀

PR #6723, opened 2026-04-28
[Multi_K8s-Plugin] Add initializer hook for Helm repo and OCI registry auth

mohammedfirdouss:
**What this PR does**:
Adds an initializer hook to the `kubernetes_multicluster` plugin, bringing it to parity with the single-cluster `kubernetes` plugin.

At plugin startup, the initializer:
1. Registers any HTTP Helm chart repositories configured in the piped config (`helm repo add` + `helm repo update`)
2. Logs in to any OCI Helm chart registries configured in the piped config (`helm registry login`)

This also introduces `KubernetesPluginConfig` (with `chartRepositories` and `chartRegistries` fields) as the plugin-level config type, replacing `sdk.ConfigNone` across the deployment, livestate, and plan-preview plugin interfaces.

**Why we need it**:
Without this, users of the multicluster plugin deploying from private Helm chart repositories or OCI registries would get a 401 Unauthorized error at manifest load time with no clear way to provide credentials. The single-cluster plugin has supported this since its initializer was added the multicluster plugin was missing the equivalent.

Fixes #6446

- **How are users affected by this change**: Users can now configure `chartRepositories` and `chartRegistries` in their piped plugin config for the multicluster plugin, enabling deployments from private Helm repositories and OCI registries.


mohammedfirdouss:
@Warashi Please look when you have time for review.

pkg/app/pipedv1/plugin/kubernetes_multicluster/provider/helm.go
```diff
@@ -47,6 +50,62 @@ func NewHelm(version, path string, logger *zap.Logger) *Helm {
...
+func (h *Helm) LoginToOCIRegistry(ctx context.Context, address, username, password string) error {
+	args := []string{
+		"registry",
+		"login",
+		"-u",
+		username,
+		"-p",
+		password,
+		address,
+	}
```
Warashi, line 61: Can we pass the password via stdin?
The command arguments can be read more easily than the stdin from other processes, so it's better to pass the password via stdin.

Warashi (commented):
Almost LGTM.
The one point I have a concern about is the password handling.

pkg/app/pipedv1/plugin/kubernetes_multicluster/provider/helm.go
mohammedfirdouss: Sure we can do that, i think the cli arguments wont be effective.

Warashi (approved):
LGTM

PR #6724, opened 2026-04-28
Support labels value for client API add application endpoint

khanhtc1202:
**What this PR does**:

Support labels value for client API - add application endpoint

**Why we need it**:

User can register applications with label values via PipeCD Web UI, they should be able to do the same using PipeCD cli (pipectl or terraform)

Related PR #6692

- **How are users affected by this change**: Possible to register applications with label values via pipectl


armistcxy:
LGTM :+1:

mohammedfirdouss:
all good👍, the branch just needs to be up-to-date. I reference this PR as a template somewhere.

Warashi (approved):
LGTM

PR #6727, opened 2026-04-28
fix: paginate ListNotCompletedDeployments in pipedv1

### LINKED ISSUE #6696: bug: piped's ListNotCompletedDeployments call ignores pagination and can miss deployments
## What happened

In both `piped` (v0) and `pipedv1`, the deployment store's `sync()` loop calls `ListNotCompletedDeployments` exactly once per sync cycle and discards the `cursor` field on the response:

- `pkg/app/piped/apistore/deploymentstore/store.go:105-111`
- `pkg/app/pipedv1/apistore/deploymentstore/store.go:105-111`

Both files carry an explicit TODO that has been there since the RPC was introduced:

```go
// TODO: Call ListNotCompletedDeployments itervally until all required deployments are fetched.
resp, err := s.apiClient.ListNotCompletedDeployments(ctx, &pipedservice.ListNotCompletedDeploymentsRequest{})
```

The server already returns a cursor on partial pages (`pkg/app/server/service/pipedservice/service.proto:278-281`), but the request message has no `cursor` / `page_size` fields, and the server (`pkg/app/server/grpcapi/piped_api.go:350-386`) does not accept one either — so there is currently no way for piped to request the next page.

## Impact

If the controlplane's datastore returns a partial page of not-completed deployments (this depends on the backing datastore's default page size), `piped` will silently skip every deployment past the first page until the next sync tick — and since sync always starts from the beginning, the tail can be starved indefinitely under steady deployment pressure.

This is the classic "cursor returned but never followed" bug; it becomes user-visible as "some deployments appear stuck in PENDING/PLANNED until piped restarts or load decreases."

## Proposed fix

1. Add `cursor` (string) and `page_size` (int32) to `ListNotCompletedDeploymentsRequest` in `service.proto`, regenerate stubs.
2. In `PipedAPI.ListNotCompletedDeployments`, forward `req.Cursor` and `req.PageSize` into `datastore.ListOptions`.
3. In both piped stores, replace the single call with a loop that re-invokes the RPC with the returned cursor until it comes back empty, accumulating deployments across pages before classifying them into `pendings` / `planneds` / `runnings`.
4. Add table-driven tests for both stores against a fake `apiClient` that returns multiple pages.

Happy to split this into a proto/server PR and a piped-client PR if maintainers would prefer smaller diffs. Will sign commits with DCO.

## Environment

- Branch: `master` at commit `78fdbeb7c`
- Affected components: `piped`, `pipedv1`, `controlplane`

If a maintainer is willing to assign this to me, I would like to pick it up.

veyron-kairo:
**What this PR does**:

Makes `pipedv1`'s deployment store actually paginate when it calls `ListNotCompletedDeployments`. The store now loops on the cursor returned by the server until the cursor comes back empty, instead of taking only the first page and discarding the cursor.

**Why we need it**:

`pipedv1`'s deployment store was calling `ListNotCompletedDeployments` once per sync tick and ignoring the cursor on the response. If the datastore returned a partial page, anything past that boundary got dropped — those deployments looked stuck in `PENDING`/`PLANNED` until the overall volume dropped enough to fit in a single page. The TODO at the top of `sync()` had been there since the RPC was introduced.

**Scope — why `pipedv0` is not touched here**:

Issue #6696 calls out the same bug in `pkg/app/piped/apistore/deploymentstore/store.go`. `pipedv0` is on the deprecation path per [RFC 0015](https://github.com/pipe-cd/pipecd/blob/master/docs/rfcs/0015-pipecd-plugin-arch-meta.md) (support intended through end of 2025), so this PR is scoped to `pipedv1` to keep the diff small. Happy to mirror the loop into `pipedv0` in a follow-up if maintainers prefer it covered.

Fixes #6696

- **How are users affected by this change**: Deployments stop being silently skipped when the not-completed list spans more than one page. Anyone running `pipedv1` under heavier deployment volume should see every pending/planned/running deployment picked up each sync tick.


veyron-kairo:
@khanhtc1202 updated the patch at #6727 — went with the proto approach as originally proposed. Added cursor and page_size to the request, forwarded them server-side, and the pipedv1 store now loops until it gets an empty cursor. v0 store untouched. Let me know if anything needs changing!

mohammedfirdouss:
@sridhar-panigrahi please follow the PR template used for pipecd. It is hard to tell why this PR/Change is needed. Look at previous PRs for any idea.
Example of a PR template: https://github.com/pipe-cd/pipecd/pull/6724

veyron-kairo:
@mohammedfirdouss thanks for the nudge — you're right, the original description didn't make the motivation clear. I've rewritten it to follow the PR template (using #6724 as the reference). Let me know if it reads better now.

mohammedfirdouss:
> @mohammedfirdouss thanks for the nudge — you're right, the original description didn't make the motivation clear. I've rewritten it to follow the PR template (using #6724 as the reference). Let me know if it reads better now.

@sridhar-panigrahi This is good, but try and remove the extra details beneath the template "A bit more detail on the changes"

veyron-kairo:
@mohammedfirdouss got it — dropped the extra section, just the template now. Thanks for the review!

mohammedfirdouss:
> @mohammedfirdouss got it — dropped the extra section, just the template now. Thanks for the review!

Remove the sign-off as well, at the bottom.

veyron-kairo:
@mohammedfirdouss , removed the signoff part too .

veyron-kairo:
Just did a fresh self-review pass and pushed updates.

1. **Dropped `page_size` from the proto.** Nothing was reading it — the pipedv1 store only forwards the cursor — so adding it to the wire format was a speculative API surface cost. Easy to add back in a follow-up if a real use case appears.

2. **Tightened the head-deployment assertions** in the new test. While writing them I noticed the overwrite order in `sync()` makes a *running* deployment win over a *pending* one for the same application, which contradicts the `Lister` docstring ("head = oldest uncompleted"). Filed that separately as #6780 — not fixing it here so this PR stays focused on the pagination bug.

3. **Updated the PR description** to spell out why `pipedv0` is intentionally not touched (deprecation track per RFC 0015). Happy to mirror the loop into v0 inside this PR if anyone wants it covered together.

@khanhtc1202 @Warashi @ffjlabo @t-kikuc when one of you has a minute — would appreciate a maintainer look.

rahulshendre (commented):
Thanks for the PR @veyron-kairo 
@mohammedfirdouss would you mind reviewing it one last time when you get a chance?

veyron-kairo:
@mohammedfirdouss , please review this once you have Time .

mohammedfirdouss:
Looks good and should fix the pagination issue @veyron-kairo

mohammedfirdouss (dismissed):

mohammedfirdouss:
@khanhtc1202 PTAL

pkg/app/pipedv1/apistore/deploymentstore/store_test.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2024 The PipeCD Authors.
```
khanhtc1202: Should be 2026

khanhtc1202:
@veyron-kairo Good catch 👍 I left a nit comment in the test file header; please update it. Otherwise, it's LGTM 👍

pkg/app/pipedv1/apistore/deploymentstore/store_test.go
khanhtc1202:
```suggestion
// Copyright 2026 The PipeCD Authors.
```

khanhtc1202 (approved):
LGTM, thank you 👍

PR #6735, opened 2026-05-02
Fix: Orphaned process and goroutine leak in ScriptRun/CustomSync leading to split-brain deployments

### LINKED ISSUE #6734: Zombie processes and goroutine leak in ScriptRun / CustomSync executors on cancellation
**What happened:**
When a deployment is cancelled or times out during a ScriptRun, CustomSync, or Kubernetes rollback stage, the Piped agent stops tracking the stage but fails to terminate the underlying shell process. This results in orphaned "zombie" processes (e.g., kubectl, terraform, sleep) and leaked goroutines on the Piped agent host. These orphaned processes continue to run and mutate cluster state even after a rollback has been initiated, leading to potential data corruption and agent resource exhaustion.

**What you expected to happen:**
When a stage is cancelled or times out, the Piped agent should immediately and definitively terminate the underlying OS process tree associated with that stage's command. This ensures that no unauthorized mutations occur after cancellation and that agent resources are correctly reclaimed.

**How to reproduce it:**
1. Create a PipeCD application with a ScriptRun stage that executes a long-running command (e.g., `run: sleep 300`).
2. Trigger a deployment for this application.
3. Once the ScriptRun stage is active, click the "Cancel" button in the PipeCD web UI.
4. Observe the PipeCD UI and logs; the stage will be reported as `CANCELLED`.
5. Check the process list on the Piped agent host (e.g., `ps aux | grep sleep`). You will find that the sleep process is still running in the background.

**Environment:**
- piped version: master / latest
- control-plane version: master / latest
- Others: This is a platform-agnostic bug affecting any Piped agent utilizing ScriptRun or CustomSync executors.

pulkitvats2007-crypto:
**What this PR does**:
- Fixes goroutine and zombie process leaks in ScriptRun, CustomSync, and Kubernetes rollback executors.
- Binds external command execution to a cancellable context using `exec.CommandContext`.
- Ensures that timeout or cancellation properly terminates underlying OS processes.
- Adds `cmd.WaitDelay = time.Second` to prevent hangs caused by inherited pipes from background shell jobs.
- Updates all affected executors:
  - pkg/app/piped/executor/scriptrun/scriptrun.go
  - pkg/app/piped/executor/customsync/customsync.go
  - pkg/app/piped/executor/customsync/rollback.go
  - pkg/app/piped/executor/kubernetes/rollback.go
- Introduces tests (`TestExecuteCommandCancellation`) to verify proper cancellation behavior and prevent regressions.

**Why we need it**:
- Prevents concurrent deployment corruption caused by orphaned scripts continuing execution after cancellation.
- Eliminates goroutine leaks and zombie processes that can exhaust system resources (CPU, memory, file descriptors).
- Fixes silent failure scenarios where pipelines report "FAILED" or "CANCELLED" but scripts continue mutating infrastructure.
- Restores correctness and reliability of PipeCD’s cancellation and timeout mechanisms.

Fixes #6734 

- No

  - Users will experience correct and deterministic cancellation behavior.
  - Cancelled or timed-out stages will no longer continue executing in the background.
  - Improved system stability under frequent cancellations or retries.

  - No

  - N/A

### DIFF
--- a/pkg/app/piped/executor/customsync/customsync.go
+++ b/pkg/app/piped/executor/customsync/customsync.go
@@ -15,6 +15,7 @@
 package customsync
 
 import (
+	"context"
 	"os"
 	"os/exec"
 	"strings"
@@ -64,10 +65,12 @@ func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
 	e.appDir = ds.AppDir
 
 	timeout := e.StageConfig.CustomSyncOptions.Timeout.Duration()
+	ctx, cancel := context.WithCancel(sig.Context())
+	defer cancel()
 
 	c := make(chan model.StageStatus, 1)
 	go func() {
-		c <- e.executeCommand()
+		c <- e.executeCommand(ctx)
 	}()
 
 	timer := time.NewTimer(timeout)
@@ -94,7 +97,7 @@ func (e *deployExecutor) Execute(sig executor.StopSignal) model.StageStatus {
 	}
 }
 
-func (e *deployExecutor) executeCommand() model.StageStatus {
+func (e *deployExecutor) executeCommand(ctx context.Context) model.StageStatus {
 	opts := e.StageConfig.CustomSyncOptions
 
 	e.LogPersister.Infof("Runnnig commands...")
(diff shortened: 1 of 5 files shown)


armistcxy:
I think this one worth reviewing

After having a quick glance, I find that the code has already used `cmd.Run()` which is  equivalent to `cmd.Start()` + `cmd.Wait()` so maybe it won't create zombie process as you stated in PR description. I guess what you mean is orphan process like the PR title.

Maybe you should change the PR & issue description first

Nice catch ^^

armistcxy:
Reference: [Zombie process vs Orphan process](https://stackoverflow.com/questions/20688982/zombie-process-vs-orphan-process)

> Orphaned children are immediately "adopted" by `init` ... `init` automatically reaps its children (adopted or otherwise). So if you exit without cleaning up your children, then they will not become zombies (at least not for more than a moment).

rahulshendre:
+1 to @armistcxy
@pulkitvats2007-crypto, please change the PR and Issue description and then we can get it reviewed 
Thanks

PR #6747, opened 2026-05-07
[Multi_K8s-Plugin] Add config validation for KubernetesApplicationSpec

mohammedfirdouss:

Implements `KubernetesApplicationSpec.Validate()` in the multicluster Kubernetes plugin. The method was previously a stub that always returned `nil`, meaning invalid configs passed silently at load time and only failed mid-deployment with cryptic errors.

Two validations are added:

1. **`helmChart` and `kustomizeOptions` are mutually exclusive** — only one templating method can be used. Specifying both previously resulted in undefined/silent behaviour.

2. **Each `multiTargets[].target.name` must not be empty** — an empty target name causes the deploy target lookup to fail at runtime. This catches it before any stage runs.

Part of #6446

**How was this tested:**

Unit tests added to `config/application_test.go` covering:
- Empty spec (valid)
- Valid multiTargets
- `multiTargets` with an empty target name (error)
- `helmChart` + `kustomizeOptions` together (error)
- Each individually (valid)

**Note for single-cluster `kubernetes` plugin:**  [pkg/app/pipedv1/plugin/kubernetes/config/application.go](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/pipedv1/plugin/kubernetes/config/application.go) has the same `Validate()` stub (introduced in #5356). The same two validations apply there, plus the single-cluster plugin does not have `multiTargets` so only the helm/kustomize mutual exclusion check is relevant.

cc @Warashi you introduced the original stub in the single-cluster plugin, flagging in case you want to track this.


Warashi (approved):
LGTM

PR #6748, opened 2026-05-07
[Multi_K8s-Plugin] Prune orphaned resources on rollback

mohammedfirdouss:

Implements rollback pruning in the `kubernetes_multicluster` plugin. When a deployment pipeline fails mid-way, the ROLLBACK stage re-applies the last known good manifests but any new resource introduced by the failed deployment (e.g. a ConfigMap or Service already applied to the cluster) was never deleted, leaving orphaned resources that cause permanent live-state drift.

After applying the running manifests and cleaning up canary/baseline variants, rollback now:

1. Loads the target (failed) manifests from `TargetDeploymentSource`
2. Identifies resources present in target but absent from running (`findOrphanedKeys`)
3. Deletes those resources from the cluster (`pruneOrphanedResources`)

Errors during target manifest loading are non-fatal, logged and skipped so a Helm/Kustomize rendering failure cannot prevent the rollback from completing.

Part of #6446

**How was this tested:**

Unit tests added to `deployment/misc_test.go` for `findOrphanedKeys` covering:
- No orphans when target equals running
- All resources orphaned when running is empty
- New resource in target not present in running is orphaned
- Resource only in running is not pruned
- Multiple orphaned resources

**Note on single-cluster plugin:**

The same TODO exists in [`pkg/app/pipedv1/plugin/kubernetes/deployment/rollback.go`](https://github.com/pipe-cd/pipecd/blob/master/pkg/app/pipedv1/plugin/kubernetes/deployment/rollback.go). That plugin is out of scope for this PR.


pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go
```diff
@@ -210,8 +210,18 @@ func (p *Plugin) rollback(ctx context.Context, input *sdk.ExecuteStageInput[kube
...
+	if err != nil {
+		lp.Infof("Failed to load target app config for pruning, skipping: %v", err)
+	} else {
+		targetManifests, err := p.loadManifests(ctx, &input.Request.Deployment, targetCfg.Spec, &input.Request.TargetDeploymentSource, provider.NewLoader(toolRegistry), input.Logger, multiTarget)
+		if err != nil {
+			lp.Infof("Failed to load target manifests for pruning, skipping: %v", err)
+		} else {
+			pruneOrphanedResources(ctx, lp, applier, manifests, targetManifests)
+		}
+	}
```
Warashi: How about using the same logic with pruning resources in the K8S_SYNC stage?
i.e., not loading the target manifest but fetching the live resources to determine what to delete.
https://github.com/pipe-cd/pipecd/blob/215ea40debc0b37d294928927cbd49b531b65080/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/sync.go#L207-L230

mohammedfirdouss: Thanks for the pointer! That makes sense @Warashi. no need to depend on the failed manifests being renderable.

pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/misc_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/pipe-cd/pipecd/pkg/app/pipedv1/plugin/kubernetes_multicluster/provider"
 )
 
+
```
Warashi: Please remove this empty line?
It would be done by running gofmt.

pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go
```diff
@@ -210,8 +210,26 @@ func (p *Plugin) rollback(ctx context.Context, input *sdk.ExecuteStageInput[kube
...
-	// This occurs when the user adds a new resource and the deployment pipeline fails.
+	lp.Info("Start finding and pruning resources that no longer exist in the running manifests")
+	namespacedLiveResources, clusterScopedLiveResources, err := provider.GetLiveResources(ctx, kubectl, deployTargetConfig.KubeConfigPath, input.Request.Deployment.ApplicationID)
+	if err != nil {
+		lp.Errorf("Failed while getting live resources (%v)", err)
+		failed = true
+	} else if len(namespacedLiveResources)+len(clusterScopedLiveResources) > 0 {
+		removeKeys := provider.FindRemoveResources(manifests, namespacedLiveResources, clusterScopedLiveResources)
+		if len(removeKeys) == 0 {
+			lp.Info("There are no live resources to prune")
```
Warashi, line 222: It's good to log this info even when `len(namespacedLiveResources)+len(clusterScopedLiveResources) == 0`.

Warashi (dismissed):

pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go
mohammedfirdouss: Yeah exactly.

mohammedfirdouss:
@Warashi this need review.

pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/rollback.go
Warashi: @mohammedfirdouss This does not seem to be resolved. Please check?

mohammedfirdouss: Lint check failed for now...

mohammedfirdouss: I will add the log too

Warashi (approved):
LGTM

PR #6749, opened 2026-05-07
[Multi_K8s-Plugin] Add per-target kustomizeVersion and kustomizeOptions

mohammedfirdouss:

Adds `kustomizeVersion` and `kustomizeOptions` as per-target overrides inside each `multiTargets` entry in `KubernetesMultiTarget`, bringing the config schema closer to the RFC-0014 spec.

Previously, `kustomizeVersion` and `kustomizeOptions` were only available at the top-level `spec.input`, meaning all targets had to share the same kustomize version and options. This is a problem when deploying to clusters with different kustomize version requirements or different per-cluster flags (e.g. `load-restrictor`).

With this change, users can override these per target:

```yaml
input:
  multiTargets:
    - target:
        name: cluster-hoge
      kustomizeVersion: "5.4.3"
      kustomizeOptions:
        load-restrictor: "LoadRestrictionsNone"
    - target:
        name: cluster-fuga
      kustomizeVersion: "5.3.0"
```

The override logic follows the same pattern already used for `manifests` in `loadManifests`: the per-target value is used when non-empty, otherwise the top-level `spec.input` value is used.

Part of #6446

**How was this tested:**

Unit tests added to [`deployment/load_manifests_test.go`](https://github.com/pipe-cd/pipecd/blob/feat/multicluster-per-target-kustomize-overrides/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/load_manifests_test.go) using a capturing mock loader that records the `LoaderInput` passed to it, covering:
- Top-level values used when multiTarget has none
- Per-target version overrides top-level version
- Per-target options override top-level options
- Both version and options overridden together
- `nil` multiTarget falls back to top-level values


Warashi (approved):
LGTM

Issue #6528, opened 2026-02-11
FindContainerImages does not detect images in initContainers and CronJob jobTemplate


safiya2610:
### Title
FindContainerImages does not detect images in initContainers and CronJob jobTemplate

### Summary
The current implementation of FindContainerImages only extracts container images from:
spec.template.spec.containers.image

However, Kubernetes manifests can reference container images in additional valid fields which are currently not handled.

### Missing Fields
The function does not detect images from:
spec.template.spec.initContainers.image
spec.jobTemplate.spec.template.spec.containers.image (CronJob)
spec.jobTemplate.spec.template.spec.initContainers.image (CronJob)
There is already a TODO comment in the code mentioning this limitation.

### Impact
Because of this limitation:
Images used in initContainers are ignored.
CronJob images may not be detected.
Image tracking or diff logic relying on this function may be incomplete.
Manifest analysis becomes inaccurate for certain workloads.

### Suggested Improvement
Extend FindContainerImages to also extract images from:
spec.template.spec.initContainers.image
spec.jobTemplate.spec.template.spec.containers.image
spec.jobTemplate.spec.template.spec.initContainers.image

This can be implemented using the existing nestedStringSlice helper for consistency.

sahil5206:
hey @safiya2610 , can i take up this issue?

safiya2610:
@eeshaanSA Please comfirm this issue, so that I can start working on this,

A-d-i-t-y:
Hi team! 

I would love to work on this issue as part of the LFX Mentorship Program.

My Plan:
1. Find the FindContainerImages function in the codebase
2. Add support to also scan initContainers inside Pod specs
3. Add support to scan CronJob jobTemplate container images
4. Write or update existing tests to cover these new cases
5. Submit a PR with test results

Could you please assign this issue to me? 
Thank you!

PR #6762, opened 2026-05-11
fix: FindReferencingConfigMaps and FindReferencingSecrets to scan Cro…

### LINKED ISSUE #6761: FindReferencingConfigMaps and FindReferencingSecrets miss CronJob references
## What's the problem

While looking at the code after #6528 was fixed, I noticed that 
`FindReferencingConfigMaps` and `FindReferencingSecrets` have the same 
gap — they don't handle CronJobs.

Both functions only look through `spec.template.spec.*`, which works 
for Deployments, StatefulSets, and DaemonSets. But CronJobs put their 
pod template under `spec.jobTemplate.spec.template.spec.*`, so any 
ConfigMap or Secret referenced inside a CronJob is completely invisible 
to these functions.

## What breaks

If a CronJob mounts a ConfigMap as a volume or pulls credentials from a 
Secret via env, PipeCD won't know about it. That means drift detection 
won't catch config changes, and re-deploys won't be triggered when those 
configs change.

## Paths that are currently missed

`FindReferencingConfigMaps`:
- spec.jobTemplate.spec.template.spec.volumes.configMap.name
- spec.jobTemplate.spec.template.spec.containers.env.valueFrom.configMapKeyRef.name
- spec.jobTemplate.spec.template.spec.containers.envFrom.configMapRef.name
- spec.jobTemplate.spec.template.spec.initContainers.env.valueFrom.configMapKeyRef.name
- spec.jobTemplate.spec.template.spec.initContainers.envFrom.configMapRef.name

`FindReferencingSecrets`:
- spec.jobTemplate.spec.template.spec.volumes.secret.secretName
- spec.jobTemplate.spec.template.spec.containers.env.valueFrom.secretKeyRef.name
- spec.jobTemplate.spec.template.spec.containers.envFrom.secretRef.name
- spec.jobTemplate.spec.template.spec.initContainers.env.valueFrom.secretKeyRef.name
- spec.jobTemplate.spec.template.spec.initContainers.envFrom.secretRef.name

The fix is straightforward — add the missing `nestedStringSlice` probes 
for the jobTemplate paths, same pattern already used for the existing paths.

Affects both `kubernetes` and `kubernetes_multicluster` provider packages.

Ram04102007:
## What this PR does / why we need it

Fixes #6761

While looking at the code after #6528 was fixed, I noticed that
FindReferencingConfigMaps and FindReferencingSecrets have the same
gap — they don't handle CronJobs.

Both functions only walked spec.template.spec.* paths, which works for
Deployments, StatefulSets, and DaemonSets. CronJob wraps its pod template
under spec.jobTemplate.spec.template.spec.*, so all ConfigMap/Secret
references inside a CronJob were silently missed.

Added the missing nestedStringSlice probes for all jobTemplate paths
in both functions, across both kubernetes and kubernetes_multicluster
provider packages. 10 new test cases added, all passing.

## Does this PR introduce a user-facing change?

Yes — PipeCD now correctly detects ConfigMap/Secret references in CronJobs,
enabling proper drift detection and re-deploy triggering for CronJob workloads.


vedantlavale:
The PR also adds `initContainers` and `jobTemplate` paths to `FindContainerImages` (deployment.go lines ~32-49), but I don't see test cases covering those new probes. The PR description mentions "10 new test cases" but the diff shows only 8 (all for `FindReferencingConfigMaps/Secrets`). Should `FindContainerImages` tests be added to cover those paths?

Ram04102007:
Thanks for the review! You're right — I noticed the same thing. 
The deployment.go diff has the FindContainerImages changes mixed in 
from when I was looking at #6528, didn't mean to include those here.

I can either strip them out and keep this PR focused purely on 
FindReferencingConfigMaps/Secrets, or keep them and just add the 
missing FindContainerImages test cases to make it complete.

What works better for you?

vedantlavale:
I would have kept them and added the test cases, but since I’m a contributor, I think it’s better to get the maintainers’ opinion on this. @khanhtc1202 what do you think works best here?
cc : @rahulshendre @Ayushmore1214

PR #6768, opened 2026-05-13
Add docs for pipectl transfer command

### LINKED ISSUE #6691: Way to transfer PipeCD data across control plane

khanhtc1202:
**What this PR does**:

As PR title

**Why we need it**:

Fixes #6691


eeshaanSA (approved):
LGTM

@rahulshendre, please take a note. We can improve on these pages later.

PR #6781, opened 2026-05-17
[Multi_K8s-Plugin] Surface partial deletion errors in deleteResources

mohammedfirdouss:
**What this PR does**:
Changes `deleteResources` in `deployment/misc.go` to return `(int, error)` instead of just `int`. All deletion errors are now collected via `errors.Join` and returned to callers. `deleteVariantResources` propagates combined errors. Stage functions (`sync`, `primaryRollout`) now fail the stage if any resource deletion fails rather than silently succeeding with a partial count.

**Why we need it**:
Previously, when deleting multiple resources (during prune or variant cleanup), any non-`ErrNotFound` error was logged but swallowed callers received only the success count and logged "Successfully deleted X resources" with no signal that some resources failed. A partially-failed cleanup appeared as a success.

Fixes #6446

- **How are users affected by this change**: A deployment stage that previously reported success despite partial resource deletion failures will now correctly report failure, allowing the pipeline to retry or alert.
- **Is this breaking change**: No, this is a correctness fix. The deletion behaviour (best-effort, continues after individual failures) is unchanged.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/misc.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/misc.go
@@ -252,20 +252,48 @@ func deleteVariantResources(ctx context.Context, lp sdk.StageLogPersister, kubec
 		clusterScoped = append(clusterScoped, r.Key())
 	}
 
-	var deletedCount int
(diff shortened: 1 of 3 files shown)


Warashi (approved):
LGTM 👍🏻

PR #6782, opened 2026-05-17
[Multi_K8s-Plugin] Extract duplicated target-building pattern into shared helpers

mohammedfirdouss:
**What this PR does**:
Extracts the repeated ~45-line target-resolution and errgroup pattern from the four stage entry-point functions (`executeK8sMultiSyncStage`, `executeK8sMultiCanaryRolloutStage`, `executeK8sMultiBaselineRolloutStage`, `executeK8sMultiPrimaryRolloutStage`) into two package-level helpers in a new file `deployment/targets.go`:

- `buildStageTargets(lp, dts, multiTargets)` resolves which deploy targets a stage should run on (all targets if multiTargets is empty, or the named subset)
- `runOnTargets(ctx, lp, targets, fn)` fans out `fn` across targets in parallel via errgroup

**Why we need it**:
All four rollout/sync entry-point functions contained identical boilerplate: build a deploy-target map, resolve the target list, fan out with errgroup, handle errors. Any bug fix or behaviour change had to be applied in four places. The helpers eliminate the duplication while keeping behaviour identical.

Fixes #6446

- **How are users affected by this change**: No user-facing change. Just internal refactor identical behaviour.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/baseline.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/deployment/baseline.go
@@ -46,59 +46,11 @@ func (p *Plugin) executeK8sMultiBaselineRolloutStage(ctx context.Context, input 
 		}
 	}
 
-	type targetConfig struct {
-		deployTarget *sdk.DeployTarget[kubeconfig.KubernetesDeployTargetConfig]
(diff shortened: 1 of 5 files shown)


Warashi (approved):
Nice refactoring. LGTM.

PR #6783, opened 2026-05-17
[Multi_K8s-Plugin] Fix livestate loadManifests to stamp tracking labels and honour kustomizeDir

mohammedfirdouss:
**What this PR does**:
Fixes `loadManifests` in `livestate/plugin.go` to match what `deployment/plugin.go` already does:

1. Honours `kustomizeDir` from the per-target `multiTarget` config previously the field was ignored, so kustomize overlay directories set per-target had no effect in livestate.
2. Stamps the standard tracking labels and annotations (`LabelManagedBy`, `LabelPiped`, `LabelApplication`, `LabelCommitHash`, `LabelOriginalAPIVersion`, `LabelResourceKey`) on every loaded manifest previously these were missing, causing a divergence between the manifests the deployment plugin applies and the manifests livestate uses for drift detection.

**Why we need it**:
The deployment and livestate plugins use separate `loadManifests` implementations that had diverged. Missing labels in the livestate version meant drift detection compared against manifests that lacked the builtin PipeCD tracking labels, potentially causing false drift reports or missed drift. Missing `kustomizeDir` meant per-target kustomize overlays were silently ignored during livestate checks.

Fixes #6446

- **How are users affected by this change**: Livestate drift detection becomes more accurate, it now compares against manifests with the same labels and overlays that were actually applied.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/livestate/plugin.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/livestate/plugin.go
@@ -320,14 +320,15 @@ type loader interface {
 	LoadManifests(ctx context.Context, input provider.LoaderInput) ([]provider.Manifest, error)
 }
 
-// TODO: share this implementation with the deployment plugin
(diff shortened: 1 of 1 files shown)


Warashi (approved):
LGTM

PR #6784, opened 2026-05-17
[Multi_K8s-Plugin] Pin Is* methods to exact Kubernetes API groups

mohammedfirdouss:
**What this PR does**:
Tightens the API group checks in `provider/manifest.go`. The `Is*` methods previously used `isBuiltinAPIGroup()` which accepted any of ~20 builtin Kubernetes API groups, a Deployment-like CRD from a non-`apps` group would pass `IsDeployment()`. Each method is now pinned to its exact API group:

- `IsDeployment`, `IsStatefulSet`, `IsDaemonSet`, `IsReplicaSet` → group must be `"apps"`
- `IsService`, `IsSecret`, `IsConfigMap`, `IsPod` → group must be `""` (core)
- `IsWorkload` → applies the per-Kind group rule above

`isBuiltinAPIGroup` and the `builtinAPIGroups` map are now unused and have been removed.

**Why we need it**:
The loose check could misidentify a custom resource with the same Kind string but a different API group as a known Kubernetes workload or service, causing incorrect variant label injection or cleanup behaviour.

Fixes #6446

- **How are users affected by this change**: No change for standard Kubernetes resources. Users with CRDs that share a Kind name with a builtin resource (e.g. a CRD named `Deployment` in a non-`apps` group) will no longer have that resource treated as a workload which is the correct behaviour.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/config/application.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/config/application.go
@@ -109,8 +109,6 @@ type KubernetesDeploymentInput struct {
 	// Default is false.
 	AutoCreateNamespace bool `json:"autoCreateNamespace,omitempty"`
 
-	// TODO: Define fields for KubernetesDeploymentInput.
-
 	MultiTargets []KubernetesMultiTarget `json:"multiTargets,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


Warashi (approved):
Sorry for late, LGTM

PR #6803, opened 2026-05-21
[PipeCD-Web]: Fix livestate UI for non-k8s plugins

armistcxy:
**What this PR does**: 
- Render resource metadata in the livestate detail panel instead of hardcode k8s-specific fields (`Kind`, `Namespace`, `API Version`)
- Show `resourceType` as fallback label on resource node cards when `Kind` metadata is not present

**Why we need it**: currently, the livestate UI was built with only for k8s plugin, causing non-k8s plugins to display empty or meaningless fields in the resource detail panel and missing type labels on resource cards.

Fixes #


khanhtc1202 (approved):
LGTM

PR #6804, opened 2026-05-21
[ECS-Plugin]: Shorter taskset and task name showed in livestate panel

armistcxy:
**What this PR does**: Shorter taskset and task name showed in livestate panel

**Why we need it**: Full-length ID and ARN are verbose, we should use a shorter name and adequate to represent identify for taskset and task

Fixes #


khanhtc1202 (approved):
LGTM

PR #6807, opened 2026-05-22
[Multi_K8s-Plugin] Implement health status for DaemonSet, ReplicaSet and Pod

mohammedfirdouss:
**What this PR does**:
Implements health status calculation for DaemonSet, ReplicaSet, and Pod resource types in the kubernetes_multicluster plugin. Previously these three types all returned `ResourceHealthStateUnknown`. Adds `IsPod()` and `IsReplicaSet()` helper methods to `Manifest` to support the new cases.

**Why we need it**:
The single-cluster kubernetes plugin already handles DaemonSet, ReplicaSet, and Pod. This brings the multicluster plugin to parity.

Fixes #6446

- **How are users affected by this change**: Health status in the UI will now correctly show Healthy/Unhealthy for DaemonSet, ReplicaSet, and Pod resources instead of always showing Unknown.


Warashi (approved):
LGTM

PR #6808, opened 2026-05-22
[Multi_K8s-Plugin] Extract images from initContainers and CronJob templates

mohammedfirdouss:
**What this PR does**:
Extends `FindContainerImages` to also extract images from `initContainers` and from CronJob `jobTemplate` paths. Previously only `spec.template.spec.containers[*].image` was checked, so init container images and all CronJob images were silently ignored.

**Why we need it**:
Applications using init containers or CronJobs would have their images missing from version tracking and drift detection.

Fixes #6446

- **How are users affected by this change**: Version tracking and drift detection now correctly include images from initContainers and CronJobs.


Warashi (approved):
LGTM

PR #6809, opened 2026-05-22
[Multi_K8s-Plugin] Add unit tests for traffic routing config and liveresources

mohammedfirdouss:
**What this PR does**:
Adds unit tests for `config/traffic.go` (`DetermineKubernetesTrafficRoutingMethod`, `Percentages`, `DisplayString`) and a test for `provider/liveresources.go`. `provider/kubernetes.go` and `config/helm.go` contain only constants and struct definitions respectively, no test files are needed for those.

**Why we need it**:
These files had no test coverage. The traffic routing helpers are used in all canary/baseline traffic-splitting stages.

Fixes #6446

- **How are users affected by this change**: tests only.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/config/traffic_test.go
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/config/traffic_test.go
@@ -1,0 +1,156 @@
+// Copyright 2025 The PipeCD Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"testing"
+
+	"github.com/pipe-cd/piped-plugin-sdk-go/unit"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDetermineKubernetesTrafficRoutingMethod(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		cfg      *KubernetesTrafficRouting
+		expected KubernetesTrafficRoutingMethod
+	}{
+		{
+			name:     "nil config returns PodSelector",
+			cfg:      nil,
+			expected: KubernetesTrafficRoutingMethodPodSelector,
+		},
+		{
+			name:     "empty Method returns PodSelector",
+			cfg:      &KubernetesTrafficRouting{Method: ""},
+			expected: KubernetesTrafficRoutingMethodPodSelector,
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/app/pipedv1/plugin/kubernetes_multicluster/config/traffic_test.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2025 The PipeCD Authors.
```
khanhtc1202:
```suggestion
// Copyright 2026 The PipeCD Authors.
```
nits

pkg/app/pipedv1/plugin/kubernetes_multicluster/provider/liveresources_test.go
```diff
@@ -0,0 +1,44 @@
...
+// returns an error when the kubeconfig path does not exist / is unreachable.
+// This exercises the error path of kubectl.GetAll and confirms that
+// GetLiveResources propagates the failure rather than silently swallowing it.
+func TestGetLiveResources_InvalidKubeconfig(t *testing.T) {
+	t.Parallel()
+
+	kubectl := NewKubectl("kubectl")
+	_, _, err := GetLiveResources(context.Background(), kubectl, "/nonexistent/kubeconfig", "app-id-123")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "failed while listing all namespace-scoped resources")
```
Warashi: This may cause a different root of error depending on the host environment.
This test asserts the behavior for invalid config as expected when the host has kubectl installed.
Conversely, this test asserts the behavior for a command not found when the host has no kubectl installed.

Warashi (commented):
How about using toolregistrytest to test the Kubectl wrapper?

pkg/app/pipedv1/plugin/kubernetes_multicluster/provider/liveresources_test.go
mohammedfirdouss: Hmm so the idea is that the test should be environment independent. The host PATH cant be dependent on. Okay going to implement, thanks.

pkg/app/pipedv1/plugin/kubernetes_multicluster/provider/liveresources_test.go
```diff
@@ -12,33 +12,30 @@
...
 // TestGetLiveResources_InvalidKubeconfig verifies that GetLiveResources
 // returns an error when the kubeconfig path does not exist / is unreachable.
-// This exercises the error path of kubectl.GetAll and confirms that
-// GetLiveResources propagates the failure rather than silently swallowing it.
 func TestGetLiveResources_InvalidKubeconfig(t *testing.T) {
 	t.Parallel()
 
-	kubectl := NewKubectl("kubectl")
-	_, _, err := GetLiveResources(context.Background(), kubectl, "/nonexistent/kubeconfig", "app-id-123")
+	registry := toolregistry.NewRegistry(toolregistrytest.NewTestToolRegistry(t))
```
mohammedfirdouss, line 33: @Warashi wdyt?

Warashi: Yes, this is exactly what I meant.

Warashi (approved):
LGTM

PR #6880, opened 2026-05-29
[Multi_K8s-Plugin] rename stages to avoid conflicts with kubernetes plugin

mohammedfirdouss:
**What this PR does**: Renames the Kubernetes multicluster plugin stage names to include a `MULTI_` prefix, avoiding collisions with the regular Kubernetes plugin stage names.

**Why we need it**: A Piped instance cannot have two plugins registering the same stage name. The multicluster plugin previously used identical names to the Kubernetes plugin (`K8S_PRIMARY_ROLLOUT`, `K8S_CANARY_ROLLOUT`, etc.), which would cause routing conflicts when both plugins are loaded.

The following stages are renamed:
- `K8S_PRIMARY_ROLLOUT` → `K8S_MULTI_PRIMARY_ROLLOUT`
- `K8S_CANARY_ROLLOUT` → `K8S_MULTI_CANARY_ROLLOUT`
- `K8S_CANARY_CLEAN` → `K8S_MULTI_CANARY_CLEAN`
- `K8S_BASELINE_ROLLOUT` → `K8S_MULTI_BASELINE_ROLLOUT`
- `K8S_BASELINE_CLEAN` → `K8S_MULTI_BASELINE_CLEAN`
- `K8S_TRAFFIC_ROUTING` → `K8S_MULTI_TRAFFIC_ROUTING`

**Which issue(s) this PR fixes**: Fixes #

- **How are users affected by this change**: Users need to update their `app.pipecd.yaml` pipeline stage names to use the new `K8S_MULTI_*` prefixed names.
- **Is this breaking change**: Yes
- **How to migrate (if breaking change)**: Replace the old stage names with the new prefixed equivalents in all `app.pipecd.yaml` files that use the Kubernetes multicluster plugin.


mohammedfirdouss:
@Warashi check here, thanks.

Warashi (approved):
LGTM, Thank you

PR #6881, opened 2026-05-30
[Docs] Expand Concepts page with additional PipeCD terminology

ObaidAbdullah16:
**What this PR does**:

Adds definitions for several commonly used PipeCD concepts and features to the Concepts page, along with links to the relevant documentation.

**Why we need it**:

Some terms are documented across different pages but are not explained in the Concepts section. Adding them here makes it easier for users and contributors to understand PipeCD terminology in one place.

Fixes #

* **How are users affected by this change**: Users can more easily find explanations for PipeCD concepts and related features.
* **Is this breaking change**: No
* **How to migrate (if breaking change)**: N/A

**docs-dev/concepts/_index.md :**

https://github.com/user-attachments/assets/c5b3d66e-c292-4346-966e-56810fbe85cc

**docs-v1.0.x/concepts_index.md :**

https://github.com/user-attachments/assets/092276d6-0397-40de-a075-cee5a481cc75


docs/content/en/docs-dev/concepts/_index.md
```diff
@@ -73,3 +73,59 @@ Currently, PipeCD is supporting these five platform providers: `KUBERNETES`, `EC
...
+`pipectl` allows users to add and sync applications, get deployment status, encrypt secrets, and more from the terminal.
+See [Command-line tool: pipectl](/docs-dev/user-guide/command-line-tool/) for usage information.
+
+### Plan Preview
+A feature that previews the expected changes of a deployment before it is actually applied.
+When integrated with a CI system such as GitHub Actions, Plan Preview posts a comment on pull requests showing which resources would be added, modified, or deleted.
+See [Plan preview](/docs-dev/user-guide/plan-preview/) for details.
+
+### Event Watcher
+A `piped` feature that watches for external events — such as a new container image pushed to a registry or a new Helm chart version published — and automatically updates files in the Git repository to trigger a new deployment.
```
Ayushmore1214: @ObaidAbdullah16  just a question :why em dashes are used here ?

docs/content/en/docs-dev/concepts/_index.md
```diff
@@ -73,3 +73,59 @@ Currently, PipeCD is supporting these five platform providers: `KUBERNETES`, `EC
...
+See [Deployment Trace](/docs-dev/user-guide/deployment-trace/) for details.
+
+### Deployment Chain
+A feature that allows triggering a sequence of deployments across multiple applications automatically.
+When a deployment in the chain completes successfully, it triggers the next deployment in the chain.
+See [Deployment chain](/docs-dev/user-guide/managing-application/deployment-chain/) for details.
+
+### Insights
+The analytics dashboard of PipeCD that displays delivery performance metrics such as deployment frequency, lead time, mean time to restore (MTTR), and change failure rate.
+See [Insights](/docs-dev/user-guide/insights/) for details.
```
Ayushmore1214: Did you verified the links before pushing this one says file not found

Ayushmore1214 (changes requested):
@ObaidAbdullah16 none of the links seems to be working here, can you run this locally and verify once ?

docs/content/en/docs-dev/concepts/_index.md
ObaidAbdullah16: I used em dashes as a stylistic choice to set off the example clause, but I can see how it looks inconsistent. I've replaced them with commas to keep it simpler.

docs/content/en/docs-dev/concepts/_index.md
ObaidAbdullah16: The links were using absolute paths (/docs-dev/user-guide/...) but the existing file uses relative paths (../user-guide/...). That's what broke them locally. Fixed all of them to use relative paths for consistency and ran it locally to confirm they work now.

Ayushmore1214: Can you share the screen recording or screenshot or them working, this is mandatory for all docs and site based PR's thanks!

ObaidAbdullah16: I have uploaded the screen recording in the pr description

rahulshendre (commented):
This only updates docs-dev, can you also add these to docs-v1.0.x/concepts/_index.md? 
@Ayushmore1214, please drop your review as well : )

ObaidAbdullah16:
I've added the same concept entries to docs-v1.0.x/concepts/_index.md. I couldn't copy everything exactly from docs-dev because a few of the referenced pages aren't available in the v1.0.x docs yet (for example, the deployment customization docs). So I updated those entries to match the existing v1.0.x documentation and only kept links that are available in that version.

docs/content/en/docs-dev/concepts/_index.md
```diff
@@ -73,3 +73,60 @@ Currently, PipeCD is supporting these five platform providers: `KUBERNETES`, `EC
...
+PipeCD continuously monitors live state and displays it on the web console, including resource status, health, and the relationship between resources (e.g., Deployment → ReplicaSet → Pod).
+See [Application live state](../user-guide/managing-application/application-live-state/) for details.
+
+### Secret Management
+PipeCD provides a mechanism for safely managing secrets used in application manifests.
+Secrets are encrypted using `pipectl` and stored in the Git repository in encrypted form. `piped` decrypts them at deployment time, so plain-text secrets never leave the deployment environment.
+See [Secret management](../user-guide/managing-application/secret-management/) for details.
+
+### Plugin
+In PipeCD v1's next-generation agent architecture (`pipedv1`), each deployment platform (Kubernetes, Terraform, ECS, etc.) is implemented as a separate plugin.
```
rahulshendre, line 132: @ObaidAbdullah16 docs-dev tracks v0, which doesn't have the plugin model. The v1 concepts page already has a plugins section, drop this plugin section from the docs-dev

rahulshendre (changes requested):
sorry for the delay @ObaidAbdullah16, thanks for the work done here.

reading through it though, I think most of these already have their own pages (Plan Preview, Event Watcher, Deployment Trace, Insights, Secret Management, Launcher, etc.), so defining them again in the concepts page mostly duplicates those intros - which essentially means we will now need to maintain two places to keep in sync as the features change, the concepts page is meant more for the terms.

The one that fits and is currently missing is Stage, so I'd suggest we trim this down to that (and any other truly core point you feel is missing), and drop the feature summaries, keeping concepts focused on what they need to be

Let me know what you think 👀

rahulshendre:
cc @Ayushmore1214 @harshitghagre please drop your review as well

harshitghagre:
agree with @rahulshendre most of these already link to their own pages so short entry plus link is enough

plugin should probably just be dropped from docs-dev since it tracks v0 and v0 has no plugin model v1 concepts page already covers it launcher also reads a bit different between the two files might be worth syncing that too

stage is fine as is thats the one actually missing elsewhere

rahulshendre:
cc @ObaidAbdullah16

rahulshendre:
cc @ObaidAbdullah16

PR #6901, opened 2026-06-07
ci: add read-only permissions to validation workflows

Yashagarwal9798:
**What this PR does**:

Adds top-level read-only GitHub token permissions to validation workflows:

- `.github/workflows/build.yaml`
- `.github/workflows/test.yaml`
- `.github/workflows/gen.yaml`
- `.github/workflows/build_tool.yaml`

**Why we need it**:

These workflows only need read access to checkout the repository and run validation jobs. Setting `permissions: contents: read` makes the token scope explicit and follows least-privilege security.

**Which issue(s) this PR fixes**:

Fixes #6899 

**Does this PR introduce a user-facing change?**:

No.

- **How are users affected by this change**:
  Users are not affected. This only changes GitHub Actions permissions.
- **Is this breaking change**:
  No.
- **How to migrate (if breaking change)**:
  No migration is needed.

PR #6730, opened 2026-04-29
Pin actions/github-script to commit SHA in first-time-contributor workflow

harshitghagre:
**What this PR does:**

Pins `actions/github-script` from the mutable tag `@v7` to its commit SHA
`f28e40c7f34bde8b3046d885e986cb6290c5673b` in the `first-time-contributor.yaml`
workflow. This follows the same convention used by other actions in this repo.

**Why we need it:**

The workflow currently fails for every first-time contributor because
`actions/github-script@v7` is not in the repository's allowed GitHub Actions list.
Pinning to a commit SHA is a prerequisite for adding it to the allowed list, and
also improves security by preventing supply-chain attacks via mutable tags.

**Which issue(s) this PR fixes:**

Fixes #6726

**Does this PR introduce a user-facing change?**:

- **How are users affected by this change**: The "First-time contributor welcome"
  bot will start working again for new contributors once the SHA is added to the
  allowed list.
- **Is this breaking change**: No
- **How to migrate (if breaking change)**: N/A

PR #6929, opened 2026-06-21
ci: fix first-time contributor welcome check

### LINKED ISSUE #6918: ci: fix first-time contributor welcome check
**What happened**:

The `First-time contributor welcome` workflow did not post a welcome comment on a user's first pull request.

For example, PR #6901 was the author's first pull request in this repository, and the workflow ran for that PR. But no welcome comment was posted.

The workflow tries to list pull requests by the PR author using:

```js
creator: prAuthor
```

However, `creator` is not a supported parameter for `github.rest.pulls.list`. Because of that, GitHub ignores it and returns the latest pull requests in the repository instead of pull requests created by that user.

Since the workflow requests `per_page: 2`, it can receive two unrelated pull requests and then incorrectly decide that the user is not a first-time contributor.

When a user opens their first pull request in this repository, the workflow should detect that correctly and post the welcome comment.

**How to reproduce it**:

1. Open a pull request from a user who has no earlier pull requests in this repository.
2. Wait for the `First-time contributor welcome` workflow to run.
3. Check the pull request conversation.
4. The welcome comment is not posted, even though it is the user's first pull request.

**Environment**:
- `piped` version: Not related.
- `control-plane` version: Not related.
- Others: GitHub Actions workflow `.github/workflows/first-time-contributor.yaml`.

Yashagarwal9798:
**What this PR does**:

Fixes the first-time contributor welcome workflow to correctly check pull requests opened by the PR author.
It now uses GitHub search with `author:<username>` instead of the unsupported `creator` parameter.

**Why we need it**:

The old check could look at unrelated recent pull requests and incorrectly skip the welcome comment.
This ensures first-time contributors get the welcome message on their first pull request.

Fixes #6918

No.

  Users are not affected. This only changes a GitHub Actions workflow.
  No.
  No migration is needed.


Yashagarwal9798:
@Warashi @Ayushmore1214 @rahulshendre Could you please review this PR?

Ayushmore1214:
@Yashagarwal9798 could you run this once locally and drop a screenshot here ?

Yashagarwal9798:
@Ayushmore1214 I tested it in my fork by opening a test PR from `test-first-time-welcome` into `fix-first-time-contributor-welcome`.

Since the fixed workflow was on the base branch, the `pull_request_target` workflow ran and posted the first-time contributor welcome comment successfully.

image
image

Ayushmore1214 (approved):
Thanks LGTM!

Warashi (commented):
@Yashagarwal9798 
The pull_request_target workflow runs on the default branch, not the PR's base branch.
Additionally, your fork repository has no PRs other than your test PR. So it's not enough to test this workflow, because `pulls.length` is less than or equal to `1` regardless of the author.

harshitghagre:
Duplicate of #6730 
cc: @Ayushmore1214 @Warashi @Yashagarwal9798

Ayushmore1214:
@harshitghagre your PR touches the SHA tags and this is about the core logic of the worklfow, so its not a duplicate as far as I unserstand

Ayushmore1214:
If you think there is some overlapping go ahead point it out specifically, thanks!

harshitghagre:
@Ayushmore1214 You're right, calling it a duplicate wasn't correct. I just meant there might be some overlap, not that both PRs solve the same thing

Yashagarwal9798:
@Ayushmore1214 @Warashi Thanks, you're right my fork only had the one test PR, so pulls.length <= 1 would pass there for any author. That test didn't really prove anything.

Since both API calls are read-only, I ran them directly against this repo instead (screenshot below):

- The current workflow's call with creator=yashagarwal9798 returns the repo's 2 latest PRs by other authors the param is silently ignored because GET /repos/{owner}/{repo}/pulls doesn't support creator ([docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests) it only exists on the [issues list endpoint](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues)). So pulls.length > 1 is always true here and the welcome is always skipped, which is what happened in #6901.

- The search call from this PR returns only my PRs with the correct total_count ([docs](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests)).

So a real first-time contributor gets total_count == 1 and the comment is posted.

Let me know if you'd like any other verification.
image

Warashi (approved):
Thank you! LGTM.

Yashagarwal9798:
happy to help!!

PR #6933, opened 2026-06-24
Update extract cookie logic to allow cookie contains equal sign

khanhtc1202:
**What this PR does**:

Update the extract cookie logic to allow cookie contains equal sign `=`

**Why we need it**:

Some cookies introduced by 3rd providers contain the `=` sign in their body. The body of the pipecd signed cookie remains, as it does not contain the equal sign (spec from jwt/v5), but currently, the pipecd auth logic checks all the cookie bodies that include cookies from 3rd providers and treats them as pipecd signed cookies.

Also, there was already a logic that verifies the token content, which will yield an error if the token string body contains an equal sign `=`, so no need to overdo it here.

ref: https://github.com/pipe-cd/pipecd/blob/master/pkg/jwt/verifier.go#L44

Fixes #

- **How are users affected by this change**: Be able to use cookie from 3rd auth providers that contains equal sign `=`

### DIFF
--- a/pkg/rpc/rpcauth/auth.go
+++ b/pkg/rpc/rpcauth/auth.go
@@ -136,10 +136,7 @@ func extractCookie(ctx context.Context) (map[string]string, error) {
 	cs := strings.Split(rawCookie[0], ";")
 	cookie := make(map[string]string, len(cs))
 	for _, c := range cs {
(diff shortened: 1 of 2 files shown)


t-kikuc (approved):
i got it

PR #6936, opened 2026-06-24
Use %w for error wrapping in controller.cancelDeployment

### LINKED ISSUE #6932: fix: for error wrapping in cancelDeployment
### What happened?

In `pkg/app/piped/controller/controller.go`, the `cancelDeployment` function formats an error using `%v` instead of `%w` when wrapping an underlying error returned from `ReportDeploymentCompleted`.

This causes the original error to be stringified rather than properly wrapped, breaking Go error chaining. As a result, callers cannot use `errors.Is()` or `errors.As()` to inspect or match the underlying error.

In contrast, other functions in the same file (e.g. `reportApplicationDeployingStatus`) correctly use `%w` for error wrapping.

---

### Affected code location

**File:** `pkg/app/piped/controller/controller.go`

**`cancelDeployment` (~line 650):**

```go
err = fmt.Errorf("failed to report deployment status to control-plane: %v", err)

### PR DESCRIPTION
**Pull request by mvanhorn**
## Summary

`cancelDeployment` now wraps the error returned by `ReportDeploymentCompleted` with `%w` instead of `%v`, so callers can use `errors.Is`/`errors.As` to inspect the underlying cause.

## Why this matters

The cancellation path formatted the reporting error with `%v`, which flattens it to a string and breaks error-chain inspection for any caller that wants to branch on the wrapped error. Switching to `%w` preserves the chain while keeping the existing message format. Reported in #6932.

## Testing

Added a regression test that drives `cancelDeployment` with a fake API client returning a sentinel error and asserts `errors.Is` matches the wrapped error, so the test fails if the wrapping regresses to `%v`. The controller package tests pass.

Closes #6932

### CODE REVIEW

**Review by rahulshendre — COMMENTED** [review]
@khanhtc1202, this looks good from the docs side and the fix also seems great
PTAL when you get time 
Thanks a lot @mvanhorn : )

**Comment by armistcxy** [comment]
LGTM, but the returned error only used for logging like this so I don't think this one is important, just good practice stuffs

```go
		if cancel {
			if err = c.cancelDeployment(ctx, d, cancelReason); err != nil {
				c.logger.Error("failed to cancel deployment",
					zap.String("deployment", d.Id),
					zap.String("app", d.ApplicationId),
					zap.Error(err),
				)
			}
			continue
		}
```

Also the change has been made in controller of piped (v0). It's not a critical one

FYI  @khanhtc1202

khanhtc1202 (approved):
LGTM

PR #6940, opened 2026-06-26
ci: add timeout-minutes to scheduled workflow jobs

### LINKED ISSUE #6902: ci: add timeout-minutes to scheduled workflow jobs
**What would you like to be added?**

Some scheduled GitHub Actions workflow jobs should define `timeout-minutes`.

These scheduled workflow jobs currently do not define one:

- `.github/workflows/codeql-analysis.yaml`: `analyze`
- `.github/workflows/stale.yaml`: `stale`

I suggest adding explicit timeouts to these jobs, for example:

```yaml
timeout-minutes: 60
```

For `stale.yaml`, a shorter timeout such as `15` minutes may be enough because it only runs `actions/stale`.

The repository already uses explicit job timeouts in other workflows, such as `.github/workflows/build_tool.yaml` and `.github/workflows/publish_tool.yaml`.

**Why is this needed?**

Scheduled workflows run without direct user interaction, so a stuck job can occupy a runner until the default GitHub Actions timeout is reached.

Adding explicit timeouts makes these jobs fail faster when they hang and keeps runner usage bounded.

kunal-yelgate:
**What this PR does:**

Adds explicit `timeout-minutes` to scheduled GitHub Actions workflow jobs that currently don't define one.

| Workflow | Job | Timeout | Reasoning |
|---|---|---|---|
| `.github/workflows/codeql-analysis.yaml` | `analyze` | `60` | CodeQL analysis can take a while; 60 minutes is reasonable |
| `.github/workflows/stale.yaml` | `stale` | `15` | Only runs `actions/stale`, which is typically fast |

**Why we need it:**

Scheduled workflows run without direct user interaction, so a stuck job can occupy a runner until the default GitHub Actions 6-hour timeout is reached. Adding explicit timeouts makes these jobs fail faster when they hang and keeps runner usage bounded.

The repository already uses explicit job timeouts in other workflows:
- `.github/workflows/build_tool.yaml` (`timeout-minutes: 15`)
- `.github/workflows/publish_tool.yaml` (`timeout-minutes: 30`)

This change follows the same existing pattern.

Fixes #6902

- **How are users affected by this change**: No impact.


Ayushmore1214 (commented):
@Yashagarwal9798 @Warashi can you help review this one ?

.github/workflows/stale.yaml
```diff
@@ -1,24 +1,25 @@
...
     runs-on: ubuntu-24.04
+    timeout-minutes: 15
     steps:
       - uses: actions/stale@v10
         with:
           operations-per-run: 1000
           # Issues
-          stale-issue-message: 'This issue is stale because it has been open 180 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
-          close-issue-message: 'This issue was closed because it has been stalled for 7 days with no activity. Feel free to reopen if still applicable.'
+          stale-issue-message: "This issue is stale because it has been open 180 days with no activity. Remove stale label or comment or this will be closed in 7 days."
```
Ayushmore1214: @kunal-yelgate One small suggestion: could we avoid the unrelated formatting changes (switching single quotes to double quotes)? I mean this is completely unecessary (it is right ?)

kunal-yelgate: @Ayushmore1214  you’re right. I have fixed it back to the previous format and removed the unnecessary formatting changes.

mohammedfirdouss:
@kunal-yelgate This case needs consideration, you can raise in next biweekly meeting

PR #6977, opened 2026-07-03
[Multi_K8s-Plugin] Update plugin README to release standard

mohammedfirdouss:
**What this PR does**:

Updates the `kubernetes_multicluster` plugin README to match the same standard as other plugin README.

- Adds **How to Build & Run** section with Makefile commands, `--force-plugin-redownload` note, and local kind cluster setup steps
- Adds **How to Test** section explaining how to run unit tests and the role of envtest
- Adds **Directory Structure** overview
- Adds full **Config Reference** tables for all fields: piped config, deploy target config, application config, all 8 stages, Helm/Kustomize options, traffic routing (PodSelector + Istio), and patch operations
- Removes the stale "only QuickSync is supported" note
- Removes the "url is a dummy" comment (no longer accurate)
- Updates all stage names to the `K8S_MULTI_*` prefix introduced in #6880

**Why we need it**:

The plugin README was a rough developer guide written during the mentorship term. Before a binary release can be triggered (via the `plugin_release` workflow), the README should be accurate and useful for contributors and users — matching the standard set by the ECS plugin.

Fixes #

- **How are users affected by this change**: Documentation only — no code changes.

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
@@ -5,7 +5,7 @@
 The Kubernetes Multicluster plugin supports deploying applications across multiple Kubernetes clusters from a single pipeline definition. It handles sync state, traffic routing, and rollback across all target clusters.
 
 > [!CAUTION]
-> The configuration format is unstable and may change in the future.
+> Currently, this is alpha status.
 
 ### Quick sync
 
--- a/pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
+++ b/pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
@@ -1,29 +1,106 @@
 # Kubernetes Multicluster Plugin
 
-## Specification
+## Overview
 
-The current specification is described in the [RFC](../../../../../docs/rfcs/0014-multi-cluster-deployment-for-k8s.md).  
-The configuration format is unstable and may change in the future.
+The Kubernetes Multicluster plugin supports deploying applications across multiple Kubernetes clusters from a single pipeline definition. It handles sync state, traffic routing, and rollback across all target clusters.
 
-Note:
-- Currently, only QuickSync is supported.
+> [!CAUTION]
+> The configuration format is unstable and may change in the future.
 
-## Try k8s multicluster plugin locally
+### Quick sync
 
-**Switch to the upstream commit of the master branch**
+Quick sync applies manifests to all target clusters immediately, without a pipeline.
 
-The whole fixes for plugins are stored in the master branch and some of the versions don't have them.
+It will be planned in one of the following cases:
+- no pipeline was specified in the application configuration file
+- the sync was triggered manually with the quick sync option
 
-```sh
-git switch master
-git pull
+### Pipeline sync
+
+You can configure a pipeline to deploy progressively across clusters with stages like canary rollout, traffic routing, primary rollout, and baseline comparison.
+
+The plugin defines three variants for each application: primary, canary, and baseline.
+- `primary` runs the current stable version.
+- `canary` runs the proposed change and receives a configurable percentage of traffic.
+- `baseline` runs the same version as primary, used as a comparison target for canary analysis.
+
+Once deployment completes, only the `primary` variant remains across all clusters.
+
+These are the stages provided by the Kubernetes Multicluster plugin:
+
+- `K8S_MULTI_SYNC`: sync all target clusters with the current manifests (used for quick sync)
+- `K8S_MULTI_PRIMARY_ROLLOUT`: roll out the new version as the PRIMARY variant to all target clusters
+- `K8S_MULTI_CANARY_ROLLOUT`: roll out the new version as the CANARY variant to all target clusters
+- `K8S_MULTI_CANARY_CLEAN`: remove CANARY variant resources from all target clusters
+- `K8S_MULTI_BASELINE_ROLLOUT`: roll out the current version as the BASELINE variant to all target clusters
+- `K8S_MULTI_BASELINE_CLEAN`: remove BASELINE variant resources from all target clusters
+- `K8S_MULTI_TRAFFIC_ROUTING`: route traffic between PRIMARY, CANARY, and BASELINE variants across all target clusters
+- `K8S_MULTI_ROLLBACK`: rollback all target clusters to the previous state (automatically added when `autoRollback` is enabled)
+
+## Directory Structure
+
+```
+kubernetes_multicluster/
+├── main.go         # Plugin entrypoint
+├── config/         # Configuration types for piped, deploy targets, application, and stages
+├── deployment/     # Stage execution and pipeline planning
+├── provider/       # Kubernetes API client, manifest loading, kubectl/kustomize/helm wrappers
+├── livestate/      # Live state fetcher (reports current resource state across clusters)
+├── planpreview/    # Plan preview (shows diff before deployment)
+├── toolregistry/   # Tool registry for kubectl, kustomize, and helm binaries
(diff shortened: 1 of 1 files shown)


mohammedfirdouss:
@rahulshendre @eeshaanSA @khanhtc1202 please review when free

Ayushmore1214 (dismissed):
it feels a little long, but honestly that's mostly repetition rather than too much info. The kind: Piped block shows up like 2 times; maybe we keep one good annotated copy and point the other spots to it? good work though!!!

rahulshendre (commented):
checked all the tables against the config structs, fields and defaults match, and the structure looks good as well, main ones are the kubeconfig order and the test command, rest are nits. 
really nice work on the tables @mohammedfirdouss : D

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
```diff
@@ -1,159 +1,419 @@
...
+```
+kubernetes_multicluster/
+├── main.go         # Plugin entrypoint
+├── config/         # Configuration types for piped, deploy targets, application, and stages
+├── deployment/     # Stage execution and pipeline planning
+├── provider/       # Kubernetes API client, manifest loading, kubectl/kustomize/helm wrappers
+├── livestate/      # Live state fetcher (reports current resource state across clusters)
+├── planpreview/    # Plan preview (shows diff before deployment)
+├── toolregistry/   # Tool registry for kubectl, kustomize, and helm binaries
+└── example/        # Example application configurations
```
rahulshendre, line 52: nit: docs/ is missing here 👀

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
```diff
@@ -1,159 +1,419 @@
...
+Refer to [cmd/pipecd/README.md](../../../../../cmd/pipecd/README.md) to set up the PipeCD control plane, and [cmd/pipedv1/README.md](../../../../../cmd/pipedv1/README.md) to run piped locally.
+
+## How to Test
+
+### Unit tests
+
+Run unit tests from the repository root:
+
+```bash
+go test ./pkg/app/pipedv1/plugin/kubernetes_multicluster/...
```
rahulshendre: I think this won't run from the repo root since the plugin has its own go.mod, and the envtest tests need KUBEBUILDER_ASSETS
make test/go MODULES=pkg/app/pipedv1/plugin/kubernetes_multicluster is the one that works, ref - [Makefile#L102](https://github.com/pipe-cd/pipecd/blob/b3a79866e0020c5e219b34ceac72df3bbeda11ce/Makefile#L102-L131)

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
```diff
@@ -1,159 +1,419 @@
...
+```
+
+Unit tests use [envtest](https://book.kubebuilder.io/reference/envtest) to spin up a real Kubernetes API server in memory. No external cluster is needed.
+
+### Integration tests
+
+To test against real clusters, build and run piped with the plugin pointing to real clusters. The plugin resolves kubeconfig in this order:
```
rahulshendre: masterURL isn't actually used anywhere in the plugin, only kubeConfigPath gets passed to kubectl, ref - [applier.go#L58](https://github.com/pipe-cd/pipecd/blob/b3a79866e0020c5e219b34ceac72df3bbeda11ce/pkg/app/pipedv1/plugin/kubernetes_multicluster/provider/applier.go#L58)
So this resolution order doesn't really exist, I think we should just say kubeConfigPath is passed to kubectl, and empty falls back to kubectl's own defaults

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
```diff
@@ -1,159 +1,419 @@
...
+      - name: cluster2
+        config:
+          masterURL: https://127.0.0.1:62082
+          kubeConfigPath: /path/to/kubeconfig/for/cluster2
 ```
 
-### app.pipecd.yaml
+|-|-|-|-|
+| deployTargets | [][DeployTargetConfig](#deploytargetconfig) | List of Kubernetes clusters to deploy to | Yes |
```
rahulshendre: there's also a config block next to deployTargets [chartRepositories + chartRegistries](https://github.com/pipe-cd/pipecd/blob/b3a79866e0020c5e219b34ceac72df3bbeda11ce/pkg/app/pipedv1/plugin/kubernetes_multicluster/config/plugin.go#L23-L28), [Initialize()](https://github.com/pipe-cd/pipecd/blob/b3a79866e0020c5e219b34ceac72df3bbeda11ce/pkg/app/pipedv1/plugin/kubernetes_multicluster/main.go#L35) uses it for helm repo add and oci login. 
I think we should add tables for that too

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
```diff
@@ -1,159 +1,419 @@
...
+        ...
+  pipeline:
+    stages:
+      ...
 ```
+
+|-|-|-|-|
+| input | [KubernetesDeploymentInput](#kubernetesdeploymentinput) | Input for deployment such as manifests, kubectl version, helm/kustomize options. | No |
+| quickSync | [K8sSyncStageOptions](#k8ssyncstageoptions) | Options for the quick sync stage. | No |
```
rahulshendre: this anchor doesn't go anywhere, I don't think there is K8sSyncStageOptions heading on this page - [try it](https://github.com/mohammedfirdouss/pipecd/blob/feat/multicluster-plugin-readme-update/pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md#k8ssyncstageoptions). 
It should be [#k8s_multi_sync](https://github.com/mohammedfirdouss/pipecd/blob/feat/multicluster-plugin-readme-update/pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md#k8s_multi_sync), what do you think?

rahulshendre:
> it feels a little long, but honestly that's mostly repetition rather than too much info. The kind: Piped block shows up like 2 times; maybe we keep one good annotated copy and point the other spots to it? good work though!!!

good catch @Ayushmore1214
yeah the repetition is there, but I don't think we should keep just one copy. the Run block is a walkthrough, someone following it just needs enough to get the plugin loaded, and Piped Config is the reference people jump to directly, so that one should stay complete.
I think we should trim the Run block down to name, port, url and a bare deployTargets, with a small "see Plugin Configuration below for all fields" link 👀

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
mohammedfirdouss: Alright, in a bit

mohammedfirdouss:
@Ayushmore1214 thanks for the catch! The full annotated block stays in one place now.

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
mohammedfirdouss: Didn't even see that, doesnt redirect, it solved now. thanks

pkg/app/pipedv1/plugin/kubernetes_multicluster/README.md
mohammedfirdouss: Hm, so to fix the test command to `make test/go MODULES=pkg/app/pipedv1/plugin/kubernetes_multicluster` with a note about the separate go.mod

rahulshendre (approved):
LGTM 👍
everything checks out, nice work @mohammedfirdouss 🙌

khanhtc1202 (approved):
LGTM, thank you!

PR #6979, opened 2026-07-04
fix(controller): stop starting the next stage after a cancel command

### LINKED ISSUE #6978: Cancelling a deployment right as a stage finishes can leave the next stage running
Cancelling sends the command then closes cancelledCh (controller/scheduler.go). If the cancel lands while a stage is finishing, the stage can still return STAGE_SUCCESS (ctx.Err() is checked before handler.Cancel() lands), so the loop ignores the consumed cancel and continues to the next stage.

The next iteration's select instantly receives nil from the closed channel, and that branch does nothing. The loop falls through without cancelling or waiting for the stage it just started, reads result unsynchronized, marks the deployment CANCELLED and breaks. With auto rollback on, rollback then runs while the abandoned stage is still executing its plugin.

Same pattern in both piped and pipedv1.

Fix I have ready: after doneCh, if a cancel command was received, end the deployment even if the stage reported success, and treat a nil receive as a cancellation. Happy to adjust if you prefer a different approach.

Ahmad-Faraj:
**What this PR does**:

A cancel that lands while a stage is finishing can be ignored, the stage still reports success and the scheduler moves on. The next iteration then receives nil from the closed cancelledCh and abandons the stage it just started while rollback begins. This makes a received cancel end the deployment even when the stage reported success, and treats a nil receive as a cancellation. Same fix in piped and pipedv1.

**Why we need it**:

A cancelled deployment should not keep executing stages, and rollback should not run in parallel with a forward stage.

Fixes #6978

- **How are users affected by this change**: cancelling now reliably stops the deployment after the current stage.


armistcxy:
The problem is we need to determine what we should do in case `StageStatus` is `STAGE_SUCCESS` and cancel command come at the same time

```go
case cmd := <-s.cancelledCh:
    if cmd != nil {
        cancelCommand = cmd
        cancelCommander = cmd.Commander
    }
    handler.Cancel()
    <- doneCh
....

if s.cancelled && result == model.StageStatus_STAGE_SUCCESS {
// What should we do here ?
// Remain STAGE_SUCCESS or change to STAGE_CANCELLED
}
```

Ahmad-Faraj:
I would keep STAGE_SUCCESS. The stage really did finish and executeStage has already committed that status by the time the loop sees the cancel. Rewriting it would misstate what ran on the target. A cancel that lands while the stage is still running already becomes STAGE_CANCELLED via determineStageStatus. So the stage status stays truthful and the cancel is honored at the deployment level with DEPLOYMENT_CANCELLED plus auto rollback when it is configured.

armistcxy:
@Ahmad-Faraj This case needs consideration, you can raise in next biweekly meeting

Ahmad-Faraj:
Will raise it in the next biweekly meeting.

rahulshendre:
@khanhtc1202 @mohammedfirdouss , PTAL

Ahmad-Faraj:
Still relevant, and the race in #6978 still reproduces. A cancel landing while a stage is finishing leaves the next stage started.

My position has not changed, for anyone arriving at this cold. The finished stage keeps `STAGE_SUCCESS`, because `executeStage` already committed that status and rewriting it would misstate what actually ran on the target. The cancel is still honoured at the deployment level and rollback still runs.

I can rebase if it has drifted.

Ahmad-Faraj:
@armistcxy I tried raising this at three biweeklies and no maintainer was on the call any of them, so bringing it back here.

My position hasn't changed. The finished stage keeps `STAGE_SUCCESS` because `executeStage` already committed that status, and rewriting it would misstate what ran on the target. The cancel is still honoured at the deployment level and rollback still runs.

Is that the behaviour you want, or should the stage flip to `STAGE_CANCELLED`?

armistcxy:
@Ahmad-Faraj sorry for late response, I'm reviewing it now

pkg/app/piped/controller/scheduler.go
```diff
@@ -373,17 +373,40 @@ func (s *scheduler) Run(ctx context.Context) error {
...
+			default:
+			}
+		}
+
+		// A received cancel command wins even if the stage finished successfully,
+		// otherwise the loop would continue to the next stage after the user cancelled.
+		if cancelCommand != nil {
+			deploymentStatus = model.DeploymentStatus_DEPLOYMENT_CANCELLED
+			statusReason = fmt.Sprintf("Cancelled by %s while executing stage %s", cancelCommander, ps.Id)
+			break
```
armistcxy, line 408: Add a short comment explaining the intended semantics, I think you can add this "cancel stops progression, it does not rewrite an already-committed stage outcome"

Ahmad-Faraj: Added, in both piped and pipedv1.

pkg/app/piped/controller/scheduler.go
```diff
@@ -373,17 +373,40 @@ func (s *scheduler) Run(ctx context.Context) error {
...
+					cancelCommand = cmd
+					cancelCommander = cmd.Commander
+				}
+			default:
+			}
+		}
+
+		// A received cancel command wins even if the stage finished successfully,
+		// otherwise the loop would continue to the next stage after the user cancelled.
+		if cancelCommand != nil {
```
armistcxy: the new block runs before the result checks, so a cancel that arrives around the same time as a stage failure, a stage timeout, ..... flips the deployment to `CANCELLED` and we lose the real reason

Cancel should only win when the alternative is keep going (`SUCCESS`, `SKIPPED`, `NOT_STARTED_YET`)

armistcxy (commented):
@Ahmad-Faraj Can you review the part I comment

pkg/app/piped/controller/scheduler.go
Ahmad-Faraj: Fixed, cancel now only wins over `SUCCESS` and `SKIPPED`. `NOT_STARTED_YET` with a cancel already ends `CANCELLED` in the branch below with the "previous stage" reason, so I left it out. Want it in this check anyway?

armistcxy: that's fine

armistcxy (approved):

PR #6981, opened 2026-07-05
fix(pipedv1): use FindSlackGroups for group mentions in trigger-failed notification

Ahmad-Faraj:
**What this PR does**:

The failed-trigger notification in pipedv1 fills the group mentions by calling FindSlackUsers, so configured Slack groups are never mentioned when a trigger fails. Looks like a copy-paste from the line above. The v0 piped has it right. This switches it to FindSlackGroups.

None

- **How are users affected by this change**: Slack group mentions now work for trigger-failed notifications in pipedv1.


armistcxy (approved):
Confirm what you have said, seems like a typo in pipedv1, v0 already correct on this

LGTM

mohammedfirdouss (approved):

PR #6983, opened 2026-07-06
fix: avoid panic on missing application spec during migration

### LINKED ISSUE #6980: pipectl migrate application-config panics on malformed application configuration
## What happened

`pipectl migrate application-config` panics when the input application configuration is malformed instead of returning a descriptive validation error.

The panic occurs when the input configuration is missing required fields such as `spec`. Instead of reporting an invalid configuration, the command terminates with a Go panic:

```text
panic: interface conversion: interface {} is nil, not map[string]interface {}
```

The panic originates from:

```text
pkg/app/pipectl/cmd/migrate/application_config.go:142
```

---

## What you expected to happen

The command should validate the input application configuration and return a descriptive error (for example, indicating that the required `spec` field is missing or invalid) instead of panicking.

---

## How to reproduce it

1. Create a file named `bad.yaml` with the following content:

```yaml
kind: KUBERNETES
```

2. Run the migration command:

```bash
go run ./cmd/pipectl migrate application-config --config-files bad.yaml
```

3. Observe that the command panics with:

```text
panic: interface conversion: interface {} is nil, not map[string]interface {}
```

Stack trace (truncated):

```text
panic: interface conversion: interface {} is nil, not map[string]interface {}

github.com/pipe-cd/pipecd/pkg/app/pipectl/cmd/migrate.(*applicationConfig).migrateApplicationConfig(...)
    pkg/app/pipectl/cmd/migrate/application_config.go:142
```

---

## Environment

- `piped` version: N/A (issue reproduced using `pipectl`)
- `control-plane` version: N/A
- Others:
  - PipeCD branch: `master`
  - Commit: `5b81c813b`
  - Go version: `go1.26.2`
  - OS: macOS (Apple Silicon)

---

## Additional context

From the investigation, the panic is caused by unchecked type assertions while reading the parsed YAML (for example, assuming the `spec` field is always present and has the expected type). Returning a validation error instead of panicking would make the migration command more robust when handling malformed input.

If this behavior is considered a bug, I'd be happy to work on a fix and submit a PR.

CodeWith-sakib:
**What this PR does**:

- Prevents `pipectl migrate application-config` from panicking when the input application configuration is missing or has an invalid `spec` field.
- Returns a descriptive validation error instead of panicking.
- Adds a regression test covering this scenario.

**Why we need it**:

Currently, the migration command performs an unchecked type assertion on the `spec` field. When the field is missing or malformed, the command terminates with a Go panic instead of reporting a validation error. This change makes the migration command more robust and improves the user experience when handling invalid input.

Fixes #6980

Yes 

  Users now receive a descriptive error message instead of a panic when running `pipectl migrate application-config` with an invalid application configuration that is missing or has an invalid `spec` field.

  No.

  N/A


rahulshendre (commented):
nice catch @CodeWith-sakib 🙌

@mohammedfirdouss @khanhtc1202 PTAL, when you have time

mohammedfirdouss:
> nice catch @CodeWith-sakib 🙌
> 
> @mohammedfirdouss @khanhtc1202 PTAL, when you have time

LGTM @rahulshendre

mohammedfirdouss:
@khanhtc1202 PTAL

CodeWith-sakib:
@rahulshendre

PR #6992, opened 2026-07-08
Update CODEOWNERS paths for plugins

khanhtc1202:
**What this PR does**:

As @mohammedfirdouss noticed me that he didn't have the permission to merge changes in plugins dir even with the plugin team members config, I think it maybe caused by this missing `/` in the CODEOWNER.

**Why we need it**:

Fixes #

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -10,5 +10,5 @@ CONTRIBUTING.md @pipe-cd/pipecd-approvers-docs @pipe-cd/pipecd-approvers
 README.md @pipe-cd/pipecd-approvers-docs @pipe-cd/pipecd-approvers
 
 # Piped plugins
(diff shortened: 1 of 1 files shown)


khanhtc1202:
@Warashi @t-kikuc @ffjlabo please approve this change when you have time

Warashi:
@khanhtc1202 
This problem is not caused by the CODEOWNERS file.
We set the branch protection rule to prevent PR merges from anyone except the pipecd-approvers, pipecd-approvers-web, and pipecd-approvers-docs.
The settings are here: "Restrict who can push to matching branches" of https://github.com/pipe-cd/pipecd/settings/branch_protection_rules/16741214

khanhtc1202:
Thank @Warashi, I have not noticed about that one. I will close this PR, thank you!

PR #7006, opened 2026-07-10
Resolve host architecture in legacy toolregistry Darwin install scripts

### LINKED ISSUE #7005: Legacy v0 tool installer hardcodes amd64 binaries, breaking darwin/arm64 hosts
**What happened**:

The legacy piped v0 tool registry hardcodes `amd64` in every Darwin (and Linux) install script for `kubectl`, `kustomize`, `helm`, and `terraform`. On an Apple Silicon (`darwin/arm64`) host, `toolregistry.InitDefaultRegistry`/`installKubectl`/`installKustomize`/`installHelm`/`installTerraform` in `pkg/app/piped/toolregistry/tool_darwin.go` unconditionally download the `darwin/amd64` build of each tool, so the resulting binaries cannot execute natively on `arm64` hosts.

This affects both local development and any test suite that exercises these code paths.

The legacy v0 tool registry should resolve the host OS/architecture dynamically (the same way the newer `pipedv1` tool registries already parameterize `{{ .Os }}` / `{{ .Arch }}` in their install scripts), so it downloads a binary compatible with the host architecture.

**How to reproduce it**:

1. Use a `darwin/arm64` machine (e.g. Apple Silicon Mac).
2. From the repository root, run:
   ```
   go test ./pkg/app/piped/platformprovider/kubernetes/...
   ```
3. Observe failures such as:
   ```
   fork/exec /tmp/piped-bin/helm: bad CPU type in executable
   ```
   in `helm_test.go` and `kustomize_test.go`, because the previously-downloaded `darwin/amd64` `helm`/`kustomize` binaries cannot run on `arm64`.
4. Inspect `pkg/app/piped/toolregistry/tool_darwin.go` (lines ~17-55) and confirm every download URL is pinned to `darwin`/`amd64`, regardless of `runtime.GOARCH`.

**Environment**:
- `piped` version: repository checkout (master)
- `control-plane` version: not applicable
- Others: reproduced on macOS `arm64` (Apple Silicon); confirmed via static inspection that `pkg/app/piped/toolregistry/tool_darwin.go` and `pkg/app/piped/toolregistry/install.go` never reference `runtime.GOARCH`, unlike the `pipedv1` tool registries (e.g. `pkg/app/pipedv1/plugin/kubernetes/toolregistry/scripts.go`, `pkg/app/pipedv1/cmd/piped/grpcapi/tool_registry.go`) which already template `{{ .Os }}`/`{{ .Arch }}` from `runtime.GOOS`/`runtime.GOARCH`.

**Suspected root cause**:

The older v0 tool registry (`pkg/app/piped/toolregistry`) uses static, OS-specific install scripts written before Apple Silicon support was considered, while the `pipedv1` tool registries were already generalized to resolve the host architecture at install time.

**Proposed direction**:

Parameterize the Darwin (and Linux) install scripts in `pkg/app/piped/toolregistry` with the host architecture (`runtime.GOARCH`), following the pattern already used by the `pipedv1` tool registries, and add regression coverage asserting the rendered install scripts reference the host architecture rather than a hardcoded value.

**Acceptance criteria**:
- Helm, kubectl, kustomize, and Terraform install scripts reference the host architecture instead of a hardcoded `amd64`.
- The legacy (`pkg/app/piped/toolregistry`) and `pipedv1` tool registries no longer diverge on architecture handling.
- Focused tests cover the rendered install scripts for a non-`amd64` architecture.

rootp1:
**What this PR does**:

Parameterizes the legacy (v0) piped tool registry's Darwin (and Linux) install scripts for `kubectl`, `kustomize`, `helm`, and `terraform` with the host architecture (`runtime.GOARCH`) instead of a hardcoded `amd64`, matching the pattern already used by the `pipedv1` tool registries. Adds a focused test asserting the rendered Darwin install scripts reference the host architecture rather than a hardcoded value.

**Why we need it**:

On `darwin/arm64` (Apple Silicon) hosts, the legacy tool registry downloaded `darwin/amd64` binaries, which cannot execute natively (`fork/exec ...: bad CPU type in executable`), breaking local development and any test relying on these tools (e.g. `pkg/app/piped/platformprovider/kubernetes`).

Fixes #7005

- **How are users affected by this change**: Developers running `piped` v0 (legacy) tooling on `darwin/arm64` now get architecture-correct `kubectl`/`kustomize`/`helm`/`terraform` binaries instead of incompatible `amd64` ones. No behavior change on `amd64` hosts.

## Verification
- `go build ./...` — passed
- `go vet ./pkg/app/piped/toolregistry/...` — passed
- `gofmt -l` on changed files — passed (no output)
- `go test ./pkg/app/piped/toolregistry/...` — passed
- `go test ./pkg/app/piped/platformprovider/kubernetes/...` (with tool cache cleared, real downloads) — `TestTemplateLocalChart_WithNamespace` and `TestKustomizeTemplate_WithHelm` (helm-based paths) now pass on `darwin/arm64`, where they previously failed with `bad CPU type in executable`. `TestKustomizeTemplate` still fails, but for an unrelated reason: the pinned default `kustomize` version (`3.8.1`) never published a `darwin_arm64` release artifact upstream (confirmed via direct HTTP HEAD, 404), so no code-level fix can download it. This is a pre-existing upstream artifact gap for that specific pinned version, not a regression from this change.
- `go vet ./...` — pre-existing unrelated lock-copy warnings across the repo (not touched by this change); no new warnings introduced by this diff.
- `make check` (full target, requires Docker-based `golangci-lint`, web build, and envtest/Docker-backed integration tests) — not run: Docker and `golangci-lint` are not available in this environment.
- `./hack/ensure-dco.sh` — reported an unrelated pre-existing failure because it diffs against the fork's stale `origin/master` (13 commits behind `upstream/master`); the actual commit on this branch carries a proper `Signed-off-by` trailer (verified via `git log`).

## Scope
- Only `pkg/app/piped/toolregistry/install.go`, `pkg/app/piped/toolregistry/tool_darwin.go`, and the new `pkg/app/piped/toolregistry/tool_darwin_test.go` were changed. No unrelated files were modified.

### DIFF
--- a/pkg/app/piped/toolregistry/install.go
+++ b/pkg/app/piped/toolregistry/install.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
 	"text/template"
 
 	"go.uber.org/zap"
@@ -58,6 +59,7 @@ func (r *registry) installKubectl(ctx context.Context, version string) error {
 			"Version":    version,
 			"BinDir":     r.binDir,
 			"AsDefault":  asDefault,
+			"Arch":       runtime.GOARCH,
 		}
 	)
 	if err := kubectlInstallScriptTmpl.Execute(&buf, data); err != nil {
@@ -105,6 +107,7 @@ func (r *registry) installKustomize(ctx context.Context, version string) error {
 			"Version":    version,
 			"BinDir":     r.binDir,
 			"AsDefault":  asDefault,
+			"Arch":       runtime.GOARCH,
 		}
 	)
 	if err := kustomizeInstallScriptTmpl.Execute(&buf, data); err != nil {
@@ -152,6 +155,7 @@ func (r *registry) installHelm(ctx context.Context, version string) error {
 			"Version":    version,
 			"BinDir":     r.binDir,
 			"AsDefault":  asDefault,
+			"Arch":       runtime.GOARCH,
 		}
 	)
 	if err := helmInstallScriptTmpl.Execute(&buf, data); err != nil {
@@ -199,6 +203,7 @@ func (r *registry) installTerraform(ctx context.Context, version string) error {
 			"Version":    version,
 			"BinDir":     r.binDir,
 			"AsDefault":  asDefault,
+			"Arch":       runtime.GOARCH,
 		}
 	)
 	if err := terraformInstallScriptTmpl.Execute(&buf, data); err != nil {
--- a/pkg/app/piped/toolregistry/tool_darwin.go
+++ b/pkg/app/piped/toolregistry/tool_darwin.go
@@ -16,7 +16,7 @@ package toolregistry
 
 var kubectlInstallScript = `
 cd {{ .WorkingDir }}
-curl -LO https://dl.k8s.io/release/v{{ .Version }}/bin/darwin/amd64/kubectl
+curl -LO https://dl.k8s.io/release/v{{ .Version }}/bin/darwin/{{ .Arch }}/kubectl
 mv kubectl {{ .BinDir }}/kubectl-{{ .Version }}
 chmod +x {{ .BinDir }}/kubectl-{{ .Version }}
 {{ if .AsDefault }}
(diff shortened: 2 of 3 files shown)


rootp1:
@mohammedfirdouss @rahulshendre @eeshaanSA @Ayushmore1214 
wdyt

PR #7012, opened 2026-07-11
fix(ci): define GO_VERSION and pin checkout action 

Ayushmore1214:
setup-go referenced ${{ env.GO_VERSION }} but the workflow never declared it, so it silently used an arbitrary Go version instead of 1.26.2. Also pin actions/checkout to the same SHA (v4.2.2) used by every other workflow in this repo.

**What this PR does**:

**Why we need it**:

Fixes #

### DIFF
--- a/.github/workflows/plugin_release.yaml
+++ b/.github/workflows/plugin_release.yaml
@@ -14,11 +14,14 @@ on:
 permissions:
   contents: write
 
+env:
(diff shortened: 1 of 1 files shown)


Ayushmore1214:
@Warashi plz take a look!!

Warashi (approved):
Thank you!

PR #7021, opened 2026-07-12
fix(piped): clean up cloned repo after skip stage check

Ahmad-Faraj:
**What this PR does**:

`shouldSkipStage` clones the repo into a temporary directory on every skip check but never removes it. Each check with skipOn conditions leaks one temp directory until the piped restarts. This adds a deferred `repo.Clean()`, the same pattern the drift detectors and the event watcher use.

None


Ahmad-Faraj:
Still applies to master. One line. The repo cloned for the skip stage check is never removed, so a piped running skip stage checks accumulates clones on disk. I can rebase if master has moved under it.

khanhtc1202 (approved):
Nice catch, thank you 👍

khanhtc1202:
@Ahmad-Faraj, could you check the v1 piped and open a PR if needed? thank you!

Ahmad-Faraj:
@khanhtc1202 checked, v1 has it too. `determineSkipStage` in `pkg/app/pipedv1/controller/skipstage.go` clones and never cleans, same as v0 did. PR is up: #7275

PR #7030, opened 2026-07-14
Add CronJob support to FindReferencingConfigMaps and FindReferencingS…

### LINKED ISSUE #6761: FindReferencingConfigMaps and FindReferencingSecrets miss CronJob references

Ankitdotraider:
**What this PR does**:
FindReferencingConfigMaps and FindReferencingSecrets only walk spec.template.spec.*, which works for Deployment, StatefulSet, and DaemonSet. CronJob nests its pod template under spec.jobTemplate.spec.template.spec.*, so ConfigMap/Secret references inside a CronJob were silently missed.

Added the missing jobTemplate probes to both functions, in both the kubernetes and kubernetes_multicluster provider packages, following the same pattern already used for the existing paths. Added test cases covering CronJob ConfigMap/Secret references in both packages.

Scoped to just these two functions no changes to FindContainerImages or anything else, to keep this focused.

**Why we need it**:
Without this fix, PipeCD's drift detection and re-deploy triggering silently ignore config/secret changes in CronJob workloads.

Fixes #6761

- **How are users affected by this change**: PipeCD now correctly detects ConfigMap/Secret references inside CronJob workloads, enabling proper drift detection and re-deploy triggering.
- **Is this breaking change**: no.
- **How to migrate (if breaking change)**:N/A


Ankitdotraider:
@mohammedfirdouss @khanhtc1202 , could you please look into this

PR #7040, opened 2026-07-16
feat(pipectl): add application enable command

Ahmad-Faraj:
**What this PR does**:

Adds an `application enable` command to pipectl.

**Why we need it**:

An application disabled with pipectl can only be enabled back from the web UI. The API already has `EnableApplication`, only the CLI command was missing.

None

- **How are users affected by this change**: `pipectl application enable` is available to enable a disabled application.
- **How to migrate (if breaking change)**: Nothing


Ahmad-Faraj:
Still applies. `EnableApplication` is already on `apiservice` and pipectl has no equivalent, so this closes that gap. I can rebase if needed.

rahulshendre (commented):
seem like a good idea @Ahmad-Faraj 

on the docs side: the update covers docs-dev but docs-v0.58.x and docs-v1.0.x both have the disable command documented and are missing the enable section, we should add it there too

Ahmad-Faraj:
@rahulshendre v1.0.x makes sense, adding it there.

v0.58.x I'm less sure. `enable` isn't in the v0.58.0 binary, that tag only has add, delete, disable, get, list and sync. Documenting it on a released version points users at a command that errors. `disable` went docs-dev only in #4123 and reached v0.58.x through the release snapshot.

v1.0.x only, or both?

rahulshendre:
> @rahulshendre v1.0.x makes sense, adding it there.
> 
> v0.58.x I'm less sure. `enable` isn't in the v0.58.0 binary, that tag only has add, delete, disable, get, list and sync. Documenting it on a released version points users at a command that errors. `disable` went docs-dev only in #4123 and reached v0.58.x through the release snapshot.
> 
> v1.0.x only, or both?

thanks for the pointer @Ahmad-Faraj, docs-dev only makes sense for v0.58.x since the release script handles it, for v1.0.x I'm not completely sure, @mohammedfirdouss what do you think 👀

Ahmad-Faraj:
@mohammedfirdouss added it to `docs-v1.0.x` too and pushed.

v1.0.0 isn't released yet, the latest tag is v0.58.0, so nothing there points at a binary that lacks the command. And `docs-v1.0.x` isn't fed by the release snapshot the way `docs-v0.58.x` is, so it won't pick up the `docs-dev` change on its own.

v0.58.x left alone.

Ahmad-Faraj:
@rahulshendre could you take another look when you get a chance?

PR #7046, opened 2026-07-20
ci: publish Helm charts in a dedicated job gated on image builds

### LINKED ISSUE #6999: ci: `publish_image_chart` pushes Helm charts from every matrix cell, before all container images exist (race condition)
**What happened**:

In the `publish_image_chart` workflow (`.github/workflows/publish_image_chart.yaml`), the `artifacts` job runs as a matrix of about 10 parallel cells (container registry × image: `helloworld`, `launcher`, `pipecd`, `piped`, `pipectl`, ...). Each cell builds and pushes one container image.

However, the last steps of the job ("Install helm" / "Login to OCI using Helm" / "Publish helm charts") are also inside that matrix job. So they do not run once they run in **every** matrix cell, and each cell pushes **all three charts** (`pipecd`, `piped`, `helloworld`).

This causes two problems:

1. **Charts can be published before the images they reference exist.**

   The matrix cells run in parallel and finish at very different times. A small image like `helloworld` builds in a few minutes, while `piped` / `pipecd` are multi-arch builds and take much longer.

   So the fast `helloworld` cell can finish first and push the `pipecd` and `piped` charts while the `pipecd` and `piped` images are still building in other cells. During that window, the registry contains a chart version whose image tag does not exist yet. Anyone who installs the chart at that moment gets `ImagePullBackOff`.

   The worse case: if a slower cell then **fails**, the charts stay published permanently while the images they point to were never pushed a broken release that the failed run cannot undo.

2. **The same chart version is pushed to the same OCI ref by up to 10 runners at the same time.**

   Every matrix cell runs `make build/chart` and `helm push` for the same three charts to `oci://ghcr.io/pipe-cd/chart`. At best this is redundant work; at worst, concurrent pushes to the same ref can conflict and fail intermittently, which shows up as flaky CI ("failed, but passed on re-run").

Note that the newer `publish_pipedv1_exp` workflow (`.github/workflows/publish_pipedv1_exp.yaml`) already avoids this problem: it builds images in parallel jobs, then merges the multi-arch manifest in a job with `needs: container_image_build`, and only pushes the Helm chart in a final `image_chart` job with `needs: container_image_merge`. The `publish_image_chart` workflow just does not follow this structure yet.

- Helm charts should be pushed only **after** all container images for that version have been pushed successfully.
- Each chart should be pushed exactly **once** per run, not once per matrix cell.
- If any image build fails, no chart for that version should be published.

A simple fix would be to move the "Install helm" / "Login to OCI using Helm" / "Publish helm charts" steps out of the `artifacts` matrix job into a separate job (e.g. `charts`) with `needs: artifacts`, similar to how `publish_pipedv1_exp.yaml` is structured.

**How to reproduce it**:

This is a latent race condition in the workflow structure, so it depends on timing, but it can be seen as follows:

1. Look at any run of the `publish_image_chart` workflow: the "Publish helm charts" step appears in every matrix cell of the `artifacts` job, so the same charts are pushed multiple times per run.
2. In a run where a fast cell (e.g. `helloworld`) finishes several minutes before a slow cell (e.g. `piped`), the charts are already published to `ghcr.io/pipe-cd/chart` while the `piped` image is still being built. Installing the chart in that window fails with `ImagePullBackOff`.
3. If a slow matrix cell fails after a fast cell has already published the charts, the registry is left with charts referencing image tags that were never pushed.

**Environment**:
- `piped` version: Not related.
- `control-plane` version: Not related.
- Others: GitHub Actions workflow `.github/workflows/publish_image_chart.yaml` (job `artifacts`, steps "Install helm" / "Login to OCI using Helm" / "Publish helm charts").

TheRealShek:
## What this does

Moves the Helm chart publish out of the `artifacts` matrix job into a dedicated `charts` job gated on `needs: artifacts`, mirroring the structure `publish_pipedv1_exp.yaml` already uses (`container_image_build` → `image_chart`).

Fixes #6999.

## Problem

In `publish_image_chart.yaml`, the `Install helm` / `Login to OCI using Helm` / `Publish helm charts` steps live **inside** the `artifacts` matrix job. That matrix runs ~10 parallel cells, so:

- Every cell runs `Publish helm charts`, and each one pushes **all three** charts (`pipecd`, `piped`, `helloworld`) to the same `oci://ghcr.io/pipe-cd/chart` ref — up to 10 redundant, concurrent pushes per run (a source of flaky "failed, passed on re-run" CI).
- A fast cell (e.g. `helloworld`) publishes the `pipecd`/`piped` charts while those images are still building in slower cells → a window where the registry serves a chart whose image tag doesn't exist yet (`ImagePullBackOff`).
- If a slow cell then fails, the charts stay published while their images were never pushed — a broken release the failed run can't undo.

## Evidence

From run [`#29642136136`](https://github.com/pipe-cd/pipecd/actions/runs/29642136136) (all on `ghcr.io`, which the default charts reference):

| Matrix cell | `Publish helm charts` done | Image push done |
|---|---|---|
| `ghcr.io/pipe-cd, launcher` | **11:15:45Z** (pushes all 3 charts) | 11:15:32Z |
| `ghcr.io/pipe-cd, piped` | 11:16:27Z | 11:16:14Z (**+29s** after launcher's chart push) |
| `ghcr.io/pipe-cd, pipecd` | 11:16:33Z | 11:16:19Z (**+34s** after launcher's chart push) |

The `launcher` cell published the `piped`/`pipecd` charts ~30s before those images finished pushing to ghcr. (On the gcr mirror the `piped` image lagged by ~7m44s, affecting installs that override `image.repository` to gcr.)

## Fix

- New `charts` job with `needs: artifacts` — GitHub starts it only after **every** matrix cell succeeds, so charts publish once, only after all images exist, and never if a build fails.
- Repoint `trigger-event-watcher` and `release-quickstart-manifests` to `needs: charts` (the latter runs `helm template` against the just-published chart, so it must wait for the chart push).

## Testing

- `actionlint` — clean, aside from the pre-existing `oracle-vm-8cpu-32gb-x86-64` custom self-hosted runner-label warning on the unchanged `artifacts` job.
- YAML validated; job graph confirmed: `artifacts` → `charts` → `trigger-event-watcher` / `release-quickstart-manifests`.

Opened as a **draft** for maintainer feedback.


Ayushmore1214:
@TheRealShek Please mind getting assigned before opening a PR, this is okay for now as it is your first contribution here but please from next time be careful!

Warashi (approved):
sorry for being late, LGTM
thank you!

PR #7054, opened 2026-07-21
fix: prevent nil-deref panic on invalid pem block (#7053)

### LINKED ISSUE #7053: Unchecked pem.Decode in pkg/crypto/key.go causes nil-deref panic (Potential DoS)
**What happened**:
A vulnerability was identified in `pkg/crypto/key.go` where `pem.Decode` is called, but the returned `block` is not checked for `nil`. Because `pem.Decode` returns a `nil` block for any non-PEM input, the subsequent access to `block.Bytes` results in a nil-pointer dereference panic. 

This panic is reachable from the control plane (`GenerateApplicationSealedSecret` -> `encrypt()` -> `NewHybridEncrypter` -> `ParseRSAPublicKeyFromPem`). With a malformed key, this will crash the shared multi-tenant server, potentially leading to a Denial of Service (DoS). It can also crash `piped` via `NewHybridDecrypter` on a bad key file.

The functions `ParseRSAPublicKeyFromPem` and `ParseRSAPrivateKeyFromPem` should validate that the decoded block is not `nil` and return a standard `error` if it is, preventing a full process panic.
```go
	data = bytes.TrimSpace(data)
	block, _ := pem.Decode(data)
	if block == nil {
		return nil, errors.New("failed to decode PEM block")
	}
	bytes := block.Bytes

How to reproduce it:
1.Provide a non-PEM input or malformed key where a PEM format is expected (e.g., in    GenerateApplicationSealedSecret or when registering a piped).
2.The pem.Decode(data) function receives the invalid data and returns nil for the block.
3.The application panics when attempting to read block.Bytes.

Environment:

- piped version: latest

- control-plane version: latest

- Others: N/A

oindrilakha12-ui:
**What this PR does**:
Adds a nil check for the decoded PEM block in `ParseRSAPublicKeyFromPem` and `ParseRSAPrivateKeyFromPem` (inside `pkg/crypto/key.go`) to safely return an error instead of crashing.

**Why we need it**:
When `pem.Decode` receives invalid or non-PEM data, it returns a `nil` block. Accessing `block.Bytes` without checking causes a process crash. This fixes a potential Denial of Service (DoS) vulnerability.

Fixes #7053

No.


oindrilakha12-ui:
@Warashi , @rahulshendre please review it.

khanhtc1202 (approved):
Nice catch, lgtm 👍 Thank you!

PR #7088, opened 2026-07-24
Cache plan preview repository routing

### LINKED ISSUE #7087: Plan preview request routing scans all active pipeds
**What happened**:

`RequestPlanPreview` currently lists all enabled pipeds in the project and scans every configured repository entry in memory to find pipeds that match the requested repository remote and base branch. This makes each interactive plan preview request scale with total active piped count and repository mappings instead of the number of matching pipeds.

## Summary
Plan preview request routing performs an O(number of pipeds x repositories per piped) in-memory scan for every request.

## Affected Area
`pkg/app/server/grpcapi/api.go`, specifically `API.RequestPlanPreview`.

## Steps to Reproduce
1. Create a test or benchmark environment with many active pipeds in one project and repository mappings on those pipeds.
2. Call `API.RequestPlanPreview` repeatedly for one repository remote and base branch.
3. Capture allocations and latency with `go test -bench` or pprof.
4. Observe that request cost grows with total piped count rather than with the matching pipeds.

Plan preview routing should avoid a full active-piped scan on every request. Repeated routing for a repository/base-branch pair should be close to O(matches), for example via an index, bounded cache, or targeted lookup.

## Expected Behaviour
Routing a plan preview request should avoid repeated full scans and should keep behavior correct when multiple pipeds match the same repository and branch.

**How to reproduce it**:

1. Seed or mock thousands of enabled pipeds for one project.
2. Give only a subset a repository entry matching the plan preview request remote and base branch.
3. Repeatedly call `RequestPlanPreview` for that remote and branch.
4. Compare latency/allocations as piped count increases.

## Actual Behaviour
Each request calls `pipedStore.List` with only `ProjectId` and `Disabled=false` filters, then iterates through every returned piped and every repository entry until it finds matching mappings.

## Technical Evidence
In `pkg/app/server/grpcapi/api.go`, `RequestPlanPreview` lists enabled pipeds for the project and then builds a `repositories` map by checking each `p.Repositories` entry against `req.RepoRemoteUrl` and `req.BaseBranch`. The code also has a TODO noting that caching may be needed to reduce database load.

## Suspected Root Cause
The API layer has no repository-to-piped routing index or cache, so it must perform repository matching after fetching all enabled pipeds.

## Proposed Direction
Add a focused benchmark for plan preview routing at small and large piped cardinalities, then avoid repeated full scans by adding a bounded cache or another targeted lookup for repository-to-piped matches.

## Acceptance Criteria
- A benchmark covering small and large piped counts is added.
- Repeated plan preview routing for the same project/repository/base-branch avoids repeated full active-piped scans.
- Correct behavior is preserved when multiple pipeds match the same repository and branch.

**Environment**:
- `piped` version: Not applicable
- `control-plane` version: Current `master` branch
- Others: Not applicable

rootp1:
## Summary
- cache plan preview repository-to-piped matches by project, repo remote, and base branch
- avoid repeated full active-piped scans for repeated requests on the same route
- add focused tests and a benchmark covering 1, 100, and 10000 pipeds

## Linked Issue
Fixes #7087

## What Changed
- added a bounded in-memory TTL cache for `RequestPlanPreview` routing matches
- reused cached repository matches before falling back to `pipedStore.List`
- added regression tests for cached matches and cached empty results
- added a benchmark verifying repeated requests stay off the datastore scan path

## Verification
- `go test ./pkg/app/server/grpcapi -run 'TestRequestPlanPreview|TestRequireAPIKey'` — passed
- `go test -bench BenchmarkRequestPlanPreviewRepositoryRouting -benchmem ./pkg/app/server/grpcapi` — passed
- `make build/go` — passed
- `make check/gen/code` — passed
- `make check` — not passed: `lint/go` uses a pinned `golangci-lint` image built with Go 1.25, but this repository targets Go 1.26.2, so lint fails before the rest of `make check` can complete
- `make check/dco` — not passed: fails on upstream commit `25ae69937e55e8330bff96c34346718f3e5e7148`, which is already contained in `upstream/master`

## Scope
- only the plan preview routing path and its focused tests were changed; unrelated files were not modified

**What this PR does**:

Caches repository routing for plan preview requests so repeated requests do not rescan every active piped in the project.

**Why we need it**:

Plan preview is an interactive path, and repeated requests for the same repository and branch should not require a full active-piped scan each time.

Fixes #7087

- **How are users affected by this change**: Plan preview routing avoids repeated full scans for the same repository and base branch, reducing control-plane work on repeated requests.

### DIFF
--- a/pkg/app/server/grpcapi/api.go
+++ b/pkg/app/server/grpcapi/api.go
@@ -80,14 +80,15 @@ type commandOutputGetter interface {
 type API struct {
 	apiservice.UnimplementedAPIServiceServer
 
-	applicationStore     apiApplicationStore
-	deploymentStore      apiDeploymentStore
-	pipedStore           apiPipedStore
-	eventStore           apiEventStore
-	deploymentTraceStore apiDeploymentTraceStore
-	commandStore         commandstore.Store
-	stageLogStore        stagelogstore.Store
-	commandOutputGetter  commandOutputGetter
+	applicationStore           apiApplicationStore
+	deploymentStore            apiDeploymentStore
+	pipedStore                 apiPipedStore
+	eventStore                 apiEventStore
+	deploymentTraceStore       apiDeploymentTraceStore
+	commandStore               commandstore.Store
+	stageLogStore              stagelogstore.Store
+	commandOutputGetter        commandOutputGetter
+	planPreviewRepositoryCache *planPreviewRepositoryCache
 
 	encryptionKeyCache cache.Cache
 	pipedStatCache     cache.Cache
@@ -116,6 +117,10 @@ func NewAPI(
 		commandStore:         commandstore.NewStore(ds, sc, logger),
 		stageLogStore:        stagelogstore.NewStore(fs, sc, logger),
 		commandOutputGetter:  cog,
+		planPreviewRepositoryCache: newPlanPreviewRepositoryCache(
+			1024,
+			time.Minute,
+		),
 		// Public key is variable but likely to be accessed multiple times in a short period.
 		encryptionKeyCache: memorycache.NewTTLCache(ctx, 5*time.Minute, 5*time.Minute),
 		pipedStatCache:     psc,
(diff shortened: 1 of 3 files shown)


Ankitdotraider (commented):
Cached routes to a disabled piped stay stale until TTL expiry upto a minute, is that acceptable, or do we need invalidation on disable? Also eviction just deletes whatever key the map hits first instead of LRU, and there's no test for expiry/eviction even though nowFunc is already injectable for it.

PR #7099, opened 2026-07-31
Skip status and kind filters when all given values are empty

### LINKED ISSUE #7098: ListDeployments panics when statuses or kinds contain only empty values
**What happened**:

`APIService.ListDeployments` panics with `index out of range [0] with length 0` when the request has `statuses` (or `kinds`) that are not empty as a list, but every value inside is an empty string.

In `pkg/app/server/grpcapi/api.go` the loop skips empty values, but `statuses[0]` is read after it without checking that anything was actually added:

```go
if len(req.Statuses) > 0 {
    statuses := []model.DeploymentStatus{}
    for _, s := range req.Statuses {
        if s != "" {                       // empty values are skipped here
            ...
            statuses = append(statuses, model.DeploymentStatus(depstatus))
        }
    }

    filters = append(filters, datastore.ListFilter{
        Field:    "Status",
        Value:    statuses[0],             // <- panics when the slice stayed empty
    })
}
```

The same pattern is in the `kinds` block right below it.

I think this is worse than a normal error because the API server is started without any recovery interceptor (`cmd/pipecd/server.go` only adds log, api key auth, request validation and prometheus), so the panic is not caught and it stops the whole control plane process, not only that one request.

If all given values are empty, no status/kind filter should be added and the request should be handled normally, like when no filter is given at all.

**How to reproduce it**:

The shortest way is with pipectl:

```
pipectl deployment list --status=,
```

`--status=,` looks strange but it is what a user gets from an empty item in a comma separated list. pflag parses the value as CSV, so `,` becomes two empty strings.

I did not want to guess how pflag behaves, so I checked it with a small test that declares the flag the same way `pkg/app/pipectl/cmd/deployment/list.go` does:

```go
var statuses []string
fs := pflag.NewFlagSet("probe", pflag.ContinueOnError)
fs.StringSliceVar(&statuses, "status", nil, "")
fs.Parse([]string{arg})
```

Result:

```
--status=          -> []               len=0                 ok
--status=,         -> ["" ""]          len=2, none usable    panics
--status=""        -> [""]             len=1, none usable    panics
--status=,,        -> ["" "" ""]       len=3, none usable    panics
--status=SUCCESS,  -> ["SUCCESS" ""]   len=2, one usable     ok
```

pipectl lets those empty values through on purpose. In `pkg/app/pipectl/cmd/deployment/list.go` the validation skips them:

```go
for _, status := range c.statuses {
    if status != "" {
        if _, ok := model.DeploymentStatus_value[status]; !ok {
            return errors.Errorf("%s is invalid deployment status", status)
        }
    }
}
```

so they pass client side validation and reach the server as they are.

Any other gRPC client can also send `statuses: [""]` directly. The field has no validate rule in `service.proto`, so `WithRequestValidationUnaryInterceptor` does not reject it either.

To see the panic itself I called the handler directly:

```go
ctx := rpcauth.ContextWithAPIKey(context.Background(), &model.APIKey{
    ProjectId: "p", Role: model.APIKey_READ_ONLY,
})
api := &API{deploymentStore: someStore, logger: zap.NewNop()}
api.ListDeployments(ctx, &apiservice.ListDeploymentsRequest{
    Statuses: []string{""},
    Limit:    10,
})
```

```
panic: runtime error: index out of range [0] with length 0

github.com/pipe-cd/pipecd/pkg/app/server/grpcapi.(*API).ListDeployments(...)
	/pkg/app/server/grpcapi/api.go:553
```

It never reaches the datastore, the panic happens while the filters are being built.

**Environment**:
- `piped` version: not related
- `control-plane` version: `master` at commit `56fb80e7`
- Others: `pipectl` on the same commit

I have a small fix and a regression test for this in #7099.

vipulpandey21:
**What this PR does**:

In `APIService.ListDeployments`, the status and kind filters are added only when at least one usable value was collected, instead of reading `statuses[0]` / `kinds[0]` on a slice that can be empty.

```go
+		// Empty values are skipped above, so the request may contain no usable
+		// status at all. In that case no status filter should be added.
+		if len(statuses) > 0 {
 			filters = append(filters, datastore.ListFilter{
 				Field:    "Status",
 				Value:    statuses[0],
 			})
+		}
```

Same guard for `kinds`. Behaviour is unchanged when a valid value is given.

**Why we need it**:

The loop right above already skips empty values, so a request like `statuses: [""]` leaves the slice empty and the index access panics with `index out of range [0] with length 0`. The API server has no recovery interceptor, so the panic stops the whole control plane process, not just that request. It is reachable with `pipectl deployment list --status=,`. Full details and how I checked it are in #7098.

I only added the two guards. I did not change how empty values are treated, since both the handler and pipectl already skip them on purpose.

**How I tested it**

Added `TestListDeploymentsFilters`, which checks which filters actually reach the datastore.

Without the fix it reproduces the panic at the old line:

```
--- FAIL: TestListDeploymentsFilters/ok:_statuses_holding_only_empty_values_adds_no_status_filter
panic: runtime error: index out of range [0] with length 0
    .../pkg/app/server/grpcapi/api.go:553
```

With the fix:

```
--- PASS: TestListDeploymentsFilters
    --- PASS: .../ok:_no_optional_filter_is_given
    --- PASS: .../ok:_a_valid_status_is_used_as_a_filter
    --- PASS: .../ok:_statuses_holding_only_empty_values_adds_no_status_filter
    --- PASS: .../ok:_kinds_holding_only_empty_values_adds_no_kind_filter
    --- PASS: .../invalid:_unknown_deployment_status
ok  	github.com/pipe-cd/pipecd/pkg/app/server/grpcapi
```

`go test ./pkg/app/server/grpcapi/...` passes, and `go build ./cmd/pipecd ./cmd/pipectl` is fine.

`go vet` on that package still prints two `copies lock value` warnings for `RegisterEvent`. Those are already on master without my change, so I left them alone.

Fixes #7098

- **How are users affected by this change**: A request whose status or kind values are all empty is handled like no filter was given, instead of crashing the control plane.

### DIFF
--- a/pkg/app/server/grpcapi/api.go
+++ b/pkg/app/server/grpcapi/api.go
@@ -548,11 +548,15 @@ func (a *API) ListDeployments(ctx context.Context, req *apiservice.ListDeploymen
 			}
 		}
 
-		filters = append(filters, datastore.ListFilter{
-			Field:    "Status",
-			Operator: datastore.OperatorEqual,
-			Value:    statuses[0],
-		})
+		// Empty values are skipped above, so the request may contain no usable
+		// status at all. In that case no status filter should be added.
+		if len(statuses) > 0 {
+			filters = append(filters, datastore.ListFilter{
+				Field:    "Status",
+				Operator: datastore.OperatorEqual,
+				Value:    statuses[0],
+			})
+		}
 	}
 	if len(req.Kinds) > 0 {
 		kinds := []model.ApplicationKind{}
@@ -566,11 +570,15 @@ func (a *API) ListDeployments(ctx context.Context, req *apiservice.ListDeploymen
 			}
 		}
 
-		filters = append(filters, datastore.ListFilter{
-			Field:    "Kind",
-			Operator: datastore.OperatorEqual,
-			Value:    kinds[0],
-		})
+		// Empty values are skipped above, so the request may contain no usable
+		// kind at all. In that case no kind filter should be added.
+		if len(kinds) > 0 {
+			filters = append(filters, datastore.ListFilter{
+				Field:    "Kind",
+				Operator: datastore.OperatorEqual,
+				Value:    kinds[0],
+			})
+		}
 	}
 	if len(req.ApplicationIds) > 0 {
 		filters = append(filters, datastore.ListFilter{
--- a/pkg/app/server/grpcapi/api.go
+++ b/pkg/app/server/grpcapi/api.go
@@ -547,11 +547,15 @@ func (a *API) ListDeployments(ctx context.Context, req *apiservice.ListDeploymen
 			}
 		}
 
-		filters = append(filters, datastore.ListFilter{
-			Field:    "Status",
-			Operator: datastore.OperatorEqual,
-			Value:    statuses[0],
-		})
+		// Empty values are skipped above, so the request may contain no usable
+		// status at all. In that case no status filter should be added.
+		if len(statuses) > 0 {
+			filters = append(filters, datastore.ListFilter{
+				Field:    "Status",
+				Operator: datastore.OperatorEqual,
+				Value:    statuses[0],
+			})
+		}
 	}
 	if len(req.Kinds) > 0 {
 		kinds := []model.ApplicationKind{}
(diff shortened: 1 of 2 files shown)


khanhtc1202:
@vipulpandey21 please resolve the conflict 👀

vipulpandey21:
Resolved sir, thanks for the heads up. Rebased onto master and re-verified the tests locally, should be clean now.

PR #4186, opened 2023-02-14
Support pipectl to list applications by Label.

funera1:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #3295 #3343

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #7101, opened 2026-07-31
Return the cursor when listing applications without label filter

### LINKED ISSUE #7100: ListApplications does not return the cursor when no label filter is given
**What happened**:

`APIService.ListApplications` does not return the cursor when the request has no label filter, so a client can never ask for the next page.

In `pkg/app/server/grpcapi/api.go` the cursor comes back from the datastore, but the early return for the no label case drops it:

```go
apps, cursor, err := a.applicationStore.List(ctx, options)
if err != nil {
    return nil, gRPCStoreError(err, "failed to list applications")
}

labels := req.Labels
if len(req.Labels) == 0 {
    return &apiservice.ListApplicationsResponse{
        Applications: apps,        // Cursor is not set here
    }, nil
}
```

The other two returns in the same function do set it, and `ListDeployments` sets it in the same situation, so this one looks like it was just missed.

This also used to work. I checked the file at the commit before #4186, which is the PR that added label filtering, and back then the function had a single return that included the cursor:

```
$ git show cf48e6db~1:pkg/app/server/grpcapi/api.go
...
	return &apiservice.ListApplicationsResponse{
		Applications: apps,
		Cursor:       cursor,
	}, nil
}
```

Cursor pagination for this RPC was added on purpose in #1892, so this looks like a regression from #4186 rather than something that was never implemented.

The response carries the cursor so it can be passed to the next call. `pipectl application list` has a documented flag for exactly this:

```
--cursor string   The cursor which returned by the previous request applications list.
```

**How to reproduce it**:

In a project that has more applications than the limit:

```
pipectl application list --limit=10
```

The response has 10 applications and `"cursor":""`, so there is no value to pass to the next call and the remaining applications cannot be reached.

Adding a label filter, for example `--label env:prod`, returns the cursor correctly. That difference is why this is easy to miss.

To confirm it is the handler and not the datastore, I called `ListApplications` with a store stub that always returns a non empty cursor:

```go
store.EXPECT().List(gomock.Any(), gomock.Any()).
    Return([]*model.Application{{Id: "app-1", ProjectId: "project-id"}}, "next-page-cursor", nil)

resp, _ := api.ListApplications(ctx, &apiservice.ListApplicationsRequest{Limit: 1})
```

```
applications = 1
response cursor = ""        (the store had returned "next-page-cursor")
```

With `Labels: {"env":"prod"}` in the same request, the response cursor is `"next-page-cursor"`.

**Environment**:
- `piped` version: not related
- `control-plane` version: `master` at commit `56fb80e7`
- Others: `pipectl` on the same commit

I have a one line fix and a test for it in #7101.

vipulpandey21:
**What this PR does**:

Sets the `Cursor` field on the response `APIService.ListApplications` returns when the request has no label filter.

```go
 	if len(req.Labels) == 0 {
 		return &apiservice.ListApplicationsResponse{
 			Applications: apps,
+			Cursor:       cursor,
 		}, nil
 	}
```

**Why we need it**:

That early return was the only path in the function that did not set the cursor, so a client without a label filter always received an empty cursor and could not ask for the next page. `pipectl application list` has a documented `--cursor` flag but there was no value to pass to it.

The other two returns in the same function set it, and `ListDeployments` sets it in the same situation. It also used to be set here before #4186 added label filtering. Details and the history I looked at are in #7100.

I kept this to the one line. The label filtering loop below it has its own paging behaviour, which I did not touch.

**How I tested it**

Added `TestListApplicationsCursor` covering both the no label and the with label case.

Without the fix the no label case fails:

```
--- FAIL: TestListApplicationsCursor/cursor_is_returned_when_no_label_filter_is_given
    Error:      Not equal:
                expected: "next-page-cursor"
                actual  : ""
```

With the fix:

```
--- PASS: TestListApplicationsCursor
    --- PASS: .../cursor_is_returned_when_no_label_filter_is_given
    --- PASS: .../cursor_is_returned_when_a_label_filter_is_given
ok  	github.com/pipe-cd/pipecd/pkg/app/server/grpcapi
```

`go test ./pkg/app/server/grpcapi/...` passes, and `go build ./cmd/pipecd ./cmd/pipectl` is fine.

One thing I wanted to be sure about before changing this: whether handing back a real cursor could make a client loop forever. It cannot, because `applicationStore.List` returns an empty cursor once there are no more records:

```go
// In case there is no more elements found, cursor should be set to empty too.
if len(apps) == 0 {
    return apps, "", nil
}
```

So a client that loops until the cursor is empty still terminates.

Fixes #7100

- **How are users affected by this change**: `ListApplications` and `pipectl application list` return a usable cursor when no label filter is given, so paging past the first page works. A client that loops until the cursor is empty now walks through all applications instead of stopping after the first page.


khanhtc1202 (approved):
Nice catch, lgtm 👍

PR #7103, opened 2026-08-01
Fix data race in MetadataStoreRegistry stores map

### LINKED ISSUE #7096: [pipedv1] Data race in MetadataStoreRegistry causes fatal panic under concurrent deployments
**Description:**

**What happened**:

`MetadataStoreRegistry` in `pkg/app/pipedv1/metadatastore/registry.go` holds a `stores map[string]*metadataStore` field with **no synchronization primitive**, causing a fatal `concurrent map read and map write` panic when multiple deployments run simultaneously.

The inner `metadataStore` correctly protects its own metadata maps with `sync.RWMutex` (`store.go:47-49`), but the **outer registry map is unprotected**. Two independent goroutine families access it without coordination:

1. **Writers** — `Register(d)` (`registry.go:44`) and `Delete(d.Id)` (`registry.go:51`) are called from planner/scheduler goroutines as deployments start and finish.
2. **Readers** — All 7 gRPC handler methods (`GetStageMetadata`, `PutStageMetadata`, `PutStageMetadataMulti`, `GetDeploymentPluginMetadata`, `PutDeploymentPluginMetadata`, `PutDeploymentPluginMetadataMulti`, `GetDeploymentSharedMetadata`) read `r.stores[req.DeploymentId]` (`registry.go:47-119`) from piped's plugin-service gRPC server goroutines.

When two or more applications deploy concurrently, a gRPC handler reading the map for deployment A races with `Register`/`Delete` for deployment B → Go runtime panics with `fatal error: concurrent map read and map write` → entire `piped` process aborts, killing all active deployments on that agent.

`MetadataStoreRegistry` should safely serialize concurrent access to the `stores` map. Multiple deployments should coexist without data races or crashes.

**How to reproduce it**:

Add the following test to `pkg/app/pipedv1/metadatastore/registry_test.go` and run with the Go race detector:

```go
package metadatastore

import (
	"context"
	"fmt"
	"sync"
	"testing"

	"github.com/pipe-cd/pipecd/pkg/model"
	service "github.com/pipe-cd/pipecd/pkg/plugin/pipedservice"
)

func TestRegistryConcurrentAccess(t *testing.T) {
	t.Parallel()
	ac := &fakeAPIClient{
		shared:  make(map[string]string),
		plugins: make(map[string]metadata),
		stages:  make(map[string]metadata),
	}
	r := NewMetadataStoreRegistry(ac)
	ctx := context.Background()

	var wg sync.WaitGroup
	const deployments = 20
	const opsPerDeployment = 8

	for i := 0; i < deployments; i++ {
		// Writer: Register + Delete
		wg.Add(1)
		go func(idx int) {
			defer wg.Done()
			id := fmt.Sprintf("deploy-%d", idx)
			d := &model.Deployment{
				Id: id,
				MetadataV2: &model.DeploymentMetadata{
					Shared:  &model.DeploymentMetadata_KeyValues{KeyValues: map[string]string{}},
					Plugins: map[string]*model.DeploymentMetadata_KeyValues{},
				},
				Stages: []*model.PipelineStage{{Id: "stage-1"}},
			}
			r.Register(d)
			r.Delete(id)
		}(i)

		// Readers: simulate gRPC handler calls
		for j := 0; j < opsPerDeployment; j++ {
			wg.Add(1)
			go func(idx int) {
				defer wg.Done()
				id := fmt.Sprintf("deploy-%d", idx)
				_, _ = r.PutStageMetadata(ctx, &service.PutStageMetadataRequest{
					DeploymentId: id, StageId: "stage-1", Key: "k", Value: "v",
				})
			}(i)
		}
	}
	wg.Wait()
}
```

Run command: `go test -race ./pkg/app/pipedv1/metadatastore/...`

**Expected output (on unpatched master)**:
```text
==================
WARNING: DATA RACE
Read at 0x... by goroutine ...:
    (*MetadataStoreRegistry).PutStageMetadata
        .../registry.go:59
Previous write at 0x... by goroutine ...:
    (*MetadataStoreRegistry).Register
        .../registry.go:44
==================
fatal error: concurrent map read and map write
```

**Root Cause & Suggested Fix**:

The `MetadataStoreRegistry` struct (`registry.go:26-32`) contains only `apiClient` and `stores` — no mutex. `Register`/`Delete` modify the map from lifecycle goroutines, while all 7 gRPC handlers read it from server goroutines, with no synchronization between them.

To fix, add a `sync.RWMutex` to the struct:
1. `RLock` in all 7 gRPC handlers for the map lookup (snapshot the pointer, release lock, then delegate to inner store).
2. `Lock` in `Register` and `Delete` for map mutations.
3. Add a dedicated race test using channel-barrier synchronization for deterministic reproduction.

This pattern mirrors the existing locking convention one level down (`metadataStore` uses `sync.RWMutex` for `sharedMu`/`pluginsMu`/`stagesMu`) and introduces zero deadlock risk (registry lock is released before any inner store method or gRPC call).

**Environment**:
- `piped` version: `master` (commit `56fb80e`)
- `control-plane` version: `master`
- Others: Go 1.26.2, `go test -race`

pujitha24:
Motivation:
MetadataStoreRegistry.stores was read by the 7 gRPC plugin-service
handler methods (GetStageMetadata, PutStageMetadata, etc.) and written
by Register/Delete, called from the deployment controller's
planner/scheduler goroutines, with no synchronization between them.
Under concurrent deployments, a handler reading the map for one
deployment could race with Register/Delete for another, triggering
Go's `fatal error: concurrent map read and map write`. This is a
runtime fatal error, not a recoverable panic, so it aborts the whole
piped process, killing every deployment running on that agent.

Approach:
Add a sync.RWMutex (storesMu) to MetadataStoreRegistry, matching the
per-field RWMutex pattern already used one level down in
metadataStore (store.go). Register/Delete take the write lock around
the map mutation; all 7 read call sites go through a new private get()
helper that takes the read lock, snapshots the *metadataStore pointer,
and releases the lock before delegating to the inner store (which has
its own locking), so the registry lock is never held across a call
into inner-store or gRPC logic.

While wiring this up, found that MetadataStoreRegistry was passed by
value from cmd/piped/piped.go into controller.NewController
(dereferencing the pointer), while grpcapi.NewPluginAPI kept the
original pointer. Since sync.RWMutex is a value type, that by-value
copy would have given the controller (writer side) and the gRPC
handlers (reader side) two independent mutex instances guarding the
same underlying map, defeating the fix for the real call graph even
though the isolated package test passed. Changed the controller struct
field and NewController parameter to *metadatastore.MetadataStoreRegistry
so both sides share the single instance created in piped.go.

Validation:
- Reproduced the exact race from the issue: added a concurrency test
  driving 20 concurrent Register+Delete deployment lifecycles against
  8 concurrent PutStageMetadata calls each, and ran it before the fix
  with `go test -race ./pkg/app/pipedv1/metadatastore/...` — it failed
  with a DATA RACE between registry.go's Register (write) and
  PutStageMetadata (read), matching the issue's reported failure mode.
- Kept this as TestRegistryConcurrentAccess in registry_test.go. Also
  added a mutex to the fakeAPIClient test double (store_test.go),
  which had its own, separate unsynchronized maps that a second
  concurrent metadataStore instance would race on independently of the
  registry fix.
- After the fix, `go test -race -count=1 ./pkg/app/pipedv1/metadatastore/...`
  passes, including TestRegistryConcurrentAccess.
- `go build ./pkg/app/pipedv1/...` and
  `go test -race -count=1 ./pkg/app/pipedv1/controller/... ./pkg/app/pipedv1/cmd/piped/...`
  pass, confirming the pointer-sharing fix compiles and the wider
  call graph is unaffected.
- `go vet ./pkg/app/pipedv1/...` shows no copylocks warnings related to
  MetadataStoreRegistry (pre-existing, unrelated copylocks warnings in
  the planpreview package are unchanged before/after this diff).

Fixes #7096

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

---
AI assistance: this change was drafted with Claude Code.

### DIFF
--- a/pkg/app/pipedv1/cmd/piped/piped.go
+++ b/pkg/app/pipedv1/cmd/piped/piped.go
@@ -414,7 +414,7 @@ func (p *piped) run(ctx context.Context, input cli.Input) (runErr error) {
 			commandReporter,
 			notifier,
 			decrypter,
-			*metadataStoreRegistry,
+			metadataStoreRegistry,
 			p.gracePeriod,
 			input.Logger,
 			tracerProvider,
--- a/pkg/app/pipedv1/controller/controller.go
+++ b/pkg/app/pipedv1/controller/controller.go
@@ -96,7 +96,7 @@ type controller struct {
 	commandReporter       commandstore.Reporter
 	notifier              notifier
 	secretDecrypter       secretDecrypter
-	metadataStoreRegistry metadatastore.MetadataStoreRegistry
+	metadataStoreRegistry *metadatastore.MetadataStoreRegistry
 
 	// The registry of all plugins.
 	pluginRegistry plugin.PluginRegistry
@@ -138,7 +138,7 @@ func NewController(
 	commandReporter commandstore.Reporter,
 	notifier notifier,
 	secretDecrypter secretDecrypter,
-	metadataStoreRegistry metadatastore.MetadataStoreRegistry,
+	metadataStoreRegistry *metadatastore.MetadataStoreRegistry,
 	gracePeriod time.Duration,
 	logger *zap.Logger,
 	tracerProvider trace.TracerProvider,
--- a/pkg/app/pipedv1/metadatastore/registry.go
+++ b/pkg/app/pipedv1/metadatastore/registry.go
@@ -17,6 +17,7 @@ package metadatastore
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/pipe-cd/pipecd/pkg/model"
 	service "github.com/pipe-cd/pipecd/pkg/plugin/pipedservice"
@@ -26,6 +27,9 @@ import (
 type MetadataStoreRegistry struct {
 	apiClient apiClient
 
+	// storesMu protects stores against concurrent access from the
+	// Register/Delete lifecycle calls and the gRPC handler methods below.
+	storesMu sync.RWMutex
 	// stores is a map of metadata store for each deployment.
 	// The key is the deployment ID.
 	stores map[string]*metadataStore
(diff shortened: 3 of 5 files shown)


pujitha24:
Just checking in on this one — it's rebased and all checks are green, happy to make any changes if something would help move review along.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #7113, opened 2026-08-03
Detect plugin process exit and reflect it in piped's /healthz

### LINKED ISSUE #7112: piped never notices when a plugin process exits, /healthz stays ok
**What happened**:

Piped runs each plugin (Kubernetes, Terraform, etc.) as its own separate program, not inside piped itself. Think of piped as a manager and each plugin as a worker running in its own room.

If one of those workers (a plugin) suddenly dies, for example it crashes or gets killed for using too much memory, the manager (piped) never checks on it again. Piped just keeps running like nothing happened. It does not print an error, it does not tell anyone, and its health-check page (`/healthz`) keeps saying everything is fine.

Meanwhile, every deployment that needed that plugin will keep failing quietly in the background, and nobody gets told why, until someone manually notices and restarts piped.

**Why this matters**:

Piped is usually run inside Kubernetes, and Kubernetes is already set up to restart piped automatically whenever its health-check page says something is wrong. So the fix does not even need new infrastructure, piped just needs to actually say "something is wrong" when a plugin dies, instead of always saying "ok".

**How I found this**:

I looked at the code that starts a plugin (`pkg/app/pipedv1/cmd/piped/piped.go`) and traced what piped does with it afterwards.

Piped does have a way to know if a plugin process died, it is right here in `pkg/lifecycle/binary.go`:

```go
go func() {
	err := cmd.Wait()   // this returns whenever the plugin process exits, for any reason
	c.result.Store(&err)
	close(c.stoppedCh)  // this is piped's only signal that the plugin is dead
}()
```

But if you look at `piped.go`, this signal is only checked in one place, when piped itself is shutting down:

```go
group.Go(func() error {
	<-ctx.Done()   // only runs when piped is shutting down on purpose
	...
	// stops all plugins here
})
```

So if a plugin dies at 3am while piped is otherwise fine, nothing in the code ever notices, because the only place that checks is the shutdown path.

I even checked if this "is it dead" signal is used anywhere else in the whole project, and it turns out it is, just not for plugins:

```
$ grep -rn "\.IsRunning()" --include=*.go .
./pkg/app/launcher/cmd/launcher/launcher.go:291
```

That one use is for supervising piped itself (for auto-upgrades). So the idea of "check if the process died" already exists and is trusted in this codebase, it is just never applied to plugins.

And the health-check endpoint really does just say "ok" no matter what:

```go
admin.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("ok"))
})
```

I also checked how piped is actually deployed, and confirmed Kubernetes is already told to watch this exact endpoint (`manifests/piped/templates/deployment.yaml`, `quickstart/manifests/pipedv1-exp.yaml`):

```yaml
livenessProbe:
  httpGet:
    path: /healthz
readinessProbe:
  httpGet:
    path: /healthz
```

So the fix here is not about adding a whole new alerting system, it is about making an endpoint that already exists, and is already being watched, actually tell the truth.

**How to reproduce it**:

1. Run pipedv1 with at least one plugin configured.
2. Kill that plugin's process directly, for example `kill -9 <plugin-pid>`, while piped keeps running.
3. Check `/healthz`, it still returns `200 ok`.
4. Try a deployment that needs that plugin, it fails, but nothing in the logs says "the plugin is dead", it just looks like a random connection error.

**Why I think it is worth fixing, not just a one-time crash**:

Piped used to run all this deployment logic inside itself. If something broke, piped itself would crash loudly, and Kubernetes would restart it (through the same health-check). The newer plugin design moved that logic out into separate programs so that one broken plugin cannot crash the rest of piped, which is good, but it also means piped needs to actually watch those programs now, and right now it does not. As more plugins get written by people outside the core team, a bug in just one of them could quietly break only that one feature for a long time before anyone notices.

**Environment**:
- `piped` version: pipedv1, `master` at commit `56fb80e7`
- `control-plane` version: not related
- Others: applies to any plugin started the normal local way (`runPlugins` / `RunBinary`)

I already have a fix and tests ready, opening the PR for it now.

vipulpandey21:
**What this PR does**:

Makes piped actually notice when a plugin's process dies on its own, and makes `/healthz` say so, instead of only reacting when piped itself is shutting down.

**Why we need it**:

Right now, if a plugin process crashes (or gets OOM-killed, or anything else that kills it unexpectedly), piped does not notice at all. It keeps running, `/healthz` keeps saying `ok`, and any deployment that needs that plugin just keeps failing quietly. Kubernetes is already set up to restart piped whenever `/healthz` says something is wrong, so this fix mostly just makes that existing health check tell the truth. Full explanation with code references is in #7112.

**What changed, in plain terms**

1. `pkg/lifecycle/binary.go` — piped already had a way to know when a plugin process exits (a channel that closes when it happens), it just was not exposed outside that file. I added two small getters:
   - `Done()` — a channel you can wait on, closes when the process exits, for any reason.
   - `Err()` — tells you the error it exited with, if any.
   Nothing existing changes, this just exposes something that was already being tracked.

2. `pkg/app/pipedv1/cmd/piped/pluginhealth.go` (new file) — a very small helper that keeps a list of "plugins that are currently known to be dead". Nothing fancy, just a map with a lock around it, and its own tests.

3. `pkg/app/pipedv1/cmd/piped/piped.go`:
   - `/healthz` now checks that list. If any plugin is dead, it returns a `503` and says which plugin(s) are down, instead of always saying `ok`.
   - Each plugin now gets watched the whole time piped is running, not just at shutdown. If a plugin dies early, piped logs it clearly and marks it as unhealthy. If piped is the one shutting down (normal case), it stops the plugin the same way as before.

I did **not** make piped try to automatically restart a crashed plugin. If a plugin keeps crashing, silently restarting it over and over could hide a real bug from whoever is running it. Instead, this just makes `/healthz` honest, and lets Kubernetes do what it is already set up to do: restart the whole pod when something is wrong.

**How I tested it**

Added tests for both new pieces:

- `pkg/lifecycle/binary_test.go` — checks that the new `Done()`/`Err()` correctly reflect a normal exit, a crash (non-zero exit code), and a graceful stop, using real short-lived processes the same way the existing tests in that file already do.
- `pkg/app/pipedv1/cmd/piped/pluginhealth_test.go` — checks marking a plugin unhealthy, marking the same one twice (should not duplicate), multiple plugins, and that it is safe if many goroutines touch it at once.

```
go test ./pkg/lifecycle/... ./pkg/app/pipedv1/cmd/piped/... -v
--- PASS: TestCommandDone
--- PASS: TestPluginHealth
--- PASS: TestPluginHealthConcurrentAccess
ok  	github.com/pipe-cd/pipecd/pkg/lifecycle
ok  	github.com/pipe-cd/pipecd/pkg/app/pipedv1/cmd/piped
```

Also ran:
- `go test -race` on both packages — passed
- `go vet` on both packages — clean
- `go build ./pkg/app/pipedv1/... ./cmd/...` — to make sure nothing else broke
- `gofmt` — clean

I did not add a full end-to-end test that starts a real plugin binary, kills it, and checks `/healthz` over HTTP. The function that wires up the whole piped agent (`run()` in `piped.go`) did not have any test coverage before this change either, so instead of trying to add the first big integration test for it in this same PR, I kept all the new logic in small pieces that are fully testable on their own (`Command.Done/Err`, `pluginHealth`).

Fixes #7112

- **How are users affected by this change**: `/healthz` now returns `503` (and lists which plugin) if a configured plugin's process has died unexpectedly, instead of always returning `200 ok`. Anyone using the standard Helm chart or quickstart manifests already has a liveness/readiness probe pointed at `/healthz`, so this means the pod now actually gets restarted automatically in that situation, instead of silently staying broken.
- **Is this breaking change**: No. Behavior is unchanged as long as all plugins keep running normally, which is the common case.


vipulpandey21:
Hi @khanhtc1202 Sir, just pushed a fix for the lint failure. It was an unused parameter flagged by golangci-lint's `unparam` check. Everything else was already green.

Also noticed the assign (labeler) check is currently failing on this and all other open PRs, so it doesn't seem specific to this PR. `actions/labeler@v7` appears to expect each label config to be an array of match objects, while `.github/labeler.yaml` is still using the flat glob-list format from v4/v5.

Not blocking this PR, just flagging it in case it's useful.

khanhtc1202:
@vipulpandey21 Thanks for mentioning. I fixed that and updated your PR so it passes the CI. Also good to see this implementation, lemme check this when I have time to look at pipedv1 👍

PR #7114, opened 2026-08-03
Rename pipecd control plane chart to controlplane

khanhtc1202:
**What this PR does**:

Rename PipeCD control plane to controlplane, previously mentioned as pipecd. Changes include:
- code (cmd/ and pkg/app)
- manifests and Dockerfiles
- workflow

After this change is merged, the container image named pipecd will be controlplane from now on. The Helm chart for the PipeCD control plane will be named controlplane. The pipecd named Helm chart will be kept for backward compatibility, but internally it refers to the controlplane-named container.

**Why we need it**:

Up to now, the pipecd control plane is referred to as pipecd in various places across code/manifest/chart/workflow, which leads to confusion for users. This change aims to rename the control plane to controlplane in all places.

Fixes #


t-kikuc (approved):
we must announce for this breaking change

PR #7116, opened 2026-08-03
Fix empty dts panic

### LINKED ISSUE #7115: Bug: Unvalidated empty DeployTargets causes out-of-bounds panic in multiple plugins and SDK
**What happened**:
Several `pipedv1` plugins (Terraform deployment, ECS plan preview, and Kubernetes plan preview) access the first element of the `DeployTargets` slice (`dts[0]`) without verifying that the slice is not empty. If a user accidentally provides an empty configuration, or if an SDK bug passes an empty slice, this results in an unhandled out-of-bounds panic that crashes the plugin process.

Additionally, there is an incomplete implementation in the SDK: `StagePluginServiceServer.ExecuteStage` in `pkg/plugin/sdk/deployment.go` hardcodes `nil` for the `deployTargets` parameter (marked with `// TODO: pass the deployTargets`). This means any stage plugins served via this interface that expect targets will receive `nil` and crash when trying to access them.

1. Plugins should check `if len(dts) == 0` and return a clean error (e.g., `sdk.StageStatusFailure` with a descriptive message) instead of panicking. (Note: The Kubernetes deployment plugins already correctly validate this).
2. The SDK's `StagePluginServiceServer.ExecuteStage` should parse and pass the `DeployTargets` correctly, matching the logic implemented in `DeploymentPluginServiceServer`.

**How to reproduce it**:
1. Trigger a Terraform deployment stage (apply/plan/rollback), ECS plan preview, or Kubernetes plan preview with an empty list of deploy targets.
2. Observe the plugin panic when it attempts to access `dts[0]`.
3. Alternatively, create a stage plugin that uses `StagePluginServiceServer` and accesses `dts[0]`. Observe the plugin panic during execution because `dts` is passed as `nil` by the SDK.

**Environment**:
- `piped` version: Latest
- `control-plane` version: Latest
- Others: 
Affected Files:
- `pkg/plugin/sdk/deployment.go`
- `pkg/app/pipedv1/plugin/terraform/deployment/apply.go

vishnukothakapu:
**What this PR does**:
Adds validation for empty `DeployTargets` across multiple `pipedv1` stage plugins to prevent index out-of-bounds panics when accessing `dts[0]`. It also fixes a bug in the SDK where `StagePluginServiceServer.ExecuteStage` passed `nil` for deploy targets instead of correctly parsing them.

**Why we need it**:
If a user misconfigures their application with an empty targets list, or if the SDK fails to pass the targets properly, affected plugins (Terraform deployment, ECS plan preview, and Kubernetes plan preview) would experience an unhandled panic and crash the process during the execution or plan preview stages. These checks ensure the plugins return a clean failure status with a descriptive log message instead.

Fixes #7115 

No. This is purely an internal bug fix and adds protective checks to prevent unhandled panics.

Users with misconfigured deploy targets will now see a descriptive failure log in the PipeCD UI instead of the plugin crashing silently or hanging.
No.
N/A


khanhtc1202:
@vishnukothakapu Nice catch, but I think instead of doing this in each plugin, this should be part of the SDK, wdyt?
https://github.com/pipe-cd/pipecd/blob/master/pkg/plugin/sdk/planpreview.go#L51
cc @mohammedfirdouss @Warashi

pkg/plugin/sdk/deployment.go
```diff
@@ -261,7 +261,19 @@ func (s *StagePluginServiceServer[Config, DeployTargetConfig, ApplicationConfigS
 		toolRegistry:      s.toolRegistry,
 	}
 
-	return executeStage(ctx, s.name, s.base, s.pluginConfig, nil, client, request, s.logger) // TODO: pass the deployTargets
```
khanhtc1202, line 264: This is StagePlugin (like wait, wait-approval, analysis, etc), which does not use the deploy target, so it should be fine to pass nil here. But I don't remember why we leave a TODO message here. Do you have any idea about this? @Warashi @ffjlabo

armistcxy:
> @vishnukothakapu Nice catch, but I think instead of doing this in each plugin, this should be part of the SDK, wdyt? https://github.com/pipe-cd/pipecd/blob/master/pkg/plugin/sdk/planpreview.go#L51 cc @mohammedfirdouss @Warashi

I also think this kind of validation should be part of the SDK, basically no deployment plugin has zero `deployTarget` right. If that the case then just check when the plugin start running to avoid check per stage (and let the stage focus on its work also)

armistcxy:
I belive the first place to parse and store deploy targets from piped configuration file is `Plugin.run()`

https://github.com/pipe-cd/pipecd/blob/3ecbefafd6d3c3f022e9deeed228149eef134ada/pkg/plugin/sdk/plugin.go#L330-L340

so just add validation after parsing deploy targets like this is ok IMO

```go
		if p.deploymentPlugin != nil && len(commonFields.deployTargets) == 0 {
			logger.Error("deployment plugin requires at least one deploy target to be configured", zap.String("name", cfg.Name))
			return fmt.Errorf("deployment plugin requires at least one deploy target to be configured")
		}
```

Update: Livestate plugin and plan preview plugin should also check this (not only deployment plugin)

vishnukothakapu:
> @vishnukothakapu Nice catch, but I think instead of doing this in each plugin, this should be part of the SDK, wdyt? https://github.com/pipe-cd/pipecd/blob/master/pkg/plugin/sdk/planpreview.go#L51 cc @mohammedfirdouss @Warashi

Yes, agreed! I’ve updated the PR to handle this centrally. Added startup validation in `pkg/plugin/sdk/plugin.go` for missing deploy targets, plus per-request checks in wrappers like `GetPlanPreview` and `ExecuteStage` to return a clean `InvalidArgument` error for empty targets. Also reverted the duplicate checks from individual plugins.

pkg/plugin/sdk/deployment.go
vishnukothakapu: I've reverted the changes on this line since StagePlugin doesn't need the deploy targets. Regarding the TODO left here, I'm happy to remove it in this PR to clean things up if you'd like, or we can just leave it as is.

vishnukothakapu:
> > @vishnukothakapu Nice catch, but I think instead of doing this in each plugin, this should be part of the SDK, wdyt? https://github.com/pipe-cd/pipecd/blob/master/pkg/plugin/sdk/planpreview.go#L51 cc @mohammedfirdouss @Warashi
> 
> I also think this kind of validation should be part of the SDK, basically no deployment plugin has zero `deployTarget` right. If that the case then just check when the plugin start running to avoid check per stage (and let the stage focus on its work also)


Yes, exactly! I've updated the PR to add this validation during plugin startup in pkg/plugin/sdk/plugin.go. This ensures the plugin won't even start if it's missing deploy targets, which keeps the individual stage logic clean and focused on its work.

vishnukothakapu:
> I belive the first place to parse and store deploy targets from piped configuration file is `Plugin.run()`
> 
> https://github.com/pipe-cd/pipecd/blob/3ecbefafd6d3c3f022e9deeed228149eef134ada/pkg/plugin/sdk/plugin.go#L330-L340
> 
> so just add validation after parsing deploy targets like this is ok IMO
> 
> ```go
> 		if p.deploymentPlugin != nil && len(commonFields.deployTargets) == 0 {
> 			logger.Error("deployment plugin requires at least one deploy target to be configured", zap.String("name", cfg.Name))
> 			return fmt.Errorf("deployment plugin requires at least one deploy target to be configured")
> 		}
> ```
> 
> Update: Livestate plugin and plan preview plugin should also check this (not only deployment plugin)


Exactly this! I have added this exact validation check in Plugin.run right after parsing the deploy targets, including the checks for Livestate and PlanPreview plugins as you suggested. Thanks for the snippet, it made the fix very straightforward!

> I belive the first place to parse and store deploy targets from piped configuration file is `Plugin.run()`
> 
> https://github.com/pipe-cd/pipecd/blob/3ecbefafd6d3c3f022e9deeed228149eef134ada/pkg/plugin/sdk/plugin.go#L330-L340
> 
> so just add validation after parsing deploy targets like this is ok IMO
> 
> ```go
> 		if p.deploymentPlugin != nil && len(commonFields.deployTargets) == 0 {
> 			logger.Error("deployment plugin requires at least one deploy target to be configured", zap.String("name", cfg.Name))
> 			return fmt.Errorf("deployment plugin requires at least one deploy target to be configured")
> 		}
> ```
> 
> Update: Livestate plugin and plan preview plugin should also check this (not only deployment plugin)


Exactly this! I have added this exact validation check in Plugin.run right after parsing the deploy targets, including the checks for Livestate and PlanPreview plugins as you suggested. Thanks for the snippet, it made the fix very straightforward!

armistcxy:
> ### ✅ Deploy Preview for _pipecd-site_ ready!
> Name	Link
> 🔨 Latest commit	[2c51b06](https://github.com/pipe-cd/pipecd/commit/2c51b067ad2f4f488b0f32be0d3835fcfe4d9ca2)
> 🔍 Latest deploy log	https://app.netlify.com/projects/pipecd-site/deploys/6a7edab09a8f580008350e11
> 😎 Deploy Preview	https://deploy-preview-7116--pipecd-site.netlify.app
> 📱 Preview on mobile	
> Toggle QR Code...
> 🤖 Make changes	[Run an agent on this branch](https://app.netlify.com/projects/pipecd-site/agent-runs#q1YqKMrPLShRslJS0lFKKkrMS84AstMyK3RTgcKVuiklxboFiXmZyUDplNSCnPxKzxSgArNE89SUxCQDy0SLNFMLAwMDC2NTg1RDQ6VaAA)
> _To edit notification comments on pull requests, go to your [Netlify project configuration](https://app.netlify.com/projects/pipecd-site/configuration/notifications#deploy-notifications)._

Hi @rahulshendre, out of scope but why this change still trigger netlify ?

armistcxy (dismissed):
LGTM overall, only need to modify the test, you can have a look here: https://github.com/pipe-cd/pipecd/actions/runs/31786725487/job/94729559336?pr=7116 @vishnukothakapu

armistcxy:
A little bit issue with the labeler check, still safe to merge nhe anh @khanhtc1202

khanhtc1202:
> A little bit issue with the labeler check, still safe to merge nhe anh @khanhtc1202 

I just merged the labeler patch, could you please help me update branch for this PR, so we can test the labeler workflow 😁

pkg/plugin/sdk/planpreview.go
khanhtc1202, line 92: Not directly related to this PR but I think we should use deployTargets built in this function instead of requests.getDeployTargets(), wdyt?
@armistcxy @vishnukothakapu

pkg/plugin/sdk/deployment.go
khanhtc1202: That's a good idea to remove the TODO, but if you can update that comment to be something like: stage plugin (like wait, wait-approval, etc) doesn't use deploy targets so pass as nil here. It could be better

pkg/plugin/sdk/planpreview.go
```diff
@@ -80,12 +84,17 @@ func (s *PlanPreviewPluginServer[Config, DeployTargetConfig, ApplicationConfigSp
...
+		dtNames = append(dtNames, dt.Name)
+	}
+
 	response, err := s.base.GetPlanPreview(ctx, s.pluginConfig, deployTargets, &GetPlanPreviewInput[ApplicationConfigSpec]{
 		Request: GetPlanPreviewRequest[ApplicationConfigSpec]{
 			ApplicationID:           request.GetApplicationId(),
 			ApplicationName:         request.GetApplicationName(),
 			PipedID:                 request.GetPipedId(),
-			DeployTargets:           request.GetDeployTargets(),
+			DeployTargets:           dtNames,
```
khanhtc1202: Sorry, I missed this. The request.GetDeployTargets() is already validated by the inner condition of the for loop at L61, so there's no need to build it again here. Let's revert this change 😓 @vishnukothakapu

khanhtc1202:
NOTE: Release sdk after this being merged 📝

pkg/plugin/sdk/plugin.go
```diff
@@ -340,6 +340,11 @@ func (p *Plugin[Config, DeployTargetConfig, ApplicationConfigSpec]) run(ctx cont
 			}
 		}
 
+		if (p.deploymentPlugin != nil || p.livestatePlugin != nil || p.planPreviewPlugin != nil) && len(commonFields.deployTargets) == 0 {
+			logger.Error("plugin requires at least one deploy target to be configured", zap.String("name", cfg.Name))
+			return fmt.Errorf("plugin requires at least one deploy target to be configured")
+		}
+
```
khanhtc1202:
```suggestion
		if len(commonFields.deployTargets) == 0 {
			if p.deploymentPlugin != nil || p.livestatePlugin != nil || p.planPreviewPlugin != nil {
				logger.Error("plugin requires at least one deploy target to be configured", zap.String("name", cfg.Name))
				return fmt.Errorf("plugin requires at least one deploy target to be configured")
			}
		}
```
this could be easier to read 👀

pkg/plugin/sdk/deployment.go
```diff
@@ -193,6 +193,10 @@ func (s *DeploymentPluginServiceServer[Config, DeployTargetConfig, ApplicationCo
 
 	// Get the deploy targets set on the deployment from the piped plugin config.
 	dtNames := request.GetInput().GetDeployment().GetDeployTargets(s.config.Name)
+	if len(dtNames) == 0 {
```
khanhtc1202, line 196: Still, do we need to validate it here, after validating deployTargets in the plugin?go (run function)? cc @armistcxy

vishnukothakapu: Yes, we still need this check here. The validation in `plugin.go` only verifies that the plugin itself has deploy targets configured.

`dtNames` comes from the individual app’s deployment request, so an app can still send an empty targets list. Without this check, we’d pass an empty `deployTargets` slice to stage execution and hit the same `dts[0]` panic this PR is meant to prevent.

So this check is needed to return a clean `InvalidArgument` error for that case.

armistcxy: @khanhtc1202 yes anh, I have checked and as what @vishnukothakapu said, we need to check in both places

The first one in the plugin's Run function is for when starting the plugin

The next ones are for each application deployment (I've just checked that you can create an application without deploy target through Web UI)

Evidence: it can cause panic when you trigger new deployment for the application with no deploy target
image
 
**TL,DR: Yes, we still need because validate in plugin running != validate in executing stage in application's deployment**

armistcxy: But there's one part I still concern @vishnukothakapu 

Your fix check the `dtNames`, but the one we actually pass to the function `executeStage` is `deployTargets`

```go
	dtNames := request.GetInput().GetDeployment().GetDeployTargets(s.config.Name)
	if len(dtNames) == 0 {
		return nil, status.Errorf(codes.InvalidArgument, "no deploy targets provided")
	}

	deployTargets := make([]*DeployTarget[DeployTargetConfig], 0, len(dtNames))
	for _, name := range dtNames {
		dt, ok := s.deployTargets[name]
		if !ok {
			return nil, status.Errorf(codes.Internal, "the deploy target %s is not found in the piped plugin config", name)
		}

		deployTargets = append(deployTargets, dt)
	}

   return executeStage(ctx, s.name, s.base, s.pluginConfig, deployTargets, client, request, s
```

So there will be a bug like this, in application configuration we specify deploy target `ecs-dev`

Image

But in piped.yaml we specify deploy target as `ecs-target`

Image

There will be mismatch => `len(deployTargets)` = 0

Maybe you should do something like if len(deployTargets) ==0 then log error like "Mismatch deploy target between plugin and application, or no specified deploy target in application configuration" 

CC anh @khanhtc1202

pkg/plugin/sdk/livestate.go
```diff
@@ -54,6 +54,10 @@ func (s *LivestatePluginServer[Config, DeployTargetConfig, ApplicationConfigSpec
 // GetLivestate returns the live state of the resources in the given application.
 func (s *LivestatePluginServer[Config, DeployTargetConfig, ApplicationConfigSpec]) GetLivestate(ctx context.Context, request *livestate.GetLivestateRequest) (*livestate.GetLivestateResponse, error) {
 	// Get the deploy targets set on the deployment from the piped plugin config.
+	if len(request.GetDeployTargets()) == 0 {
```
armistcxy, line 57: consider what I have said above

pkg/plugin/sdk/planpreview.go
```diff
@@ -50,6 +50,10 @@ func (s *PlanPreviewPluginServer[Config, DeployTargetConfig, ApplicationConfigSp
 // GetPlanPreview returns the plan preview of the resources in the given application.
 func (s *PlanPreviewPluginServer[Config, DeployTargetConfig, ApplicationConfigSpec]) GetPlanPreview(ctx context.Context, request *planpreview.GetPlanPreviewRequest) (*planpreview.GetPlanPreviewResponse, error) {
 	// Get the deploy targets set on the deployment from the piped plugin config.
+	if len(request.GetDeployTargets()) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "no deploy targets provided")
```
armistcxy, line 54: consider what I have said above

armistcxy (changes requested):

pkg/plugin/sdk/deployment.go
vishnukothakapu: Thanks for checking this! @armistcxy .

Actually, if there is a mismatch (e.g., the application asks for `ecs-dev` but the plugin only has `ecs-target`), the code inside the for loop will hit the !ok condition and immediately return an error:

`return nil, status.Errorf(codes.Internal, "the deploy target %s is not found in the piped plugin config", name)`

Because it returns this error immediately, it never proceeds to `executeStage` with an empty deployTargets slice. `If len(dtNames) > 0, len(deployTargets)` is guaranteed to be greater than 0, otherwise it fails safely before execution. So we are completely covered for the mismatch scenario!

pkg/plugin/sdk/livestate.go
vishnukothakapu: @armistcxy, as discussed in the thread above, this is also covered! If there is a mismatch between the requested targets and the plugin's configured targets, the for loop immediately below this check will hit the !ok condition and return an Internal error early. It will never proceed with an empty slice, so we are perfectly safe here as well!

pkg/plugin/sdk/planpreview.go
vishnukothakapu: Same here! Just like in the deployment and `livestate` methods, the `!ok` check inside the loop below immediately catches any missing deploy targets and safely returns an Internal error. It will never pass an empty slice to the core logic, so we are completely covered!

pkg/plugin/sdk/deployment.go
armistcxy: @vishnukothakapu  Oh I see, sorry for false alert, then I think it's all good now

rahulshendre:
> Hi @rahulshendre, out of scope but why this change still trigger netlify ?

great catch @armistcxy - our netlify.toml has no ignore build rule, so Netlify runs a preview on every PR regardless of what changed, even a Go-only one.
I think we should add an ignore command to skip builds that don't touch docs, what do you think?

armistcxy (approved):

armistcxy:
@rahulshendre Yeah, I think we should ignore non-docs changes

rahulshendre:
> @rahulshendre Yeah, I think we should ignore non-docs changes

@harshitghagre, could you please open a PR for this? 
Since you own the Netlify migration (#6930), you'd be the best fit for this one 😄
Fix = add an ignore build command to docs/netlify.toml so PRs that don't touch docs/ skip the preview build.

rahulshendre (approved):
LGTM 
Great work @vishnukothakapu, thanks : )

harshitghagre:
@rahulshendre thanks for flagging this opened the fix in #7179 netlify now skips the build if there are no changes under /docs.

rahulshendre:
> @rahulshendre thanks for flagging this opened the fix in #7179 netlify now skips the build if there are no changes under /docs.

that was fast @harshitghagre, thanks a lot : )

khanhtc1202 (approved):

PR #7133, opened 2026-08-05
Add developer guide explaining APIService, WebService, and PipedService interactions

vanshika2720:
**What this PR does**:

Adds a contributor-facing guide explaining how PipeCD's internal services interact with each other.

The guide documents the three gRPC services, their roles and communication paths, and traces concrete cross-service flows such as command polling and live-state updates based on the current implementation.

**Why we need it**:

The interactions between PipeCD's services are not clearly documented, making it harder for contributors to understand how requests and data move between the Control Plane, Web/API services, and Piped.

This guide provides a central reference based on the actual implementation and helps contributors understand the architecture when working on or extending these services.

* **How are users affected by this change**: No user-facing behavior changes. This PR only adds contributor documentation.

### DIFF
--- a/docs/content/en/docs-dev/contribution-guidelines/service-interactions.md
+++ b/docs/content/en/docs-dev/contribution-guidelines/service-interactions.md
@@ -1,0 +1,81 @@
+---
+title: "APIService, WebService, and PipedService"
+linkTitle: "APIService, WebService, and PipedService"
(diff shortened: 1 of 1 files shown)


vanshika2720:
@rahulshendre @yashisrani @mohammedfirdouss @khanhtc1202

PR #7137, opened 2026-08-05
Fix HasDiff returning false when OwnerIds or ParentIds actually differ

### LINKED ISSUE #7135: HasDiff returns false when a resource's OwnerIds or ParentIds actually changed
**What's happening**

`KubernetesResourceState.HasDiff` is supposed to tell piped whether a resource's state changed, so it knows whether to report an update. For every field except `OwnerIds`/`ParentIds`, a mismatch correctly returns `true`. For `OwnerIds` and `ParentIds`, a mismatch returns `false` instead — the opposite of what it should do:

```go
for i := range s.OwnerIds {
    if s.OwnerIds[i] != a.OwnerIds[i] {
        return false // should be true, a diff was found here
    }
}
```

**Why it matters**

If a resource's owner or parent changes (e.g. a Pod picked up by a new ReplicaSet during a rollout), `HasDiff` says "nothing changed", so piped never reports the update, and the live state shown to users keeps the old relationship.

**Where**

`pkg/model/application_live_state.go`, lines 47-57.

**How I found it**

I was tracing how piped reports Kubernetes resource state changes. This function has no test coverage at all, which is likely why this went unnoticed — it's been like this since 2020.

I have a fix and a test ready, opening a PR for it.

vipulpandey21:
**What this PR does**:

Fixes `KubernetesResourceState.HasDiff` so a mismatch in `OwnerIds` or `ParentIds` returns `true` instead of `false`.

```go
- return false
+ return true
```
(both loops, same one-line change)

**Why we need it**:

Every other check in this function returns `true` on a mismatch — that's the whole point of `HasDiff`. For `OwnerIds`/`ParentIds` it did the opposite, so when a resource's owner or parent changes, piped never reports it as an update. Details in #7135.

**How I tested it**:

Added `TestKubernetesResourceState_HasDiff`. Fails on current code:
```
--- FAIL: TestKubernetesResourceState_HasDiff/owner_ids_differ_in_content,_same_length
--- FAIL: TestKubernetesResourceState_HasDiff/parent_ids_differ_in_content,_same_length
```
Passes with the fix. `go test ./pkg/model/...` passes, `go build ./...` is clean.

Fixes #7135

- **How are users affected by this change**: a resource whose owner or parent changes is now correctly reported as changed in the live state, instead of appearing unchanged.


pkg/model/application_live_state.go
```diff
@@ -46,13 +46,13 @@ func (s KubernetesResourceState) HasDiff(a KubernetesResourceState) bool {
 
 	for i := range s.OwnerIds {
 		if s.OwnerIds[i] != a.OwnerIds[i] {
-			return false
+			return true
```
khanhtc1202, line 49: Maybe the reason the first implementation goes this way is that we're not sure about the order of OwnerIds/ParentIds in the list. Should we implement logic to ensure the order before comparison?

khanhtc1202:
Good catch, I left a comment 👍

pkg/model/application_live_state.go
khanhtc1202: Never mind, we've already sorted it here, so this comparison logic should be fine.
https://github.com/pipe-cd/pipecd/blob/master/pkg/app/piped/platformprovider/kubernetes/state.go#L44

khanhtc1202 (approved):
Good catch, I almost forgot about this line 😅

vipulpandey21:
Hi Khanh sir, thanks for double-checking the sort order instead of just taking the fix at face value. Appreciate you tracing it back to state.go rather than trusting the comparison blind. Good to have this one closed out, especially since it had been sitting unchanged since 2020.

PR #7138, opened 2026-08-05
Advance the tracked snapshot version while applying live state events

### LINKED ISSUE #7136: Older live-state event can silently overwrite a newer one in the same batch
**What's happening**

`PatchKubernetesApplicationLiveState` (`pkg/app/server/applicationlivestatestore/store.go`) applies a batch of Kubernetes resource-state events to a snapshot. Each event carries a version, and the code is supposed to skip any event older than what's already stored:

```go
if ev.SnapshotVersion.IsBefore(*snapshot.Version) {
    continue
}
```

The problem: `snapshot.Version` is fetched once at the top of the function and never updated while the loop applies events. So every event in the batch gets compared against that same original version, not against the version the previous event in the loop just moved the state to.

**Why it matters**

If a batch has two events for the same resource where an older one arrives after a newer one (a redelivered retry is a normal way this can happen), the check above doesn't catch it. The older event silently overwrites the newer state — no error, no log.

**How I found it**

Traced how piped assigns event versions (`pkg/app/piped/livestatestore/kubernetes/appnodes.go`) through to how the server applies them. Wrote a test with two out-of-order events for the same resource — the final state ends up on the stale value.

I have a fix and a test ready, opening a PR for it.

vipulpandey21:
**What this PR does**:

In `PatchKubernetesApplicationLiveState`, advances `snapshot.Version` to the applied event's version after each event is applied, instead of leaving it at the version fetched once at the start of the loop.

```go
+ snapshot.Version = ev.SnapshotVersion
```

**Why we need it**:

The staleness check `ev.SnapshotVersion.IsBefore(*snapshot.Version)` only protects against events older than the very first version fetched from storage. Within one batch, if an older event for the same resource arrives after a newer one (e.g. a redelivered/retried event), it passes the check and silently overwrites the newer state. Details and how I traced this are in #7136.

**How I tested it**:

Added `TestPatchKubernetesApplicationLiveState_OutOfOrderEvents`, using two events for the same resource delivered out of order. Fails on current code (final state ends up on the stale value), passes with the fix.

`go test ./pkg/app/server/applicationlivestatestore/...` passes, `go build ./...` is clean.

Fixes #7136

- **How are users affected by this change**: the live state server no longer lets a delayed or duplicate event overwrite newer resource state within the same request.


khanhtc1202 (dismissed):
Nice catch, lgtm 👍

khanhtc1202:
There was a lint error on the test file, could you fix? We have `make lint/go FIX=true` command that you can use, but be sure to fetch the latest master branch code to ensure the golangci-lint version is correct @vipulpandey21

vipulpandey21:
Hi Khanh sir, fixed the gofmt alignment issue in store_test.go and pushed. Tests pass locally, should be clean now, let me know if anything else needs a look sir.

PR #7141, opened 2026-08-06
Skip piped-config volume when launcher loads config from git

### LINKED ISSUE #7140: piped chart: piped-config ConfigMap mounted unconditionally, breaks launcher configFromGitRepo
**What happened**:

Installing the `piped` Helm chart with the launcher enabled and its config loaded from a git repository fails when `config.create: false`:

```
Deployment.apps "piped" is invalid:
  spec.template.spec.volumes[1].configMap.name: Required value
  spec.template.spec.containers[0].volumeMounts[1].name: Not found: "piped-config"
```

The chart unconditionally renders a `piped-config` ConfigMap volume and volumeMount in `manifests/piped/templates/deployment.yaml`, even though the launcher does not read `/etc/piped-config/` in that mode (it uses `--config-from-git-repo=true` and the `--git-*` flags per `piped.launcherArgs` in `_helpers.tpl`).

When `launcher.enabled` and `launcher.configFromGitRepo.enabled` are both true, the local `piped-config` ConfigMap is unused, so the chart should not require it. `helm upgrade` should succeed with `config.create: false`.

**How to reproduce it**:

`values.yaml`:

```yaml
launcher:
  enabled: true
  configFromGitRepo:
    enabled: true
    repoUrl: git@github.com:<org>/<repo>.git
    branch: main
    configFile: piped/config.yaml
    sshKeyFile: /etc/piped-secret/ssh-key
config:
  create: false
secret:
  create: false
  name: piped-secret
```

Then:

```bash
helm upgrade -i piped oci://ghcr.io/pipe-cd/chart/piped \
  --namespace piped -f values.yaml
```

**Root cause**:

- `manifests/piped/templates/deployment.yaml` — the `piped-config` volumeMount and volume are rendered outside any `{{ if }}` guard.
- `manifests/piped/templates/_helpers.tpl` — `piped.configMapName` returns `.Values.config.name` when `config.create: false`; if `config.name` is empty, the volume's `configMap.name` renders as an empty string, which the API server rejects.
- `manifests/piped/templates/configmap.yaml` — the ConfigMap resource itself is only rendered when `config.create: true`.

**Proposed fix**:

Guard the `piped-config` volume and volumeMount so they only render when the local ConfigMap is actually consumed — i.e. not when the launcher is loading config from git:

```yaml
{{- if not (and .Values.launcher.enabled .Values.launcher.configFromGitRepo.enabled) }}
- name: piped-config
  ...
{{- end }}
```

**Environment**:
- `piped` version: chart from `oci://ghcr.io/pipe-cd/chart/piped`, master HEAD (verified on commit `5edb3b65a`)
- `control-plane` version: v0.57.x
- Others: helm 3.x, kubernetes 1.29 (kind)

I'd like to work on this — I have a small fix ready.

kishoredurai:
**What this PR does**:

Only render the `piped-config` ConfigMap volume and volumeMount in the piped Deployment when a local ConfigMap is actually consumed. Specifically, skip them when `launcher.enabled` and `launcher.configFromGitRepo.enabled` are both true — in that mode the launcher fetches piped config from git (via `--config-from-git-repo=true` and the `--git-*` flags) and does not read `/etc/piped-config/`.

**Why we need it**:

With `launcher.enabled: true` + `launcher.configFromGitRepo.enabled: true`, the local ConfigMap is unused. Users naturally set `config.create: false` in that mode, but that produces an invalid Deployment (empty `configMap.name`) and `helm upgrade` fails:

```
Deployment.apps "piped" is invalid:
  spec.template.spec.volumes[1].configMap.name: Required value
  spec.template.spec.containers[0].volumeMounts[1].name: Not found: "piped-config"
```

Fixes #7140

- **How are users affected by this change**: Users running the launcher with `configFromGitRepo` can now set `config.create: false` and install the chart cleanly. All other modes (piped without launcher, and launcher with a local ConfigMap) render identically to before.

**Verification** — `helm template` in three modes:

- Default (no launcher): `piped-config` volume + mount present. ✓
- `launcher.enabled=true`, `configFromGitRepo.enabled=false`: `piped-config` still present (launcher reads from mounted ConfigMap). ✓
- `launcher.enabled=true` + `configFromGitRepo.enabled=true` + `config.create=false`: `piped-config` absent, launcher uses `--git-*` flags. ✓

`helm lint manifests/piped` passes.


khanhtc1202 (approved):
Nice catch, thank you 👍

PR #7145, opened 2026-08-10
test(web): add unit tests for isDeploymentRunning utility

vikash7485:
**What this PR does**:

Add unit tests for `isDeploymentRunning` in `web/src/utils/is-deployment-running.ts`, covering all 7 `DeploymentStatus` enum values and the `undefined` edge case.

**Why we need it**:

This utility has zero test coverage despite being a critical decision function called in:
- `DeploymentDetailPage` (controls polling interval)
- - `DeploymentDetail` (controls cancel button visibility)
- - `Pipeline` component (determines if stages are interactive)
Without tests, future enum additions to `DeploymentStatus` could silently break polling behavior with no regression signal.

Fixes #

No. This PR only adds test coverage for an existing utility function.

- **How are users affected by this change**: No user-facing impact.
- - **Is this breaking change**: No.
- - **How to migrate (if breaking change)**: N/A.


rahulshendre (commented):
the 7 cases could fold into one it.each table, we can trim the repetition.
Also Fixes #6915 is not really correct - that's the Snyk alpine upgrade PR, please fix it.

khanhtc1202 (approved):

PR #7147, opened 2026-08-11
fix: paginate ListNotCompletedDeployments in legacy piped

### LINKED ISSUE #1937: Call ListNotCompletedDeployments itervally until all required deployments are fetched.

Goyamjain06:
## What this PR does

Updates the legacy `piped` deployment store to follow the pagination cursor returned by `ListNotCompletedDeployments`.

Previously, `sync()` fetched only the first page and ignored the returned cursor. It now continues fetching until the cursor is empty, ensuring all pending/planned/running deployments are included.

## Changes

- Update `pkg/app/piped/apistore/deploymentstore/store.go` to paginate `ListNotCompletedDeployments` using `resp.Cursor`.
- Add `store_test.go` with coverage for:
  - empty responses
  - deployment status classification
  - multi-page pagination
  - 3+ page pagination
  - `ROLLING_BACK` classification
  - errors during pagination
- Fix the server-side pagination contract in `pkg/datastore/deploymentstore.go` (see the Update section below).

## Tests

- `go test ./pkg/app/piped/apistore/deploymentstore/...` — pass
- `go test ./pkg/datastore/... ./pkg/app/pipedv1/apistore/deploymentstore/... ./pkg/app/server/grpcapi/...` — pass (incl. `-race`)
- `go vet ./pkg/app/piped/apistore/deploymentstore/...` — pass
- `make check` — Go build and web build pass. The golangci-lint step is blocked by the pinned lint image using Go 1.25.0 while the repository requires Go 1.26.2.

## Update: fixing the server-side pagination contract

Review feedback correctly pointed out that following the cursor is unsafe against the current server. `PipedAPI.ListNotCompletedDeployments` builds its datastore query with no `Orders` set. In that case `deploymentStore.List` was still asking the iterator for a cursor, and `Iterator.Cursor()` base64-encodes an empty ordering-key map to a *non-empty* string (`"e30="`). So the first page always came back with a non-empty cursor, the new client loop issued a second request carrying it, and both the Firestore and MySQL backends reject that with `opts.Cursor also requires Orders to be set` — meaning every sync tick would fail for any piped with at least one active deployment.

Rather than introduce an ordering plus a new `PipedId` composite Firestore index (a much larger change that is hard to verify without a live datastore), this fixes the contract at the source: `deploymentStore.List` now returns an empty cursor whenever the query has no `Orders`. Ordered, paginated callers (`WebAPI` / `API` `ListDeployments`) always set `Orders`, so their behaviour is unchanged. Unordered callers — `ListNotCompletedDeployments`, the inner lookup in `ListDeploymentTraces`, and the deployment-chain updater — already discard the cursor and receive the full result set in a single call.

The same loop was added to the pipedv1 deployment store in #6727 against this same unchanged handler, so this fix also removes a latent break there. A regression test is included on the v1 store.

### Additional changes

- `pkg/datastore/deploymentstore.go`: return an empty cursor for unordered `List` queries.
- `pkg/datastore/deploymentstore_test.go`: cover the unordered case (no cursor, `Iterator.Cursor()` not called) and the ordered case (cursor passed through).
- `store_test.go` (v0 and v1): the fake API client now records each request's cursor, and the table tests assert the full `"" -> "page2" -> "page3"` threading.

Fixes #1937

### DIFF
--- a/pkg/app/piped/apistore/deploymentstore/store.go
+++ b/pkg/app/piped/apistore/deploymentstore/store.go
@@ -103,23 +103,30 @@ func (s *store) Lister() Lister {
 }
 
 func (s *store) sync(ctx context.Context) error {
-	// TODO: Call ListNotCompletedDeployments itervally until all required deployments are fetched.
-	resp, err := s.apiClient.ListNotCompletedDeployments(ctx, &pipedservice.ListNotCompletedDeploymentsRequest{})
-	if err != nil {
-		s.logger.Error("failed to list unhandled deployment", zap.Error(err))
-		return err
-	}
-
 	var pendings, planneds, runnings []*model.Deployment
-	for _, d := range resp.Deployments {
-		switch d.Status {
-		case model.DeploymentStatus_DEPLOYMENT_PENDING:
-			pendings = append(pendings, d)
-		case model.DeploymentStatus_DEPLOYMENT_PLANNED:
-			planneds = append(planneds, d)
-		case model.DeploymentStatus_DEPLOYMENT_RUNNING, model.DeploymentStatus_DEPLOYMENT_ROLLING_BACK:
-			runnings = append(runnings, d)
+	cursor := ""
+	for {
+		resp, err := s.apiClient.ListNotCompletedDeployments(ctx, &pipedservice.ListNotCompletedDeploymentsRequest{
+			Cursor: cursor,
+		})
+		if err != nil {
+			s.logger.Error("failed to list unhandled deployment", zap.Error(err))
+			return err
 		}
+		for _, d := range resp.Deployments {
+			switch d.Status {
+			case model.DeploymentStatus_DEPLOYMENT_PENDING:
+				pendings = append(pendings, d)
+			case model.DeploymentStatus_DEPLOYMENT_PLANNED:
+				planneds = append(planneds, d)
+			case model.DeploymentStatus_DEPLOYMENT_RUNNING, model.DeploymentStatus_DEPLOYMENT_ROLLING_BACK:
+				runnings = append(runnings, d)
+			}
+		}
+		if resp.Cursor == "" {
+			break
+		}
+		cursor = resp.Cursor
 	}
 
 	headDeployments := make(map[string]*model.Deployment)
--- a/pkg/app/piped/apistore/deploymentstore/store.go
+++ b/pkg/app/piped/apistore/deploymentstore/store.go
@@ -123,6 +123,12 @@ func (s *store) sync(ctx context.Context) error {
 				runnings = append(runnings, d)
 			}
 		}
+		// Safety guard against a misbehaving server: if a page carries no
+		// deployments, stop paging even when a non-empty cursor is returned.
+		// Without this, such a response would spin this loop forever.
+		if len(resp.Deployments) == 0 {
+			break
+		}
 		if resp.Cursor == "" {
 			break
 		}
--- a/pkg/app/piped/apistore/deploymentstore/store_test.go
+++ b/pkg/app/piped/apistore/deploymentstore/store_test.go
@@ -36,14 +36,20 @@ import (
 // If err is set, it is returned instead of a page on the call index
 // identified by errAt (0-based), letting a test simulate a failure that
 // happens partway through pagination rather than on the very first call.
+//
+// gotCursors records the Cursor field of every request received, in order,
+// so tests can assert that each response cursor is threaded into the next
+// request ("" -> "page2" -> "page3").
 type fakeAPIClient struct {
-	pages []*pipedservice.ListNotCompletedDeploymentsResponse
-	call  int
-	err   error
-	errAt int
+	pages      []*pipedservice.ListNotCompletedDeploymentsResponse
+	call       int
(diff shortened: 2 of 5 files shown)


Goyamjain06:
Hi maintainers, just checking in on this one. All checks are passing and it's ready for review whenever you get a chance — no rush at all. Happy to make any changes if needed!

pkg/app/piped/apistore/deploymentstore/store_test.go
```diff
@@ -0,0 +1,184 @@
...
+				"app-4": rollingBack,
+			},
+		},
+		// Three pages forces at least two follow-up requests, which a
+		// "call once more if the first cursor is non-empty" shortcut
+		// would not satisfy — only a genuine loop passes this case.
+		{
+			name: "paginates_across_three_or_more_pages",
+			pages: []*pipedservice.ListNotCompletedDeploymentsResponse{
+				{Deployments: []*model.Deployment{pending}, Cursor: "page2"},
```
vanshika2720, line 150: Could we assert the request cursor in these tests? The fake currently ignores req.Cursor, so the tests don't verify that each response cursor is actually passed to the next request ("" → "page2" → "page3").

pkg/app/piped/apistore/deploymentstore/store.go
```diff
@@ -103,23 +103,30 @@ func (s *store) Lister() Lister {
...
-		case model.DeploymentStatus_DEPLOYMENT_PENDING:
-			pendings = append(pendings, d)
-		case model.DeploymentStatus_DEPLOYMENT_PLANNED:
-			planneds = append(planneds, d)
-		case model.DeploymentStatus_DEPLOYMENT_RUNNING, model.DeploymentStatus_DEPLOYMENT_ROLLING_BACK:
-			runnings = append(runnings, d)
+	cursor := ""
+	for {
+		resp, err := s.apiClient.ListNotCompletedDeployments(ctx, &pipedservice.ListNotCompletedDeploymentsRequest{
+			Cursor: cursor,
```
vanshika2720, line 110: This pagination assumes the server returns a cursor that can be passed back to this RPC, but ListNotCompletedDeployments currently doesn't set Orders/Limit on the datastore request. This can produce a non-empty cursor that fails on the next request with opts.Cursor also requires Orders to be set. Could we fix the server-side pagination contract before adding the client loop here?

pkg/app/piped/apistore/deploymentstore/store_test.go
Goyamjain06: Good call — done. `fakeAPIClient` now records `req.Cursor` on every call, and each table case carries a `wantCursors` slice the test asserts against, so the `"" -> "page2" -> "page3"` threading is verified explicitly (and the mid-pagination error case checks that `page2` is still forwarded to the failing request). Applied the same change to the pipedv1 store test. Thanks for the catch!

pkg/app/piped/apistore/deploymentstore/store.go
Goyamjain06: You're right, thank you for catching this. `ListNotCompletedDeployments` builds its query with no `Orders`, and in that case `deploymentStore.List` was still returning a cursor — `Iterator.Cursor()` encodes an empty ordering-key map to `"e30="`, which is non-empty — so the follow-up request hit the `opts.Cursor also requires Orders to be set` check in both the Firestore and MySQL backends.

I've fixed it at the source: `deploymentStore.List` now returns an empty cursor when the query has no `Orders`. Ordered, paginated callers (`ListDeployments`) are unaffected since they always set `Orders`; unordered callers get the whole result set in one call, which is what this RPC already relied on. I went this route rather than adding an ordering plus a new `PipedId` composite Firestore index, which would be a much larger and harder-to-verify change. Full write-up is in the PR description's Update section.

Note the same loop landed for pipedv1 in #6727 against this same handler, so this also fixes a latent break there — I've added a regression test on the v1 store.

pkg/app/piped/apistore/deploymentstore/store.go
```diff
@@ -103,23 +103,30 @@ func (s *store) Lister() Lister {
...
+				planneds = append(planneds, d)
+			case model.DeploymentStatus_DEPLOYMENT_RUNNING, model.DeploymentStatus_DEPLOYMENT_ROLLING_BACK:
+				runnings = append(runnings, d)
+			}
+		}
+		if resp.Cursor == "" {
+			break
 		}
+		cursor = resp.Cursor
 	}
```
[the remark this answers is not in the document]
Goyamjain06: Good catch, thanks! You're right — the loop only inspected the cursor, so a server bug that returned an empty page with a non-empty cursor would spin forever. I've added a safety guard that breaks out as soon as a page comes back with zero deployments, regardless of the cursor value. I also added a test case (`stops_on_empty_page_with_non_empty_cursor`) that reproduces exactly that scenario and asserts the bogus cursor is never threaded into a follow-up request.

PR #7148, opened 2026-08-11
fix(web): guard undefined hash in commit detail view

vikash7485:
**What this PR does**:

Guard against `undefined` hash in the deployment detail commit section by adding a null-coalescing fallback (`?? ""`) before calling `.slice(0, 7)`.

**Why we need it**:

When `deployment.trigger.commit.hash` is `undefined` (possible from older Piped v0 protobuf messages where the field was unpopulated), the current code calls `.slice(0, 7)` directly on `undefined`, causing a runtime crash:

```
TypeError: Cannot read properties of undefined (reading 'slice')
```

The outer guard at L254 (`deployment.trigger?.commit &&`) only checks that `commit` exists, not that `hash` is populated. This fix adds a safe fallback so the link renders with an empty string instead of crashing.

Fixes #

No. This prevents a crash for an edge case. Users with valid commit hashes see no difference.

- **How are users affected by this change**: Deployments with missing commit hashes no longer crash the detail page.
- - **Is this breaking change**: No.
- - **How to migrate (if breaking change)**: N/A.

### DIFF
--- a/web/src/components/deployments-detail-page/deployment-detail/index.test.tsx
+++ b/web/src/components/deployments-detail-page/deployment-detail/index.test.tsx
@@ -1,8 +1,9 @@
 import userEvent from "@testing-library/user-event";
 import { dummyDeployment } from "~/__fixtures__/dummy-deployment";
+import { dummyTrigger } from "~/__fixtures__/dummy-trigger";
 import { render, screen, MemoryRouter, waitFor, act } from "~~/test-utils";
 import { DeploymentDetail } from ".";
-import { DeploymentStatus } from "~~/model/deployment_pb";
+import { Deployment, DeploymentStatus } from "~~/model/deployment_pb";
 import * as deploymentsApi from "~/api/deployments";
 import { server } from "~/mocks/server";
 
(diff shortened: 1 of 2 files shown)


rahulshendre (dismissed):
LGMT, thanks @vikash7485 

one thing I am curious about - commit.hash is typed string from the protobuf (proto3 defaults an unset string to "", not undefined), so I couldn't see where it'd actually be undefined. 
If you saw it, please let use know where, mostly so we know if message/url next to it need the same

vikash7485:
@rahulshendre Thanks for the review and approval

The reason `hash` was specifically at risk here is that it had a method call (`hash.slice(0, 7)`), whereas `message` and `url` are passed directly into JSX (`<Typography>` and `href`), which React safely handles even if they are `undefined`.

In standard protobuf generated classes, proto3 does default strings to `""`. However, in frontend tests/mocks, partial payloads, or when objects are deserialized from JSON without `emitDefaults: true`, omitted properties can end up as `undefined` in TypeScript. `(hash ?? "").slice(0, 7)` ensures that even in those edge cases, the component renders safely without throwing a runtime TypeError.

khanhtc1202 (dismissed):
Nice catch, thank you 👍

khanhtc1202:
@vikash7485 same here, need to run make lint/web on your local to fix this

khanhtc1202:
@vikash7485 the DCO check has failed 👀 You need to sign-off the commit 😁

khanhtc1202 (approved):

PR #7149, opened 2026-08-11
fix(web): replace .map() with .forEach() in formalizePoliciesList

vikash7485:
**What this PR does**:

Replace 4 misused `.map()` calls with `.forEach()` in `formalizePoliciesList` (`web/src/utils/formalize-policies-list.ts`).

**Why we need it**:

`.map()` was used at 4 locations purely for side effects (`.push()` into outer arrays). Each `.map()` call silently creates and discards an intermediate `[undefined, undefined, ...]` array, wasting memory. Using `.forEach()` is semantically correct and avoids the unnecessary allocation.

Fixes #

No. This is a semantic code correctness fix with no behavioral change.

- **How are users affected by this change**: No user-facing impact.


rahulshendre (commented):
nice catch @vikash7485
btw what are your thoughts on using map().join() here 👀

vikash7485:
@rahulshendre  I think using `map().join()` is definitely cleaner and more idiomatic. It eliminates the mutable arrays (`policies`, `resources`, `actions`) and the manual trailing comma slicing (`rsc.slice(0, -1)`). 

I'd be happy to refactor it to a fully declarative `.map().join()` pattern in this PR if you prefer that approach. Let me know and I can update the branch!

rahulshendre:
> I'd be happy to refactor it to a fully declarative `.map().join()` pattern in this PR if you prefer that approach. Let me know and I can update the branch!

sounds good, go for it 🙌 the map().join() version reads much cleaner and it drops that slice(0, -1) too
since there's no unit test on this util yet, please double check the output 

cc @armistcxy @mohammedfirdouss

mohammedfirdouss (approved):
LGTM

Issue #6679, opened 2026-04-16
[LFX Mentorship T-2 2026] Plugin Development Book, Documentation, and Community Growth


rahulshendre:
## Overview

This issue tracks the LFX Mentorship Term 2 2026 project focused on delivering a practical PipeCD Plugin Development Book as the primary outcome, with targeted adoption work across docs experience, technical content, and community growth.

## Background

PipeCD's pipedv1/plugin ecosystem is expanding quickly. Community discussions repeatedly highlighted the need for a structured, hands-on learning path for plugin contributors as well as maintainers, plus better discoverability and contributor 
enablement around docs.

## Primary Goal

Translate and publish the [PipeCD Plugin Development Book](https://zenn.dev/warashi/books/try-and-learn-pipecd-plugin) (chapter-based, practical, maintainable):

* Beginner-to-implementation progression
* Real examples aligned with current plugin architecture
* Contributor-focused workflows and references
* Video guides around the same

### 1. Docs Experience (DX)
* Improve plugin/onboarding docs flow and discoverability
* Prioritize high-impact UX improvements discussed by the community:
  * Docs live preview workflow
  * Dark mode/theming consistency
  * Navigation and minor bug fixes

### 2. Technical Content
* Produce focused support content tied to book chapters:
  * Walkthrough videos (target: 2–4)
  * Practical blogs/tutorials
* Reuse workflow: book/docs → blog → short video → social snippet

### 3. Talks and Outreach
* Submit talks to relevant meetups/events
* Target: submit multiple CFPs, deliver talks where accepted
* Track accepted/given talks as adoption outcomes

### 4. Contributor Growth
* Implement Contributor Ladder and recognition flow (with maintainers discussion)
* Encourage contribution pathways from docs/content/community touchpoints

## Tracking and Reporting

* Public tracking via issue checklist / project board
* Bi-weekly updates with mentors/community sync
* Monthly KPI snapshot, including:
  * Book chapters published
  * Docs PRs merged
  * Videos/blogs shipped
  * Talk CFPs submitted / accepted / delivered

## Expected Outcomes

* Publishable Plugin Development Book in PipeCD docs
* Improved first-time contributor and adopter experience
* Better docs usability and content discoverability

## Mentors

* Eeshaan Sawant [@eeshaanSA](https://github.com/eeshaanSA)
* Khanh Tran [@khanhtc1202](https://github.com/khanhtc1202)

## Timeline

[...]

PR #7153, opened 2026-08-11
Docs V1: Add plugin tutorial Chapter 4 - Lifecycle methods

rahulshendre:
**What this PR does**:
Adds chapter 4 of creating a plugin tutorial, covers how to implement the lifecycle methods FetchDefinedStages, DetermineVersions, and DetermineStrategy.

**Why we need it**:
4th chapter of the tutorial

Part of #6679

### DIFF
--- a/docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
+++ b/docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
@@ -1,0 +1,62 @@
+---
+title: "Lifecycle methods: stages, versions, and strategy"
+linkTitle: "Lifecycle methods: stages, versions, and strategy"
+weight: 4
+description: >
+  Implement FetchDefinedStages, DetermineVersions, and DetermineStrategy.
+---
+
+In this chapter you replace the first three empty methods with real ones. Together they tell `piped` which stages the plugin provides, what version it is deploying, and which deployment strategy to use. None of them run a deployment yet; they give `piped` the information it needs before a deployment starts.
+
+## Declare the stages
+
+`FetchDefinedStages` returns the names of the stages the plugin provides. The file plugin defines three: `FILE_DIFF` and `FILE_SYNC` for the two deployment stages, and `FILE_ROLLBACK` for the stage that runs when a deployment fails and needs to roll back.
+
+Define the stage names as constants, because other methods refer to them too, and return them from `FetchDefinedStages`:
+
+```go
+const (
+	stageDiff     = "FILE_DIFF"
+	stageSync     = "FILE_SYNC"
+	stageRollback = "FILE_ROLLBACK"
+)
+
+func (p *plugin) FetchDefinedStages() []string {
+	return []string{stageDiff, stageSync, stageRollback}
+}
+```
+
+## Report the deployed version
+
+`DetermineVersions` returns the version of what the plugin deploys, which `piped` records and shows in the UI. The deployment source describes the files being deployed, including the Git commit they come from, so the file plugin reports that commit hash as its version. It does not set a name or URL:
+
+```go
+func (p *plugin) DetermineVersions(ctx context.Context, _ *sdk.ConfigNone, input *sdk.DetermineVersionsInput[applicationConfig]) (*sdk.DetermineVersionsResponse, error) {
+	return &sdk.DetermineVersionsResponse{
+		Versions: []sdk.ArtifactVersion{{Version: input.Request.DeploymentSource.CommitHash}},
+	}, nil
+}
+```
+
+## Choose a deployment strategy
+
+`DetermineStrategy` decides how a deployment runs. There are two strategies:
+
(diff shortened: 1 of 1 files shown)


khanhtc1202:
@mohammedfirdouss @armistcxy could you help me check this docs 👀

docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "Lifecycle methods: stages, versions, and strategy"
+linkTitle: "Lifecycle methods: stages, versions, and strategy"
+weight: 4
+description: >
+  Implement FetchDefinedStages, DetermineVersions, and DetermineStrategy.
```
armistcxy: wrap function name with backticks

docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
```diff
@@ -0,0 +1,62 @@
...
+)
+
+func (p *plugin) FetchDefinedStages() []string {
+	return []string{stageDiff, stageSync, stageRollback}
+}
+```
+
+## Report the deployed version
+
+`DetermineVersions` returns the version of what the plugin deploys, which `piped` records and shows in the UI. The deployment source describes the files being deployed, including the Git commit they come from, so the file plugin reports that commit hash as its version. It does not set a name or URL:
```
armistcxy: When skimming through this part, user can misunderstand "the version of what the plugin deploys" with "version of the plugin" so I suggest that we change to "`DetermineVersions` reports the versions of the artifacts a deployment uses (computed from the deployment source for each deployment)"

docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+## Choose a deployment strategy
+
+`DetermineStrategy` decides how a deployment runs. There are two strategies:
+
+- **Pipeline sync** runs the deployment pipeline the user defines in the application configuration.
+- **Quick sync** skips the user's pipeline and runs a fixed set of stages the plugin defines, which is useful for small changes that do not need a full pipeline.
+
+A plugin only needs to implement `DetermineStrategy` when it has its own logic for choosing between the two. The file plugin does not, so it returns `nil, nil` and lets `piped` decide:
```
armistcxy: I think this part is wrong, in `plugin/sdk/deployment.go`

```go
	if response == nil {
		// If the plugin does not have specific logic to determine strategy,
		// use PipelineSync by default.
		response = &DetermineStrategyResponse{
			Strategy: SyncStrategyPipelineSync,
			Summary:  "Use PipelineSync because no other logic was matched",
		}
	}
```
=> This mean that it always returns pipeline sync if response is nil, nil (piped does not decide anything here)

I think you can change to "Returning nil, nil means no custom logic: the SDK answers piped on the plugin's behalf with pipeline sync"

armistcxy: But you should check if the other chapters use pipeline sync or quick sync

armistcxy (changes requested):
@rahulshendre please check my comments, I think you should treat the determine strategy carefully

docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
rahulshendre: fixed

docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
rahulshendre: good catch, updated the wording to avoid the ambiguity

docs/content/en/docs-v1.0.x/plugins/creating-a-plugin/chapter-04-lifecycle-methods.md
rahulshendre: done, traced it through plugin/sdk/deployment.go, the SDK intercepts nil and substitutes PipelineSync itself, so "lets piped decide" was actually wrong..while doing that I also added a note that DetermineStrategy is only called after piped's own checks don't resolve the strategy

rahulshendre:
> @rahulshendre please check my comments, I think you should treat the determine strategy carefully

thanks @armistcxy, i have made the changes, PTAL

armistcxy:
Hi @rahulshendre, you forgot to sign your commit

rahulshendre:
> Hi @rahulshendre, you forgot to sign your commit

my bad, it's done now 😄

armistcxy (approved):
LGTM now

PR #7168, opened 2026-08-12
fix(web): guard zero startedAt timestamp in application detail

vikash7485:
**What this PR does**:

Guard against `deployment.startedAt === 0` in the `MostRecentlySuccessfulDeployment` component (`web/src/components/application-detail-page/application-detail/index.tsx`). When `startedAt` is zero, render an em-dash instead of calling `dayjs(0)`.

**Why we need it**:

`0` is the protobuf default for an unset `int64` field. When an application is newly registered but has never been deployed, `startedAt` is `0`. The current code passes this to `dayjs(0 * 1000)` which renders as "January 1, 1970" / "56 years ago" -- misleading the user into thinking a deployment happened decades ago.

Yes. Applications with no prior deployments now show an em-dash instead of "56 years ago".

- **How are users affected by this change**: Users no longer see a confusing epoch date for apps that were never deployed.


khanhtc1202 (dismissed):
LGTM

khanhtc1202:
@vikash7485, can you run `make lint/web FIX=true` on your local? The change has not passed our lint check.

vikash7485:
Hi @khanhtc1202 , I have fixed the lint issues and updated the PR. Could you please take another look?

rahulshendre (approved):
LGTM, nice catch on the epoch-date bug 👍
Thanks @vikash7485 : )

khanhtc1202 (approved):
Thank you!

PR #7169, opened 2026-08-12
Return 0 from sortFunc when values are equal

Ujjwal-Gowda:
**What this PR does**:

`sortFunc` never returned 0. For equal values `a > b` is false, so it fell through to `-1` in ASC and `1` in DESC, reporting both `a < b` and `b < a` for identical input.

`Array.prototype.sort` is only stable when the comparator returns 0 for equal elements, so equal entries were reordered instead of keeping their original order. With 12 entries where 11 share a name, the equal ones came out fully reversed. The result also varied with list length, because V8 switches sorting strategy for larger arrays.

This adds the `a === b` short-circuit, and adds `web/src/utils/common.test.ts` (the file did not exist before) covering ordering, equality, and the stability regression.

**Why we need it**:

This function sorts the piped and application lists in the four application forms and in the encrypt secret drawer, where duplicate names are normal. `sortDateFunc` delegates to `sortFunc` and had the same problem for equal timestamps.

The visible effect is that the dropdown order for same-named entries is unstable: it changes depending on how many items are in the list, so a user picking "the second `api` piped" is not picking a consistent thing.

None — found while reading the application form code. Happy to open an issue first if that is preferred.

- **How are users affected by this change**: Piped and application dropdowns keep same-named entries in their original order instead of reversing them.


khanhtc1202 (approved):
good catch 👍

Issue #6266, opened 2025-09-24
Prepare examples for pipedv1


khanhtc1202:
**What would you like to be added**:

Prepare examples for pipedv1 (plugin-arch piped) on https://github.com/pipe-cd/examples and make it live on https://play.pipecd.dev

**Why is this needed**:

Better understand how to use pipecdv1

shivansh-gohem:
Hi @khanhtc1202, I see this issue is open and unassigned. I'd like to help by creating the first example for pipedv1.

To make sure I'm on the right track, what would be a good, simple plugin to use for the first example? I'm happy to read the pipedv1 documentation and get started.

I'll also look at the pipe-cd/examples repository to understand the existing structure.

Nesar976:
Hi @khanhtc1202, I’d like to work on this issue. I’ll start by creating a simple pipedv1 example (plugin-arch based) and ensure it runs on play.pipedcd.dev. Please let me know if there’s any preferred example to begin with.

khanhtc1202:
Hi @Nesar976 this issue will be implement during the LFX mentorship program 2026 term 2, please do not open any PR toward this. If you feel interest in contributing to PipeCD, consider to check good first issue, pipectl command refinement or add tests in the meantime. Thank you

PriiJai:
Hi @khanhtc1202! I'm Priyanka, a CSE undergraduate applying for the LFX Term 2 Mentorship for this project.

I saw your note about keeping this specific issue reserved for the official mentorship term, so I will absolutely respect that and hold off on opening any PRs here!

I am super excited about the prospect of translating the Plugin Book, working on the pipedv1 examples, and helping grow PipeCD's global developer adoption. To get a head start and familiarize myself with the codebase, I'm setting up my local dev environment today and will look for a few open good first issue tags or small docs updates to contribute to first.

Thank you for the guidance, and I look forward to the application review!

RishiAnand108:
Hi,  @khanhtc1202 I recently applied for the mentee program and was wondering if you could please review my application when you have time. I would really appreciate any feedback or suggestions. Thank you!

mohammedfirdouss:
Reviews for LFX applications are made after the application is closed not by requesting or pinging the mentors. Thanks for understanding.

khanhtc1202:
Note: We plan to rename the current pipe-cd/examples as pipe-cd/examples-v0 and create a new pipe-cd/examples repo for pipedv1 examples as we want to promote the pipecd v1
[...]

PR #7173, opened 2026-08-13
Split examples into v0 and v1

rahulshendre:
**What this PR does**:
Restructures the examples/ directory into v0/ and v1/ subdirectories. All the current examples move under v0/, and I added a v1/ directory for the upcoming PipeCD v1 (plugin architecture) examples. The shared README.remote.md at the root now points to both, and the existing catalog moved into v0/README.md.

**Why we need it**:
Start of adding v1 examples and a lot of users still reference the v0 ones, so we keep both side by side instead of replacing them.

Related to #6266


aamirpatel (commented):
Reviewed

armistcxy (commented):
agree, split them to avoid confuse

armistcxy:
@rahulshendre test fail due to moving examples, I think you should update the test also

Suggestion from Copilot when explaining action error

image

armistcxy (changes requested):
I think we need to update `docs/content/en/docs-dev/examples/_index.md` also @rahulshendre

rahulshendre:
> I think we need to update `docs/content/en/docs-dev/examples/_index.md` also @rahulshendre

thanks for the mention, @armistcxy. those links will need updating, they point to the separate pipe-cd/examples repo (not the examples/ dir here), so they still resolve fine today and would only break if I switch them to v0/ before that repo is restructured. Per #6266 we're restructuring pipe-cd/examples the same way, and I'll update index.md to v0/ as part of that. 
 I was thinking of keeping this PR scoped to the main-repo move, what do you think 👀

rahulshendre:
> @rahulshendre test fail due to moving examples, I think you should update the test also

Thanks @armistcxy, nice catch. That test hardcodes the old examples/kubernetes/simple path, so the move to examples/v0/ broke it. I've updated both paths in sync_test.go to point at examples/v0/. PTAL

armistcxy:
> I was thinking of keeping this PR scoped to the main-repo move, what do you think 👀

Agree, thanks for pointing out those links belong to another repo

rahulshendre:
@khanhtc1202 @mohammedfirdouss, PTAL when you have time

khanhtc1202 (approved):
LGTM, let's check and update all docs link after this example be synced 👍

armistcxy (approved):

PR #7174, opened 2026-08-13
fix: use pointer receivers for piped protobuf models

srinivasr:
**What this PR does**:
generated protobuf structs like applicationsyncstate embed a sync.mutex. because their methods were using value receivers, the lock state was being copied every time they were called or passed around (flagged by go vet).

fix:
- change methods on applicationsyncstate and applicationlivestateversion to use pointer receivers
- update all caller sites in driftdetector and livestatestore to pass pointers
- convert range loops in tests to index loops to avoid copying test structs

**Why we need it**:
passing mutexes by value creates silent data races under concurrent execution because the copies do not block each other.

Fixes # N/A

- **How are users affected by this change**: no user-facing change.
- **Is this breaking change**: no.


areebahmeddd:
i don’t think this branch compiles .. you should run `go build ./...` and `make check` locally

pkg/model/application.go
```diff
@@ -27,12 +27,11 @@ const (
 	applicationConfigFileExtention      = ".pipecd.yaml"
 )
 
-// GetApplicationConfigFilePath returns the path to application configuration file.
```
areebahmeddd, line 30: i don't see why this is removed

pkg/app/piped/livestatestore/kubernetes/reflector_test.go
```diff
@@ -36,10 +36,10 @@ func TestResourceMatcher(t *testing.T) {
 			name: "empty config",
 			cfg:  config.KubernetesAppStateInformer{},
 			gvks: map[schema.GroupVersionKind]bool{
-				{"pipecd.dev", "v1beta1", "Foo"}:       false,
-				{"", "v1", "Foo"}:                      false,
-				{"", "v1", "Service"}:                  true,
-				{"networking.k8s.io", "v1", "Ingress"}: true,
+				{Group: "pipecd.dev", Version: "v1beta1", Kind: "Foo"}:       false,
```
areebahmeddd: `schema.GroupVersionKind` has no lock in it, so this is an unrelated `composites` fix

srinivasr:
> i don’t think this branch compiles .. you should run `go build ./...` and `make check` locally

you're right, i missed a caller in builder.go when i was switching the models to use pointer receivers. i just pushed a fix and verified it builds cleanly locally now 
image

pkg/model/application.go
srinivasr: gopls dropped the doc comment when I changed the receiver. added it back

pkg/app/piped/livestatestore/kubernetes/reflector_test.go
srinivasr: gopls auto-keyed the struct literals (go vet composites warning) but yeah that's out of scope for this PR. reverted to positional

srinivasr:
sorry for the delay on this. merged master in (it has #7137 now which touches the same HasDiff func, both changes work together fine) and squashed the branch while I was at it since the history had piled up fixups, so commit dates show today. doc comment is back and the unrelated reflector_test change is reverted as discussed

also pulled the metadatastore race fix out of this PR into its own branch to keep this one focused on what was reviewed.

areebahmeddd:
looks good. maybe @khanhtc1202 will give a final look and merge 🙂

pkg/app/piped/livestatestore/kubernetes/store.go
```diff
@@ -289,8 +290,7 @@ func (s *store) getAppLiveState(appID string) (AppState, bool) {
 		resources      = make([]*model.KubernetesResourceState, 0, len(nodes))
 	)
 	for i := range nodes {
-		state := nodes[i].state
-		resources = append(resources, &state)
+		resources = append(resources, nodes[i].state)
```
armistcxy, line 294: add note for this, since this will change the behavior, the old code only pass in ref for a copy, but the new one will pass the ref of the origin object. So maybe write to this could lead to change in other place (currently no write)

Just add a comment above this line: "Note: The returned states are pointers shared with the store's internal data, treat them as read-only"

armistcxy (commented):
LGTM, there one place I think you should add comment on top of that line

pkg/app/piped/livestatestore/kubernetes/store.go
srinivasr: added the note

armistcxy:
Special thanks to @areebahmeddd for helping with the review 💯

armistcxy (approved):
LGTM

PR #7178, opened 2026-08-15
Allow label scoped resources in POLICIES_STRING_REGEX

### LINKED ISSUE #7172: Label scoped RBAC policies are rejected by the role dialog, so roles that use them cannot be edited
`POLICIES_STRING_REGEX` in `web/src/constants/project.ts` does not allow the
`{key:value}` label block that a resource can carry, so the Add Role and Edit
Role dialogs reject any policy that scopes a resource by labels:

```
resources=application{env:prod};actions=get        -> "Invalid policy format"
resources=application{env:prod,team:foo};actions=get -> "Invalid policy format"
resources=application{env:prod},piped;actions=get  -> "Invalid policy format"
```

The rest of the stack supports labels:

- `ProjectRBACResource` has a `labels` map in `pkg/model/project.proto`
- `parseRBACPolicies` parses the `{...}` block (`RESOURCES_LABELS_REGEX`)
- `formalizePoliciesList` writes the `{...}` block back out

The worst part is editing. `EditRoleDialog` fills the textarea with
`formalizePoliciesList(...)`, which emits the label block, and validates it
with `POLICIES_STRING_REGEX`. So opening a role that uses labels shows a form
that is already invalid according to its own validation, and the role cannot
be saved even if nothing is changed.

The dialogs accept the same policy syntax that `parseRBACPolicies` and
`formalizePoliciesList` already handle, so label scoped roles can be created
and edited.

**How to reproduce it**:

Round tripping through the two helpers works fine:

```
"resources=application{env:prod};actions=get"
  -> parseRBACPolicies -> formalizePoliciesList
  -> "resources=application{env:prod};actions=get"
```

but the same string fails the form's validation schema:

```js
yup.string().matches(POLICIES_STRING_REGEX).isValidSync(
  "resources=application{env:prod};actions=get"
) // false
```

In the UI: Settings > Project > Roles > Add Role, enter
`resources=application{env:prod};actions=get`, and the field shows
"Invalid policy format" so the submit button stays disabled.

Goyamjain06:
- Fixes #7172
- `POLICIES_STRING_REGEX` now accepts an optional `{key:value,...}` label block after each resource type, using the same shape as `RESOURCES_LABELS_REGEX` (used in `parseRBACPolicies`).
- This stops the Add/Edit Role dialogs from rejecting label-scoped RBAC policies.
- The label block requires at least one valid `key:value` pair and excludes `;`, `{`, `}`, `:`, `,` from key/value characters, so a semicolon inside a label can no longer escape into the resources/actions separator (e.g. `application{a=b;c:d}` is rejected instead of being silently dropped by `parseRBACPolicies`), and empty (`{}`) or malformed (`{garbage}`) label blocks are rejected instead of passing validation.
- Added `web/src/constants/project.test.ts` with 14 cases: no-label regression guard, label-scoped valid cases, and invalid formats (including the semicolon-escape and empty-label cases from review).
- Verified: full `make test/web` suite passes (50 suites / 185 tests), including the `add-role-dialog` and `edit-role-dialog` tests.


web/src/constants/project.ts
```diff
@@ -65,11 +65,11 @@ export const TEXT_TO_RBAC_ACTION_TYPE: Record<
   delete: ProjectRBACPolicy.Action.DELETE,
 };
 
-// example: resources=(\*|application|deployment|event|piped|deploymentChain|project|apiKey|insight|,)+;\s*actions=(\*|get|list|create|update|delete|,)+
+// example: resources=(\*|application(\{[^}]*\})?|deployment(\{[^}]*\})?|...|,)+;\s*actions=(\*|get|list|create|update|delete|,)+
 export const POLICIES_STRING_REGEX = new RegExp(
   "resources=(" +
     rbacResourceTypes()
-      .map((v) => v.replace(/\*/, "\\*"))
+      .map((v) => v.replace(/\*/, "\\*") + "(\\{[^}]*\\})?")
```
areebahmeddd: `[^}]*` also allows `;`, which the parser can't handle. `resources=application{a=b;c:d};actions=get` passes validation but gets silently dropped by `parseRBACPolicies`, leaving the role with no policies

`\\{[^;{}]*\\}` fixes this while keeping the valid cases above. similarly, `{}` / `{garbage}` parse to empty labels, so the policy gets saved un-scoped. i think we should guard against that too

web/src/constants/project.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  // already supported them.
+  it.each([
+    "resources=application{env:prod};actions=get",
+    "resources=application{env:prod,team:foo};actions=get",
+    "resources=application{env:prod},piped;actions=get",
+    "resources=application{env:prod},piped{env:prod,team:foo};actions=get,list",
+  ])("matches policy with label scoped resources: %s", (policy) => {
+    expect(POLICIES_STRING_REGEX.test(policy)).toBe(true);
+  });
+
```
areebahmeddd, line 25: could you add `resources=application{a=b;c:d};actions=get` and `resources=application{};actions=get` to the invalid cases once the pattern is tightened? these are the cases where form validation and `parseRBACPolicies` currently disagree

areebahmeddd:
>label block accept karta hai

Could you rewrite the description in English? a few lines are in Hinglish .. 🤔

web/src/constants/project.ts
Goyamjain06: Thanks for catching both of these — the semicolon-in-label escape and the empty/malformed label case. Fixed: the label block now excludes `;`, `{`, `}`, `:`, `,` from key/value characters (so `application{a=b;c:d}` no longer escapes into the resources/actions separator), and it requires at least one valid `key:value` pair, so `{}` and `{garbage}` are rejected too. Pushed in ce7d41d.

web/src/constants/project.test.ts
Goyamjain06: Added both cases to the invalid-policy test list — `resources=application{a=b;c:d};actions=get` and `resources=application{};actions=get` — and confirmed the full suite still passes (50 suites / 185 tests) with all existing valid/invalid cases intact.

Goyamjain06:
Apologies for that — the description had a few lines in Hinglish. I've rewritten it fully in English while keeping the same technical content.

Goyamjain06:
Hi @areebahmeddd, just checking in on this one — I've updated the description as requested. Let me know if there are any further changes needed, or if this is ready for another look. Thanks for the feedback so far!

areebahmeddd:
looks good. maybe @khanhtc1202 will give a final look and merge 🙂

rahulshendre (dismissed):
LGTM @Goyamjain06, thank you : ) 

Thanks @areebahmeddd for the review.

rahulshendre:
cc @armistcxy @mohammedfirdouss

web/src/constants/project.ts
```diff
@@ -65,11 +65,19 @@ export const TEXT_TO_RBAC_ACTION_TYPE: Record<
   delete: ProjectRBACPolicy.Action.DELETE,
 };
 
-// example: resources=(\*|application|deployment|event|piped|deploymentChain|project|apiKey|insight|,)+;\s*actions=(\*|get|list|create|update|delete|,)+
+// A single label pair, e.g. "env:prod". Key/value exclude the characters
+// that are structurally significant elsewhere in the grammar (`;{}:,`) so a
+// label block can't accidentally swallow the resources/actions separator.
+const LABEL_PAIR = "[^;{}:,]+:[^;{}:,]+";
```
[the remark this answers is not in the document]
Goyamjain06: Good catch — fixed. `LABEL_PAIR` now excludes `\s` alongside `;{}:,`, so a `\n\n` inside a label key/value can no longer act as a policy separator. Added a regression case for this (newline inside a label value) and reran the full suite — 186/186 passing. Pushed in 20dd7ac87.

web/src/constants/project.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it.each([
+    "resources=application{env:prod;actions=get", // unclosed brace
+    "resources=bogus;actions=get", // unknown resource type
+    "resources=application;actions=bogus", // unknown action
+    "resources=application", // missing actions
+    "resources=application{a=b;c:d};actions=get", // semicolon escapes the label block
+    "resources=application{};actions=get", // empty label block
```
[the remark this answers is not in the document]
Goyamjain06: Good catch — fixed. `LABEL_PAIR` now excludes `\s` alongside `;{}:,`, so a `\n\n` inside a label key/value can no longer act as a policy separator. Added a regression case for this (newline inside a label value) and reran the full suite — 186/186 passing. Pushed in 20dd7ac87.

PR #7201, opened 2026-08-21
Fix git worktree race condition issue

khanhtc1202:
**What this PR does**:

- Add a mutex for worktree to the repo structure, call it in Copy, and places that work with git worktree command to avoid race condition when multiple worktree add command calls to the same repo
- Change some repo method from value receiver to pointer receiver to keep the repo type methods consistent

**Why we need it**:

There was a race condition reported by users, when the piped runs multiple workers (approximately 13) that execute git worktree add in parallel on the same repository. Running multiple add commands simultaneously causes the underlying git directories to vanish mid-process, resulting in `failed to read common dir` or `could not open gitdir` errors.

Follow PR #5412 
Fixes #

### DIFF
--- a/pkg/git/repo.go
+++ b/pkg/git/repo.go
@@ -24,6 +24,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


rahulshendre (approved):
LGTM 👍

Warashi (approved):

PR #7214, opened 2026-08-22
Fix goyaml 1.19.2

mohammedfirdouss:
**What this PR does**:
Bumps `github.com/goccy/go-yaml` from `v1.9.8` to `v1.19.2` in the `kubernetes` plugin and updates `yamlprocessor` tests to match that release's (corrected) parsing/serialization behavior.

**Why we need it**:

The original Dependabot PR (#7199) only bumped the dependency version, but v1.19.2 changed two behaviors: `"::"` now parses as a valid plain scalar instead of erroring, and replacing a value with an empty string now serializes as `foo:\n` instead of `foo: \n` (no trailing space). 

Fixes #

- **How are users affected by this change**:only test expectations were updated to match the new library version's corrected output.


mohammedfirdouss:
@armistcxy ptal

armistcxy (approved):
LGTM

PR #7217, opened 2026-08-22
fix(crypto): prevent AES decrypt panic

### LINKED ISSUE #7215: Prevent panic when decrypting short AES-GCM ciphertext
**What happened**:

# Prevent panic when decrypting short AES-GCM ciphertext

## Description

The `Decrypt` function in `pkg/crypto/aes.go` does not validate the length of the decoded ciphertext before extracting the AES-GCM nonce.

The function currently accesses the nonce using `encrypted[:gcm.NonceSize()]`. If the decoded ciphertext is shorter than the required nonce size, this slice operation can cause a runtime panic instead of returning an error.

## Expected Behavior

Invalid or truncated ciphertext should be rejected gracefully by returning an error rather than causing a panic.

## Current Behavior

After base64 decoding, the function directly extracts the nonce without checking whether the ciphertext contains enough bytes:

```text
encrypted[:gcm.NonceSize()]
```

Therefore, insufficient ciphertext can result in an out-of-range slice panic.

## Impact

Malformed or truncated encrypted input can trigger an unexpected runtime panic in the AES decryption path.

Adding input validation would allow invalid ciphertext to be handled safely and consistently as an error.

## Proposed Resolution

Validate that the decoded ciphertext is at least `gcm.NonceSize()` bytes long before extracting the nonce.

If the ciphertext is shorter than the required nonce size, `Decrypt` should return an appropriate error.

## Tests

Add regression tests covering:

* Empty ciphertext
* Ciphertext shorter than the GCM nonce size
* Valid ciphertext to ensure existing decryption behavior is unchanged

The tests should verify that invalid ciphertext returns an error rather than panicking.

## Evidence

The attached screenshot shows the current `Decrypt()` implementation, where the ciphertext length is not checked before the nonce is extracted.

Image

## Scope

This issue is limited to handling insufficient ciphertext length in `pkg/crypto/aes.go`. No unrelated refactoring is required.

anubhavsingh2106:
## What does this PR do?

Prevents `AESEncryptDecrypter.Decrypt()` from panicking when the decoded ciphertext is shorter than the AES-GCM nonce size.

The current implementation slices the ciphertext to extract the nonce without first validating its length. This can cause a slice-bounds panic for empty or truncated ciphertext.

This change validates the ciphertext length before extracting the nonce and returns an error for invalid input instead.

## Changes

- Add ciphertext length validation before extracting the GCM nonce.
- Add regression tests for empty and truncated ciphertext.

## Testing

- `go test ./pkg/crypto/... -count=1` ✅
- `git diff --check` ✅
- `make check` ⚠️

`make check` could not complete because of unrelated existing build errors in `pkg/app/piped/toolregistry/install.go`:

- `undefined: kubectlInstallScript`
- `undefined: kustomizeInstallScript`
- `undefined: helmInstallScript`
- `undefined: terraformInstallScript`

These errors are unrelated to the changes in this PR.

## Related Issue

Closes #7215


anubhavsingh2106:
Hi @armistcxy, could u pls take a look?

armistcxy:
Valid concern, thanks @anubhavsingh2106

armistcxy (approved):
LGTM 👍

PR #7222, opened 2026-08-23
fix: use min requested index for ecs rollback stage

armistcxy:
**What this PR does**:  
- Fixes `buildPipelineStages` in ECS plugin to reuse the smallest requested index for the appended rollback stage
- Adds unit tests covering the pipeline/quick-sync stage builders

**Why we need it**: `sdk.PipelineStage.Index` requires each returned index to be one of the indexes from the request. The previous implementation returned an invented index for `ECS_ROLLBACK`, so piped's `validateStageIndexes` rejected the plugin response

### DIFF
--- a/pkg/app/pipedv1/plugin/ecs/deployment/pipeline.go
+++ b/pkg/app/pipedv1/plugin/ecs/deployment/pipeline.go
@@ -96,14 +96,12 @@ func buildPipelineStages(input *sdk.BuildPipelineSyncStagesInput) ([]sdk.Pipelin
 	}
 
 	if input.Request.Rollback {
-		// Guard against a rollback request without any stage, which would
-		// panic on stages[0] below.
+		// Guard against a rollback request without any stage
 		if len(stages) == 0 {
 			return nil, ErrRollbackRequiresStages
 		}
 		// The rollback stage must reuse one of the requested indexes and
 		// piped runs rollback stages as a trail sorted by index.
-		// Use the smallest requested index so our rollback runs first among all plugins' rollbacks.
 		minIndex := stages[0].Index
 		for _, s := range stages[1:] {
 			if s.Index < minIndex {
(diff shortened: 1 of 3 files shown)


pkg/app/pipedv1/plugin/ecs/deployment/pipeline.go
```diff
@@ -90,9 +90,18 @@ func buildPipelineStages(input *sdk.BuildPipelineSyncStagesInput) []sdk.Pipeline
...
 	if input.Request.Rollback {
+		// The rollback stage must reuse one of the requested indexes and
+		// piped runs rollback stages as a trail sorted by index.
+		// Use the smallest requested index so our rollback runs first among all plugins' rollbacks.
+		minIndex := stages[0].Index
+		for _, s := range stages[1:] {
+			if s.Index < minIndex {
+				minIndex = s.Index
+			}
+		}
```
Warashi, line 110: We can reuse the same strategy with the Kubernetes plugin.
We need to check `len(stages)` to avoid a panic.
https://github.com/pipe-cd/pipecd/blob/dc6a6a9a6f5b571985d77c192955d637de16970d/pkg/app/pipedv1/plugin/kubernetes/deployment/pipeline.go#L122-L128

Warashi (commented):
We need this logic in all of the deployment plugins, so it's good to port it to the SDK?

pkg/app/pipedv1/plugin/ecs/deployment/pipeline.go
armistcxy: make sense

armistcxy:
> We need this logic in all of the deployment plugins, so it's good to port it to the SDK?

At this point I don't have my answer yet, but I think this part is actually repetitive among plugins. So let me investigate first

Warashi (approved):
We can open another PR to make this part of the SDK, so I approve this as-is.

PR #7226, opened 2026-08-23
Close datastore iterators to release underlying rows and Firestore resources on decode errors

### LINKED ISSUE #7219: datastore.Iterator leaks database/Firestore resources on decode errors
# Title

MySQL/Firestore `datastore.Iterator` leaks connections on decode error

# Description

* `pkg/datastore/datastore.go` (`Iterator` interface has no `Close()`), `pkg/datastore/mysql/iterator.go:38-52`
* `List()` calls across the datastore layer can return early on a row/document decode error without releasing the underlying `sql.Rows`/Firestore iterator.
* No `SetMaxOpenConns` is configured, so repeated failures can accumulate unreleased database resources. `ListNotCompletedDeployments` and similar methods are polled continuously by every piped instance, meaning a single malformed/incompatible row could eventually exhaust available MySQL connections and impact datastore operations across the control plane.
* This is distinct from the related but narrower open issue [[#6699](https://github.com/pipe-cd/pipecd/issues/6699)](https://github.com/pipe-cd/pipecd/issues/6699), which concerns a swallowed `json.Unmarshal` error in the same MySQL iterator but does not address iterator/resource cleanup.

## Suggested fix

Add `Close() error` to the `datastore.Iterator` interface and ensure every caller closes the iterator immediately after obtaining it:

```go
iterator, err := ...
if err != nil {
    return err
}
defer iterator.Close()
```

This ensures the underlying database/Firestore iterator is released both on successful iteration and when iteration exits early because of a decode error.

## Verification

The MySQL implementation can return from iteration when row decoding fails without closing the underlying `sql.Rows`. The Firestore implementation should be verified for the same lifecycle issue.

vanshika2720:
**What this PR does**:

Adds explicit resource cleanup to `datastore.Iterator` so underlying MySQL and Firestore iterators are closed when datastore iteration completes or exits early due to an error. All `List()` callers now defer iterator cleanup, and tests verify that `Close()` is called on both successful and error paths.

**Why we need it**:

Previously, `datastore.Iterator` had no `Close()` method, so an early return caused by a decode/scan error could leave the underlying `sql.Rows` or Firestore `DocumentIterator` open. Repeated failures could accumulate unreleased database resources and potentially exhaust available connections.

Fixes #7219

* **How are users affected by this change**: No direct user-facing behavior change. This fixes datastore resource cleanup and prevents database/Firestore resources from remaining open after iterator errors.

### DIFF
--- a/pkg/datastore/apikey.go
+++ b/pkg/datastore/apikey.go
@@ -85,6 +85,7 @@ func (s *apiKeyStore) List(ctx context.Context, opts ListOptions) ([]*model.APIK
 	if err != nil {
 		return nil, err
 	}
+	defer it.Close()
 	ks := make([]*model.APIKey, 0)
 	for {
 		var k model.APIKey
--- a/pkg/datastore/apikey_test.go
+++ b/pkg/datastore/apikey_test.go
@@ -86,6 +86,7 @@ func TestListAPIKeys(t *testing.T) {
 			opts: ListOptions{},
 			ds: func() DataStore {
 				it := NewMockIterator(ctrl)
+				it.EXPECT().Close().Return(nil)
 				it.EXPECT().
 					Next(&model.APIKey{}).
 					Return(ErrIteratorDone)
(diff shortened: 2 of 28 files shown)


vanshika2720:
@mohammedfirdouss

PR #7231, opened 2026-08-24
Prevent commandstore cache poisoning when backend Get fails after UpdateStatus

### LINKED ISSUE #7230: `commandstore` cache poisoning after transient `Get` failure following successful status update
# Description

`commandstore` can cache an invalid zero-value `Command` when `backend.Get` fails immediately after `UpdateStatus` succeeds.

### Details

* `pkg/app/server/commandstore/store.go:106-120`
* `pkg/app/server/commandstore/cache.go:29-48`

If `backend.Get` fails after `UpdateStatus` has already succeeded, the error is effectively swallowed and the resulting `nil` command is passed to the cache.

The `nil` value is marshaled to JSON as `null`. When `null` is later unmarshaled into a non-pointer `Command` struct, Go treats it as a no-op, leaving a zero-value `Command`, including `PipedId=""`.

Subsequent `Get` calls can therefore return a valid-looking zero-value command instead of the actual command.

This can cause `ReportCommandHandled` to return `PermissionDenied` for the legitimate piped. Since `ReportCommandHandled` is also the RPC that could otherwise repair the cache, this can leave manual-approval/cancel commands stuck indefinitely.

### Suggested fix

Either:

* return the `backend.Get` error instead of continuing and caching the invalid result, or
* skip `cache.Put` when `cmd == nil`.

Additionally, add a nil guard in `commandCache.Put` so a nil command cannot be cached accidentally.

### Tests

There currently does not appear to be a test file for this package.

Add a regression test covering the following sequence:

1. `UpdateStatus` succeeds.
2. The subsequent `backend.Get` fails.
3. The failed result is not cached.
4. A later `Get` does not return a zero-value `Command`.
5. `ReportCommandHandled` continues to work for the legitimate piped.

vanshika2720:
**What this PR does**:

Prevents `commandstore` from caching a nil command when the datastore `Get` fails after a successful command status update. It also adds a defensive nil check in the cache and regression tests covering the failure path.

**Why we need it**:

A transient `Get` failure could previously result in `nil` being cached as JSON `null`. On a later cache read, this could become a zero-value `Command` with an empty `PipedId`, causing `ReportCommandHandled` to return `PermissionDenied` for the legitimate piped.

Fixes #7230 

**Does this PR introduce a user-facing change**:

* **How are users affected by this change**: No direct user-facing behavior change. This prevents a transient datastore error from poisoning the command cache and incorrectly rejecting subsequent command-handling requests.

### DIFF
--- a/pkg/app/server/commandstore/cache.go
+++ b/pkg/app/server/commandstore/cache.go
@@ -52,12 +52,19 @@ func (c *commandCache) Put(commandID string, command *model.Command) error {
 	if command == nil {
 		return fmt.Errorf("command must not be nil")
 	}
+	if err := command.Validate(); err != nil {
+		return fmt.Errorf("command is invalid: %w", err)
+	}
 	key := cacheKey(commandID)
 	data, err := json.Marshal(command)
 	if err != nil {
 		return err
 	}
 	return c.backend.Put(key, data)
+}
+
+func (c *commandCache) Delete(commandID string) error {
+	return c.backend.Delete(cacheKey(commandID))
 }
 
 func cacheKey(commandID string) string {
--- a/pkg/app/server/commandstore/cache.go
+++ b/pkg/app/server/commandstore/cache.go
@@ -40,6 +40,9 @@ func (c *commandCache) Get(commandID string) (*model.Command, error) {
 }
 
 func (c *commandCache) Put(commandID string, command *model.Command) error {
+	if command == nil {
+		return fmt.Errorf("command must not be nil")
+	}
 	key := cacheKey(commandID)
 	data, err := json.Marshal(command)
 	if err != nil {
--- a/pkg/app/server/commandstore/cache.go
+++ b/pkg/app/server/commandstore/cache.go
@@ -36,6 +36,15 @@ func (c *commandCache) Get(commandID string) (*model.Command, error) {
 	if err := json.Unmarshal(item.([]byte), &s); err != nil {
 		return nil, err
 	}
+	// A cached entry that doesn't satisfy the same validation required to
+	// store a command (e.g. a zero-value Command decoded from a JSON null
+	// written before Put started rejecting nil commands) must not be
+	// treated as a valid cache hit. Evict it on a best-effort basis and
+	// report a miss so the caller falls back to the datastore.
+	if err := s.Validate(); err != nil {
+		_ = c.backend.Delete(key)
+		return nil, cache.ErrNotFound
+	}
 	return &s, nil
 }
 
--- a/pkg/app/server/commandstore/store_test.go
+++ b/pkg/app/server/commandstore/store_test.go
@@ -140,3 +140,73 @@ func TestGetCommand_NotPoisonedAfterUpdateHandledGetFailure(t *testing.T) {
 	assert.Equal(t, realCommand, got)
 	assert.NotEmpty(t, got.PipedId, "cached/returned command must not be a poisoned zero-value with an empty PipedId")
 }
+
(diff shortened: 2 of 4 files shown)


pkg/app/server/commandstore/store_test.go
```diff
@@ -0,0 +1,212 @@
...
+	getCalls := 0
+	backend := &fakeCommandStore{
+		updateStatusFunc: func(ctx context.Context, id string, status model.CommandStatus, metadata map[string]string, handledAt int64) error {
+			return nil
+		},
+		getFunc: func(ctx context.Context, id string) (*model.Command, error) {
+			getCalls++
+			if getCalls == 1 {
+				// First Get (triggered by UpdateCommandHandled) fails transiently.
+				return nil, errors.New("transient datastore error")
```
areebahmeddd, line 132: `return nil` stops the poisoning, but the cache still keeps the pre-update command

`ReportCommandHandled` calls `GetCommand` right before this, so the old entry (`status: NOT_HANDLED_YET`, `handled_at: 0`) is already cached. after this returns, the datastore has `SUCCEEDED` while the cache still says unhandled for up to the 5m ttl

that's what `pipectl application sync` polls via `API.GetCommand`, so it can keep waiting even though the sync already succeeded

i would invalidate the stale entry instead :

```go
cmd, err := s.backend.Get(ctx, id)
if err != nil {
    s.logger.Error("failed to get command from datastore", zap.Error(err))
    // Drop the stale entry so readers fall back to the datastore.
    if err := s.cache.Delete(id); err != nil {
        // ...
    }
    return nil
}
```

`cache.Cache` already embeds `Deleter`, so `commandCache` would just need a thin `Delete` method

pkg/app/server/commandstore/cache.go
```diff
@@ -36,10 +36,22 @@ func (c *commandCache) Get(commandID string) (*model.Command, error) {
...
 		return nil, err
 	}
+	// A cached entry that doesn't satisfy the same validation required to
+	// store a command (e.g. a zero-value Command decoded from a JSON null
+	// written before Put started rejecting nil commands) must not be
+	// treated as a valid cache hit. Evict it on a best-effort basis and
+	// report a miss so the caller falls back to the datastore.
+	if err := s.Validate(); err != nil {
+		_ = c.backend.Delete(key)
+		return nil, cache.ErrNotFound
```
areebahmeddd, line 46: `Get` now rejects anything failing `Validate()`, but `Put` only rejects `nil`. so a command with `CreatedAt == 0` can still be cached, then gets evicted on every read :

```text
Get → validate fails → Delete → datastore → Put → repeat
```

that means a permanent cache miss plus an extra redis round-trip on every read

`realCommand` in `store_test.go` also has no `CreatedAt` / `UpdatedAt`, so it wouldn't survive its own `Get`. i would make `Put` use `command.Validate()` instead of only checking `nil`, keeping `Put` and `Get` consistent the `Get` validation can stay as cleanup for already-poisoned entries

also minor: fix those two fixtures to use valid timestamps, since the current shapes aren't commands the datastore could produce 🙂

pkg/app/server/commandstore/store_test.go
vanshika2720: I've updated this to evict the stale entry when the post-UpdateStatus Get fails, with a regression test covering the datastore fallback path.

pkg/app/server/commandstore/cache.go
vanshika2720: Agreed. Put and Get should enforce the same validation invariant. I've updated Put to reject invalid commands, kept the validation in Get for cleanup of existing poisoned entries, and fixed the test fixtures to use valid timestamps.

PR #7234, opened 2026-08-24
Page datastore queries that fetched all applications

### LINKED ISSUE #7051: bug: piped ListApplications silently truncates paginated results
---
name: Bug Report
about: Report a bug encountered
labels: kind/bug

---

**What happened**:
`PipedAPI.ListApplications` only returns the first datastore page for a piped's applications. When `applicationStore.List` returns a non-empty cursor, the RPC ignores it and returns immediately, so applications beyond the first page are silently dropped.

## Summary
`pkg/app/server/grpcapi/piped_api.go` fetches a single page from `applicationStore.List` and discards the returned cursor. A piped that manages more applications than fit in one datastore page receives an incomplete application list.

## Affected Area
- `pkg/app/server/grpcapi/piped_api.go`
- callers of `pipedservice.PipedService/ListApplications`

`ListApplications` should return a complete application set for the piped. With the current RPC contract, that means iterating datastore pages internally until the cursor is exhausted before returning the response.

## Expected Behaviour
`ListApplications` collects all pages from `applicationStore.List` and returns the full application list without requiring client-side pagination support.

**How to reproduce it**:
1. Mock `applicationStore.List` for `PipedAPI.ListApplications` so the first call returns one application plus a non-empty cursor.
2. Make the second call return another application and an empty cursor.
3. Invoke `PipedAPI.ListApplications`.
4. Observe on current `master` that only the first application is returned because the second page is never fetched.

## Steps to Reproduce
1. Seed or mock a project/piped with more applications than the datastore returns in one page.
2. Call `PipedAPI.ListApplications`.
3. Observe that only the first page is returned and no continuation cursor is exposed to the piped client.

**Environment**:
- `piped` version: current `master` as of July 21, 2026
- `control-plane` version: current `master` as of July 21, 2026
- Others: Not applicable

## Actual Behaviour
Only the first datastore page is returned, and remaining applications are silently omitted.

## Technical Evidence
- `pkg/app/server/grpcapi/piped_api.go` contains `TODO: Support pagination in ListApplications` immediately before `apps, _, err := a.applicationStore.List(ctx, opts)`.
- The method ignores the returned cursor and responds with that single page.
- The public API path already supports pagination separately, so this bug is isolated to the piped gRPC path rather than the general application listing API.

## Suspected Root Cause
`PipedAPI.ListApplications` does not loop on `datastore.ListOptions.Cursor` and has no way to return a continuation cursor in the existing response type.

## Proposed Direction
Keep the current RPC contract and paginate internally in `PipedAPI.ListApplications` until the datastore cursor is empty, with a focused unit test that proves multi-page results are fully aggregated.

## Acceptance Criteria
- A mocked multi-page `applicationStore.List` test for `PipedAPI.ListApplications` returns all applications across pages.
- A piped with more than one datastore page of applications receives a complete list.
- Existing single-page callers continue to behave the same.

srinivasr:
**What this PR does**:

Fixes unbounded datastore queries in `ListApplications` (piped and web) and `CreateDeploymentChain`. These endpoints ran one query with no limit and returned every matching row; now they fetch pages of 100 and combine them before responding. Cursor paging needs a stable sort, so these queries order by `Id`.

`ListEvents` gets the same treatment, with a default of newest-first when a request leaves the order unset — without that default, paging would error on the second page.

Adds the composite indexes these query shapes need to `firestoreindexensurer`.

**Why we need it**:

A project with thousands of applications pays for all of them on every piped sync. The `TODO` in `ListApplications` flagged this; until the reads are chunked, the query grows with the project and eventually risks datastore timeouts.

Fixes #7051

No breaking changes. Responses still contain the complete aggregated list — they're just fetched from the backend in chunks.

- **How are users affected by this change**: Firestore users: `firestoreindexensurer` creates the new composite indexes automatically on control-plane startup, but GCP index builds take time — calls fail with `FAILED_PRECONDITION` until each index finishes building. MySQL users need no action.


areebahmeddd:
this probably needs discussion before merge 🤔

1. firestore index rollout

the description says `firestoreindexensurer` creates the indexes on control-plane startup but it actually runs in the ops component. that means the server can start serving the new queries before the indexes exist, causing `ListApplications` to fail with `FAILED_PRECONDITION` until the builds finish

could the index changes land in a separate, earlier pr so operators can let them build before the query changes ship? that would avoid a hard upgrade ordering constraint

2. application index shapes

`CreateDeploymentChain` always filters on `ProjectId` and `Disabled`, with `Name`/`Kind` being optional. but four of the six new application indexes omit `Disabled`:

```text id="x2c7pz"
ProjectId | Id
ProjectId | Kind | Id
ProjectId | Name | Id
ProjectId | Name | Kind | Id
```

the indexes that seem to match the actual query shapes would be :

```text id="6q3v8m"
ProjectId | Disabled | Name | Id
ProjectId | Disabled | Kind | Id
ProjectId | Disabled | Name | Kind | Id
```

could you walk through the query-shape to index mapping again? the two event indexes look correct, so this may just be a slip on the application side

pkg/app/server/grpcapi/piped_api.go
```diff
@@ -237,9 +243,12 @@ func (a *PipedAPI) ListApplications(ctx context.Context, req *pipedservice.ListA
...
 			},
 		},
+		Limit: listApplicationsPageSize,
 	}
-	// TODO: Support pagination in ListApplications
-	apps, _, err := a.applicationStore.List(ctx, opts)
+	// Page through the datastore so a large project cannot be served by one
+	// unbounded query. The RPC response has no cursor field, so all pages are
+	// aggregated here before returning.
+	apps, err := listAllApplications(ctx, a.applicationStore, opts)
```
areebahmeddd, line 251: `ListApplications` is called by every piped on each app-store sync and again by the event watcher. with 3,000 apps, this change turns one query into 30 sequential round-trips per sync, while the response is still fully aggregated in memory.

the description also says paging avoids paying for all applications on every sync, but the total read is still the same, just split across queries.

could `listApplicationsPageSize` be increased substantially, say to `1000`, to keep the bounded-query benefit without adding 10x the round-trips? 🤔

srinivasr:
> this probably needs discussion before merge 🤔
> 
>     1. firestore index rollout
> 
> 
> the description says `firestoreindexensurer` creates the indexes on control-plane startup but it actually runs in the ops component. that means the server can start serving the new queries before the indexes exist, causing `ListApplications` to fail with `FAILED_PRECONDITION` until the builds finish
> 
> could the index changes land in a separate, earlier pr so operators can let them build before the query changes ship? that would avoid a hard upgrade ordering constraint
> 
>     2. application index shapes
> 
> 
> `CreateDeploymentChain` always filters on `ProjectId` and `Disabled`, with `Name`/`Kind` being optional. but four of the six new application indexes omit `Disabled`:
> 
> ```
> ProjectId | Id
> ProjectId | Kind | Id
> ProjectId | Name | Id
> ProjectId | Name | Kind | Id
> ```
> 
> the indexes that seem to match the actual query shapes would be :
> 
> ```
> ProjectId | Disabled | Name | Id
> ProjectId | Disabled | Kind | Id
> ProjectId | Disabled | Name | Kind | Id
> ```
> 
> could you walk through the query-shape to index mapping again? the two event indexes look correct, so this may just be a slip on the application side

makes sense, split the indexes out into #7274 so operators can let them build first without hitting `FAILED_PRECONDITION`.

also fixed the app index shapes to include `Disabled` across the `CreateDeploymentChain` and `ListApplications` queries (`ProjectId | Disabled | Name | Id`, `ProjectId | Disabled | Kind | Id`, `ProjectId | Disabled | Name | Kind | Id`), and dropped the redundant `ProjectId | Id` entry. updated this branch with the same fixes so both stay in sync.

pkg/app/server/grpcapi/piped_api.go
srinivasr: makes sense, bumped `listApplicationsPageSize` to 1000 to keep the round-trips low while still keeping each individual query bounded.

PR #7275, opened 2026-08-26
fix(pipedv1): clean up cloned repo after skip stage check

Ahmad-Faraj:
**What this PR does**:

Same leak as #7021, on the pipedv1 side. `determineSkipStage` in `pkg/app/pipedv1/controller/skipstage.go` clones the application repo on every skip check and never removes it, so each check with skipOn conditions leaks one temp directory until the piped restarts. `Clean()` is not called anywhere in that file. This adds the deferred `repo.Clean()`, one line, mirroring what merged in #7021.

Follows up on @khanhtc1202's request in https://github.com/pipe-cd/pipecd/pull/7021#issuecomment-5379348990

None


vanshika2720 (approved):

mohammedfirdouss (approved):
LGTM

PR #7278, opened 2026-08-31
fix: prioritize oldest uncompleted deployment for head selection

### LINKED ISSUE #6780: bug: pipedv1 deployment store returns newest deployment as head, not oldest
**What happened**:

While reviewing the deployment store sync logic in `pipedv1`, the head-deployment selection in `pkg/app/pipedv1/apistore/deploymentstore/store.go:125-134` looks inverted relative to the `Lister` interface's docstring.

The interface promises:

> `ListAppHeadDeployments` returns the map from application ID to its head of deploying deployments.
> Head deployment is same with the oldest uncompleted one.

But the implementation writes `pendings → planneds → runnings` into the same `map[string]*model.Deployment`, so for an application with both a pending and a running deployment, the *running* one wins by overwrite order:

```go
for _, d := range pendings { headDeployments[d.ApplicationId] = d }
for _, d := range planneds { headDeployments[d.ApplicationId] = d }
for _, d := range runnings { headDeployments[d.ApplicationId] = d }
```

A `PENDING` deployment is logically *newer* than a `RUNNING` one for the same application (it was created later, after the running one started), so this code returns the *newest* uncompleted deployment, not the oldest.

The same inversion exists at the corresponding location in the v0 store (`pkg/app/piped/apistore/deploymentstore/store.go`).

Either:
1. The docstring is updated to match the implementation (\"head = newest uncompleted\"), or
2. The implementation is corrected to match the docstring — either reverse the write order or pick the head explicitly by `created_at`.

I don't know which side is correct, since I haven't traced where `ListAppHeadDeployments` is consumed downstream. Whichever way it ends up, the docstring and the code should agree.

**How to reproduce it**:

Unit test against the store with two `*model.Deployment` instances sharing one `ApplicationId`, one `DEPLOYMENT_PENDING` and one `DEPLOYMENT_RUNNING`. `Lister.ListAppHeadDeployments()[appID]` returns the running one — the opposite of what the docstring promises.

**Environment**:

- `piped` version: master @ `78fdbeb7c` (both v0 and v1 affected)
- `control-plane` version: same
- Others: surfaced during self-review of #6727 while writing tests for `ListAppHeadDeployments`

Vishakha7-Kumari:
### What this PR does / Why we need it
This PR fixes a bug in the synchronization logic of the `pipedv1` deployment store. 

Currently, `sync` builds the `headDeployments` map by iterating sequentially over pending, planned, and running deployments. As a result, running deployments (or newer deployments processed later in the sequence) always overwrite other uncompleted ones. This contradicts the contract defined in `ListAppHeadDeployments()`:
> `"Head deployment is same with the oldest uncompleted one"` (Issue #6780).

To fix this, we introduce an `updateHead` helper which compares the `CreatedAt` timestamps of the deployments. It only populates or updates `headDeployments` if a candidate deployment has a smaller `CreatedAt` timestamp (i.e. is older), guaranteeing that the oldest uncompleted deployment is always selected as the head.

### Changes
- **Store Logic**: Updated [store.go](file:///c:/Users/singh/pipecd/pkg/app/pipedv1/apistore/deploymentstore/store.go) to select the oldest uncompleted deployment by comparing `CreatedAt`.
- **Tests**: Added test cases in [store_test.go](file:///c:/Users/singh/pipecd/pkg/app/pipedv1/apistore/deploymentstore/store_test.go) covering scenarios where either the running or the pending deployment is the oldest.

Fixes #6780


armistcxy:
Hi @khanhtc1202, as far as I read the codebase, this function has not been used any place

do you have any idea or insight on why this function exists in the first place ? 😓

Vishakha7-Kumari:
Hi @armistcxy, you're right. Thanks for checking this. I verified it locally as well, and  ListAppHeadDeployments() has no call sites in the repo.
I originally opened this PR based on #6780 and the method's documented contract: "Head deployment is same with the oldest uncompleted one." The current sync() logic doesn't satisfy that contract, so I treated it as a correctness issue and added regression tests.
But it looks like this method is leftover from the old drift-detector flow and is no longer used after the controller started reporting deploying status through ReportApplicationDeployingStatus.
I'm happy to adjust the PR:
 -If it's dead code, I can remove the method and related `headDeployments` logic.
 -If it's still intended to be used, I'd appreciate a pointer to where it should be wired in.
@khanhtc1202, Could you please let me know if ListAppHeadDeployments() is still meant to be used somewhere?

PR #7283, opened 2026-08-31
fix(plugin/kubernetes): prevent data race on applications map in livestate store

Vishakha7-Kumari:
### What this PR does / Why we need it
This PR fixes a concurrency bug in deployTargetResources.getApplicationResources in the pipedv1 Kubernetes plugin live state store.

Previously, getApplicationResources attempted to initialize and insert missing entries into s.applications while only holding a read lock (s.mu.RLock()). In Go, mutating a map under a read lock creates a data race that can crash the piped process with atal error: concurrent map writes when multiple goroutines access application resources concurrently.

This change:
1. First looks up the application in s.applications under s.mu.RLock().
2. If absent, releases the read lock, acquires s.mu.Lock() (write lock), and double-checks existence before initializing and storing 
ewApplicationResources(s.deployTarget).

### Changes
- Updated pkg/app/pipedv1/plugin/kubernetes/livestate/store/store.go to acquire a write lock before modifying s.applications.
- Added pkg/app/pipedv1/plugin/kubernetes/livestate/store/store_test.go with a concurrent test (TestGetApplicationResources_Concurrent) spawning 100 parallel workers to ensure thread safety.
- Verified all tests pass.

Signed-off-by: Vishakha7-Kumari <singhdaisy669@gmail.com>

### DIFF
--- a/pkg/app/pipedv1/plugin/kubernetes/livestate/store/store.go
+++ b/pkg/app/pipedv1/plugin/kubernetes/livestate/store/store.go
@@ -223,9 +223,16 @@ func (s *deployTargetResources) initialize() {
 // getApplicationResources returns the application resources by the application ID.
 func (s *deployTargetResources) getApplicationResources(appID string) *applicationResources {
 	s.mu.RLock()
-	defer s.mu.RUnlock()
-
 	app, ok := s.applications[appID]
+	s.mu.RUnlock()
+	if ok {
+		return app
+	}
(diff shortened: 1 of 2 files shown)


armistcxy (approved):
LGTM, nice catch 👍

Vishakha7-Kumari:
Thanks @armistcxy for reviewing my fix.

PR #7319, opened 2026-09-04
Prevent nil pointer panic in deployment metadata updates

### LINKED ISSUE #7228: Bug: Prevent nil pointer panic in deployment metadata updates

omlahore:
Fixes #7228

## What was wrong

`UpdateSharedMetadata` and `UpdatePluginMetadata` in `pkg/datastore/deploymentstore.go` guard the inner fields but never the pointer that holds them:

```go
return s.update(ctx, id, func(d *model.Deployment) error {
    if d.MetadataV2.Shared == nil {   // panics here when MetadataV2 is nil
        d.MetadataV2.Shared = &model.DeploymentMetadata_KeyValues{}
    }
    ...
```

So `d.MetadataV2.Shared` and `d.MetadataV2.Plugins` dereference a nil pointer.

## Why a nil gets there

Two paths, both normal:

- pipedv0's trigger, `pkg/app/piped/trigger/deployment.go`, builds a `model.Deployment` without setting `MetadataV2`. Only the pipedv1 trigger sets it.
- The deployment collection's `Factory()` decodes into `&model.Deployment{}`, so any stored document without the field arrives with it nil.

Either one leaves a deployment in the datastore that panics the moment something updates its metadata.

## The change

Both methods initialise `MetadataV2` before touching its fields. Six lines. I did not touch the existing inner guards or `mergeMetadata`, since they are correct once the outer pointer exists.

## Tests

Three added to `pkg/datastore/deploymentstore_test.go`:

- `TestUpdateSharedMetadataWithNilMetadataV2`
- `TestUpdatePluginMetadataWithNilMetadataV2`
- `TestUpdateSharedMetadataMergesExisting`, so the fix cannot quietly start replacing metadata instead of merging it

They drive the real updater through a `MockDataStore` whose `Update` invokes the callback. There is also a small `validDeployment()` helper, because `s.update` runs `Validate()` after the updater and an empty deployment fails on `ApplicationId` before reaching the code under test.

I checked they fail without the change. Stashing just `deploymentstore.go` gives:

```
--- FAIL: TestUpdateSharedMetadataWithNilMetadataV2
panic: runtime error: invalid memory address or nil pointer dereference
```

`go build`, `go test ./pkg/datastore/` green. `go vet ./pkg/datastore/` reports only the pre-existing copylock warnings in `deploymenttracestore.go`, `eventstore.go`, `projectstore.go` and the older test cases at lines 198, 210, 297 and 309, none of which this PR touches.

### DIFF
--- a/pkg/datastore/deploymentstore.go
+++ b/pkg/datastore/deploymentstore.go
@@ -249,6 +249,9 @@ func (s *deploymentStore) UpdateStageMetadata(ctx context.Context, deploymentID,
 
 func (s *deploymentStore) UpdateSharedMetadata(ctx context.Context, id string, metadata map[string]string) error {
 	return s.update(ctx, id, func(d *model.Deployment) error {
+		if d.MetadataV2 == nil {
+			d.MetadataV2 = &model.DeploymentMetadata{}
+		}
 		if d.MetadataV2.Shared == nil {
 			d.MetadataV2.Shared = &model.DeploymentMetadata_KeyValues{}
 		}
@@ -261,6 +264,9 @@ func (s *deploymentStore) UpdateSharedMetadata(ctx context.Context, id string, m
 
 func (s *deploymentStore) UpdatePluginMetadata(ctx context.Context, id string, pluginName string, metadata map[string]string) error {
 	return s.update(ctx, id, func(d *model.Deployment) error {
+		if d.MetadataV2 == nil {
+			d.MetadataV2 = &model.DeploymentMetadata{}
+		}
 		if d.MetadataV2.Plugins == nil {
 			d.MetadataV2.Plugins = make(map[string]*model.DeploymentMetadata_KeyValues)
 		}
--- a/pkg/datastore/deploymentstore_test.go
+++ b/pkg/datastore/deploymentstore_test.go
@@ -529,3 +529,109 @@ func TestMergeMetadata(t *testing.T) {
 		})
 	}
 }
+
+// validDeployment returns the minimum deployment that passes Validate, so the
+// metadata tests below exercise the updater rather than tripping validation.
+// MetadataV2 is deliberately left nil.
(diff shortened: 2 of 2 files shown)


omlahore:
@anubhavsingh2106 this duplicates your #7229 and I did not see it before opening. You filed #7228 and opened #7229 the same day, 23 August, and mine is twelve days later with the same two nil guards in the same file. Yours has priority on every count.

Happy for this to be closed in favour of #7229. @Warashi, say the word and I will close it, or close it yourself.

The only thing here that might be worth carrying over is the test shape: mine leaves `MetadataV2` nil on a deployment that otherwise passes `Validate`, so the updater is what gets exercised rather than validation rejecting the object first. If that is useful to #7229 take it, no attribution needed.

anubhavsingh2106:
@omlahore Thanks for pointing this out! I hadn’t noticed the overlap either. Since #7229 was opened earlier, I agree that it makes sense to continue with that PR. I also appreciate you sharing the test approach. I’ll take a look and incorporate the useful parts into #7229. Thanks again for catching this and for the helpful contribution!

omlahore:
Closing in favour of #7229, which came first. Thanks for picking up the test shape.

PR #7360, opened 2026-09-10
fix(pkg/filematcher): escape regex special characters in pattern matcher

Norway-02:
**What this PR does**:
Fixes an issue in `pkg/filematcher` where regular expression metacharacters (`+`, `(`, `)`, `{`, `}`, `|`, `^`) outside character classes `[...]` were not escaped when building internal regular expressions. As a result, file paths containing these characters (such as `a+b.txt`, `foo(1).txt`, `foo{1}.txt`, `foo|bar.txt`, or `^foo.txt`) failed to match their corresponding patterns.

**Why we need it**:
Without escaping regex metacharacters outside `[...]`, pattern matching fails on standard file paths and names containing parentheses, plus signs, braces, pipes, or carets.

Fixes # NONE

N/A

### DIFF
--- a/pkg/filematcher/filematcher.go
+++ b/pkg/filematcher/filematcher.go
@@ -181,11 +181,23 @@ func (p *Pattern) regexpString() string {
 		escSL += `\`
 	}
 
+	inClass := false
+
 	for scan.Peek() != scanner.EOF {
 		ch := scan.Next()
 
 		switch ch {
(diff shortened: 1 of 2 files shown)


Norway-02:
cc @mohammedfirdouss @TheRealShek @Warashi - Please review when you get a chance. Thanks!

PR #7378, opened 2026-09-12
fix(pkg/yamlprocessor): guard GetValue, ReplaceString, and Bytes against nil processor

Norway-02:
**What this PR does**:
Adds defensive nil-receiver and uninitialized AST file pointer guards (`p == nil || p.file == nil`) to `GetValue()`, `ReplaceString()`, and `Bytes()` in `pkg/yamlprocessor`.

**Why we need it**:
Calling `GetValue()`, `ReplaceString()`, or `Bytes()` on an uninitialized or nil `*Processor` previously caused an unhandled `runtime error: invalid memory address or nil pointer dereference` panic. Adding these checks prevents panics and safely returns clean error messages or nil byte slices.

Fixes # NONE

N/A

### DIFF
--- a/pkg/yamlprocessor/yamlprocessor.go
+++ b/pkg/yamlprocessor/yamlprocessor.go
@@ -50,6 +50,9 @@ func NewProcessor(data []byte) (*Processor, error) {
 //
 // e.g. "$.foo.bar[0].baz"
 func (p *Processor) GetValue(path string) (interface{}, error) {
+	if p == nil || p.file == nil {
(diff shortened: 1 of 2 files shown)


Norway-02:
cc @mohammedfirdouss @TheRealShek @Warashi - Please review when you get a chance. Thanks!

armistcxy:
Thanks for the PR, but I don't think this change is warranted. `NewProcessor()` is the only way to construct a `*Processor` and it either returns `(nil, err)` on failure or a fully-initialized `&Processor{file: f}` on success

